PR #9, opened 2018-01-18
Fix build on gcc6

notaz:
Fixes build with mingw from ubuntu 17.10.
Missing C++17 features I guess.


doitsujin:
Looks good. I'm using GCC 7.2 so I wasn't aware of this.

PR #19, opened 2018-01-20
Added script to setup dlls in wine

Raffarti:
Out for review.


Raffarti:
This does not create the dlls links. I'm not sure whether this has to be implemented in meson.

doitsujin:
Just to make sure I understand correctly, this way of setting up the DLLs would work best if the DLLs were named `d3d11_vk.dll` and `dxgi_vk.dll`, right? In that case I'd just change the library names so that we don't need any symlinks at all. It kind of breaks the current way of using the libraries, but with the setup script making things easier anyway I believe it's worth it.

In any case I'd like one minor change, and that is to move `wine_utils` out of the `src` directory into the top level directory. `src` should only contain code that is required to build the libraries.

Raffarti:
renaming the dlls would also do

doitsujin:
Alright, will tend to that later, thanks.

PR #21, opened 2018-01-21
[dxvk] fix missing renderPassEnd calls

libcg:
vkCmdBlitImage and vkCmdResolveImage need to be called outside a render
pass. flatten logic.

this fixes mipmap generation with Trackmania.


doitsujin:
Good catch :+1:

PR #40, opened 2018-01-29
Add an option to set up dxvk through overrides

Raffarti:
Minimal implementation.

### DIFF
--- a/wine_utils/dlls_setup.sh.in
+++ b/wine_utils/dlls_setup.sh.in
@@ -35,19 +35,32 @@ function redirect {
     fi
 }
 
+function addToPATH {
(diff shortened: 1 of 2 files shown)


Raffarti:
Ups I've cleaned up a little too much, need add a commit.

doitsujin:
What exactly does this do?

Raffarti:
`sh dlls_setup.sh overrides` can be used to symlink dxvk dlls to system32/syswow64 and set them to native. This doesn't prevent wine updates from breaking links (#40), but it's very simple and provides a minimal help to those who want to test wine-vulkan. Rest of the script behaves the same.

PR #59, opened 2018-02-05
Setup script: added an option to check status

Raffarti:
I mentioned this in #57.
It could help troubleshooting when things go wrong. Not really necessary, hopefully will be handy.

### DIFF
--- a/wine_utils/setup_dxvk.sh.in
+++ b/wine_utils/setup_dxvk.sh.in
@@ -44,6 +44,23 @@ function removeOverride {
     fi
 }
 
+function checkOverride {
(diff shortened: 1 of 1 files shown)


Raffarti:
forgot to update help

Raffarti:
Added some work which aims to improve usage of this script within another script:
- Added return value 2 to check to distinguish between script failing and check not passed.
- Added -q|--quite to output to null
- Added -y|-n to avoid user interaction

Also added a test to check validity of the WINEPREFIX directory.

Pretty standard shiny stuff awaiting judgement.

Issue #32, opened 2018-01-24
"dlls_setup.sh" not executable and with too generic a name


ssorgatem:
I just uploaded a PKGBUILD for dxvk and noticed  what the title says.

It's easily fixable from the packaging point of view, but i think I'd be better to just do it upstream, i.e., set the executable bit on the script and rename it to something like "dxvk_dlls_setup.sh", to make it evident that it's a script belonging to dxvk.

What do you think?

If I weren't completely unfamiliar with meson, I'd be submitting a pull request ;)

Raffarti:
I second the name change. I'm not familiar with mason either though, I'm not sure which is the proper way to fix permissions.

Raffarti:
Also note that the script only sets the dlls for the build that generates it, so you need one for 32 and one for 64 bit or to merge them together. Specifically, the dlls redirection is common but the PATH environment modification is not, as it have to point to the proper architecture.

doitsujin:
I think a good name would be `setup_dxvk.sh`.

On a related note, the script currently hardcodes the path where the DLLs are installed, which causes trouble with any sort of binary release (as you probably have noticed). I considered changing it to `$PWD` but that is not an optimal solution either. Any thoughts on this?

Raffarti:
I didn't used `$PWD` because it would mess things up if the script is not used in its folder. Packaging is no easy matter in my opinion, for example having the setup script in the same folder of dlls not what I'd expect in a package (windows dlls and executables shouldn't be in bin folder the same as the linux bash script). I was thinking about adding an meson option to change the hardcoded path to dlls, that would be quite clean and easy for packagers.

doitsujin:
The thing with binary releases in general is that people should not be forced to put them into a specific directory, so the script **must** determine the install path at runtime anyway. `$PWD` is bad, but better than what we have now, and checking whether the two DLLs exist within the given directory is trivial.

Alternatively we could allow/force users to specify the full path as a command line argument as well, but then have to make sure that the path is absolute.

Raffarti:
> Alternatively we could allow/force users to specify the full path as a command line argument as well, but then have to make sure that the path is absolute.

I though about that, but since wine is used by any kind of user I'd keep it as simple as possible.
Sadly, packaging is a nasty task: if you want to give packagers absolute freedom (allow them to place the script in a different directory than dlls) and to allow it to be specified at runtime common ways are:
- Environment variable (but it's not a clean solution)
[...]

PR #63, opened 2018-02-06
Fixes for the setup script

Raffarti:
I wonder why I can't type it correctly.


Raffarti:
Last commit should fix #32. Deleting the script in the build dir *might* be necessary.

Raffarti:
Shouldn't the script be named `dxvk_setup.sh` rather than the other way around? That would make it easier to find when added to PATH.

doitsujin:
There's two scripts anyway (one for 32-bit and one for 64-bit), and with my recent changes to the script they are required to be in the same location as the DLL files, so adding it to `$PATH` is not an option anyway.

PR #72, opened 2018-02-12
Fix for RTV and DSV not handling DXGI_FORMAT_UNKNOWN

ZeroFault:
Some Unreal Engine 4 games need this for render target views to be created.


doitsujin:
This is a rather dirty hack. I implemented a proper fix in 4965f8c31868b2c093b3c0eda047cae3a4dd7da9, please let me know if that works.

ZeroFault:
Yes, that works. Thank you.

PR #100, opened 2018-02-24
Throw error if requested layer is not installed

GabrielMajeri:
This pull request will make missing layers crash the application (specifically, the standard validation layer).

Somebody might enable `DXVK_DEBUG_LAYERS`, and yet forget to install the validation layers. In that case, they would see no errors.


doitsujin:
I'm not entirely sure if throwing a hard error is the best solution, but since layers have to be enabled manually anyway, it certainly makes sense.

PR #105, opened 2018-02-27
[WIP] Configurable HUD

vesim987:
Added configurable HUD

TODO:
- Add more HUD elements: CPU usage, GPU usage.
- Clean-up code.


doitsujin:
Having thought about this for a bit, I think this is the kind of feature creep that is a bit premature, and also something I don't want to have to maintain. I don't think the DXVK HUD needs to be able to display all sorts of system information; some additional performance metrics and maybe some sort of benchmarking capability/frametime measurements would be okay, but anything beyond that is unnecessary bloat in my opinion.

Another thing I don't like is that commit 96f56ada7fce98422a0c48a1bcb6164695d1a9a1 makes changes to something completely unrelated, and `hud_cpu.cpp` seems to be missing (or I'm just too stupid to find it, which certainly is a possibility).

vesim987:
Ouch, i will remove https://github.com/doitsujin/dxvk/commit/96f56ada7fce98422a0c48a1bcb6164695d1a9a1 later. 
`hud_cpu.cpp`, it is here because i was trying to find a way to get CPU usage under wine, but all "Windows methods" failed, only parsing `/proc/stat` will be work. And i forget to remove that.

If you doesn't want any system information's, then I can add minimum, average, maximum frame rates, like `Performance Metrics` from DOOM's 2016(without CPU and GPU usage):

doitsujin:
I think @shmerl has expressed some interest in a generic Vulkan HUD/performance overlay as well.

shmerl:
Yep. I was thinking about making something of the sort, that would use Vulkan layer. As such, it can be used with any Vulkan driver in theory. But I never really used Vulkan in general, so it's something I'd need to learn :)

volca02:
Here's how amd implemented a loadable layer FPS overlay for vulkan - https://github.com/GPUOpen-Tools/OCAT/blob/master/GameOverlay/vulkan/src/

shmerl:
Ah, if they already have it, can it be used as a standalone layer?

volca02:
I think it's windows only, to be honest. It serves as an example, though.

shmerl:
Yeah, as an example it's good.

doitsujin:
Added some new stuff and config options to the HUD in 1c22ae97965b1f6a8cdc67c607304437beb67f36.

nim:
I am setup DXVK_HUD=1 environment variable, but i am not see anything

nim:
I am setup DXVK_HUD=1 environment variable, but i am not see anything. I have windows 10

PR #115, opened 2018-03-02
Debug logging for compute shaders

ZeroFault:
This will make it easier to figure which compute shaders are failing.

### DIFF
--- a/src/dxvk/dxvk_compute.cpp
+++ b/src/dxvk/dxvk_compute.cpp
@@ -52,7 +52,7 @@ namespace dxvk {
 
     if (Logger::logLevel() <= LogLevel::Debug) {
       Logger::debug("Compiling compute pipeline..."); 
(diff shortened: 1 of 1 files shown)


src/dxvk/dxvk_compute.cpp
```diff
@@ -49,6 +49,11 @@ namespace dxvk {
   
   void DxvkComputePipeline::compilePipeline() {
     std::vector<VkDescriptorSetLayoutBinding> bindings;
+
+    if (Logger::logLevel() <= LogLevel::Debug) {
+      Logger::debug("Compiling compute pipeline..."); 
+      if(m_cs != nullptr) Logger::debug(str::format("  cs  : ", m_cs ->debugName()));
```
doitsujin: that null pointer check is actually not necessary, compute pipelines cannot be created without a compute shader.

PR #122, opened 2018-03-04
[d3d11] add stub for ID3DUserDefinedAnnotation

Mixaill:
This patch prevents insane amount of logs (about 80-100 MiB/min) from World of Tanks 1.0.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -67,6 +67,7 @@ namespace dxvk {
     COM_QUERY_IFACE(riid, ppvObject, ID3D11DeviceContext);
     
     Logger::warn("D3D11DeviceContext::QueryInterface: Unknown interface query");
+    Logger::warn(str::format(riid));
     return E_NOINTERFACE;
   }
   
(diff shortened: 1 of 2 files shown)


doitsujin:
Interesting, never knew that this interface even existed.

Question is, do we actually *need* this? As I see it, wine does not implement this interface either. In order to prevent log spam, we can just return `E_INVALIDARG` from the `QueryInterface` method *without* logging a message if this specific interface is queried.

The reason why I think this might be better is that there are a couple of other interfaces as well that some applications like to query, but don't actually use if the implementation doesn't provide them. One other example would be `ID3D11Debug`, and I've even seen applications query RenderDoc-specific D3D11 interfaces which of course we don't have to implement.

As for the code itself, two things:
- `d3d11_user_defined_annotation.(h|cpp)` seem like awfully lengthy file names, unless there is potential confusion I'd suggest to chnage them to `d3d11_annotation.(h|cpp)`.
- Please keep the formatting somewhat consistent with the rest of the code (i.e. don't put `{` and `);` on new lines).

Mixaill:
>we can just return E_NOINTERFACE from the QueryInterface method without logging a message if this specific interface is queried.

fixed in that way

Yardanico:
Is it possible to do the same for err:   D3D11DeviceContext::SetPredication: Not implemented ?
Watch Dogs spams that like crazy

Issue #60, opened 2018-02-05
Build with wineg++ wrapper


pchome:
I'm trying to build project w/ **wineg++** but it seems only tests/* executables wrapped correctly. Both **d3d11.dll.so** and **dxgi.dll.so** are useless and produce errors when I'm trying to replace original libraries :
```
...
uild.w64_5\tests\dxgi\dxgi-factory.exe.so: malloc.c:2399: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.
err:seh:call_stack_handlers invalid frame 7ffea8fc25c0 (0x132000-0x230000)
err:seh:NtRaiseException Exception frame is not in stack limits => unable to dispatch exception.
```
`[221151.875587] d3d11-triangle.[6933]: segfault at 366 ip 0000000000000366 sp 00007ffc3ed84bf8 error 14 in bash[55f36b668000+e4000]`

So the question is: Is it possible to build WINE's **native linux libraries** to replace original ones? Or i'm doing this totally wrong and really need to configure crossdev environment w/ mingw32?

Here is the list of changes was made :
`build-win64-wine.txt`
```meson
[binaries]
c = '/usr/bin/winegcc'
cpp = '/usr/bin/wineg++'
ar = '/usr/bin/ar'
strip = '/usr/bin/strip'
exe_wrapper = 'sh'

[properties]
c_args = ['-m64', '-DNOMINMAX', '-DWINEBUILD']
c_link_args = ['-mwindows', '-lpthread']

cpp_args = ['-m64', '-DNOMINMAX', '-DWINEBUILD']
cpp_link_args = ['-mwindows', '-lpthread']

[host_machine]
system = 'linux'
cpu_family = 'x86_64'
cpu = 'x86_64'
endian = 'little'
```
actually **-mwindows** required only by tests/* and **-lpthread** by d3d11.dll.so and dxgi.dll.so

`meson.build.wine`
```meson
project('dxvk', ['c', 'cpp'], default_options : ['cpp_std=gnu++1z'])

cpu_family = target_machine.cpu_family()

dxvk_compiler = meson.get_compiler('cpp')
dxvk_include_path = include_directories('./include')

lib_d3d11          = dxvk_compiler.find_library('d3d11')
lib_dxgi           = dxvk_compiler.find_library('dxgi')
[...]

PR #123, opened 2018-03-05
Add MSVC support

Mixaill:
This patch set contains various improvements which allows to build DXVK on MSVC compiler.

To build DXVK you also need:

* apply this patch to meson https://github.com/mesonbuild/meson/pull/2836
* get MSVC-compatible SPIRV-Tools (you can use https://github.com/microsoft/vcpkg with this PR https://github.com/Microsoft/vcpkg/pull/2959 ) and place them into lib/lib32 folder


meson.build
```diff
@@ -1,8 +1,12 @@
-project('dxvk', ['c', 'cpp'], default_options : ['cpp_std=c++1z'])
+project('dxvk', ['c', 'cpp'], default_options : ['cpp_std=c++17'])
```
qwertychouskie: This should probably only be changed for MSVC as changing it for GCC could cause some older versions to not compile it even though they can.

src/dxvk/dxvk_memory.cpp
```diff
@@ -34,7 +34,7 @@ namespace dxvk {
   DxvkMemory& DxvkMemory::operator = (DxvkMemory&& other) {
     m_chunk   = std::exchange(other.m_chunk,  nullptr);
     m_heap    = std::exchange(other.m_heap,   nullptr);
-    m_memory  = std::exchange(other.m_memory, VkDeviceMemory(nullptr));
+    m_memory  = std::exchange(other.m_memory, VkDeviceMemory(0));
```
doitsujin: this should be `VK_NULL_HANDLE` instead of 0 or `nullptr`.

doitsujin: Settng the C++ version to c++17 will break GCC 6.x, which is what currently ships with Ubuntu. Breaking that is questionable, but I'm not sure if there is a way to set different values for different compilers.

src/d3d11/d3d11_context.h
```diff
@@ -7,6 +7,10 @@
 #include "d3d11_context_state.h"
 #include "d3d11_device_child.h"
 
+#ifndef WINBOOL
+typedef BOOL WINBOOL;
```
doitsujin: I think `WINBOOL` shouldn't be a thing in the first place, if it's the same as `BOOL` then I think we should rather get rid of all uses of `WINBOOL` instead.

Mixaill:
*  `WINBOOL` replaced with `BOOL`
* 0 replaced with `VK_NULL_HANDLE`
* `cpp_std=c++17` only for MSVC.

doitsujin:
Had to revert the changes because it breaks mingw builds. mingw does not support `.def` files and exits with an error message when linking the libraries. Please open a new PR once that is sorted out.

pchome:
@Mixaill 
I see this changes mostly echoed with those one from #60 . Now I also using `.def` files and redefined `meson.build` files but in a bit different way ( https://github.com/pchome/dxvk ).

Last changes was to create `./linux/` directory and move every specific change there to less interfere w/ upstream changes. It mostly WIP and POC but maybe some ideas would be useful for you.

And maybe you'll find even better way to merge both linux and MSVC builds together.

pchome:
@Mixaill 
Also there is ability to create custom meson build file set and patches with [wrap feature](http://mesonbuild.com/Wrap-dependency-system-manual.html). As last chance to extend some feature.

pchome:
BTW: `subdir()` it's mesons `include` - includes everything from specified dir in place it defined.
So possible to do something like this: https://github.com/doitsujin/dxvk/commit/b771995958e2f8a942872495349279c043e17e4c

Mixaill:
I wonder, if CMake will be easier for that crossplatform crosscompiler mess?

doitsujin:
One thing just came to mind, can you try compiling DXVK on MSVC++ with the MinGW D3D11 headers instead of the system headers? Maybe that way we could avoid using `.def` files. As far as I'm aware, the main issue is that the Windows headers are meant for application developers who *import* the D3D libs, not for crazy Linux people trying to re-implement the entire API.

Mixaill:
>can you try compiling DXVK on MSVC++ with the MinGW D3D11 headers instead of the Windows SDK headers? 

It will require some nontrivial black magic.

-----

BTW, my MinGW form MSYS2 project has all the `D3D11_` header and doesn't have _countof macro =\

PR #125, opened 2018-03-05
Small fixes

pchome:


doitsujin:
I don't see why the first three are necessary, and I don't see what good a2e237c2e8db2db9607f25a273821364f6b29737 would do except making the whole include even more inconsistent than it already is. Originally I intended to use `<>` style includes across the different modules (which *does* already work), but in retrospect I think it was a bad idea to begin with and we should just stick with the file paths as they are.

In general, it would be nice to have some sort of explanation what you did and **why** you did it. After all, that's what the description field for pull requests is for.

I'm willing to merge 940cabd2690ab165151572fa8fd242147bf179bb separately though.

pchome:
> In general, it would be nice to have some sort of explanation what you did and why you did it.

I thought it's obvious for such small patches. There is small explanations inside commits ([...]). But ok:
1 - dxbc_decoder.h in dxbc_decoder.h
2 - spirv_module.h in dxgi_presenter.h included two rows earlier
3 - Cosmetic, `#include "spirv/spirv_module.h"` looks better than `#include "../spirv/spirv_module.h"` as for me

pchome:
> making the whole include even more inconsistent than it already is

So yes, first 3 are cosmetic changes. But the last one uses `#include "util/com/com_pointer.h"` based on previous one. Need to be fixed if you reject previous patch.

Or it's ease to change all "inconsistent" includes at once with simple regexp. If you wish I can do this in any way you like them to be.

PR #130, opened 2018-03-05
Add support for MSVC, attempt 3

Mixaill:
since previous PR (#123):

* fixed .def files for MinGW. They should just works fine
* remove `DLLEXPORT` macro. Symbols export are handled by `.def` files now.


doitsujin:
This still breaks 32-bit mingw builds because there's some stdcall name mangling going on:
```
src/d3d11/d3d11@sha/d3d11_main.cpp.obj: In function `D3D11CreateDevice@40':
/home/philip/dxvk.mixail/build.32/../src/d3d11/d3d11_main.cpp:39: undefined reference to `CreateDXGIFactory@8'
/home/philip/dxvk.mixail/build.32/../src/d3d11/d3d11_main.cpp:118: undefined reference to `DXGICreateDevicePrivate@12'
```
It works again if I add back the `DLLEXPORT` macro. A quick and dirty solution might be to restore `DLLEXPORT` and to just leave the define empty on MSVC.

meson.build
```diff
@@ -1,8 +1,15 @@
...
+project('dxvk', ['c', 'cpp'])
 
 cpu_family = target_machine.cpu_family()
 
 dxvk_compiler = meson.get_compiler('cpp')
+if dxvk_compiler.get_id() == 'msvc'
+  add_global_arguments('-DNOMINMAX', language : 'cpp')
+  dxvk_cpp_std='c++17'
+else
+  dxvk_cpp_std='c++1z'
```
libcg: leaving this as is might break MSVC if we starting using new features. might be safer to use c++17 everywhere.

roothorick: `std=c++17` breaks GCC 6 *now*. This was changed from using c++17 everywhere for that specific reason.

ZeroFault: MSVC is behind GCC in supporting some c++17 features anyway, so we already have to be careful not to break it. Maybe use c++latest instead for MSVC?

Mixaill: fixed

Mixaill:
>It works again if I add back the DLLEXPORT macro. A quick and dirty solution might be to restore DLLEXPORT and to just leave the define empty on MSVC.

copypaste from Discord:

Looks like,
* `--add-stdcall-alies` creates "demangled" symbols but not vice versa
* `--enable-stdcall-fixup` does not always tries to strip @ 
* GCC/ld does not support to link against "demangled" stdcall at all.

So, the most hackless way for MinGW is:
- use ` __declspec(dllexport)` to generate stdcall mangled export symbols to link against it
- use `--add-stdcall-alias` to generate demangled export symbols which will be used by applications.

oscarbg:
@doitsujin @Mixaill please make the effort to integrate ASAP..
have locally some branch with MSVC support (old one patched by me) and have some made some enhancements to it (for ex. have a fix as said in other thread Intel IGPU Win Vulkan enter infinite loop of by calling DXGICrateDXGIFactory dxgi function from his vulkan library and hangs)..
but is getting old fast..
thanks for your work..

oscarbg:
thanks!

PR #137, opened 2018-03-09
Add Custom PCI Vendor and Device ID Support

Guy1524:
This commit allows the user to configure a custom GPU Vendor and Device PCI ID

This is useful for getting programs to behave similarly on different graphics cards.


Yardanico:
We need to add this to readme/wiki

PR #143, opened 2018-03-10
[WIP] [dxbc] partially implemented dcl_hs_max_tessfactor

Mixaill:
Implemented value reading but i am not sure where i should apply it. Looks like tessellation factors are set up in `DxbcCompiler::emitHsSystemValueStore`.

I'm confused with using of `uint32_t` type because tessellation factor has `float32` type in DXBC.


Mixaill:
Oh, it implemented now :)

doitsujin:
Some unfortunate timing there as I had already started working on this (see 28880d0fa899bb8dcc9ba644f3f3a315e04ba9ab)

Tessellation factors are floats in SPIR-V as well, but those `uint32_t` values represent SPIR-V IDs, not the actual values (since they can be computed dynamically).

Issue #86, opened 2018-02-17
Galactic Civilizations III requires DxgiOutput::FindClosestMatchingMode


roothorick:
### Software information
Galactic Civilizations III, current Steam version

Crashes on start with "Fatal error: Failed to initialize display."

### System information
- GPU: Radeon RX 480
- Driver: Mesa 17.3.2
- Wine version: 2.21-staging
- DXVK version: git fd1fd40

### Apitrace file(s)
todo

### Log files
- d3d11.log: (empty)
- dxgi.log: https://pastebin.com/Ks1gYbyS

pingubot:
BFBC2 has the same issue, see #16

doitsujin:
Fixed by #153.

Issue #16, opened 2018-01-20
Battlefield Bad Company 2 DX11: Works now without MSAA 


pingubot:
Hi,

the game starts, but dosesn't display anything. I used wine in virtual desktop mode with 1024x768, same resolution in game. Disabling wine virtual desktop also leeds to the same issue:

System: OpenSuse Tumbleweed, i3570k, 16GB RAm, NVidia Gtx 970, 390.12 driver
[d3d11.log](https://github.com/doitsujin/dxvk/files/1648888/d3d11.log)
[dxgi.log](https://github.com/doitsujin/dxvk/files/1648889/dxgi.log)

Debug Log:

ParameterValidation(ERROR): object: 0x0 type: 0 location: 1890 msgCode: 352323990: vkCmdSetViewport 0: width (0.000000) exceeds permitted bounds (0,16384). The spec valid usage text states 'width must be greater than 0.0 and less than or equal to VkPhysicalDeviceLimits::maxViewportDimensions[0]' (https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VUID-VkViewport-width-01227)

Cu,
Christian
[d3d11.log](https://github.com/doitsujin/dxvk/files/1648890/d3d11.log)
[dxgi.log](https://github.com/doitsujin/dxvk/files/1648891/dxgi.log)

pingubot:
Since DxgiOutput::FindClosestMatchingMode was implemented the game now starts but freezes/crashes during loading a savegame/map. This happens in both branches. It does not happen in wined3d.

Logs with debug layer attched:

[console.log](https://github.com/doitsujin/dxvk/files/1812855/console.log)
[BFBC2Game_dxgi.log](https://github.com/doitsujin/dxvk/files/1812856/BFBC2Game_dxgi.log)
[BFBC2Game_d3d11.log](https://github.com/doitsujin/dxvk/files/1812857/BFBC2Game_d3d11.log)

pingubot:
@doitsujin : i tried to get an apitrace and i got one done (with wine 3.3 without dxvk). The problem is that trying to replay the trace with dxvk does not work here. I only get a grey screen and the replay crashes with a lot of vulkan errors displayed.In the real game i come to the menu and only fail when loading a level. Any hint what could causing this ? Should i upload the trace nonetheless ?

pingubot:
[...]

PR #153, opened 2018-03-12
[dxgi] implement DxgiOutput::FindClosestMatchingMode()

Mixaill:
looks like some Frostbite titles requires it.

should help with #86 #16

### DIFF
--- a/src/dxgi/dxgi_output.cpp
+++ b/src/dxgi/dxgi_output.cpp
@@ -46,8 +46,92 @@ namespace dxvk {
     const DXGI_MODE_DESC *pModeToMatch,
           DXGI_MODE_DESC *pClosestMatch,
           IUnknown       *pConcernedDevice) {
-    Logger::err("DxgiOutput::FindClosestMatchingMode: Not implemented");
-    return E_NOTIMPL;
+    if (pModeToMatch == nullptr) {
+      Logger::err("DxgiOutput::FindClosestMatchingMode: pModeToMatch is nullptr");
+      return DXGI_ERROR_INVALID_CALL;
+    }
+        
+    if (pClosestMatch == nullptr) {
+      Logger::err("DxgiOutput::FindClosestMatchingMode: pClosestMatch is nullptr");
+      return DXGI_ERROR_INVALID_CALL;
+    }
+
+    if (pModeToMatch->Format == DXGI_FORMAT_UNKNOWN && pConcernedDevice == nullptr) {
+      Logger::err("DxgiOutput::FindClosestMatchingMode: no pointer to device was provided for DXGI_FORMAT_UNKNOWN format");
+      return DXGI_ERROR_INVALID_CALL;
+    }
+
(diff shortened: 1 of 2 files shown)


src/dxgi/dxgi_output.cpp
```diff
@@ -46,8 +46,77 @@ namespace dxvk {
...
+      Logger::err("DxgiOutput::FindClosestMatchingMode: pClosestMatch is nullptr");
+      return DXGI_ERROR_INVALID_CALL;
+    }
+
+    if (pModeToMatch->Format == DXGI_FORMAT_UNKNOWN && pConcernedDevice == nullptr) {
+      Logger::err("DxgiOutput::FindClosestMatchingMode: no pointer to device was provided for DXGI_FORMAT_UNKNOWN format");
+      return DXGI_ERROR_INVALID_CALL;
+    }
+
+    if (pModeToMatch->Format == DXGI_FORMAT_UNKNOWN) {
```
doitsujin: What exactly is supposed to happen when `Format` is `UNKNOWN`? I can't find any documentation about it.

It's fine to leave this open for now, but since this is a legal thing to do, please add some kind of `// TODO implement` comment.

src/dxgi/dxgi_output.cpp
```diff
@@ -46,8 +46,77 @@ namespace dxvk {
...
+    DXGI_MODE_DESC modeToMatch = *pModeToMatch;
+    UINT modesCount = 0;
+    GetDisplayModeList(pModeToMatch->Format, 0, &modesCount, nullptr);
+
+    if (modesCount == 0) {
+      Logger::err("DxgiOutput::FindClosestMatchingMode: no device formats were found");
+      return DXGI_ERROR_NOT_FOUND;
+    }
+
+    DXGI_MODE_DESC * modes = new DXGI_MODE_DESC[modesCount];
```
doitsujin: I'd like to see an `std::vector<DXGI_MODE_DESC>` here instead of a raw array.

src/dxgi/dxgi_output.cpp
Mixaill: There's some obfuscated stuff in MSDN:
https://msdn.microsoft.com/en-us/library/windows/desktop/bb174547(v=vs.85).aspx

>If `pConcernedDevice` is `NULL`, `Format` **CANNOT** be `DXGI_FORMAT_UNKNOWN`.

using some basic laws of logic, we can read this like

>If `pConcernedDevice` is NOT `NULL`, `Format` **COULD** be `DXGI_FORMAT_UNKNOWN`.

But `Format` in structures from `GetDisplayModeList()` cannot be `DXGI_FORMAT_UNKNOWN` by definition. 

There is way in case of `DXGI_FORMAT_UNKNOWN` and `pDevice != nullptr` we should perform additional format matching. It may be just ignoring of `Format` field or using some range of formats but MSDN nothing says about of criteria. :(

src/dxgi/dxgi_output.cpp
Mixaill: OK, 'll fix.

PR #156, opened 2018-03-12
[d3d11] D3D11Device::QueryInterface() suppress warning for unknown GUID

Mixaill:
there is no public information about this interface

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -67,6 +67,10 @@ namespace dxvk {
 
     if (riid == __uuidof(ID3D11Debug))
       return E_NOINTERFACE;      
+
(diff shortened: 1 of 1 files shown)


src/d3d11/d3d11_device.cpp
```diff
@@ -67,6 +67,10 @@ namespace dxvk {
 
     if (riid == __uuidof(ID3D11Debug))
       return E_NOINTERFACE;      
+
+    //d56e2a4c-5127-8437-658a-98c5bb789498, from GTA V, no occurrences in Google
+    if (riid == GUID{0xd56e2a4c,0x5127,0x8437,{0x65,0x8a,0x98,0xc5,0xbb,0x78,0x94,0x98}})
```
MaceWindu: google shows that it is IWineD3DDevice

PR #157, opened 2018-03-12
[dxgi] fix DxgiOutput::FindClosestMatchingMode() implementation

Mixaill:
* remove useless checks
* fixed refresh rate matching
* fixed unspecified refresh rate detection

### DIFF
--- a/src/dxgi/dxgi_output.cpp
+++ b/src/dxgi/dxgi_output.cpp
@@ -93,21 +93,14 @@ namespace dxvk {
     std::vector<DXGI_MODE_DESC> modes(modesCount);
     GetDisplayModeList(pModeToMatch->Format, 0, &modesCount, modes.data());
 
-    //filter out modes with different scanline ordering if it was set
-    if (modeToMatch.ScanlineOrdering != DXGI_MODE_SCANLINE_ORDER_UNSPECIFIED) {
-      for (auto it = modes.begin(); it != modes.end();) {
-        if (it->ScanlineOrdering != modeToMatch.ScanlineOrdering)
-            it = modes.erase(it);
-        else
-            ++it;
-      }
-    }
+    /* TODO: add scaling and scanline filter when we implement they */
 
     //filter out modes with different refresh rate if it was set
-    if (modeToMatch.RefreshRate.Denominator != 0 || modeToMatch.RefreshRate.Numerator != 0) {
+    if (modeToMatch.RefreshRate.Denominator != 0 && modeToMatch.RefreshRate.Numerator != 0) {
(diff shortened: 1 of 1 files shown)


src/dxgi/dxgi_output.cpp
```diff
@@ -93,21 +93,42 @@ namespace dxvk {
...
 
+    //filter out modes with different scaling if it was set
+    if (modeToMatch.Scaling != DXGI_MODE_SCALING_UNSPECIFIED) {
+      if (modeToMatch.Scaling != DXGI_MODE_SCALING_CENTERED) {
+        Logger::warn("DxgiOutput::FindClosestMatchingMode: DXVK currently supports only DXGI_MODE_SCALING_CENTERED. Fall back to it.");
+        modeToMatch.Scaling = DXGI_MODE_SCALING_CENTERED;
+      }
+
+      for (auto it = modes.begin(); it != modes.end();) {
+         if (it->Scaling != modeToMatch.Scaling)
```
doitsujin: Why is this needed? Can't we just *ignore* the Scaling? If we're falling back to CENTERED anyway, I really don't see the point in filtering out any modes based on the scaling.

src/dxgi/dxgi_output.cpp
```diff
@@ -93,21 +93,42 @@ namespace dxvk {
...
+
     //filter out modes with different scanline ordering if it was set
     if (modeToMatch.ScanlineOrdering != DXGI_MODE_SCANLINE_ORDER_UNSPECIFIED) {
+      if (modeToMatch.ScanlineOrdering != DXGI_MODE_SCANLINE_ORDER_PROGRESSIVE) {
+        Logger::warn("DxgiOutput::FindClosestMatchingMode: DXVK currently supports only DXGI_MODE_SCANLINE_ORDER_PROGRESSIVE. Fall back to it.");
+        modeToMatch.ScanlineOrdering = DXGI_MODE_SCANLINE_ORDER_PROGRESSIVE;
+      }
+
       for (auto it = modes.begin(); it != modes.end();) {
         if (it->ScanlineOrdering != modeToMatch.ScanlineOrdering)
```
doitsujin: Same thing here, I don't think we need this (or the clunky log message) until DXVK actually reports the scaling modes and scanline modes properly.

src/dxgi/dxgi_output.cpp
Mixaill: sure, removed

src/dxgi/dxgi_output.cpp
Mixaill: same, removed

PR #162, opened 2018-03-14
Some changes to the packaging script

ssorgatem:
I've added an option for skipping the building of test binaries, and I've added that to the packaging script.
I really didn't want to add the meson_options.txt file... but it seems there's no way around it (that I could find, at least...).

I've also enabled "unity builds" there, as recommended by the Meson documentation. This makes a couple of warnings appear, but so far it seems to work (not tested extensively, though)

I've also added support for relative output directory paths to the packaging script.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,0 +1,1 @@
+option('enable_tests', type : 'boolean', value : false)
--- a/package-release.sh
+++ b/package-release.sh
@@ -8,7 +8,7 @@ fi
 DXVK_VERSION="$1"
 DXVK_SRC_DIR=`dirname $(readlink -f $0)`
 DXVK_TMP_DIR="/tmp/dxvk-$DXVK_VERSION"
-DXVK_ARCHIVE_PATH="$2/dxvk-$DXVK_VERSION.tar.gz"
+DXVK_ARCHIVE_PATH=$(realpath "$2")"/dxvk-$DXVK_VERSION.tar.gz"
 
 function build_arch {
   cd "$DXVK_SRC_DIR"
--- a/package-release.sh
+++ b/package-release.sh
@@ -18,6 +18,7 @@ function build_arch {
         --prefix "$DXVK_TMP_DIR/install.$1"           \
 	--unity on                                    \
         --strip                                       \
+        -Denable_tests=false                          \
         "$DXVK_TMP_DIR/build.$1"
 
   cd "$DXVK_TMP_DIR/build.$1"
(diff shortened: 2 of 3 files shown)


meson_options.txt
```diff
@@ -0,0 +1 @@
+option('skip_tests', type : 'boolean', value : false)
```
doitsujin: I think it would be better to have a "positive" option (like `build_tests`) that defaults to true, as it may be a bit more intuitive to use. We might even consider *disabling* them by default since they are mostly useless for the end user.

package-release.sh
```diff
@@ -8,15 +8,17 @@ fi
...
-DXVK_ARCHIVE_PATH="$2/dxvk-$DXVK_VERSION.tar.gz"
+DXVK_ARCHIVE_PATH=$(realpath "$2")"/dxvk-$DXVK_VERSION.tar.gz"
 
 function build_arch {
   cd "$DXVK_SRC_DIR"
 
   meson --cross-file "$DXVK_SRC_DIR/build-win$1.txt"  \
         --buildtype "release"                         \
         --prefix "$DXVK_TMP_DIR/install.$1"           \
+	--unity on                                    \
```
doitsujin: I'd like to do more extensive testing before enabling this by default, and have time to fix potential problems.

People may (well, some actually **do**) use the script to build DXVK, no reason to potentially break it for them.

package-release.sh
```diff
@@ -8,15 +8,17 @@ fi
...
         --strip                                       \
+        -Dskip_tests=true                             \
```
doitsujin: `-Dbuild_tests=false` if you change the option. Obvious, but this is the kind of line that I would almost certainly forget to change.

meson_options.txt
ssorgatem: Done

package-release.sh
ssorgatem: I agree, although in theory there shouldn't be issues with it from what I've read.

ssorgatem:
Well, as a bonus, the test binaries can now be built with MSVC... or at least it can be attempted by enabling them when using MSVC, no idea if they will build or not.

doitsujin:
I'll give it some thought, although I don't feel like this is a change that needs to be addressed immediately. 

I will merge your changes as soon as I finish messing around with my texture stuff.

ssorgatem:
I have a branch with the split scripts, I don't know whether to add that commit to this PR or to make a new one (which will depend on this one).

PR #163, opened 2018-03-14
split the package-release.sh script into 2 scripts to allow building without making a tarball

ssorgatem:
The script 'package.sh' can be used in the same way the previous single script was, and does the same; but using the script 'release.sh' alone creates a directory with the built libraries instead of a tarball

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+source release.sh
+
+function package {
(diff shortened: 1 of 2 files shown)


doitsujin:
I'll be honest, i'm not a huge fan of splitting the script into two parts. I'm also not sure if this is actually *necessary*. If people want to build dxvk from source, they are better off doing it manually in my opinion. Packagers might also need more control over the build process than what the script provides.

doitsujin:
873deed4a31c884f1d54420f0b42a2c0d3dc4be0 adds an option to disable creating the tarball.

PR #165, opened 2018-03-15
[WIP] [d3d11] stub implementation of D3D11.1 and DXGI1.2

Mixaill:
Still to do:
* format code in `d3d11_context.cpp`
* new color formats

test cases:
* Mass Effect: Andromeda (D3D11.1)
* Mirror's Edge: Catalyst (D3D11.1)
* FiveM (DXGI 1.2)

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -65,6 +65,7 @@ namespace dxvk {
     COM_QUERY_IFACE(riid, ppvObject, IUnknown);
     COM_QUERY_IFACE(riid, ppvObject, ID3D11DeviceChild);
     COM_QUERY_IFACE(riid, ppvObject, ID3D11DeviceContext);
+    COM_QUERY_IFACE(riid, ppvObject, ID3D11DeviceContext1);
     
     if (riid == __uuidof(ID3DUserDefinedAnnotation))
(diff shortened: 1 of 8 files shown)


oscarbg:
Hi Mixaill,
even with "still to do" do you mean this three games won't work without your patches and are working ok with your patches currently?
thanks..

oscarbg:
Also related, must say, it will be interesting the new color formats/spaces things as here comes the native Windows HDR support (well perhaps is DXGi 1.3 or 1.4) and that should be mappable to VK_khr swapchain colorspace ext and perhaps even hdr-metadata ext.. sadly no Vulkan drivers implementing it.. but hope NV once ships it will be on Linux and Windows drivers.. should later this year I hope..
Many Win first HDR games use NVAPI, AMD AGS HDR ext.. and many of the few which use native Win DXGI HDR are for Windows Store(?).. anyway we will always have FFXV demo or game which I think uses native Win HDR APIs even on Steam version(?)

Mixaill:
>even with "still to do" do you mean this three games won't work without your patches and are working ok with your patches currently?

It's a just stubs for some interfaces.

It may help a bit. For example, it allows to bypass FiveM check for DXGI 1.2.

Mixaill:
Commited via other PRs

PR #170, opened 2018-03-16
Fix tests for MSVC

Mixaill:
* replaced `WinMain()` with `main()`/`wmain()` for console applications.
* added missing `<string>` header in test_dxbc_disasm
* fixed `d3dcompiler` detection for MSVC


doitsujin:
> WinMain is only for application with /SUBSYSTEM=Windows. Console applications should use main/tmain/wmain.

Did you test this on MinGW? The only reason WinMain is being used in those demo applications is that MinGW actually required it, don't know if that has changed.

Mixaill:
Well, there is bug in meson. For some reason it tries to compile some executables with as GUI apps with `Windows` subsystem even if `gui_app` is set to `false`.

doitsujin:
The interest in supporting this seems to be low, and with the MinGW issues being there I don't think I'm going to support this.

PR #172, opened 2018-03-16
[dxgi] added stub implementation of IDXGIDevice2

Mixaill:


Mixaill:
`dxvk::createDxgiFactory` is about `IDXGIFactory`, but this PR is about `IDXGI*Device*`

doitsujin:
You're right, I should probably go to bed.

PR #179, opened 2018-03-18
Add possible override of wine binary in installer

gillg:


ssorgatem:
can't this be already done just by modifying $PATH?

Why is setting/modifying a new environment variable (WINE) better than using a standard one?

doitsujin:
The exact same thing can be achieved by prepending your custom wine install directory to `PATH`, I don't really see the point either to be honest.

gillg:
Yes, I'm according with you, I made it before to test my DLL install. But lot of people, not necessary friendly with bash system Path etc, are using for exemple PlayOnLinux and/or winetricks. Just to be more clear and consistent in wine sphere (like it's done on winetricks), I think it's not bad to just add a "special" environment variable to specify which wine use.

I'm not totally convinced of my choice, but the goal is just to simplify documentation and risks of path corruption for lamba users.

gillg:
Maybe, just set PATH in install script regarding a possible WINEPATH environment variable could be a better solution...

ssorgatem:
To me, it's overcomplicating things.
This looks like something PlayOnlinux and similar software should take care of.

Besides, if you are not overriding PATH but just the wine binary, that can still lead to potential conflicts if using, say, winecfg, regedit or some other wine command. This is all avoided if overriding the PATH.

So I don't really see any point in that.

Chances are, if someone is using some management software because they are unfamiliar with command line... they shouldn't have an easier time with new extra variables to set.

AFAIK, winetricks doesn't support that either and it relies on the user prepending the desired wine version to the PATH.


Besides... it doesn't matter if the wine version used in the setup script is different than whatever you use afterwards to run a game in the same prefix. It just adds a couple registry keys.

doitsujin:
> Besides... it doesn't matter if the wine version used in the setup script is different than whatever you use afterwards to run a game in the same prefix.

It will reconfigure your prefix if it's different from the wine version that you are trying to run, or maybe you don't have *any* wine in your `$PATH` at all. Running a similar setup with multiple different wine versions myself, I can certainly see the problem.

But I have to agree, using `$PATH` makes much more sense than adding (and documenting) yet another environment variable.

Thaodan:
>AFAIK, winetricks doesn't support that either and it relies on the user prepending the desired wine version to the PATH.
nope, winetricks uses the exact same variable, if you want to set the wine for winetricks you do the exact same thing.

see here:
https://github.com/Winetricks/winetricks/blob/master/src/winetricks#L4981

gillg:
Ok, you are not favorable to this. I could understand, even if I'm not totaly agree with you.
But, as @Thaodan shows, winetricks use really this sort of env trick.

Maybe if someone else see any interest to this, we will see.

Thaodan:
Its just using ${WINE:-wine} instead of wine.
Even on a 64bit calling just wine is ok.
Except maybe for pure 64bit wine.

See here for some guidelines to handle this if you want to add more.
But I think everything else except using $WINE is the task for an external script.

See here for an example:
https://github.com/Thaodan/usenew/blob/master/src/libuse/base.in.sh#L93

PR #180, opened 2018-03-18
[dxvk] print shader names on compilation failure

libcg:


doitsujin:
Implemented in 097eb89cfbed677cef7f2f716891b25a1b4ccf4a.

PR #204, opened 2018-03-25
Added DXVK_PIPECACHE_DIR environment variable for pipecache directory.

jmis:


doitsujin:
Why do we need *yet another* environment variable to change something that the user isn't supposed to see in the first place?

jmis:
Yeah, I agree.  Let me go look into an issue I was having with the pipecache and figure out if there's a better solution.

PR #210, opened 2018-03-26
[d3d11] Make D3D11Device share DXGIDevice's refcount

notaz:
If we query IDXGIDevice interface from d3d11 device as follows:
  D3D11CreateDevice(..., &device, NULL, NULL);
  device->QueryInterface(IDXGIDevice, (void **)&dxgi_device);
then manipulating dxgi_device's refcount using AddRef()/Release() will
also change device's refcount (and vice versa). In fact they seem to match.

Also, according to wine's tests, code like
  D3D11CreateDevice(..., &device, NULL, NULL);
  device->QueryInterface(IDXGIDevice, (void **)&dxgi_device);
  device->Release();
does not destroy the device. Make the refcount shared to implement this.

---
Unsure if it's the right approach, but it makes wine's tests progress much further.

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -79,8 +79,19 @@ namespace dxvk {
     Logger::warn(str::format(riid));
     return E_NOINTERFACE;
   }
-    
-  
+
+
+  // D3D11Device shares DXGIDevice's refcount
+  ULONG STDMETHODCALLTYPE D3D11Device::AddRef(void) {
+    return m_dxgiDevice->AddRef();
+  }
+
+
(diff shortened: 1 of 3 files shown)


doitsujin:
The main issue here is that DXVK does not follow COM rules properly, so messing around with the ref count will probably cause more issues than it solves.

A better solution would be to create some sort of container object which stores all the objects, implements `QueryInterface` and the reference counting methods, and `D3D11Device` and `DXGIDevice` would store a pointer to that object.

doitsujin:
ae88f83b8691fad6432409c09e017606b747b547 should fix this.

PR #231, opened 2018-04-01
[dxgi] Replace MSVC _countof macro with std::size

pchome:
`std::size` (**c++17**) support required

**pro**: portable
**con**: may not work for older clang/gcc/mingw

http://en.cppreference.com/w/cpp/iterator/size
For example `Run this code` mode produces errors for GCC-5.2(C++17) and clang-3.8(C++17). Local GCC-7.3 and clang-7 versions are ok. 
Not tested w/ MinGW64.

### DIFF
--- a/src/dxgi/dxgi_adapter.cpp
+++ b/src/dxgi/dxgi_adapter.cpp
@@ -118,7 +118,7 @@ namespace dxvk {
     }
     
     std::memset(pDesc->Description, 0, sizeof(pDesc->Description));
-    std::mbstowcs(pDesc->Description, deviceProp.deviceName, _countof(pDesc->Description) - 1);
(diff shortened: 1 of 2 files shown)


doitsujin:
does GCC 6 work? I can't afford to break that yet.

pchome:
Sorry, I have no MinGW or GCC <7.3 installed.

pchome:
https://gcc.gnu.org/onlinedocs/gcc-6.4.0/libstdc++/manual/manual/status.html#status.iso.201z says `N4280 | Non-member size() and more | Y` but maybe someone with actually installed MinGW GCC 6.x can test/confirm.

ShadowsFriend:
On my end, dxvk with this pull request compiles successfully with gcc 6.4.0 and mingw64 5.0.3.

PR #235, opened 2018-04-03
[WIP] Added Pipeline Cache Size to DXVK HUD info

Guy1524:
This can be useful to both alarm people of a abnormally large pipeline cache, and also potentially help them confirm the cause of stutters.
Currently, the size only shows what is written on disk, but this could be easily changed.
Also, there is a framework to output in a more readable format, but right now it outputs a large number of bytes.

### DIFF
--- a/src/dxvk/dxvk_device.h
+++ b/src/dxvk/dxvk_device.h
@@ -77,6 +77,16 @@ namespace dxvk {
     }
     
     /**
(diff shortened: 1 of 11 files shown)


doitsujin:
See 1c22ae97965b1f6a8cdc67c607304437beb67f36.

PR #247, opened 2018-04-06
[d3d11] Fix compilation with WINE headers

pchome:
All this structures are defined in WINE headers but `D3D11_FORMAT_SUPPORT2`.

### DIFF
--- a/src/d3d11/d3d11_include.h
+++ b/src/d3d11/d3d11_include.h
@@ -39,6 +39,7 @@ typedef enum D3D11_FORMAT_SUPPORT2 {
   D3D11_FORMAT_SUPPORT2_SHAREABLE                                     = 0x400,
   D3D11_FORMAT_SUPPORT2_MULTIPLANE_OVERLAY                            = 0x4000
 } D3D11_FORMAT_SUPPORT2;
+#ifndef __WINE__
 typedef struct D3D11_FEATURE_DATA_FORMAT_SUPPORT2 {
   DXGI_FORMAT InFormat;
   UINT        OutFormatSupport2;
@@ -88,4 +89,5 @@ typedef struct D3D11_FEATURE_DATA_SHADER_MIN_PRECISION_SUPPORT {
 typedef struct D3D11_FEATURE_DATA_D3D9_SHADOW_SUPPORT {
   BOOL SupportsDepthAsTextureWithLessEqualComparisonFilter;
 } D3D11_FEATURE_DATA_D3D9_SHADOW_SUPPORT;
-#endif
+#endif // __WINE__
+#endif // _MSC_VER


Guy1524:
Phcome, your branch has a bunch of compilation errors present, do you have any intention of getting your winegcc build to work again?

pchome:
@Guy1524 
As you can see I'm proposing those changes to be merged upstream. Since some experimental patches was dropped the only changes remained are this ones, wide-string fixes and build system patches.

My current `winebuild` branch have build order problem, you have to do `$ ninja || ninja` several times if you see errors like "emty/missing filename.o". This was introduced as result of an experiment with GCC + options build.

Previously used `winegcc` have another problem: hardcoded path to wine libs. Some (all ?) distros usually providing `winegcc` for current arch only, so there are problems to build x86 binaries on x86_64 systems (linking x86_64 libwinecrt0.a and so on).

So either one of this variants will be merged upstream as experimental, or will be transformed to build system patches. Or maybe script to convert entire DXVK as an patch for wine-staging, not sure for now.

Thaodan:
Can't we just have something like build arch= <arch>wine or <arch>mingw and so on?

About the winegcc is this a packaging problem or a upstream problem?

I don't think it's a good way patching this into wine as this is separate. I think a better way would be something like with mono or wine gecko.

pchome:
@Thaodan 
> Can't we just have something like build arch= wine or mingw and so on?

my winebuild vs master first patch : https://github.com/doitsujin/dxvk/commit/e7102ccbe465810bf6b0edfef9552f5bdd63f1f6 (https://github.com/doitsujin/dxvk/compare/master...pchome:winebuild)
There was `build-win64-wine.txt` cross-file to build with `winegcc` in my first version. Generally this patch shows what should be changed for `winegcc` support.

> About the winegcc is this a packaging problem or a upstream problem?

Packaging problem, winegcc requires some system paths to be build
```bash
winegcc_EXTRADEFS = \
    -DINCLUDEDIR="\"${includedir}\"" \
    -DDLLDIR="\"${dlldir}\"" \
    -DLIBDIR="\"${libdir}\"" \
    -DCC="\"$(CC)\"" \
    -DCPP="\"$(CPPBIN)\"" \
    -DCXX="\"$(CXX)\"" \
    -DLD="\"$(LD)\"" \
    -DPRELINK="\"$(PRELINK)\""
```
and than uses those LIBDIR and DLLDIR to build wine libraries. While WINE builds itself with winegcc compiled for x86 and x86_64 accordingly, distros providing only one of them named winegcc.
So we need something like `x86_64-winegcc` and `i686-winegcc`. 

Also there is some options to skip passing WINE libraries to linker in winebuild/winegcc, but finding correct path to wine is another quest in different distros (e.g. in gentoo I have several WINE versions at the same time).

> I don't think it's a good way patching this into wine as this is separate.

I don't propose to do this for DXVK, but talking about how I can play with this (theoretically).

pchome:
#60

Thaodan:
>and than uses those LIBDIR and DLLDIR to build 
wine libraries. While WINE builds itself with winegcc 
compiled for x86 and x86_64 accordingly, distros 
providing only one of them named winegcc.
>So we need something like `x86_64-winegcc` and 
`i686-winegcc`. 
So this an upstream bug or should we just use m32 
or m64 like with gcc?

pchome:
@Thaodan 
`winegcc` calls `winebuild` with parameters, which calls linker or compiler with parameters, one of them is `path/to/wine/lib64/libwinectr0.a` despite you -m32 attr, for example.

Ok for x86_64 but causes linker error for x86.

mdrost:
Did you report bug in wine about missing struct? Some time ago I report bug about missing constant that is needed by dxvk and it was fixed https://bugs.winehq.org/show_bug.cgi?id=44501 Same bugs should be also reported in mingw.

PR #258, opened 2018-04-08
[wine-utils] allow 32 bit setup_dxvk.sh to work alongside 64 bit

GloriousEggroll:
This allows the 32 bit dxvk to be installed alongside 64 bit in syswow64 in 64 bit prefixes, in case of wine cross compiled x86_64, thus allowing a 64 bit prefix to run both 64 and 32 bit dxvk


GloriousEggroll:
closed per discord convo

PR #264, opened 2018-04-10
Added instructions for installing from the easy build script.

Managor:
I think this would lessen the amount of questions in the discord.


ssorgatem:
Maybe it'd be better not to create the archive (which has to be unpacked later), just the folder containing the build.

That is done with the `--no-package` argument to the package-release.sh script.

PR #268, opened 2018-04-11
[build] Experimental winegcc support

pchome:
* Allow DXVK to be built as WINE builtin dlls (#60)
* .spec and .def files not provided by WINE are placed
  into ./linux/ directory in src root
* wine-staging dll-redirects support added as an option
* An attempt to fix wide-strings problem, but need to be reviewed
* Unstable, can either run or not
* Performance comparable to native .dlls, so I'm asking people to
  not use this, if you don't know why you need this

### DIFF
--- a/src/d3d11/meson.build
+++ b/src/d3d11/meson.build
@@ -25,15 +25,25 @@ d3d11_src = [
   'd3d11_view_srv.cpp',
   'd3d11_view_uav.cpp',
 ]
-
+if not dxvk_wine_builtin
 d3d11_dll = shared_library('d3d11', d3d11_src,
   name_prefix         : '',
   link_with           : [ util_lib ],
-  dependencies        : [ dxvk_dep, dxgi_dep, dxbc_dep ],
+  dependencies        : [ declare_dependency(link_args: ['-ldxgi']), dxbc_dep, dxvk_dep ],
   include_directories : dxvk_include_path,
   install             : true,
   vs_module_defs      : 'd3d11.def',
   override_options    : ['cpp_std='+dxvk_cpp_std])
+else
+d3d11_dll = shared_library('d3d11'+dll_suffix, d3d11_src,
+  name_prefix         : '',
+  link_with           : [ util_lib ],
+  dependencies        : [ declare_dependency(link_args: ['-ldxgi'+dll_redirect_suffix]), dxbc_dep, dxvk_dep ],
+  include_directories : dxvk_include_path,
+  install             : true,
+  objects             : wine_spec_path + 'd3d11'+dll_redirect_suffix+'.spec',
+  override_options    : ['cpp_std='+dxvk_cpp_std])
+endif
 
 d3d11_dep = declare_dependency(
   link_with           : [ d3d11_dll ],
--- a/src/d3d11/meson.build
+++ b/src/d3d11/meson.build
@@ -29,19 +29,19 @@ if not dxvk_wine_builtin
 d3d11_dll = shared_library('d3d11', d3d11_src,
   name_prefix         : '',
   link_with           : [ util_lib ],
-  dependencies        : [ declare_dependency(link_args: ['-ldxgi']), dxbc_dep, dxvk_dep ],
+  dependencies        : [ dxvk_dep, dxgi_dep, dxbc_dep ],
   include_directories : dxvk_include_path,
   install             : true,
   vs_module_defs      : 'd3d11.def',
   override_options    : ['cpp_std='+dxvk_cpp_std])
 else
-d3d11_dll = shared_library('d3d11'+dll_suffix, d3d11_src,
+d3d11_dll = shared_library('d3d11.dll', d3d11_src,
   name_prefix         : '',
   link_with           : [ util_lib ],
-  dependencies        : [ declare_dependency(link_args: ['-ldxgi'+dll_redirect_suffix]), dxbc_dep, dxvk_dep ],
+  dependencies        : [ declare_dependency(link_args: ['-ldxgi']), dxbc_dep, dxvk_dep ],
   include_directories : dxvk_include_path,
   install             : true,
-  objects             : wine_spec_path + 'd3d11'+dll_redirect_suffix+'.spec',
+  objects             : wine_spec_path + 'd3d11.spec',
   override_options    : ['cpp_std='+dxvk_cpp_std])
 endif
 
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -16,10 +16,9 @@ namespace dxvk::env {
     }
     
     result.resize(len);
-    return str::fromws(result);
+    return str::fromws(result.c_str());
   }
   
-#ifndef __WINE__  
   std::string getExeName() {
     std::wstring exePath;
     exePath.resize(MAX_PATH + 1);
@@ -27,27 +26,13 @@ namespace dxvk::env {
     DWORD len = ::GetModuleFileNameW(NULL, &exePath.at(0), MAX_PATH);
     exePath.resize(len);
     
-    std::string fullPath = str::fromws(exePath);
+    std::string fullPath = str::fromws(exePath.c_str());
     auto n = fullPath.find_last_of('\\');
     
     return (n != std::string::npos)
       ? fullPath.substr(n + 1)
       : fullPath;
   }
-#else
-  std::string getExeName() {
-    char exePath[MAX_PATH + 1];
-
-    DWORD len = ::GetModuleFileNameA(NULL, exePath, MAX_PATH);
-
-    std::string fullPath(exePath);
-    auto n = fullPath.find_last_of('\\');
-
-    return (n != std::string::npos)
-      ? fullPath.substr(n + 1)
-      : fullPath;
-  }
-#endif  
   
   std::string getTempDirectory() {
     WCHAR windowsTempDir[MAX_PATH] = {0};
@@ -61,7 +46,7 @@ namespace dxvk::env {
         return std::string();
     }
     
-    return str::fromws(dxvkTempDir);
+    return str::fromws(dxvkTempDir.c_str());
   }
   
 }
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -19,7 +19,7 @@ namespace dxvk::env {
     return str::fromws(result);
   }
   
-  
+#ifndef __WINE__  
   std::string getExeName() {
     std::wstring exePath;
     exePath.resize(MAX_PATH + 1);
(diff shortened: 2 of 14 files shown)


oscarbg:
just a comment: I don't know why I need it what I want to know your motivation anyway :-)

src/d3d11/meson.build
```diff
@@ -25,15 +25,25 @@ d3d11_src = [
   'd3d11_view_srv.cpp',
   'd3d11_view_uav.cpp',
 ]
-
+if not dxvk_wine_builtin
 d3d11_dll = shared_library('d3d11', d3d11_src,
   name_prefix         : '',
   link_with           : [ util_lib ],
-  dependencies        : [ dxvk_dep, dxgi_dep, dxbc_dep ],
+  dependencies        : [ declare_dependency(link_args: ['-ldxgi']), dxbc_dep, dxvk_dep ],
```
doitsujin: Why is this change needed, i.e. just using `dxgi_dep` not enough?

src/util/util_env.cpp
```diff
@@ -34,7 +34,20 @@ namespace dxvk::env {
...
+    DWORD len = ::GetModuleFileNameA(NULL, exePath, MAX_PATH);
+
+    std::string fullPath(exePath);
+    auto n = fullPath.find_last_of('\\');
+
+    return (n != std::string::npos)
+      ? fullPath.substr(n + 1)
+      : fullPath;
+  }
+#endif
```
doitsujin: Does the old function not work with wine? If that's the case, it might even be better to use this one for everything. Having two almost identical functions around is poor style.

src/util/util_string.h
```diff
@@ -21,9 +25,18 @@ namespace dxvk::str {
...
-  
+#ifdef __WINE__
+  inline std::string fromws(const std::wstring& ws) {
+    char *ret = nullptr;
+    if (ws.size() > 0) {
+        ret = new char[ws.size() + 1];
+        WideCharToMultiByte( CP_UTF8, 0, ws.c_str(), -1, ret, ws.size() + 1, NULL, NULL );
+    }
+    return ret ? ret : "";
+  }
```
doitsujin: Why is this needed? Also, you're leaking memory.

src/util/util_string.h
```diff
@@ -5,6 +5,10 @@
 #include <string>
 #include <sstream>
 
+#ifdef __WINE__
+#include "windows.h"
```
doitsujin: If anything needs windows.h, please don't include `windows.h` directly but use the `com_include.h` header which sets up a few defines beforehand.

pchome:
@oscarbg 
> just a comment: I don't know why I need it what I want to know your motivation anyway :-)

tl;dr: "why not" and "because I can"
I have some problems with setting MinGW build environment earlier, so decided to try this. Also I used this for a long time with vulkan implementation from wine-staging-2.21, before \*Template\* extensions support was introduced in DXVK. Now with winevulkan this builds became more unstable, but I hope it will be fixed soon.
<hr>

@doitsujin 
> I'm afraid I don't understand what the "wide string problem" is and why those new functions are needed.

I think it's multi-byte char strings manipulation, e.g. taking character position sometimes produces weird results. Also std::string is valid container for unicode (they say) so better to avoid any manipulations with strings (like substr) in cross-platform code. That's how I understand this.
`std::codecvt_utf8` conversion don't work for me, and produces exceptions like this:
```c++
dxgi-factory.exe.so: malloc.c:2399: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.
```

I don't used C or C++ for a long time and completely "forgot" everything, so this changes are rather hacks and I hope more eyes will help to fix "bad parts".

Also I hope you finally implement configuration file support, than all this conversion/environment code will gone. Especially I'm worried my hacks running thousand times here : https://github.com/doitsujin/dxvk/blob/master/src/d3d11/d3d11_shader.cpp#L63
<hr>

> I also don't really understand why explicit link args have to be used in the meson build files rather than generic dependency objects.

`winegcc` have some magic, taking filenames and extensions it produces different command arguments for `winebuild` and than `winebuild` does similar things for `gcc` and `ld`. Passing `dxgi.dll.so` or `-ldxgi` give different results. First variant adds `dxgi.dll.so` to direct dependency for `d3d11.dll.so`. I tested second variant on MinGW build and mentioned no problems so far.
<hr>

> Second, I think it's better to have the dll-redirect stuff removed - it makes the meson files hard to maintain and having duplicate .spec files isn't great either, and the gain is questionable.

dll-redirects currently have no effect on other build types, but for winegcc builds it helps to not mess with original WINE libs. Setup similar to dll-overrides: place libraries to path set in WINEDLLPATH and switch redirect settings for an .exe in regedit.

src/d3d11/meson.build
pchome: I don't tested dll-redirects with MinGW build, but if MinGW links libraries similar to winegcc than it should help (will check later). I have noticed no difference with generic MinGW build.

src/util/util_env.cpp
pchome: Those functions are redefined because they don't work with winebuild. While other do work (`getEnvVar`, function `getTempDirectory` - don't but currently unused),  . Have no idea why.

This expected to not work with unicode filenames, your example produces `?????_dxgi.log` for renamed `dxgi-factory.exe`. But it's not big problem, as for me.

src/util/util_string.h
pchome: Current `std::codecvt_utf8` conversion in `fromws` don't work with wine, maybe it should be changed to use different encoding type (UTF-16 ?), but with lack of knowledge I'm trying to do my best to make this conversion work with winegcc generated libraries.

Also std::codecvt_* are marked as deprecated in C++17, maybe something like shown in [example](http://en.cppreference.com/w/cpp/locale/codecvt) should be used to work for both MinGW and winebuild.

doitsujin:
> dll-redirects currently have no effect on other build types, but for winegcc builds it helps to not mess with original WINE libs

I still don't like it because it makes the whole thing much more messy than it needs to be, and it *only* works with a wine-staging patch that may never make it into mainline. Surely the whole winebuild thing works *without* it since you made it optional as well.

Users can still set `d3d11.dll` and `dxgi.dll` to `builtin` in order to disable DXVK for a certain application.

> Also std::string is valid container for unicode

Sure it is, but Windows APIs only work with UCS-2 so we have to do this conversion *somewhere*.

src/util/util_env.cpp
doitsujin: What this basically means is that file paths containing anything non-Latin1 won't work. This is unacceptable. I might take a look at this later but this needs to be resolved.

Guy1524:
Just tested your commits on Ubuntu 16.04, and it worked as expected other than 

-`meson configure` failed but was not necessary

-the arguments for glslangValidator were incorrectly generated, leading to an invalid variable name (@basename) in some headers.

pchome:
@Guy1524 
#270 ?

Guy1524:
@pchome yeah, that's and issue on my end then, nevermind.

pchome:
6995e80 will produce `ファイル名_dxgi.log` for `ファイル名.exe`

ssorgatem:
So how is the status of this?

I'm mainly interested in it because it drops the requirements for a cross-compiling environment for those not planning to use DXVK on Windows...

pchome:
@ssorgatem 
Hmm, status: it's compiles :)

I have MinGW GCC 6.4 now to test everything is ok with regular builds after changes applied, and I basically use MinGW build for my WINE prefixes. 
This patches needs more eyes to compete with mingw as full replacement.

Guy1524:
Hey @pchome , I just tried your patches with latest master.  Some of your string work has caused me to crash but returning a fake value from the getExeName function fixed that issue.  After that, your build worked correctly with the cascaded shadow maps sample.  

My problem now is that GTA V doesn't open with your build.  Nothing happens for a few seconds after launching the application, then the application crashes from within.  To load the dll I am using dll redirects with the output dll.so s inside /opt/wine-staging/lib64/wine and $prefix/drive_c/windows/system32.  The correct redirected dlls show up on the crash report, and GTA V works fine with the mingw build.  Oh, and nothing is loggedit to the log files, although they are created.

mdrost:
Hey @pchome, we the people invites you to dxvk discord channel https://discord.gg/b29gDTn as some people want to talk with you in real time about your pull requests and other problems :P

sdack:
@oscarbg
```text
just a comment: I don't know why I need it what I want to know your motivation anyway :-)
```
All DLLs in WINE are build with `winegcc`. Therefore it's essential to have `winegcc` support for DXVK, because it means it opens up the possibility for DXVK to become a standard component of WINE. I believe most WINE users will agree having DXVK as a builtin component to WINE would be awesome. It's kind of a no-brainer really.

Windows user of course can benefit from DXVK, too, however they do have Microsoft's native DX11 available of course and shouldn't even need DXVK. That DXVK works under Windows proves its working independent of WINE and allows for a direct comparison of DXVK to Microsoft's DX11. However, in the long run will DXVK most likely establish itself more with the WINE/Linux community as we can already see, than with Windows users.

This just to add to pchome's "why not" and "because I can" argument.

saboya:
DXVK is not going to be part of Wine anytime soon since Wine does not accept C++ code in its codebase.

ssorgatem:
Well, there's wine-staging for that...

sdack:
```text
DXVK is not going to be part of Wine anytime soon since Wine does not accept C++ code in its codebase.
```
Doesn't matter. Code can change, so can policies, and there will always be a bit of drama. What people will agree on is that DXVK is awesome and they will find a way. Support for winegcc is getting it one step closer.

sl1pkn07:
seems need rebase

pchome:
I have spent a bit of time for debugging string conversion and now I see what happens.
* WINE using `-fshort-wchar` compiler option to use 2 byte `wchar_t` (unsigned short), but system (linux) `wchar_t` is 4 byte.
This causes `std::wstring_convert<std::codecvt_utf8<wchar_t>>().to_bytes(ws);` conversion problem and undefined behaviour for other `std::wstring` manipulation since system libraries (libstdc++/libgcc) used.
* `winegcc` do not support static linking with `-static`
  >  Static  linking  is  not currently supported against Wine DLLs. As a result, the -static, --static, and -Wl,-static options will generate an error.
so building/using statics like MinGW does is not an option.
* `-fno-short-wchar` compiler option partially solves this problem, but additional conversion from 32 bit `wchar_t` (string literals) to `WCHAR` (unsigned short) required [here](https://github.com/doitsujin/dxvk/blob/9ff17b03f24c7a5d0a1825214f1697e5f6e48f1e/src/util/util_env.cpp#L7).

**If anyone want to help** and can properly implement such conversions - **URW**.

Here is somehow working example from my testing:
```c++
  std::string getEnvVar(const wchar_t* name) {
    std::string mb_name = std::wstring_convert<std::codecvt_utf8<wchar_t>>{}.to_bytes(name);

    DWORD slen = MultiByteToWideChar(CP_UNIXCP, 0, mb_name.c_str(), -1, nullptr, 0);
    WCHAR ret[slen];
    MultiByteToWideChar(CP_UNIXCP, 0, mb_name.c_str(), -1, ret, slen);

    DWORD len = ::GetEnvironmentVariableW(ret, nullptr, 0);
...
```

_FYI: #392, using e.g. `wine` (/usr/lib/wine-staging-3.9/bin/wine64) as "executable" parameter for `gdb` compatible debugger lets `gdb` automatically determine system type an load symbols from `*.so` shared libraries._

doitsujin:
Superseded by #501, which I'm probably going to merge.

PR #271, opened 2018-04-12
[dxgi] Use GetMonitorInfoW function explicitly

pchome:
No `std::mbstowcs` string conversion needed than.
W/o this change I have different output when running `dxgi-factory.exe` compiled with MinGW:
* `Output 0:`
  ```
  \\.\DISPLAY1 (default)
  \\.\DISPLAY1 (DXVK MinGW)
  \            (DXVK winebuild)
  ```

With this patch all three variants are identical (`\\.\DISPLAY1`)

p.s. same problem in dxgi_adapter.cpp, but `deviceProp.deviceName` is vulkan structure parameter (char deviceName[VK_MAX_PHYSICAL_DEVICE_NAME_SIZE]).

### DIFF
--- a/src/dxgi/dxgi_output.cpp
+++ b/src/dxgi/dxgi_output.cpp
@@ -150,16 +150,15 @@ namespace dxvk {
     if (pDesc == nullptr)
       return DXGI_ERROR_INVALID_CALL;
     
-    ::MONITORINFOEX monInfo;
+    ::MONITORINFOEXW monInfo;
     monInfo.cbSize = sizeof(monInfo);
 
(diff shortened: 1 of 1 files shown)


mdrost:
IMHO similarly code like that https://github.com/doitsujin/dxvk/blob/master/src/util/util_env.cpp should use ansi functions version to avoid all that conversions.

doitsujin:
Didn't even know that these functions had Unicode equivalents. Praise Microsoft documentation!

@mdrost Thing is, ANSI functions really only support some 8-bit character set, not UTF-8 like any sane environment does. Hence the W functions. It *might* be an option for the environment variable function, but for stuff like file names that can realistically contain non-ANSI characters it's definitely a no-go.

PR #279, opened 2018-04-14
[dxvk] Added memory heap listing to adapter info

ShadowsFriend:
In the progress of fiddling around because of my memory allocation issue, I have decided to output the available memory heaps during the logging of the the adapter info. Although this is just a very small addition, I thought it might be useful to someone sometime.


doitsujin:
Nested is fine. Especially since now you don't actually need to print the heap index - for heap 1, the heap indices of all memory types on that heap will obviously be 1, so compacting the output to something like `info:      Memory Type[6]: Flags = 0x6` should be possible without losing any information.

ShadowsFriend:
Well, I was not sure about removing the index which is why I left it although you are obviously right that it is completely redundant. Regardless, now it is removed and the output looks like [1]. I also replaced the newlines with calls to Logger:: such that the logs look more uniform and subsequent changes to the log string do not influence the indentation.

[1] [dxvk_memory_output.log](https://github.com/doitsujin/dxvk/files/1911475/dxvk_memory_output.log)

PR #290, opened 2018-04-16
[dxvk] Add basic error checking to the heap allocator

ShadowsFriend:
Another small by-product resulting from my issue tracking. I decided to make the output debug level instead of error level, although my understanding is that an allocation failure at that stage means we have a problem most of the time.


doitsujin:
I added a bit of error reporting lately and while this was probably helpful for debugging, we now know the only reason why mapping memory can fail, there's no need for this one anymore.

PR #295, opened 2018-04-17
[d3d11] nullptr check on Begin and End

ZeroFault:
Checks for nullptr on pAsync. It is needed for replaying a Star Citizen apitrace which tries to create a ID3D11Counter object and using it with the Begin and End functions.


mdrost:
So shouldn't be fixed ID3D11Counter object creation?

mdrost:
Those checks will stay forever or will be removed when ID3D11Counter be implemented?

PR #296, opened 2018-04-17
[dxbc] Moved all operators in dxbc_names to dxvk namespace

vesim987:


doitsujin:
Might be a good idea to move the others (in dxvk, d3d11, dxgi) into `namespace dxvk` as well in the future.

PR #301, opened 2018-04-18
[build] Fix build with MinGW GCC 6.4

pchome:
GCC 7.1 is minimal version for `std::clamp`:
`An algorithm to "clamp" a value between a pair of boundary values | P0025R0 | 7.1`
https://gcc.gnu.org/onlinedocs/gcc-7.3.0/libstdc++/manual/manual/status.html#status.iso.2017

### DIFF
--- a/src/dxvk/hud/dxvk_hud_fps.cpp
+++ b/src/dxvk/hud/dxvk_hud_fps.cpp
@@ -96,8 +96,8 @@ namespace dxvk::hud {
       minMs = std::min(minMs, uint32_t(us / 100.0f));
       maxMs = std::max(maxMs, uint32_t(us / 100.0f));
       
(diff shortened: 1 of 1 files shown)


doitsujin:
Not sure where the non-std `clamp` is supposed to come from, it seems to break the build. 8125d53e58f0c08394e2e730bc88cbc57234a147 replaces `std::clamp` by a min/max pair.

One more reason to drop GCC 6.x support soon...

pchome:
> Not sure where the non-std clamp is supposed to come from

https://github.com/doitsujin/dxvk/blob/master/src/util/util_math.h#L7
```c++
template<typename T>
  T clamp(T n, T lo, T hi) {
    if (n < lo) return lo;
    if (n > hi) return hi;
    return n;
  }
```

PR #305, opened 2018-04-19
[d3d11] add some resource validation for copying resources

ZeroFault:
Some games, such as Saints Row IV and Star Citizen, use invalid parameters for resource copying. These invalid parameters can cause Device Lost errors and other undefined problems.


doitsujin:
Not sure if this is actually the correct thing to do.

Any chance these games are trying to copy from a compressed to an uncompressed image or vice versa? Because that **is** legal, I don't think this case is being handled properly yet.

ZeroFault:
In this case of saints row IV and Starcitizen they don't copy between uncompressed and compressed. This code should still allow uncompressed and compressed copying, just without validation. I can add the validation for it later if you want me to.

PR #307, opened 2018-04-19
Make hashes use correct types and fix narrowing warnings in spirv module.

Joshua-Ashton:
In d3d11_state.cpp there were several implicit casts of floats to size_t which were then being hashed. I do not believe this is intentional. I have made them use the hash functions of their respective types. There was also one the other way around.

In spirv_module.cpp several enums were being narrowly casted implicitly as their type width was 64 bits wide and they were going to a uint32. This has been resolved with a static cast before their usage in the std::array.


Joshua-Ashton:
Alright, will remember in future.
I prefer those too but I noticed static_casts elsewhere in code.

doitsujin:
Yeah, that's stuff to clean up at some point because I wasn't aware that these constructor-like casts were a thing. for these types.

mdrost:
>Alright, will remember in future.  

Just rebase you commits before they got merged.

Issue #309, opened 2018-04-20
Wine Assumptions Result in Build Failure


Enverex:
Spent the last hour or so trying to track down why this wouldn't build. It would make the 32bit libraries fine, but not the 64bit libraries.

Turns out, the build process appears to just use the user's default Wine prefix (so ~/.wine). The problem with this is that it could be configured in any way. In my case, my default prefix was a 32bit prefix which meant it couldn't run the 64bit test binary (as a 32bit Wine prefix cannot run 64bit binaries, but vice versa can). Thus it resulted in this error...

	[02:28:58] arcade@arcade /var/tmp/makepkg/dxvk-git/src/dxvk-git $ ./package-release.sh master ../built --no-package
	The Meson build system
	Version: 0.45.1
	Source dir: /var/tmp/makepkg/dxvk-git/src/dxvk-git
	Build dir: /var/tmp/makepkg/dxvk-git/src/built/dxvk-master/build.64
	Build type: cross build
	Project name: dxvk
	wine: Bad EXE format for Z:\var\tmp\makepkg\dxvk-git\src\built\dxvk-master\build.64\meson-private\sanitycheckc_cross.exe.

	meson.build:1:0: ERROR: Executables created by c compiler /usr/bin/x86_64-w64-mingw32-gcc are not runnable.

	A full log can be found at /var/tmp/makepkg/dxvk-git/src/built/dxvk-master/build.64/meson-logs/meson-log.txt
	ninja: error: loading 'build.ninja': No such file or directory
	cp: cannot stat '/var/tmp/makepkg/dxvk-git/src/built/dxvk-master/install.64/bin/d3d11.dll': No such file or directory
	cp: cannot stat '/var/tmp/makepkg/dxvk-git/src/built/dxvk-master/install.64/bin/dxgi.dll': No such file or directory
	cp: cannot stat '/var/tmp/makepkg/dxvk-git/src/built/dxvk-master/install.64/bin/setup_dxvk.sh': No such file or directory
	rm: cannot remove '/var/tmp/makepkg/dxvk-git/src/built/dxvk-master/install.64': No such file or directory

Forcing a new prefix (which will be 64bit by default) allowed the process to complete:

	[03:54:23] arcade@arcade /var/tmp/makepkg/dxvk-git/src/dxvk-git $ WINEPREFIX=~/wintest ./package-release.sh master ../built --no-package
	The Meson build system
	Version: 0.45.1
[...]

PR #311, opened 2018-04-20
Don't use system Wine prefix

Enverex:
Create a Wine prefix specifically for the purpose of building the libraries. Fixes issue https://github.com/doitsujin/dxvk/issues/309.


doitsujin:
As mentioned in #309, already fixed by f52cc3859b4d77fe3a1477ca4c512e134db26c33.

PR #322, opened 2018-04-25
[clang-tidy] performance-for-range-copy fix

pchome:
> Finds C++11 for ranges where the loop variable is copied in each iteration but it would suffice to obtain it by const reference.

> The check is only applied to loop variables of types that are expensive to copy which means they are not trivially copyable or have a non-trivial copy constructor or destructor.

https://clang.llvm.org/extra/clang-tidy/checks/performance-for-range-copy.html

### DIFF
--- a/src/d3d11/d3d11_cmdlist.cpp
+++ b/src/d3d11/d3d11_cmdlist.cpp
@@ -52,7 +52,7 @@ namespace dxvk {
   void D3D11CommandList::EmitToCommandList(ID3D11CommandList* pCommandList) {
     auto cmdList = static_cast<D3D11CommandList*>(pCommandList);
     
-    for (auto chunk : m_chunks)
+    for (const auto& chunk : m_chunks)
       cmdList->m_chunks.push_back(chunk);
     
     cmdList->m_drawCount += m_drawCount;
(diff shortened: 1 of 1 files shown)


doitsujin:
It would be nice if you could announce your plans first, and more importantly, keep code formatting consistent.

If you are planning on messing around with the entire code base with `clang-tidy` just for the sake of it, I'm not going to merge it unless there are some actual benefits. Looking at the other PRs I can see the point of fixing some minor stuff even though none of this is performance-critical, but I need to know the extent of what you are actually doing.

pchome:
I'm just playing with code and clang-tidy. Feel free to reject this PRs if you don't like them.

Those clang-tidy-* changes are auto-generated by `clang-tidy -checks='...' -fix` command.

doitsujin:
Okay, that makes sense. I don't like the "auto-generated" part of it though since some issues may require more thought than just that.

pchome:
As for "code formatting consistent" I think it can be configured via `-format-style=` option or `clang-format` tool. But it's rather for regular usage.

Thaodan:
Talking about coding style what about using something like this?
http://editorconfig.org/

pchome:
@Thaodan 
AFAIK editorconfig is for IDE/editor. I'm using VSCode and it detects code formatting for existent files automatically, and the only thing where editorconfig can be helpful is creating new files or as a guide.
Tools like `clang-*` have nothing to do with editorconfig but uses `.clang-format` file in source directories.

PR #324, opened 2018-04-25
[clang-tidy] performance-move-const-arg fix

pchome:
> The check warns
> *    if std::move() is called with a constant argument,
> *    if std::move() is called with an argument of a trivially-copyable type,
> *    if the result of std::move() is passed as a const reference argument.
>
> In all three cases, the check will suggest a fix that removes the std::move().

https://clang.llvm.org/extra/clang-tidy/checks/performance-move-const-arg.html

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -2577,7 +2577,7 @@ namespace dxvk {
     }
     
     // Create and bind the framebuffer object to the context
-    EmitCs([cAttachments = std::move(attachments)] (DxvkContext* ctx) {
+    EmitCs([cAttachments = attachments] (DxvkContext* ctx) {
       ctx->bindRenderTargets(cAttachments);
     });
   }
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -1025,7 +1025,7 @@ namespace dxvk {
       return S_FALSE;
     
     *ppVertexShader = ref(new D3D11VertexShader(
-      this, std::move(module)));
+      this, module));
     return S_OK;
   }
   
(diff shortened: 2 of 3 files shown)


src/d3d11/d3d11_context.cpp
```diff
@@ -2577,7 +2577,7 @@ namespace dxvk {
     }
     
     // Create and bind the framebuffer object to the context
-    EmitCs([cAttachments = std::move(attachments)] (DxvkContext* ctx) {
+    EmitCs([cAttachments = attachments] (DxvkContext* ctx) {
```
doitsujin: Why is a copy better than a move in this case?

pchome:
This change auto-generated by `clang-tidy`, for sure it can be wrong, feel free to reject PR.

doitsujin:
The other cases are certainly valid but I'm really not sure about this one. Base on my understanding, move-capture should work.

pchome:
Warning was:
```c++
d3d11_context.cpp:2580:28: warning: passing result of std::move() as a const reference argument; no move will actually happen [performance-move-const-arg]
    EmitCs([cAttachments = std::move(attachments)] (DxvkContext* ctx) {
                           ^~~~~~~~~~~~~~~~~~~~~~~~
```
But this warnings are intended for review, so it can be false-positive.

mdrost:
I would also point that expression `[cAttachments = attachments]` is nonsensical, if you want capture variable by copy just type `[attachments]`. In case of move capture there is no need for new name, just type `[attachments = std::move(attachments)]`.

doitsujin:
@mdrost as for the declared destructor, fair point. I guess the proper fix for that would be declaring default move and copy constructors then.

As for the names, I know how lambda capture works. Assiging a new name there might be a relic (I'm doing this a lot because most lambdas do *not* just capture local variables), but ultimately it's a non-issue.

mdrost:
So it is merged without fixing mentioned eg missing move constructor?

doitsujin:
yeah, in the end i don't really care because the whole render target binding stuff will see some refactoring soon anyway.

PR #328, opened 2018-04-27
[dxvk] Fix native build on *nix x86_64 systems

pchome:
> /usr/include/c++/v1/algorithm:2633:1:
> note: candidate template ignored: deduced conflicting types for parameter '_Tp' ('unsigned long' vs. 'unsigned long long')

winegcc, clang-tidy and other native build types/tools are affected.

http://en.cppreference.com/w/cpp/language/types#Data_models
> 64 bit systems:
> * LLP64 or 4/4/8 (int and long are 32-bit, pointer is 64-bit) 
>    * Win64 API 
> * LP64 or 4/8/8 (int is 32-bit, long and pointer are 64-bit) 
>    * Unix and Unix-like systems (Linux, Mac OS X) 

http://en.cppreference.com/w/cpp/types/integer#Function_macros_for_minimum-width_integer_constants
Macro `UINT64_C(1)` from `stdint.h` should literally interpret `1` to `1UL` or `1ULL`
```c
# if __WORDSIZE == 64
#  define UINT64_C(c)	c ## UL
# else
#  define UINT64_C(c)	c ## ULL
# endif
```

### DIFF
--- a/src/dxvk/hud/dxvk_hud_stats.cpp
+++ b/src/dxvk/hud/dxvk_hud_stats.cpp
@@ -47,7 +47,7 @@ namespace dxvk::hud {
     const Rc<DxvkContext>&  context,
           HudRenderer&      renderer,
           HudPos            position) {
-    const uint64_t frameCount = std::max(m_diffCounters.getCtr(DxvkStatCounter::QueuePresentCount), 1ull);
+    const uint64_t frameCount = std::max(m_diffCounters.getCtr(DxvkStatCounter::QueuePresentCount), UINT64_C(1));
     
     const uint64_t gpCalls = m_diffCounters.getCtr(DxvkStatCounter::CmdDrawCalls)       / frameCount;
     const uint64_t cpCalls = m_diffCounters.getCtr(DxvkStatCounter::CmdDispatchCalls)   / frameCount;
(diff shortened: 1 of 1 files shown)


mdrost:
Wouldn't `std::max<uint64_t>(…, 1);` work?

doitsujin:
I'd prefer the explicit template argument as well if `1ull` messes up the build for some reason.

pchome:
`std::max<uint64_t>(…, 1);` looks ok for both MinGW and winegcc builds.

pchome:
There is also some other places in code where `*ll` suffix used, which may lead to problems:
`$ grep -R -E "[0-9]u?ll?" ./*`
```c++
./dxvk/dxvk_cmdlist.cpp:        1'000'000'000ull);
./dxvk/hud/dxvk_hud_fps.cpp:      const int64_t fps = (10'000'000ll * m_frameCount) / elapsedFps.count();
./dxvk/dxvk_memory.h:    std::atomic<VkDeviceSize>        m_memoryAllocated = { 0ull };
./dxvk/dxvk_memory.h:    std::atomic<VkDeviceSize>        m_memoryUsed      = { 0ull };
./util/com/com_object.h:    std::atomic<ULONG> m_refCount = { 0ul };
```

pchome:
WINE windef.h
```c
#ifdef _MSC_VER
typedef long                                   *LPLONG;
typedef unsigned long   DWORD,      *PDWORD,   *LPDWORD;
typedef unsigned long   ULONG,      *PULONG;
#else
typedef int                                    *LPLONG;
typedef unsigned int    DWORD,      *PDWORD,   *LPDWORD;
typedef unsigned int    ULONG,      *PULONG;
#endif
```
long is 64 bit for *nix systems
uint64_t is unsigned long

winegcc itself uses definitions `-D__int64=long` and `-D__int64=long long`  for 64 and 32 bit builds

... and so on.
It can potentially lead to unpredictable results.

_As I understand all this cross-platform situation._

mdrost:
`… = { 0ul };` really c++ couldn't compile it without suffix?

PR #329, opened 2018-04-27
Allow DXUP to work (D3D10->D3D11)

Joshua-Ashton:


pchome:
I can launch **SleepingDogs: DefinitiveEdition** with this patch.
There was white screen and only output in ` sdhdship_dxgi.log`:
```
warn:  DxgiFactory::QueryInterface: Unknown interface query
warn:  50c83a1c-e072-4c48-87b0-3630fa36a6d0
err:   DXGI: CheckInterfaceSupport: No D3D10 support
```

Not sure is it requires dxup, but it successfully loading to game options screen. Also I have no chance to start new game w/o crash neither with regular wine not with dxvk, it's some kind of game bug reported many times by steam community.

doitsujin:
`50c83a1c-e072-4c48-87b0-3630fa36a6d0` is `IDXGIFactory2`, so apparently we have to implement a proper check for DXGI interfaces here.

Not sure about the D3D10 ones because DXVK doesn't support D3D10, and that stuff should be handled in DXUP instead.

ssorgatem:
Maybe DXUP should be part of DXVK?

oscarbg:
Yep make part of DXVK.. the question is: can it run Crysis in DX10 mode? some photos  showing it please..

Joshua-Ashton:
@oscarbg Crysis doesn't work in Win 10 anymore so I need to find an opportunity to test it perhaps under WINE.

oscarbg:
@Joshua-Ashton Are you sure? I have no problems running Crysis (demo) on Win10 either on dx9 or DX10.. are you sure is not an issue of your own system? Can you test demo at least it should work..

Joshua-Ashton:
I got Crysis warhead working.

baka0815:
I'm sure this patch is needed to get DXUP running, but what would be the behaviour if DXUP is not installed? Is there a possibility to check if DXUP is installed and return ``S_OK`` in that case, ``DXGI_ERROR_UNSUPPORTED`` otherwise?

Thaodan:
How is dxup called?
Maybe dxvk should try to load it if enabled.

Joshua-Ashton:
@baka0815 The behaviour would be, to my knowledge, that it knows that D3D10 is supported by the device. I wouldn't have thought this would cause issues with games but it'd be great for people to test this.

@Thaodan DXUP isn't "called." DXUP calls DXVK.

PR #333, opened 2018-04-28
[d3d11] implement stateblock normalization

ZeroFault:
Implement d3d11 stateblock normalization to match what d3d11 does in state object creation.

### DIFF
--- a/src/d3d11/d3d11_blend.cpp
+++ b/src/d3d11/d3d11_blend.cpp
@@ -149,110 +149,67 @@ namespace dxvk {
   
   HRESULT D3D11BlendState::NormalizeDesc(D3D11_BLEND_DESC1* pDesc) {
     const D3D11_BLEND_DESC1 defaultDesc = DefaultDesc();
-    if (pDesc->AlphaToCoverageEnable != 0) {
-      pDesc->AlphaToCoverageEnable = 1;
+    
+    if (pDesc->AlphaToCoverageEnable) {
+      pDesc->AlphaToCoverageEnable = TRUE;
     }
 
-    if (pDesc->IndependentBlendEnable != 0) {
-      pDesc->IndependentBlendEnable = 1;
+    if (pDesc->IndependentBlendEnable) {
+      pDesc->IndependentBlendEnable = TRUE;
     }
 
-    D3D11_RENDER_TARGET_BLEND_DESC1* rt = &pDesc->RenderTarget[0];
-    if (rt->BlendEnable != 0) {
-      rt->BlendEnable = 1;
-      //Can not have Blend enabled the same time as LogicOp
-      if (rt->LogicOpEnable == 1) {
-        Logger::err(str::format("D3D11BlendState: Logic Op must be disabled if Blend is enabled: "));
-        return E_INVALIDARG;
-      }
-      
-      if (!ValidateBlendOperations(
+    const uint32_t numRenderTargets = pDesc->IndependentBlendEnable ? 8 : 1;
+
+    for (uint32_t i = 0; i < numRenderTargets; i++) {
+
+      D3D11_RENDER_TARGET_BLEND_DESC1* rt = &pDesc->RenderTarget[i];
+      if (rt->BlendEnable) {
+        rt->BlendEnable = TRUE;
+        //Can not have Blend enabled the same time as LogicOp
+        if (rt->LogicOpEnable) {
+          return E_INVALIDARG;
+        }
+
+        if (!ValidateBlendOperations(
           rt->SrcBlend, rt->SrcBlendAlpha,
           rt->DestBlend, rt->DestBlendAlpha,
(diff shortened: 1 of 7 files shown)


mdrost:
I see many things like
```
if (pDesc->AlphaToCoverageEnable != 0) {
      pDesc->AlphaToCoverageEnable = 1;
}
```
1. Isn't AlphaToCoverageEnable already 1 if not 0?
2. It's better to use TRUE and FALSE.

ZeroFault:
It's to fix applications who don't initialize all the members so they are stuck with garbage data. D3D11 will enable those variables if they aren't equal to 0. So I force it back 1 to make it do what D3D11 does. It's also very stupid because random garbage data could accidentally be legal values. D3D11 is to blame for allowing values other than 1 to enable something.

ZeroFault:
I mean I guess I could, but by the time the code get's there I already corrected for it by setting it to 1. Also, D3D11 resets it back to 1. I'm only matching what it does.

oscarbg:
This series of patches have immediate effect of fixing rendering issues on any game to test?

ZeroFault:
This patch mainly fixes reference count issues that usually show up in apitraces. Though, it _may_ help games as well.

doitsujin:
@ZeroFault 
> I mean, I guess I could, but by the time the code get's there I already corrected for it by setting it to 1

But then there's even less reason to pretend that these booleans are integers. This `== 0` and `== 1` is something I really don't like either.

src/d3d11/d3d11_blend.cpp
```diff
@@ -148,8 +148,130 @@ namespace dxvk {
...
+    if (rt->LogicOpEnable == 0) {
+      rt->LogicOp = defaultDesc.RenderTarget[0].LogicOp;
+    }
+
+    if (rt->RenderTargetWriteMask > D3D11_COLOR_WRITE_ENABLE_ALL) {
+      Logger::err(str::format("D3D11BlendState: Invalid RenderTargetWriteMask: ", rt->RenderTargetWriteMask));
+      return E_INVALIDARG;
+    }
+
+    //for the rest of the rendertargets
```
doitsujin: This loop looks like unnecessary code duplication, or what are the differences between the first and all following render targets?

src/d3d11/d3d11_blend.cpp
```diff
@@ -148,8 +148,130 @@ namespace dxvk {
   
   
   HRESULT D3D11BlendState::NormalizeDesc(D3D11_BLEND_DESC1* pDesc) {
-    // TODO validate
-    // TODO clear unused values
+    const D3D11_BLEND_DESC1 defaultDesc = DefaultDesc();
+    if (pDesc->AlphaToCoverageEnable != 0) {
```
doitsujin: As mentioned, please no `!= 0`, `== 1` etc. for booleans.

src/d3d11/d3d11_blend.cpp
```diff
@@ -148,8 +148,130 @@ namespace dxvk {
...
+      pDesc->AlphaToCoverageEnable = 1;
```
doitsujin: Assign `TRUE` instead of `1`

src/d3d11/d3d11_rasterizer.cpp
```diff
@@ -151,7 +151,50 @@ namespace dxvk {
...
+    if (pDesc->AntialiasedLineEnable != 0) {
+      pDesc->AntialiasedLineEnable = 1;
+    }
+
+    if (pDesc->ForcedSampleCount != 0
+     && pDesc->ForcedSampleCount != 1
+     && pDesc->ForcedSampleCount != 2
+     && pDesc->ForcedSampleCount != 4
+     && pDesc->ForcedSampleCount != 8
+     && pDesc->ForcedSampleCount != 16) {
```
doitsujin: You can do the same thing by checking the result of `DecodeSampleCount`.

ZeroFault:
I'll fix it, but BOOL is literally defined as an integer anyway. It's not pretending.

src/d3d11/d3d11_blend.cpp
```diff
@@ -235,5 +314,56 @@ namespace dxvk {
...
     return VK_LOGIC_OP_NO_OP;
   }
+
+  bool D3D11BlendState::ValidateBlend(const D3D11_BLEND blend) {
+    return blend >= D3D11_BLEND_ZERO
+        && blend <= D3D11_BLEND_INV_SRC1_ALPHA;
+  }
+
+  bool D3D11BlendState::ValidateBlendAlpha(const D3D11_BLEND blendAlpha) {
+    //can't be color operations
```
mdrost: Instead of blacklist, with weird ranges outside possible enums, wouldn't whitelist be better?

ZeroFault: Well, in this situation,  a whitelist would have 13 comparisons versus 8 comparisons on the blacklist.

PR #345, opened 2018-05-05
Add checks for DXUP

Joshua-Ashton:


kiljacken:
Nice addition, but it would probably be preferable to have DXVK export a function to configure this, than having it check for DXUP explicitly. Just my two cents

Joshua-Ashton:
Not possible.

PR #346, opened 2018-05-05
Adding env var for opting into 24-bit Vulkan formats.

saboya:
I guess this is the fastest way to make this optional. The performance boost is relevant on Nvidia cards, so I thought it's worth a shot.


saboya:
Yeah I figured you wouldn't want to merge an env var way to do this, I understand your reasoning. I just made this to test the impact locally. I realize the correct way to do this is to request for formats having the device as a parameter, but like you said, it would require some refactoring.

And also, like you said, might still cause some issues on some games, so being able to toggle it on/off would be desirable.

oscarbg:
@saboya just curiosity what % speedup gains are we talking about with this patch?
Can share some benchmark results?

doitsujin:
Fixed in 757be61b700c56c49f7ba32545a3b1f464e2d2bd.

PR #356, opened 2018-05-07
emplace_back() usually faster than push_back()

CochainComplex:


doitsujin:
See #358.

neVERberleRfellerER:
in this case you create a temporary that is then passed to emplace_back. Purpose of emplace_back is to avoid temporaries when constructing object from arguments of emplace_back. Totally useless and performance difference is zero.
For this to be useful you would have to add constructor to Entry and use emplace_back this way:
`emplace_back(buffer, slice)`

PR #358, opened 2018-05-07
emplace_back() usually faster than push_back()

CochainComplex:


doitsujin:
If you want this merged, please sumit it as **one** PR that I can review properly, there's no need to submit five that all just change one line of code. I'm going to close the rest of your PRs.

From what I've seen, none of your changes affect performance-critical code in any way so the title might be misleading.

CochainComplex:
sry .. ;)

neVERberleRfellerER:
There is absolutely no performance difference in this case, since inserted value is pointer.

Issue #259, opened 2018-04-09
clang-tidy linter usage


pchome:
I'm currently playing with [clang-tidy](https://clang.llvm.org/extra/clang-tidy/), just for fun. So here is some useful (?) notes how to use it with DXVK:
* go to project build directory
* copy `/usr/lib/llvm/7/share/clang/run-clang-tidy.py` or https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/run-clang-tidy.py python script, it will use `ninja`'s `compile_commands.json` file to analyse project files
* create script or use command as shown below :
```sh
#!/bin/sh

./run-clang-tidy.py \
-extra-arg="-std=c++17" \
-extra-arg="-stdlib=libc++" \
-extra-arg="-isystem/usr/include/wine/windows" \
-extra-arg="-isystem/usr/include/c++/v1" \
-extra-arg="-DNOMINMAX" \
-extra-arg="-DWINE_UNICODE_NATIVE" \
-extra-arg="-D__WINE__" \
-extra-arg="-fshort-wchar" \
-header-filter='.*,-vulkan.h,-vulkan_platform.h' \
-checks='-*,modernize-*' 

```

### Output example:
```c++
/tmp/build.w64/../dxvk/src/util/util_env.cpp:60:49: warning: use nullptr [modernize-use-nullptr]
    if (::CreateDirectoryW(dxvkTempDir.c_str(), 0) == 0) {
                                                ^~
                                                nullptr                                                                                                
Suppressed 6433 warnings (6433 in non-user code).
```
```c++
/tmp/build.w64/../dxvk/src/dxvk/dxvk_graphics.cpp:218:5: warning: use range-based for loop instead [modernize-loop-convert]
    for (uint32_t i = 0; i < 4; i++)
    ^   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
        (float & blendConstant : cbInfo.blendConstants)                                                                                                
Suppressed 14428 warnings (14427 in non-user code, 1 with check filters).
```

```c++
/tmp/build.w64/../dxvk/src/dxvk/hud/dxvk_hud_text.cpp:75:5: warning: use auto when initializing with a cast to avoid duplicating the type name [modernize-use-auto]
    HudTextVertex* vertexData = reinterpret_cast<HudTextVertex*>(
    ^~~~~~~~~~~~~~
    auto                                                                                                                                               
Suppressed 14082 warnings (14082 in non-user code).
```

[...]

PR #373, opened 2018-05-14
Remove whitespaces

N00byKing:
This PR, as the title says, removes all trailing whitespaces.

Whitespaces were removed using `git apply --whitespaces=fix [...]`


ssorgatem:
Statistically, this will be the biggest  external code contribution so far :laughing:

N00byKing:
lel, true. However, Git(-hub) counts contributions by number of commits, and this is only one \:3

pchome:
> There's no harm in having those spaces and it's what my editor does by default.

This is one of the most annoying thing while preparing PR.
While "_Remove trailing whitespaces_" usually an option you should set manually, "_Add new line at the end of file_" usually default for editors.

So at least you can support those poor people trying to contribute by enabling such options on your side.

Or I can advertise **VSCode** to you, I know ist's hard to change environment and it's not what developer should do for an random reason, but I spent a lot of time with this earlier, with only theme (darkula) and `llvm-cland` plugins. That's was enough even for me, despite `llvm-cland` is in active development now.

FYI, #259 and `compile_flags.txt` for note.

PR #455, opened 2018-06-25
package-release.sh: use mkdir -p

mati865:
This silences error when directory already exists.

You can use it for master also.

### DIFF
--- a/package-release.sh
+++ b/package-release.sh
@@ -27,7 +27,7 @@ function build_arch {
   cd "$DXVK_BUILD_DIR/build.$1"
   ninja install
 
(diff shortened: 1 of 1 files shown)


ssorgatem:
I'm not sure this is a good thing.
You probably want to make sure that directory is empty if it already exists, or you risk mixing up files from different builds.
I think it's just easier to complain about the directory already existing rather than silently ignore it or risking inadvetently deleting or overwriting files.

mati865:
Right now the script doesn't use `set -e` so it doesn't exit despite the error.
It's just a cosmetic change.

ssorgatem:
All the more reason not to silence it, IMO.

PR #465, opened 2018-06-29
[dxvk] Remove unnecessary Parameter

N00byKing:
Obsoleted in 79a1703aea52d7b48558d391ad81b7fbb872086b


doitsujin:
That's being awfully picky (this parameter was used until i removed the log messages), but whatever.

N00byKing:
Code gotta be so clean you can see your reflection in it ;)

Issue #488, opened 2018-07-08
Feature enhancement : DXVK HUD place and DXVK version reporting in DXVK HUD 1 option


Leopard1907:
Hi , i wish i could help you with that request but i'm just an end-user. So these are just suggestions actually.

1-) DXVK hud appears on top left on screen , that makes some info pop up-s on games ( GTA5 ) nearly invisible. Due to i'm a new GTA5 user and still need guidance on some missions , yet i have also a bad habit like checking fps always that situation makes also gaming and checking fps unbearable.

2-) I sometimes grab screenshots at some certain areas with games. I'm also a member of a local Linux community site and planning to inform about " What is DXVK and how you can use it simply ? " to help some people who are willing to use Linux but can't stop dual booting because of Windows only games that runs really slow with WineGL. I'm planning to keep that topic up-to-date just like our native Linux games topic.

So i'm gonna report every new release with performance improvements etc. As you can guess , people have short attention span usually. So reporting visually makes it more visible rather than text.

I know you have lot of tasks going on , maybe you can consider them in future.

Thanks.

jarrard:
Can the commit version be shown even in the HUD? I was pretty sure DXVK is completely unaware of what version or commit it is.

Bfgeshka:
It's actually pretty easy to do, the only thing needed is line in HUD and `-D`-variable from build system.

pchome:
Simplified variant (prints e.g. `DXVK:  d91041f`):
```diff
diff --git a/meson.build b/meson.build
index 5349c64..03b7dc6 100644
--- a/meson.build
+++ b/meson.build
@@ -5,6 +5,10 @@ cpu_family = target_machine.cpu_family()
 dxvk_compiler = meson.get_compiler('cpp')
 dxvk_cpp_std  = 'c++1z'
 
+# DXVK version
+dxvk_ver = run_command('git', 'rev-parse', '--short', 'HEAD').stdout().strip()
+add_project_arguments('-DDXVK_VERSION="' + dxvk_ver + '"', language : 'cpp')
+
 # MSVC
 if dxvk_compiler.get_id() == 'msvc'
   add_project_arguments('-DNOMINMAX', language : 'cpp')

```

```diff
diff --git a/src/dxvk/hud/dxvk_hud_devinfo.cpp b/src/dxvk/hud/dxvk_hud_devinfo.cpp
index 4b1604d..964784f 100644
--- a/src/dxvk/hud/dxvk_hud_devinfo.cpp
[...]

PR #490, opened 2018-07-09
[build] detect and report DXVK version

pchome:
#488
Version reported in log and HUD like `DXVK: 0.61 (56f7ef5)`

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,12 +11,18 @@ else
 endif
 
 # DXVK version
-dxvk_ver = run_command('head', '-n1', 'RELEASE').stdout().strip()
-dxvk_git_ver = run_command('git', 'rev-parse', '--short', 'HEAD').stdout().strip()
-if dxvk_git_ver != ''
-  dxvk_ver += ' (' + dxvk_git_ver + ')'
+is_head = find_program('head', required : false).found()
+is_git  = find_program('git', required : false).found()
+if is_head
+  dxvk_ver = run_command('head', '-n1', 'RELEASE').stdout().strip()
+  if dxvk_ver != '' and is_git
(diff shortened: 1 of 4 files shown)


pchome:
> Has this been tested on Windows?

no
> What happens if the command to retrieve the version fails?

`meson.build:9:0: ERROR:  Program or command 'head123' not found or not executable`
or, I suppose, `run_command().stdout()` will return nothing if command exists but prints to `.stderr()`
> I'm really not a fan of those `#ifdefs`, they should all be removed.

they was added for cases when DXVK compiled from sources (.tar.gz) and not possible to detect git version, but they don't work as expected (since there is no corresponding condition checks)
> and I'd prefer to add it as a separate option rather than adding it to the `devinfo` thing.

"Software info block" logic used, not sure it needs separate option, but...

feel free to reject this PR

pchome:
Last commits should check commands and fix general logic of `#ifdefs`. For `.tar.gz` builds will show only `DXVK: 0.61`. On Windows or fail should not show DXVK version (linux only PR).

doitsujin:
Thanks for the update. It's fine if the version display doesn't work on Windows builds, but it shouldn't break them. I'll take a closer look at this tomorrow.

GloriousEggroll:
What if we make it give the DXVK version number by default, and add the git commit number only if it's not null? that way it always has a version number, doesn't require ifdefs, and shouldnt spit out an error because it wouldnt add anything to the version number if it was found to be null

doitsujin:
After reading some documentation, Meson already has a farily convenient way of forwarding git tags and commit hashes with the [vcs-tag](http://mesonbuild.com/Reference-manual.html#vcs_tag) command, so I went with that. In any case, thanks for the effort.

PR #501, opened 2018-07-18
Winelib build

jacekcw:
This pull request is partially based on https://github.com/doitsujin/dxvk/pull/268

With those patches I'm able to do a working 64-bit winelib builds. 32-bit builds need some more work.

Tests can be built as well, but Wine needs to be patched to provide d3dcompiler_47 importlib by adding:
IMPORTLIB = d3dcompiler_47
to dlls/d3dcompiler_47/Makefile.in. I till take care of that on Wine side once this is merged.

### DIFF
--- a/src/d3d11/d3d11_context_imm.cpp
+++ b/src/d3d11/d3d11_context_imm.cpp
@@ -527,7 +527,7 @@ namespace dxvk {
       
       // TODO implement properly in DxvkDevice
       while (Resource->isInUse())
-        std::this_thread::yield();
+        dxvk::this_thread::yield();
     }
     
     return true;
--- a/src/d3d11/d3d11_context_imm.cpp
+++ b/src/d3d11/d3d11_context_imm.cpp
@@ -3,6 +3,9 @@
 #include "d3d11_device.h"
 #include "d3d11_texture.h"
 
+constexpr static uint32_t MinFlushIntervalUs = 1250;
+constexpr static uint32_t MaxPendingSubmits  = 3;
+
 namespace dxvk {
   
   D3D11ImmediateContext::D3D11ImmediateContext(
--- a/src/util/thread.h
+++ b/src/util/thread.h
@@ -1,0 +1,63 @@
+#pragma once
+
+#include "util_error.h"
+
+#include <windef.h>
+#include <winbase.h>
+
+/*
+ * This is needed mostly  for winelib builds. Wine needs to setup each thread that
+ * calls Windows APIs. It means that in winelib builds, we can't let standard C++
+ * library create threads and need to use Wine for that instead. We use a thin wrapper
+ * around Windows thread functions so that the rest of code just has to use
+ * dxvk::thread class instead of std::thread.
+ */
+namespace dxvk {
+  class thread {
+  private:
+    HANDLE m_handle;
+
+    template<typename T>
+    class thread_proc {
+    private:
+      T proc;
+    public:
+      thread_proc(T &proc) : proc(proc) {}
+
+      static DWORD WINAPI nativeProc(void *arg) {
+        thread_proc *proc = reinterpret_cast<thread_proc*>(arg);
+        proc->proc();
+        delete proc;
+        return 0;
+      }
+    };
+
+  public:
+    template<class T>
+    explicit thread(T &&func) {
+        thread_proc<T> *proc = new thread_proc<T>(func);
+        m_handle = ::CreateThread(nullptr, 0, thread_proc<T>::nativeProc, proc, 0, nullptr);
+        if (!m_handle) {
+            delete proc;
+            throw DxvkError("Failed to create thread");
+        }
+    }
+
+    thread() : m_handle(nullptr) {}
+
+    ~thread() {
+      if (m_handle)
+        ::CloseHandle(m_handle);
+    }
+
+    void join() {
+      ::WaitForSingleObject(m_handle, INFINITE);
+    }
+  };
+
+  namespace this_thread {
+    inline void yield() {
+      Sleep(0);
+    }
+  }
+}
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -36,22 +36,6 @@ namespace dxvk::env {
   }
   
   
-  std::string getTempDirectory() {
-    WCHAR windowsTempDir[MAX_PATH] = {0};
-    UINT ret = ::GetTempPathW(MAX_PATH, windowsTempDir);
-    if (ret > MAX_PATH || ret == 0)
-      return std::string();
-    
-    auto dxvkTempDir = std::wstring(windowsTempDir) + L"dxvk\\";
-    if (::CreateDirectoryW(dxvkTempDir.c_str(), 0) == 0) {
-      if (::GetLastError() != ERROR_ALREADY_EXISTS)
-        return std::string();
-    }
-    
-    return str::fromws(dxvkTempDir);
-  }
-
-
   void setThreadName(const wchar_t* name) {
     using SetThreadDescriptionProc = void (WINAPI *) (HANDLE, PCWSTR);
 
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -16,7 +16,7 @@ namespace dxvk::env {
     }
     
     result.resize(len);
-    return str::fromws(result);
+    return str::fromws(result.data());
   }
   
   
@@ -27,7 +27,7 @@ namespace dxvk::env {
     DWORD len = ::GetModuleFileNameW(NULL, &exePath.at(0), MAX_PATH);
     exePath.resize(len);
     
-    std::string fullPath = str::fromws(exePath);
+    std::string fullPath = str::fromws(exePath.data());
     auto n = fullPath.find_last_of('\\');
     
     return (n != std::string::npos)
@@ -51,4 +51,24 @@ namespace dxvk::env {
       (*proc)(::GetCurrentThread(), name);
   }
   
+}
+
+#ifdef CP_UNIXCP
+static constexpr int cp = CP_UNIXCP;
+#else
+static constexpr int cp = CP_ACP;
+#endif
+
+namespace dxvk::str {
+    std::string fromws(const WCHAR *ws) {
+        size_t len = WideCharToMultiByte(cp, 0, ws, -1, NULL, 0, NULL, NULL);
+        if (len <= 1)
+            return "";
+
+        len--;
+        std::string result;
+        result.resize(len);
+        WideCharToMultiByte(cp, 0, ws, -1, &result.at(0), len, NULL, NULL);
+        return result;
+    }
 }
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -1,4 +1,5 @@
 #include "util_env.h"
+#include <vector>
 
 #include "./com/com_include.h"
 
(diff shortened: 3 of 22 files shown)


src/d3d11/d3d11_context_imm.cpp
```diff
@@ -3,6 +3,9 @@
 #include "d3d11_device.h"
 #include "d3d11_texture.h"
 
+constexpr static uint32_t MinFlushIntervalUs = 1251;
```
doitsujin: Is there a reason the value was changed from 1250 to 1251?

src/util/thread.h
```diff
@@ -0,0 +1,51 @@
...
+        delete proc;
+        return 0;
+      }
+    };
+
+  public:
+    template<class T>
+    explicit thread(T &&func) {
+        thread_proc<T> *proc = new thread_proc<T>(func);
+        m_handle = CreateThread(NULL, 0, thread_proc<T>::nativeProc, proc, 0, NULL);
```
doitsujin: Please use `nullptr` instead of `NULL` for consistency.

src/util/thread.h
```diff
@@ -0,0 +1,51 @@
+#pragma once
+
+#include "util_error.h"
+
+#include <windef.h>
+#include <winbase.h>
+
+namespace dxvk {
+  class thread {
```
doitsujin: Can you add a small comment explaining why this class is needed? It's obvious from the commit message, but it might not be obvious when reading the code that this is needed for winelib builds.

src/util/util_env.cpp
```diff
@@ -68,3 +53,23 @@ namespace dxvk::env {
...
+namespace dxvk::str {
+    std::string fromws(const WCHAR *ws) {
+        size_t len = WideCharToMultiByte(cp, 0, ws, -1, NULL, 0, NULL, NULL);
+        if (len <= 1)
+            return "";
+
+        len--;
+        std::string result;
+        result.resize(len);
+        WideCharToMultiByte(cp, 0, ws, -1, &result.at(0), len, NULL, NULL);
```
doitsujin: Will this work with things like CJK file names when using either the mingw build or the winelib build?

src/util/thread.h
```diff
@@ -0,0 +1,51 @@
...
+    };
+
+  public:
+    template<class T>
+    explicit thread(T &&func) {
+        thread_proc<T> *proc = new thread_proc<T>(func);
+        m_handle = CreateThread(NULL, 0, thread_proc<T>::nativeProc, proc, 0, NULL);
+        if (!m_handle) {
+            delete proc;
+            throw DxvkError("Failet to create thread");
```
doitsujin: Typo: "Faile**d**"

src/d3d11/d3d11_context_imm.cpp
jacekcw: That wasn't intentional, I will change it back.

src/util/thread.h
jacekcw: Sure. I will also use ::CreateThread while I'm at this.

src/util/thread.h
jacekcw: Right, that sounds useful, I will add it.

src/util/thread.h
jacekcw: Good catch, fixing.

src/util/util_env.cpp
jacekcw: I admit that I didn't test it too extensively, but I'd expect it to work (to the degree that Windows allows it using non-unicode APIs). I may do more testing, but here is some more background:

- on mingw build
When you do std::ofstream(str, ...), standard c++ library will end up calling one of msvcrt.*open variants, passing your string argument. Windows internally uses utf-16 for things like file names, so it needs to be converted to utf-16 at some point. In this case, msvcrt will then convert the string to urf-16 using default system codepage (CP_ACP) and pass that to lower level functions (CreateFileW). So if A->utf16 uses CP_ACP, utf16->A should also use CP_ACP for things to work reliably.

Obviously, current codepage may or may not support CJK. But if it doesn't, there is no way we may support that using ansi-based APIs. This is general limitation of the way those APIs work on Windows (and Wine needs to be compatible with Windows in this aspect). Passing utf8 string will not help in this case, because Windows will interpret that as different codepage anyway (well, unless utf8 is the system codepage).

The only way to avoid the problem is to avoid all conversions. It would require using utf16 APIs like kernel32 CreateFileW or msvcrt _wopen. There is no portable C++ way of doing that. MSVC C++ library allows passing wchar_t strings to constructors and functions, but GCC stdc++ does not. There are some tricks like what Mozilla does:
https://hg.mozilla.org/mozilla-central/file/tip/mfbt/FStream.h#l32
but I don't think that's worth it in case of dxvk.

- on winelib build
std::ofstream just ends up calling glibc's open. There are no Windows paths nor Windows encoding involved. It's just plain UNIX path name. Linux treats file names just as byte strings and does not attempt to do any conversions or decoding. It's standard practice that those byte strings are interpreted using configured encoding and that's what Wine's CP_UNIXCP extension uses. Those days it probably means utf8 for majority users, but it depends on configuration.

pchome:
What about linking **d3d11** with **dxgi**?
I my PR I'm used `dependencies        : [ declare_dependency(link_args: ['-ldxgi']), dxbc_dep, dxvk_dep ],` for d3d11 linking to avoid direct **dxgi.dll.so** dependency
```
$ ldd d3d11.dll.so 
        linux-vdso.so.1 (0x00007ffd9b7ac000)
        dxgi.dll.so => not found
        libwine.so.1 => not found
        libstdc++.so.6 => /usr/lib/gcc/x86_64-pc-linux-gnu/8.1.0/libstdc++.so.6 (0x00007f60a02f4000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f609ff26000)
        libgcc_s.so.1 => /usr/lib/gcc/x86_64-pc-linux-gnu/8.1.0/libgcc_s.so.1 (0x00007f609fd0e000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f609f974000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f60a0700000)
```
Not sure if this is a big deal, but for wine builds this produces smaller library and simplifies dll_redirects variant for wine-staging (by not depend on `dxgi.dll.so` but using `link_args: ['-ldxgi-vk']`).

jacekcw:
You're right, I meant to look at that. We have lib_dxgi for that, I will update patches.

If you wanted to add dll redirection on top of that, it should be simple to do in top level meson.build on top of this series.

pchome:
Thanks, it looks like dll redirection support patch can be simplified to:
```diff
diff --git a/meson.build b/meson.build
index 931ba15..771fccf 100644
--- a/meson.build
+++ b/meson.build
@@ -27,10 +27,10 @@ if meson.get_cross_property('winelib', false)
   add_project_link_arguments('-mwindows', language : 'cpp')
   lib_vulkan  = declare_dependency(link_args: [ '-lvulkan-1' ])
   lib_d3d11   = declare_dependency(link_args: [ '-ld3d11' ])
-  lib_dxgi    = declare_dependency(link_args: [ '-ldxgi' ])
+  lib_dxgi    = (get_option('enable_dll_redirection')) ?  declare_dependency(link_args: [ '-L'+dxvk_library_path, '-ldxgi-vk' ]): declare_dependency(link_args: [ '-ldxgi' ])
   lib_d3dcompiler_47 = declare_dependency(link_args: [ '-ld3dcompiler_47' ])
   exe_ext = '.exe.so'
-  dll_ext = '.dll'
+  dll_ext = (get_option('enable_dll_redirection')) ? '-vk.dll' : '.dll'
   def_spec_ext = '.spec'
 else
   lib_vulkan  = dxvk_compiler.find_library('vulkan-1', dirs : dxvk_library_path)
diff --git a/meson_options.txt b/meson_options.txt
index 6e0f05b..174f903 100644
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1 +1,4 @@
 option('enable_tests', type : 'boolean', value : false)
+
+option('enable_dll_redirection', type : 'boolean', value : false,
+  description: 'support wine-staging dll redirection feature')

```
and modified `libdxgi.def` as `libdxgi-vk.def` placed to `lib/` and `lib32/` directories.

If @doitsujin don't mind can you please add this into your PR?

doitsujin:
@pchome I don't think we need to bring back the DLL redirect thing. Just adds more stuff that can break for hardly any benefit.

ssorgatem:
Can this make a wine32 build too?

doitsujin:
> With those patches I'm able to do a working 64-bit winelib builds. **32-bit builds need some more work.**

pchome:
@doitsujin 
> I do wonder what exactly _more work_ will amount to, however.

AFAIK and what I have faced with:
1. You need 32 bit `winegcc`, because it uses predefined path to your `libwinecrt0.a` library and 64 bit one will try to link your exe/lib with e.g. `lib64/libwinecrt0.a`. Most (all?) distributions provides only one `winegcc` variant. Using `-nostartfiles` option and manually defining paths to required libraries probably can solve this problem, but different distribution have different paths to `wine` and it's hard to guess/determine  which one is correct (e.g. we can have multiple wine versions installed at the same time in gentoo, switching by `eselect` command and paths looks like `/usr/lib64/wine-staging-3.12` or `/usr/lib64/wine-vanilla-3.11`).
2. Produced 32 bit libraries don't work for unknown reason. Crashing even with tests. But I have tested this a while ago, and if this still true some debugging work can help to understand what going on there.

jacekcw:
1. Yes, that's a problem. winegcc/wineg++ comes from 64-bit build of Wine and does now know how 32-bit version is configured. It tries to find it in runtime, but it may or may not be able to do that. For example this /usr/lib64/wine-vanilla-3.11 layout can't work this way. For some other setups, this patch may help:
https://source.winehq.org/patches/data/148781
The problem may be worked around by passing paths with -L. In your example this would be something like -L/usr/lib32/wine-vanilla-3.11 -L/usr/lib32/wine-vanilla-3.11/wine

2. The reason of crashes is that Vulkan structs expose differences in Windows and Linux ABIs in regards to alignment. Wine had to deal with this problem quite a bit, see struct conversions in dlls/winevulkan in Wine sources. I could get 32-bit winelib build working with some pretty bad hacks, I will look for a cleaner solution.

pchome:
Also (FYI) some distribution have `lib|lib32` or `lib|lib64` directory naming, where `lib/` is 64 bit or 32 bit libraries path. E.g. gentoo users now using `lib->lib64` simlink and `lib32` for "profile" <17.1 and `lib|lib64` for >=17.1 (experimental). So `lib/` can be either 64 bit or 32 bit.

With 17.1 profile I have 32 bit libraries in ` $ ls /usr/lib/wine-staging-3.12/`
```
bin  libwine.so  libwine.so.1  libwine.so.1.0  wine
```
https://wiki.gentoo.org/wiki/Project:AMD64/Multilib_layout

pchome:
Unfortunately I don't know any other way to use this builds other than using **dll redirection** feature from wine-staging. While using e.g. `WINEDLLPATH=/home/$USER/.local/lib64/wine`, where `d3d11-vk.dll.so` and `gxgi-vk.dll.so` placed and `d3d11->d3d11-vk|dxgi->dxgi-vk` redirections you can control multiple prefixes/apps to use original wine or dxvk libraries.

That's why I'm pushing this **dll redirection** support, but it's true this feature will not work as expected for every app (tested on dxvk tests and TW3 as working). E.g. `$ lsof | grep -i d3d11 | awk '{print $1" "$11}' | sort | uniq` sometimes shows both `d3d11.dll.so` and `d3d11-vk.dll.so` used.
Maybe I'm going to create non-mergable PR as a guide how to use this winelib builds with dll redirection, but if anyone know any better/simpler way to use this ...

Or for an specified app you can replace (I suppose) wine's original `d3d11.dll.so` and `gxgi.dll.so` with `dxvk's` ones and use `WINESERVER=`|`WINE=`|`WINELOADER=` env vars to specify where your modified wine exists.

doitsujin:
@pchome you can just rename the SOs to d3d11.dll and dxgi.dll and literally use them in exactly the same way as the MinGW DLLs.

pchome:
:) nice trick, I wonder if this will work. So renaming and set them to `native` will work?

PR #506, opened 2018-07-21
[build] added appveyor support

Mixaill:
Added CI for MSVC. It could be used for:
- checking compatibility with MSVC 
- one of the way to retrieving DXVK builds in addition to https://haagch.frickel.club/files/dxvk/

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,44 @@
+image: Visual Studio 2017
+
+clone_folder: C:\Projects\DXVK
+
+environment:
+  NINJA_VERSION: "1.8.2"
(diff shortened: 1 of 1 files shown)


doitsujin:
What else would be necessary to set this up? I know pretty much nothing about Appveyor.

vlakam:
Any updates? It would be great to have autobuilds for windows-compatible dlls

Mixaill:
>And I'm not sure if maintaining MSVC compatibility is really a priority

This really should not be a priority, but it gives another possibility to get ready-to-work binaries without installing the compiler stuff on the user side.

Issue #38, opened 2018-01-27
Wine staging is getting more and more outdated, is Dxvk usable with wine-vulkan ?


pingubot:
Hi,

as wine-staging is becoming more and more outdated and for me it looks pretty dead atm (no changes since 3 month), i am wondering if dxvk is designed that it also will work with wine-vulkan ?

Many thanks !
Christian

Raffarti:
Long story short, it does, but requires more work.

wine-vulkan requires the user to install vulkan-sdk and to set it up properly, it doesn't support dlls redirections thus the current script to setup dxvk into a wineprefix will not work.

pingubot:
ok, thank you very much for the info. I will try to get it working by simply copying the correct d3d11 and dxgi dlls into the correct directories like it worked before using the setup script..

Raffarti:
I was looking into how gallium nine worked w/o staging. Their solution is to set d3d9 to native and replace d3d9.dll in system32/syswow64 with a symlink to the correct library. It will work fine as long as no other dll is loaded from a game folder or something.

doitsujin:
With wine-staging being more or less dead we don't really have a choice but to support wine-vulkan. It also seems to be in a good enough shape right now that it should be able to run DXVK, and if it supports debug callbacks (which wine-staging doesn't), it might also make development easier.

The elephant in the room is obviously getting the DLLs to load in a reliable way. Not a huge fan of creating symlinks in the wine prefix since it **will** break with every single wine update, but if there is no less invasive way of doing it I guess we have to do the same thing.

Raffarti:
@doitsujin I've successfully tested a .bat which deletes d3d11.dll when `wineboot -u` windows Run registry key (so it should run automatically when the wine version changes). Would adding that to the current script be a viable solution?

doitsujin:
Is that guaranteed to run *before* wine updates all the DLL links? If so, it's at least a temporary solution. Again I'm not a huge fan of this, but I don't have any better ideas either.

pingubot:
Did anyone of you had luck with wine-vulkan already ? I have 64 bit vulkan apps working, but not the 32-bit ones. And i guess that is also necessary for making dxvk working for 32 bit apps, correct ?

Raffarti:
@pingubot I had it working a couple of days ago, now it suddenly stopped working and I can't get why.

Raffarti:
[...]

PR #541, opened 2018-08-09
[WIP] Direct3D 9 support

GabrielMajeri:
## D3D9 API support based on DXVK

DXVK does a great job for Direct3D 11 games, however there is a huge amount of Direct3D 9 games which could certainly benefit from the extra performance / lower CPU overhead of Vulkan.

This pull request is my work-in-progress D3D9-to-D3D11 API converter, built on top of DXVK's D3D11-to-Vulkan converter.

Here is a diagram which explains what's going on:

D3D9 gets converted to D3D11/DXGI, which gets converted to Vulkan

## Why not directly convert D3D9 to Vulkan

I know that the [VK9 project](https://github.com/disks86/VK9) exists, but I believe converting the very high-level D3D9 API directly to Vulkan is way too hard, which is why I'm leveraging DXVK's extensive support for D3D11 features.

A lot of D3D9 concepts map almost directly to D3D11, it's really easy to port a lot of the features. D3D9 surfaces are nothing more than `D3D11Texture2D`, the DXBC shader compiler is already here, etc.

And since this project is built on top of DXVK, we could in certain cases use Vulkan directly, although the D3D11 API is still the one I'm building on.

## Organisation

The code follows the general DXVK coding style. I've tried to avoid modifiying existing DXVK interfaces. All of my code is located in the `src/d3d9` library, and it gets compiled to a `d3d9.dll` which can be then used normally.

## Help wanted

The D3D9 API is pretty big, including both a fixed-function pipeline (like OpenGL 1) and a partially programmable pipeline (like OpenGL 2 or newer). The [`IDirect3DDevice9`](https://docs.microsoft.com/en-us/windows/desktop/api/d3d9/nn-d3d9-idirect3ddevice9) is so big I had to break its implementation into multiple files.

I've created this pull request because I wanted to get feedback on my work and help from the community. I don't know if we should merge this WIP into DXVK yet.

Contributions to [my repo](https://github.com/GabrielMajeri/dxvk/tree/d3d9) are welcome, and there is a lot of work to be done.

*P.S: Excuse the huge number of messy commits. I can squash them once I get an idea how the final code organisation will work.*


K0bin:
If you want to run D3D9 on **top** of D3D11/DXVK, why not keep it as a separate project?

GabrielMajeri:
@K0bin This project wouldn't exist or make sense without DXVK.

- Some games make use of D3D9-D3D10-D3D11 interop. There already is a discussion to [merge DXUP (D3D10) into DXVK (D3D11)]. This project would be an extension of that.

- On Windows, converting D3D9 to D3D11 is useless, since D3D9 drivers are already well optimized.
  On Linux (what we're interested in), using this *without* DXVK would be pointless, since WineD3D's implementation of D3D11 is **slow** and built on top of OpenGL.

- I'm relying on some of DXVK's internal behavior. Having access to the source code of the lower-level API is beneficial.

- I'm using some of DXVK's utilities. Although I could copy them in a new project, I think it's better to reduce duplication.

mirh:
Aside of what has been already said, I can see why you'd want to go down the quick road..
But just like DXUP it seems redundant, if not counterproductive, in the long run (or, I mean, at least that might have its fair point as a standalone endeavor)

Were you not following all the discussing on #38?
Not to mention sooner or later this would clash with @disks86 work.

K0bin:
> I'm relying on some of DXVK's internal behavior. Having access to the source code of the lower-level API is beneficial.

You should clarify on what you're going to do:

Is this going to be an implementation of D3D9 on top of the DXVK state tracker or on top of D3D11.
Going for something in between is probably going to turn out to be a huge mess.

> This project wouldn't exist or make sense without DXVK.

I just said that it might be useful for testing because it allows you to find out if a specific issue is caused by your code or DXVK.

> Some games make use of D3D9-D3D10-D3D11 interop.

I don't think D3D9 -> 10/11 interop is supported on Windows. 10 -> 11 is supported and some games use that as mentioned in the DXUP discussion. 

It would be fine as a D3D9 fork (if it uses the DXVK state tracker) or an entirely separate project (similar to DXUP if not) but getting it merged would require @doitsujin to maintain it and he's already busy enough with 'just' D3D11.

pchome:
> but getting it merged would require @doitsujin to maintain it and he's already busy enough with 'just' D3D11.

Maybe [git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules) will help?
And meson option to enable third-party code.

doitsujin:
I can't really say I'm in favour if this. Quite the contrary, in fact.

First of all - please let me know what you're working on **before** dropping such a massive PR. This one came literally out of nowhere.

Second - as others have said, why not contribute to VK9 instead? Sure, it's not in a state to run games yet, but it already implements large portions of the API **including** the fixed-function stuff. I'd argue that the time would be better spent helping Mr. Schaefer out, but please, *talk* to him first.

Also, as others have said, if your goal is to create a Dx9->Dx11 wrapper to use on DXVK, it should be maintained as such, rather than being part of a Dx11->Vulkan translation layer.

It's not impossible to add D3D9 support to DXVK, but then t should target the DXVK state tracker directly instead of D3D11. Please note however that I currently **do not** intend to support or maintain such a thing - right now, DXVK's D3D11 support is not feature-complete yet, and it still doesn't have any built-in D3D10 support.

And last but not least, there are some major issues with the entire concept, most importantly the shader compiler:

> the DXBC shader compiler is already here, etc.

Direct3D 11 only support Shader Model 4 and 5, and that's what DXVK's shader compiler implements. It **does not** support Shader Model 1-3, and feeding Dx9 shaders to Dx11 plain and simply does not work. Part of the reason is that D3D9 and 11 have incompatible binding models and different concepts of shader constants.

> I'm relying on some of DXVK's internal behavior.

Which is always a bad thing because internal behaviour can change at **any** time. The goal is to emulate native D3D behaviour as closely as possible and deviations from that are usually a bug, not something to be relied upon.


@mirh DXUP has the advantage of being a very layer on top of D3D11, since the APIs are largely indentical if we ignore the new features that are missing from D3D10. While it might end up being somewhat redundant in the future, it is not a complete waste of time. This is most definitely not the case for D3D9.

@K0bin Surface sharing between DXGI and D3D9 (Ex) [is possible](https://docs.microsoft.com/de-de/windows/desktop/direct3darticles/surface-sharing-between-windows-graphics-apis#interoperability-between-direct3d-9ex-and-dxgi-based-apis), but DXVK currently does not provide the means to implement it. Implementing DXGI surface sharing in general would probably require wine support for Vulkan's external memory extensions as well, which is not trivial.

GabrielMajeri:
> First of all - please let me know what you're working on before dropping such a massive PR. This one came literally out of nowhere.

I created this PR to discuss my idea - as I mentioned initially, I didn't have any plans to get this merged directly into `master` - I was looking for feedback, and I've received it now.


> Direct3D 11 only support Shader Model 4 and 5, and that's what DXVK's shader compiler implements. 

Hmm, I though that might be too good to be true. I didn't realise feature level 9.0 != compatible with 9.0 shaders.


I will close this pull request and instead try to see if I can keep working on the D3D9 to D3D11 wrapper, or help Vk9 instead.

Thank you for your understanding.

doitsujin:
> I didn't realise feature level 9.0 != compatible with 9.0 shaders.

There are special [shader profiles](https://docs.microsoft.com/en-us/windows/desktop/direct3dhlsl/specifying-compiler-targets) for that, namely `xs_4_0_level9_3`. The resulting shaders actually contain both Shader Model 4 **and** Shader Model 2/3 code, as well as some information on how to map the D3D9 bindings to D3D11 bindings.

But I think most of that information exists only for shader reflection purposes, D3D11 requires the SM4 code in order to work.

PR #543, opened 2018-08-11
[build] Add a Docker-based build system

cheald:
This adds a Dockerfile which builds DXVK in an isolated Docker context (built on Debian Buster). It's straightforward - it just sets up the dependencies, copies in the source tree, and kicks off the build.

Assuming you have Docker installed and running, just run `./make_docker.sh` and it'll build the container and then run the build process with `./out` mounted to the output directory in the container. Once the run completes, your build artifacts will be in ./out, and your host system's environment remains pristine.

May be useful for people who have trouble wrangling dependencies or a build environment for whatever reason - or who just want to run one command and have everything handled for them!

The resultant image is quite large (about 2.21GB on my system), but once it's set up, it can be reused for subsequent builds easily, or torn down and thrown away without any consequence.


Zeioth:
It's a cool hack for personal use but I don't think it belongs to the main repo.

cheald:
Fair enough! I'll adjust it to be a complimentary standalone project then. Thanks!

cheald:
I ended up with it here: https://github.com/cheald/dxvk-docker

Works great - I've got a docker image published on dockerhub, and the included build script will just grab that and then build whatever gitref you want (and make a best attempt at installing it in Lutris!)

Not the sort of thing that devs will find useful, but might be handy for folks just looking to get new fixes into play quickly.

PR #560, opened 2018-08-15
[build] reduce the number of d3dcompiler versions

Mixaill:
use only d3dcompiler_43 for DXVK builds


doitsujin:
Is it **really** necessary to use _47 on MSVC? not that i have too many objections, but the two GUIDs are kind of ugly.

Mixaill:
>Is it really necessary to use _47 on MSVC? not that i have too many objections, but the two GUIDs are kind of ugly.

Other option is to redistribute `.lib` file from `d3dcompiler_43` with DXVK.

Mixaill:
@doitsujin added `.lib` files to DXVK, it should be much nicer.

doitsujin:
Hmm. I would prefer this solution, but wouldn't that violate licenses?

Mixaill:
@doitsujin replaced libs files with generated myself via `lib` tool:
```
lib /DEF:lib/libd3dcompiler_43.def /OUT:lib/d3dcompiler_43.lib /MACHINE:X64
lib /DEF:lib32/libd3dcompiler_43.def /OUT:lib32/d3dcompiler_43.lib /MACHINE:X86
```

doitsujin:
Great, thanks for getting this to work.

PR #563, opened 2018-08-15
Add script to install wine and DXVK on Ubuntu

nextgenthemes:
I thought I share my install script after it took me some time to get DXVK running on Ubuntu.

### DIFF
--- a/ubuntu-wine-dxvk-install.sh
+++ b/ubuntu-wine-dxvk-install.sh
@@ -5,7 +5,7 @@
 # This script will install wine staging and setup dxvk in the default .wine folder in the home directory. Run again to build the last cutting edge DXVK
 #
 # install wine staging via https://wiki.winehq.org/Ubuntu
-cd ~/tmp
+cd /tmp
 sudo dpkg --add-architecture i386
 wget -nc https://dl.winehq.org/wine-builds/Release.key
 sudo apt-key add Release.key
--- a/ubuntu-wine-dxvk-install.sh
+++ b/ubuntu-wine-dxvk-install.sh
@@ -1,0 +1,41 @@
+#/bin/bash
+#
+# Tested only on (K)Ubuntu 18.04 may work on other Ubuntu versions as well and very likely work on all kinds of distros based on Ubuntu 18.04.
+#
+# This script will install wine staging and setup dxvk in the default .wine folder in the home directory. Run again to build the last cutting edge DXVK
+#
+# install wine staging via https://wiki.winehq.org/Ubuntu
+cd ~/tmp
+sudo dpkg --add-architecture i386
+wget -nc https://dl.winehq.org/wine-builds/Release.key
+sudo apt-key add Release.key
+sudo apt-add-repository https://dl.winehq.org/wine-builds/ubuntu/
+sudo apt install -y winehq-staging
+# show installed version
+wine --version
+# install meson (only staight forward thing here sadly)
+sudo apt install -y meson
+# install specific versions of mingw needed to build dxvk
+sudo apt install -y mingw-w64
+printf "$(tput bold)\n>>>>>> Select posix on all the following dialogs. <<<<<<\n\n$(tput sgr 0)"
+sudo update-alternatives --config i686-w64-mingw32-g++
+sudo update-alternatives --config i686-w64-mingw32-gcc
+sudo update-alternatives --config x86_64-w64-mingw32-g++
+sudo update-alternatives --config x86_64-w64-mingw32-gcc
+# install glslang
+wget https://github.com/KhronosGroup/glslang/releases/download/7.8.2850/glslang-master-linux-Release.zip
+sudo unzip glslang-master-linux-Release.zip -d /usr
+# Build Dxvk
+if cd ~/dxvk; then
+	git pull
+else
+	git clone https://github.com/doitsujin/dxvk.git ~/dxvk
+	cd ~/dxvk
+fi
+./package-release.sh master ~/dxvk-target --no-package
+
+# set wine to use dxvk
+cd ~/dxvk-target/dxvk-master/x32
+WINEPREFIX=~/.wine bash setup_dxvk.sh
+cd ~/dxvk-target/dxvk-master/x64
+WINEPREFIX=~./wine bash setup_dxvk.sh
--- a/ubuntu-wine-dxvk-install.sh
+++ b/ubuntu-wine-dxvk-install.sh
@@ -24,7 +24,8 @@ sudo update-alternatives --config x86_64-w64-mingw32-g++
 sudo update-alternatives --config x86_64-w64-mingw32-gcc
 # install glslang
 wget https://github.com/KhronosGroup/glslang/releases/download/7.8.2850/glslang-master-linux-Release.zip
-sudo unzip glslang-master-linux-Release.zip -d /usr
+mkdir ~/.local
+unzip glslang-master-linux-Release.zip -d ~/.local
 # Build Dxvk
 if cd ~/dxvk; then
 	git pull


tingvarsson:
Pretty sure @doitsujin don't want to maintain additional scripts, and not within this repo. Suggestion would be to create your own repo for the script and share it from there. Same suggestion was made to the recent dxvk-docker image PR.

The mentioned dxvk-docker image (https://github.com/cheald/dxvk-docker) might be something you want to utilize in your script so you don't have to install the whole dev env on your desktop setup. Would turn your script into a fraction of what it is now as well probably.

nextgenthemes:
I don't think there is much to maintain, maybe the version number of glslang. I would pull it and link to it from the readme for Ubuntu users instead of getting flooded by issues from people like me who do not read the readme to the end. Apart form that the troubleshooting section is incomplete and not well thought out. Its basically lets people run into the issue to then fix is later instead of telling them form the get to they need to run this 4 commands to set mingw to posix.

pchome:
YDTW
1. You should not put anything in `/usr` as regular user, you can simply break your system in some cases. `/usr/local` is better but not much. Use your `~/.local` (or whatever else w/o `sudo`) for that.
2.  `glslangValidator` is a part of [glslang-tools](https://packages.ubuntu.com/cosmic/amd64/glslang-tools/filelist). See `1.` (conflict/breakage).
3. [HOWTO](https://www.reddit.com/r/linux_gaming/comments/8bvrab/howto_vulkan_and_dxvk_in_wine_34/)

nextgenthemes:
@pchome yes I kind if knew /usr is the wrong place but I just went with it because i wanted to get this going. The benefit of having this script here is actually show right now that it gets the attention it needs rather then sitting in some gist or repo of me where nobody sees it. Will be updating it to extract to ~/.local

You however linked to glslang-tools in cosmic (18.10) but I am using 18.04 and it seems the package does not exist here. Also did not look into the "conflict/breakage" part so it seems extracting the release to ~/.local should still be in my script.

Given how quick releases are coming out yes it may be better to just download the release. With something cutting edge like this I actually prefer to have the very latest, the compilation is not that long so I am perfectly fine with compiling it myself. And for the script its actaully easier as its just need to `git pull` and run the build again rather then figuring out how to download the latest release, get its filename right and extract that ... I am sure its possible but I don't know how.

I say it again but that readme really needs some improvement its really not good for people who just want to jump in. It should probably just link the the reddit you just linked or something. Its more geared towards devs I guess but there will be more people who just want to try it out and play games who are not devs.

nextgenthemes:
Anyone knows if the sudo update-alternatives commands can be automated to select posix without user input?

doitsujin:
I don't like the idea of having something distro-specific in this repository. It's great that your script helps out Ubuntu users, but something like this should be maintained separately, like the Docker build system proposed in #543.

If you put it into your own repo, I can put link to it into the README and/or Wiki so that users are aware that it exists.

Thaodan:
I suggest add a git repository for the glslang package , require the exact package.
There should be no user interaction needed.
(If there's no way to use the other mingw during build you can pass one to update alternatives by using a pipe, like this:
echo 1 | update-alternatives).
Everything you could be handled by packaging.

Either use a ppa or open suse build service.
You can add the message about how to install dxvk in a post install hook.

ssorgatem:
Besides... the script is pulling this repository. 
That makes no sense for something that is (or wants to be) part of it, since if you have it you already have the rest of the repository...

Overall, I think it makes more sense as a separate repository. 
See, for example, my AUR DXVK packages: https://aur.archlinux.org/packages/?O=0&SeB=nd&K=dxvk&outdated=&SB=m&SO=d&PP=50&do_Search=V%C3%A9s

Maybe you could set up your own PPA?

nextgenthemes:
> Besides... the script is pulling this repository. That makes no sense for something that is (or wants to be) part of it.

If someone ends up here like me because "Github is the new Homepage" TM and there is actually then it absolutely makes sense because like me many users will end up reading the readme and actually do not have pulled it at that point, besides what the big deal if the script pulls and builds it in a different dir again when the already have it?

The readme totally sucks and I think this will help and I can follow up and implement improvements when people suggest them like I already did. I just took a very long time for me to get this going and its far from straight forward listing things without their package names, or they do not even have packages in Ubuntu or suggesting just to install the release instead of building at all, the readme fails at all that ... I wish you many many issues like mine of people having trouble installing or building then in case you will not pull it not my problem. I have this script that works perfectly probably on way more systems then just Ubuntu 18.04. I was thinking even adding the graphics drivers PPA and optionally installing the latest Nvidia drivers as well for people who need it. I had the idea of the PR even b4 someone specifically asked me how I did install glslang btw.

> Maybe you could set up your own PPA?
No I do not because it makes no sense to me it way to complicated. Probably not even possible without a PPA for GLSLANG so I would need to do that as well if it not already exists. Also using the default .wine prefix and things like this are opinionated things not really good for a ppa so there would be a script needed or commands anyway. A PPA could just install it to some folder but that would not mean its ready to run.

Anyway, again I do not think it will be getting enough attention when I run my own repo for it. Ubuntu and derivatives are **by far** the most used Distros out there, and I would bet even more for gamers. So what the problem with providing a bunch of people a straight forward solution instead if sending then into a trail and error hell with this crap readme? Not sure whats the issue with it. I thought of it like a "And Ubuntu users can run this commands to build ..." section in the readme (very very common) but instead of listing a bunch of commands, just link the the script (or actually truely put it in the readme) If someone wants to let this script die somewhere else you can do so I hereby declare it public domain but I won't maintain this on my own, at least not public-ally if not tied to the project it belongs to. This is enough of me arguing for this its not that I desperately want to spend time helping.

leafi:
Linking to a PPA, if one exists, seems like a great possible solution.

Also (linking to a script that) builds in a Docker container is a great idea. Allows for a known cross-distro compilation environment, and doesn't have the disadvantage of the script in this PR of modifying the system in potentially surprising ways.

Doesn't Lutris ship DXVK in a bunch of wine versions these days? That's a really user friendly tool, and probably a great way for someone who doesn't want to start building stuff to get access to modern DXVK quite quickly.

@nextgenthemes Your idea of having something in the README providing guidance to people who've heard cool things about DXVK, but will find performing a build themselves frustrating, is something I can agree with.

Your script isn't a good solution here. Building in a Docker container would erase all the concerns I have about leaving files hanging around, making update-alternatives modifications that persist without telling the user, hardcoding the glslang release version when you could scrape it from github, .. etc. But you don't seem to even be interested in improving the script, Docker or not...

(Whether the script should go in or not has nothing to do with how hard you worked on it. It should only go in if there's a problem to be solved (maybe, yes!) and it's the right solution (not really).)

Also, stop being rude to people on GitHub. It doesn't get you what you want and other people will not want to work with you. I read your rant, and frankly I want my time back.

ssorgatem:
Maybe building DXVK is not an endeavour most Ubuntu and derivatives users are likely to want to pursue, anyway. I think their users are better off leveraging Lutris or winetricks, which include easier ways to install DXVK.

In general, Ubuntu and derivatives are not too friendly to building and installing your own software, specially if it's bleeding edge software with dependencies newer than what is included in the repos.

Those distros are simply not meant for that.

This is actually the reason I ended up switching to Manjaro (for Xorg/Mesa drivers specifically) years ago. Maybe you could give it a try.

I don't think it's DXVK's resposibility to support setup and install scripts for specific distros.
Even less so scripts that setup the build environment for a specific distro.
Pointing to a docker container *may* be the right solution, though.

ssorgatem:
And no, pulling *again* the repository the script is in makes no sense.

IarnGreiper:
Actually Valve included DXVK as a part of their WINE  fork called Proton.
You just have to use Steam beta from now.

And gratz doitsujin for the job at Valve :)

doitsujin:
Closing because I fully agree with @ssorgatem here.

PR #564, opened 2018-08-15
Reworked patch from https://github.com/jomihaka/dxvk-poe-hack for discard shaders until they be compiled

tuxmaniac:
Adopted to dxvk master branch (new options in config file) and tested with games:
- Path of Exile - playing without this patch just a pain in the ass, especially bosses with many visual effects (elder and his guardians, uber elder)
- No Man's Sky - some shutters while compiling shaders can cause sound glitches and sometimes need to restart game to fix audio (maybe hardware or wine dependent, but this patch fixed it)

### DIFF
--- a/src/dxvk/dxvk_graphics.cpp
+++ b/src/dxvk/dxvk_graphics.cpp
@@ -93,6 +93,14 @@ namespace dxvk {
     
     m_common.msSampleShadingEnable = fs != nullptr && fs->hasCapability(spv::CapabilitySampleRateShading);
     m_common.msSampleShadingFactor = 1.0f;
+
+    DxvkShader* vert;
+    DxvkShader* frag;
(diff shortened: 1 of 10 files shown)


doitsujin:
Upstreaming this is not an option.

First of all, applying it to games **will** break rendering in undefined ways. This is fine for some titles, but it is most definitely **not** fine for most of them, especially multiplayer games where rendering all parts of the scene is crucial. I don't want to add an option that breaks games, people will find ways to use it inappropriately.

Secondly, and I'm sorry to put it this way, but the solution presented here is awful. This is probably due to lack of familiarity with the code base, but that doesn't change that it is absolutely insane.

The only thing that this hack would need to do is skip the pipeline compilation that stalls the rendering thread. That's what my original hack did, and you can do that with three lines of code. No need for "placeholder shaders" that do nothing, no need to modify a total of **ten** files and scatter dirty code everywhere. Totally ridiculous for a workaround that is *designed to break games*.

PR #575, opened 2018-08-18
[build] Don't run wine during winelib build process

pchome:
This will tell meson to not run initial compiler test on binaries generated by `winegcc`.
http://mesonbuild.com/Cross-compilation.html#defining-the-environment

Since `winegcc` produces `sanitycheckc_cross.exe` wrapper script, which will run command like `wine sanitycheckc_cross.exe.so` to test generated binary, we can skip this test and don't summon `wine` at all during build process.

This also can help to package DXVK in some distribution (like gentoo), where running wine in sandboxed environment can be a problem.

### DIFF
--- a/build-wine64.txt
+++ b/build-wine64.txt
@@ -5,6 +5,7 @@ ar = 'ar'
 strip = 'strip'
 
 [properties]
+needs_exe_wrapper = true
 winelib = true
 
 c_args=['-m64']
--- a/build-wine32.txt
+++ b/build-wine32.txt
@@ -3,13 +3,13 @@ c = 'winegcc'
 cpp = 'wineg++'
 ar = 'ar'
 strip = 'strip'
-exe_wrapper = 'wine'
 
 [properties]
+needs_exe_wrapper = true
 winelib = true
 
 c_args=['-m32']
-cpp_args=['-m32', '--no-gnu-unique']
+cpp_args=['-m32', '--no-gnu-unique', '-Wno-attributes']
 cpp_link_args=['-m32', '-mwindows']
 
 [host_machine]


pchome:
Also you can test this with mingw builds, AFAIK wine used only to test if compiler can produce valid executable (by executing sanitycheckc_cross.exe). Creating whole new (300MB+/500MB+ for 32/64 bit) prefix to test this is a bit redundant way.

Than you can remove `WINEPREFIX` and `WINEARCH` exports from build helper, If this working.

cjacek:
I considered the same thing when I was adding build-wine32.txt, but decided to be consistent with mingw setup. I think the change is fine.

pchome:
I have working `ebuild` repository for gentoo users here: https://github.com/pchome/dxvk-gentoo-overlay

If someone want to test all this `winelib` stuff on gentoo use 
`# USE="test utils abi_x86_32" emerge -1 games-util/dxvk:9999 -q`

Also some patches included in https://github.com/pchome/dxvk-gentoo-overlay/tree/master/games-util/dxvk/files to get things work.

Currently there:
* `dxvk-0.70-fix-32bit-build.patch` - patch for `0.70` , actually to current git master state
* `dxvk-0.70-winelib-fix.patch` - **this PR**
* `dxvk-0.70-option-for-utils.patch` - using custom setup scripts we don't want original one, disabled by option
* `dxvk-0.70-fix-tests.patch` - some fixes for tests to be built, `hlsl-compiler` temporary disabled due to linkage error

_NOTE: experimental and WIP, things changing actively, expect random breakage (I mean `use` flags and so on, not your systems ;) )_

pchome:
Proton using similar patch for a while : https://github.com/ValveSoftware/dxvk/commit/cf8264ffc3c781bd9e4629ea35cfc3eef05e592b

build-wine64.txt
```diff
@@ -5,6 +5,7 @@ ar = 'ar'
 strip = 'strip'
 
 [properties]
+needs_exe_wrapper = true
```
doitsujin: Why does this need to be set when we actually *don't* want wine to execute?

pchome: If I correctly understand meson documentation -- this will stop meson from guessing/using wrapper:

> One important thing to note, if you did not define an exe_wrapper in the previous section, is that Meson will make a best-effort guess at whether it can run the generated binaries on the build machine. It determines whether this is possible by looking at the system and cpu_family of build vs host. There will however be cases where they do match up, but the build machine is actually not compatible with the host machine. Typically this will happen if the libc used by the build and host machines are incompatible, or the code relies on kernel features not available on the build machine. One concrete example is a macOS build machine producing binaries for an iOS Simulator x86-64 host. They're both darwin and the same architecture, but their binaries are not actually compatible. In such cases you may use the needs_exe_wrapper property to override the auto-detection:
```
[properties]
needs_exe_wrapper = true

```

PR #588, opened 2018-08-21
[d3d11/d3d10] Store stream output offsets and return in d3d10 if requested

Joshua-Ashton:
D3D10's SOGetTargets has the parameter to return the offsets of each SO.

I know SO support isn't there in DXVK yet but it should at least get it prepared for D3D10.

I have added the offsets to the SO context state (which may be useful for when SO gets implemented anyway) and made it return them in the function SOGetTargets on d3d10.

### DIFF
--- a/src/d3d10/d3d10_device.cpp
+++ b/src/d3d10/d3d10_device.cpp
@@ -1561,7 +1561,7 @@ namespace dxvk {
     }
 
     if (pOffsets != nullptr)
-      Logger::warn("D3D10: SOGetTargets: Reporting buffer offsets not supported");
+      m_context->SOGetOffsets(NumBuffers, pOffsets);
   }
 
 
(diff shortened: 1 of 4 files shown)


doitsujin:
I don't like this one for two reasons:
a) what offsets is this function supposed to return? The byte offset that was passed to `SOSetTargets`, or the internal counter value?
b) I'm not aware of any game that needs this, and
c) I'd like to get Dx11 Stream Output done first before messing around with internal APIs.

Joshua-Ashton:
a) It's meant to return the original byte offset.
b) It'll probably crop up eventually.
c) That's understandable, I just figured it may have been a step in the right direction. :smiley:

Joshua-Ashton:
[https://docs.microsoft.com/en-us/windows/desktop/api/d3d10/nf-d3d10-id3d10device-sogettargets](The docs) also don't reference to this parameter being optional.

doitsujin:
Resolved internally.

PR #604, opened 2018-08-26
[build] Partially fix tests for winelib build

pchome:
This change was present in initial winelib patch, but was missed.

The only broken test is ` hlsl-compiler`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -30,7 +30,7 @@ if meson.get_cross_property('winelib', false)
   lib_d3d11   = declare_dependency(link_args: [ '-ld3d11' ])
   lib_dxgi    = declare_dependency(link_args: [ '-ldxgi' ])
   lib_d3dcompiler_43 = declare_dependency(link_args: [ '-L'+dxvk_library_path, '-ld3dcompiler_43' ])
-  lib_d3dcompiler_47 = declare_dependency(link_args: [ '-ld3dcompiler_47' ])
+  lib_d3dcompiler_47 = declare_dependency(link_args: [ '-ld3dcompiler' ])
   exe_ext = '.exe.so'
   dll_ext = '.dll'
   def_spec_ext = '.spec'


doitsujin:
Do the HLSL compiler and the DXBC disassembler **work** with this change? Because if not, making this change is kind of pointless and I remember having issues with that before.

pchome:
With this change I can compile all (except hlsl-compiler) tests.
`hlsl-compiler` failing with undefined reference error to `dxvk::Logger::s_instance`. I'm still going to find out how to fix this, disabling this test in my build at the moment.

DXBC disassembler do work.

pchome:
Also it's a `meson unity` build problem, I can build and run all tests with `--unity off` option.

jacekcw:
I didn't send this commits mostly because when I got it to Wine, a dependency on d3dcompiler_43 was introduced and I meant to look at it. Shipping .def file does not seem ideal, but works, so just fixing d3dcompiler_47 should be enough.

I can't see your other linker problem here. Could you provide more info why do you need dxbc-compiler changes?

I submitted a cleaner solution for dxgi factory tests: https://github.com/doitsujin/dxvk/pull/607.

pchome:
@jacekcw 
> I can't see your other linker problem here. Could you provide more info why do you need dxbc-compiler changes?

You mean `hlsl-compiler`? 
Error occurs during compilation with `meson --cross-file ... -Denable_tests=true --unity on ...` 
See http://mesonbuild.com/Unity-builds.html#unity-builds

pchome:
With proposed patch all tests are compile/work fine now.
Added as temporary workaround for my `games-util/dxvk[test]` and `games-util/dxvk-tests` [ebuilds](https://github.com/pchome/dxvk-gentoo-overlay/tree/master/games-util).

pchome:
Closing as superseded by #607 .

Patch from https://github.com/doitsujin/dxvk/pull/604#issuecomment-416283000 still valuable.

PR #607, opened 2018-08-27
Fix winelib tests builds.

jacekcw:


pchome:
@pchome
> hack

Looks like [here](https://source.winehq.org/git/wine.git/blob/0c790adf300dd37f624029023bc6de53cb21373c:/dlls/ntdll/loader.c#l2488) is the right place to start.
`LO_DEFAULT:`

jacekcw:
That's a good question. In case it's configured as native, it will be handled by this:
https://source.winehq.org/git/wine.git/blob/0c790adf300dd37f624029023bc6de53cb21373c:/dlls/ntdll/loader.c#l2502
When it's set to builtin, but its file is found during filesystem lookup, Wine will try to load it as builtin first. The attached hack works around that. [loader.txt](https://github.com/doitsujin/dxvk/files/2337814/loader.txt) Wine will not use its builtin when configured with b or b,n, but leave current behaviour for default settings. A solution for upstream would need more thoughts.

There is also WINEDLLPATH environment variable. It currently does not work for replacing existing Wine builtin DLLs, but that could probably be changed. (If we could have a solution not requiring DLL overrides at all that would be even better, but I don't have much hopes for that, at least short term).

pchome:
@jacekcw 
Thanks, I'll try this patch. I've thought about hardcoding DXVK dll names for my local usage, so any solution is fine atm.

PR #610, opened 2018-08-28
Stop cross-compile attempting to change rpath

flukejones:
Attempts to fix cross-compilation when using macOS.

Meson tries to change the rpath when using `d3d11_dep` on macOS. This results in an error since the tool is being used on a non-native build.

### DIFF
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -4,7 +4,8 @@ d3d10_main_src = [
 ]
 
 d3d10_deps = [ lib_d3dcompiler_43, lib_dxgi ]
-d3d10_deps += meson.get_cross_property('winelib', false) ? lib_d3d11 : d3d11_dep
+cross_build = meson.get_cross_property('winelib', false) ? true : meson.is_cross_build()
+d3d10_deps += cross_build ? lib_d3d11 : d3d11_dep
 
 d3d10_core_dll = shared_library('d3d10core'+dll_ext, d3d10_main_src,
   name_prefix         : '',


Luke-Nukem:
It may also be possible to change:

```
d3d10_deps += meson.get_cross_property('winelib', false) ? lib_d3d11 : d3d11_dep
```

to:

```
d3d10_deps += meson.is_cross_build() ? lib_d3d11 : d3d11_dep
```

The above would also catch the winelib I think?

Luke-Nukem:
As it turns out, this breaks the w32 build (I was testing only w64):

```
[148/176] Linking target src/d3d10/d3d10.dll.
FAILED: src/d3d10/d3d10.dll 
i686-w64-mingw32-g++  -o src/d3d10/d3d10.dll 'src/d3d10/src@d3d10@@d3d10@sha/d3d10_main.cpp.obj' 'src/d3d10/src@d3d10@@d3d10@sha/d3d10_reflection.cpp.obj' ../../dxvk/src/d3d10/d3d10.def -Wl,--as-needed -Wl,-O1 -shared ../../dxvk/src/d3d10/d3d10.def -Wl,--start-group -Wl,--out-implib=src/d3d10/libd3d10.dll.a src/util/libutil.a src/dxbc/libdxbc.a src/dxvk/libdxvk.a src/spirv/libspirv.a /Users/luke/src/Proton/dxvk/lib32/d3dcompiler_43.lib -ldxgi -ld3d11 /Users/luke/src/Proton/dxvk/lib32/vulkan-1.lib -pthread -lkernel32 -luser32 -lgdi32 -lwinspool -lshell32 -lole32 -loleaut32 -luuid -lcomdlg32 -ladvapi32 -Wl,--end-group '-Wl,-rpath,$ORIGIN/../../../../dxvk/lib32' -static -static-libgcc -static-libstdc++ -Wl,--add-stdcall-alias,--enable-stdcall-fixup 
/usr/local/Cellar/mingw-w64/5.0.4_1/toolchain-i686/bin/i686-w64-mingw32-ld: src/d3d10/src@d3d10@@d3d10@sha/d3d10_main.cpp.obj:d3d10_main.cpp:(.text+0x89): undefined reference to `D3D11CoreCreateDevice@24'
collect2: error: ld returned 1 exit status
[157/176] Compiling C++ object 'src/d3d11/src@d3d11@@d3d11@sha/d3d11_context.cpp.obj'.
ninja: build stopped: subcommand failed.
```

Luke-Nukem:
Given that 32-bit is going to be depreciated on macOS, the breakage of 32-bit builds is probably a non-issue. But it would be much better to restrict that to macOS only.

doitsujin:
We can't support Mac OS upstream.

PR #611, opened 2018-08-29
[util] don't exit when setup script is successful

Lctrs:
If dxvk's setup script is called inside another script, it prevents the code following the call from running.


doitsujin:
Why is this necessary? There is literally no difference between `exit 0` and doing nothing.

PR #612, opened 2018-08-29
[util] correctly resolves the setup script's path

Lctrs:
If the setup script is called from another script, current code will return the path of the callee script, and not the setup script's one.

### DIFF
--- a/utils/setup_dxvk.sh.in
+++ b/utils/setup_dxvk.sh.in
@@ -2,7 +2,7 @@
 
 export WINEDEBUG=-all
 
-dlls_dir=`dirname "$(readlink -f $0)"`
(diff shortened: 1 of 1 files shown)


Thaodan:
You can use ${VAR%/*} instead of dirname to avoid  an external command.

Lctrs:
@edmondo I'm not really good with Bash. :/

I took this from https://stackoverflow.com/questions/59895/getting-the-source-directory-of-a-bash-script-from-within.

Thaodan:
Except that I don't see why you 
want to source that scripts its 
fine.

edmondo:
Yes, same question here. It would be nice to know, if it's an existing practical and real issue where something is failing because of it.

doitsujin:
Obsolete due to new setup script. And I do agree, there's no reason to source the script instead of running it directly.

PR #618, opened 2018-09-02
[utils] Add option to control utils installation

pchome:
Previous behaviour preserved for those using `package-release.sh` script.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -62,12 +62,10 @@ dxvk_version = vcs_tag(
 
 subdir('src')
 
-enable_tests = get_option('enable_tests')
-
-if enable_tests
+if get_option('enable_tests')
   subdir('tests')
 endif
 
-if dxvk_compiler.get_id() != 'msvc'
+if get_option('enable_utils')
   subdir('utils')
 endif
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,1 +1,2 @@
 option('enable_tests', type : 'boolean', value : false)
+option('enable_utils', type : 'boolean', value : false)
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,2 +1,2 @@
 option('enable_tests', type : 'boolean', value : false)
-option('enable_utils', type : 'boolean', value : false)
+option('enable_utils', type : 'boolean', value : true)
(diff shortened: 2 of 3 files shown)


doitsujin:
What is the point?

pchome:
By default `ninja install` command will install setup script into `$prefix/bin` directory.

With option it's easier to control if custom setup script used, or no setup script needed. Especially when `ninja install` used to install files directly into system.

pchome:
Ok, maybe I should extend my answer.

I think the only reason [all this](https://github.com/ValveSoftware/Proton/blob/proton_3.7/build_proton.sh#L447) `cp .../x.dll .../y.dll` constructions exist in [build-scripts](https://github.com/doitsujin/dxvk/blob/master/package-release.sh#L36) is that DXVK always install `setup_dxvk.sh` into `bin/` directory. 

_While writing this text I realized that `winelib` and `mingw` installation differs, and while `winelib` libraries are installed into `$libdir`, `mingw` dlls installed into `$bindir` and *.a static libs into $libdir. This also can be controlled w/ e.g. `static_libs` option for those don't want them to be installed._

The point is to use common practice, when you have to properly configure meson build and then use `ninja install` .

Something like this (simplified example for Proton)
```sh
meson --strip --buildtype="release" --prefix="$DST_DIR" --bindir=lib64/wine/dxvk ...
cd "$TOP"/build/dxvk.win64
ninja install
```

pchome:
Fixed this PR with following changes:
* `enable_utils` option set to `true` by default, so nothing will change for e.g. https://haagch.frickel.club/files/dxvk/latest/ and those who expect `setup_dxvk.sh` still there
* descriptions added for meson options, will be visible using `$ meson configure` command in current build directory

Also I wanted to add `'install_implibs` option, to control import libraries installation (lib/libd*.dll.a), but I have no MinGW to test this. This can be done using `implib` [keyword argument](http://mesonbuild.com/Reference-manual.html#executable) ([libd3d11.dll.a.patch.txt](https://github.com/doitsujin/dxvk/files/2368229/libd3d11.dll.a.patch.txt)), or by adding custom targets with `build_by_default: false` keyword argument and `install: get_option('install_implibs'),` for MinGW builds.

pchome:
Nevermind, this issue don't worth more time than it deserves.

PR #619, opened 2018-09-03
[dxvk] Added "DXVK_HUD_OFFSET" env var and usage doc to README

ivanpepelko:
I am by no means a c++ developer, but it was really bugging me that HUD obscured parts of game UI.


doitsujin:
Not a fan, see https://github.com/doitsujin/dxvk/issues/488#issuecomment-403622075 for details. My stance on that hasn't changed, the HUD is the last thing I want to be bloated with unnecessary features in this project since it's mostly a debugging and performance monitoring tool.

Issue #580, opened 2018-08-19
Batman Arkham Knight, game crash in detective mode


jp7677:
Batman Arkham Knight runs perfectly well with DXVK, no graphically issues and performance is really well. Thanks a lot for the hard work.
Unfortunately the game crashes on my machine when entering detective mode i.c.w. certain objects. This was also mentioned here https://github.com/doitsujin/dxvk/issues/335. The game usually crashes with a message box from the game itself.
![screenshot from 2018-08-20 22-02-40](https://user-images.githubusercontent.com/17219293/44363620-f5d93c00-a4c4-11e8-83ef-85465a7543ad.png)
I have also seen unhandled page faults in wine with the same scene. Running in fullscreen mode shows identical behavior, though I guess the message box is just behind the window. That makes it appear  like a system freeze.

### Software information
Batman Arkham Knight
Everything enabled or high, Max FPS: 60, Texture Filtering: 2x Anisotrpic.
Nvidia Gameworks features disabled

### System information
- GPU: Nvidia GTX 1080
- Driver: 396.51
- Wine version: 3.13 (Staging) from official Fedora repositories
- DXVK version: v0.70

### Apitrace file(s)
Obtained on Windows 10 with the procedure described here https://github.com/doitsujin/dxvk/wiki/Common-issues#apitrace:
- See https://github.com/doitsujin/dxvk/issues/580#issuecomment-417849484

### Log files
- d3d11.log: [BatmanAK_d3d11.log](https://github.com/doitsujin/dxvk/files/2303784/BatmanAK_d3d11.log)
- dxgi.log: [BatmanAK_dxgi.log](https://github.com/doitsujin/dxvk/files/2303785/BatmanAK_dxgi.log)
- Console output with Vulkan validation layers enabled: [vulkan-validation-output.log](https://github.com/doitsujin/dxvk/files/2307833/vulkan-validation-output.log)

### Save Game
- https://www.dropbox.com/s/cu80w2052vftvi2/SaveData.zip?dl=0
Though not sure if this works.

Don't know if this is relevant, but apitrace logging looks like
```
apitrace: loaded into C:\Program Files (x86)\Steam\steamapps\common\Batman Arkham Knight\Binaries\Win64\BatmanAK.exe
apitrace: warning: D3D11CreateDevice: unexpected IDXGIAdapter pointer 0000000007F67180
[...]

PR #627, opened 2018-09-08
Spoof AMD card for Batman: Arkham Knight

jp7677:
This prevents game crashes in "Detective Mode" for NVidia cards.
Fixes https://github.com/doitsujin/dxvk/issues/580


jp7677:
Thanks a lot!

PR #630, opened 2018-09-09
[util] Improve logging for configuration settings.

jp7677:
Log when default settings are found.
Log effective configuration settings.

### DIFF
--- a/src/dxvk/dxvk_adapter.cpp
+++ b/src/dxvk/dxvk_adapter.cpp
@@ -274,6 +274,8 @@ namespace dxvk {
     VkPhysicalDeviceMemoryProperties memoryInfo = this->memoryProperties();
     
     Logger::info(str::format(deviceInfo.deviceName, ":"));
+    Logger::info(str::format("  Vendor ID: ", deviceInfo.vendorID));
+    Logger::info(str::format("  Device ID: ", deviceInfo.deviceID));
     Logger::info(str::format("  Driver: ",
       VK_VERSION_MAJOR(deviceInfo.driverVersion), ".",
       VK_VERSION_MINOR(deviceInfo.driverVersion), ".",
--- a/src/dxvk/dxvk_adapter.cpp
+++ b/src/dxvk/dxvk_adapter.cpp
@@ -274,8 +274,8 @@ namespace dxvk {
     VkPhysicalDeviceMemoryProperties memoryInfo = this->memoryProperties();
     
     Logger::info(str::format(deviceInfo.deviceName, ":"));
-    Logger::info(str::format("  Vendor ID: ", deviceInfo.vendorID));
-    Logger::info(str::format("  Device ID: ", deviceInfo.deviceID));
+    Logger::info(str::format("  Vendor ID: 0x", std::hex, deviceInfo.vendorID));
+    Logger::info(str::format("  Device ID: 0x", std::hex, deviceInfo.deviceID));
     Logger::info(str::format("  Driver: ",
       VK_VERSION_MAJOR(deviceInfo.driverVersion), ".",
       VK_VERSION_MINOR(deviceInfo.driverVersion), ".",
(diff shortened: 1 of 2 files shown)


jp7677:
Yes, sure, sorry ;)
Currently there are two ways for setting configuration options, built-in and by config file. Settings from both way are merged and applied. I thought it might be useful to see:
1) That default settings are applied.
2) What the result is when default settings and settings from the config file are merged.

Furthermore I added logging for vendor and device id. E.g. if I don't want to use the built-in settings for vendor and device ID Arkham Knight, I would know which ID's I would have to use in a dxvk config file.

jp7677:
Logging with this PR would look like:
```
info:  Game: BatmanAK.exe
info:  DXVK: v0.71
info:  Found built-in config: BatmanAK.exe
info:  Effective config settings:
info:    dxgi.customVendorId = 1002
info:    dxgi.customDeviceId = E366
warn:  OpenVR: Failed to locate module
info:  Enabled instance extensions:
info:    VK_KHR_get_physical_device_properties2
info:    VK_KHR_surface
info:    VK_KHR_win32_surface
info:  GeForce GTX 1080:
info:    Vendor ID: 4318
info:    Device ID: 7040
info:    Driver: 396.54.0
info:    Vulkan: 1.1.70
...
```

src/util/config/config.cpp
```diff
@@ -111,6 +111,14 @@ namespace dxvk {
   void Config::merge(const Config& other) {
     for (auto& pair : other.m_options)
       m_options.insert(pair);
+
+    if (!m_options.empty()) {
+      // Inform the user about effective configuration settings
+      Logger::info("Effective config settings:");
+      for (auto& pair : m_options) {
```
doitsujin: Just a bit of nitpicking, but please don't use `{ }` around single-line blocks for consistency.

src/dxvk/dxvk_adapter.cpp
```diff
@@ -274,6 +274,8 @@ namespace dxvk {
     VkPhysicalDeviceMemoryProperties memoryInfo = this->memoryProperties();
     
     Logger::info(str::format(deviceInfo.deviceName, ":"));
+    Logger::info(str::format("  Vendor ID: ", deviceInfo.vendorID));
+    Logger::info(str::format("  Device ID: ", deviceInfo.deviceID));
```
doitsujin: These should be reported as hex numbers (`std::hex` is your friend).

jp7677:
Should be good now.

src/dxvk/dxvk_adapter.cpp
jp7677: Done.

src/util/config/config.cpp
jp7677: Done.

jp7677:
Nitpicking alert: std::hex results in lower case hex value. I guess it would read slightly nicer in the logs if the values for the built-in settings for device ID and vendor ID were also lower case. Thus 

`{ "dxgi.customDeviceId",              "e366" },`
instead of:
`{ "dxgi.customDeviceId",              "E366" },`

Let me know if you want me to include casing adjusting for the built-in settings in this PR.

doitsujin:
I added this functionality in d4947261c66132e1dee8c9b08a137505f27dc00b based on your work.

jp7677:
Thanks a lot, nice to see that this PR turned out somewhat useful.

I think you missed the logging statements for the actual vendor and device ID. May I ask you to include these too?

jp7677:
I think (as a general note) if you can override a setting, it is useful to know the default value.
Let say I want to undo the built-in overrides for GTA5 for whatever reason (troubleshooting, see if things got improved with nvapi, test if the built-in override is still needed etc). From the logs I know the built-in override IDs and I know that I would have to use a configuration file to put my actual/original values in there (because settings in the config file override the built-in settings, as far as I understood the code). In that case it would be cool if I can just copy the ID's (excluding 0x) from the dxvk log file instead of searching through /proc or somewhere else and then ev. having to think about hex-dec conversion.

I fully agree that these ID's are way to much details for the HUD, though I think they wouldn't hurt in the logs.

PR #674, opened 2018-09-28
Minor speed optimizations & overhead reductions.

Nisk:
This contribution is very trivial, but it is side-effect free & reduced overhead by a little bit in a lot of places.

It's just something I noticed when browsing the code & couldn't overlook.

### DIFF
--- a/src/d3d10/d3d10_blend.cpp
+++ b/src/d3d10/d3d10_blend.cpp
@@ -64,7 +64,7 @@ namespace dxvk {
     pDesc->DestBlendAlpha         = D3D10_BLEND   (d3d11Desc.RenderTarget[0].DestBlendAlpha);
     pDesc->BlendOpAlpha           = D3D10_BLEND_OP(d3d11Desc.RenderTarget[0].BlendOpAlpha);
 
-    for (uint32_t i = 0; i < 8; i++) {
+    for (uint32_t i = 0; i < 8; ++i) {
       uint32_t srcId = d3d11Desc.IndependentBlendEnable ? i : 0;
       pDesc->BlendEnable[i]           = d3d11Desc.RenderTarget[srcId].BlendEnable;
       pDesc->RenderTargetWriteMask[i] = d3d11Desc.RenderTarget[srcId].RenderTargetWriteMask;
--- a/src/d3d10/d3d10_device.cpp
+++ b/src/d3d10/d3d10_device.cpp
@@ -427,7 +427,7 @@ namespace dxvk {
 
     std::vector<D3D11_SO_DECLARATION_ENTRY> d3d11Entries(NumEntries);
 
-    for (uint32_t i = 0; i < NumEntries; i++) {
+    for (uint32_t i = 0; i < NumEntries; ++i) {
       d3d11Entries[i].Stream          = 0;
       d3d11Entries[i].SemanticName    = pSODeclaration[i].SemanticName;
       d3d11Entries[i].SemanticIndex   = pSODeclaration[i].SemanticIndex;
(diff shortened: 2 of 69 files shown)


ah-:
Does this really change anything? Do you see different code being generated?

I'd have thought that modern compilers do this automatically.

doitsujin:
@ah- This looks like a genuine trolling attempt. There is literally no difference between `i++` and `++i` when you don't use the result. In the context of a loop it's just a single `add` instruction at the end, anyway.

doitsujin:
Just enter a silly snippet like this into https://gcc.godbolt.org and see for yourself:
```
int sumArray(int* arr, int n) {
    int sum = 0;
    for (int i = 0; i < n; i++)
        sum += arr[i];
    return sum;
}
```
For integer types, there is absolutely no semantic difference between `++i` and `i++` if you discard the result. And if you *do* use the result, at worst you'll get one extra move instruction (and you don't have a choice anyway because using `++i` over `i++` would likely be incorrect).

pchome:
@Thinkaboutmin 
> For non-boolean operands, the expression ++x is exactly equivalent to x += 1, and the expression --x is exactly equivalent to x -= 1, that is, the prefix increment or decrement is an lvalue expression that identifies the modified operand.

-- VS --
> The result is prvalue copy of the original value of the operand. As a side-effect, for non-boolean operands, the expression x++ modifies the value of its operand as if by evaluating x += 1, and the expression x-- modifies the value of its operand as if by evaluating x -= 1.

https://en.cppreference.com/w/cpp/language/operator_incdec

> Notes
> 
> Because of the side-effects involved, built-in increment and decrement operators must be used with care to avoid undefined behavior due to violations of sequencing rules.
>
> Because a temporary copy of the object is constructed during post-increment and post-decrement, pre-increment or pre-decrement operators are usually more efficient in contexts where the returned value is not used.

Thinkaboutmin:
@doitsujin for sure it will be the same, at least for us, humans I mean. The round about code could differ in someway. Maybe the compiler would run it in a somewhat slow way, which is obviously not the case here. But I see that there's no difference, if not better, to use i++ here with @pchome shoutouts. Thanks for the answers guys!

Nisk:
It wasn't a trolling attempt. The prefix operator is the better one to use as far as what programmers intent should be (since we don't care about the value) - not relying on the compiler to do the smart thing. 

However, I didn't realise gcc optimizes for it when return value is not used (I don't think that was always the case), so there's no difference in generated code - so that's fair enough, even though objectively I think prefix-increment should be used - it wouldn't justify the extra work involved in merging the code if it doesn't yield any benefit.

I was just browsing the code & this was something that stood out to me. I really like & appreciate this project, my intention was not to troll or waste anyones time, I appologize for that.

doitsujin:
There's no actual "compiler being smart" involved. More like "compiler not being the dumbest piece of shit on the planet". If any compiler actually generates a `mov` instruction for this `i++` case that it doesn't even use the result of, you can safely assume that it cannot be entrusted with any project more complex than a "Hello World" program.

I personally don't like using `++i` unless it's necessary, it's kind of like writing the constant first when comparing integers (i.e. `0 == x`) because you *could* accidentally omit one of the `=` - yes, some people do it, but IMHO it's more a matter of (in)consistency rather than actual merits.

Nisk:
Well, as a programmer by using `i++`, you're telling you want one thing (create a copy, increment original value by one, return copy) - where as you really mean another (just incriment by one). It is the case that the  compiler **is** being smart - as it identifies that what you meant isn't what you actually asked it to do, so it compensates for the programmer. Using prefix, any compiler, however basic would generate the correct (or maybe it's better to say - the desired) behaviour.

I think it's just a matter of treating post & pre increments completely non-interchangeably, not related to any notion of consistency. Probably getting a little off-topic (that's closed anyway), but I think it's the difference between schools of thought - some people effectively say "whatever, the compiler/tech-stack should be smart enough to handle the tedious bits so I can get on with solving actual problems & coding without having to worry about the simple, tedious things" - while the other believes all responsibility (good and bad) lies with the coder. Anyhow, I'm looking forward to the future of this project, keep up the good work.

PR #679, opened 2018-09-30
[build] Indicate the working tree status in the version string.

jp7677:
This adds -dirty to the version string when the working copy was, well, dirty.
See https://git-scm.com/docs/git-describe

Please reject this PR if this would not fit in your own workflow.


jp7677:
@tingvarsson Yes, exactly.
I have local changes in my tree, git status show changes to be committed. Thus:
`git describe` returns `v0.80-25-g0d1c167`. whereas 
`git describe --dirty` returns `v0.80-25-g0d1c167-dirty`.
The output of both commands is identical when there are no local changes.

This PR should indeed have no effect at all on the release builds, assuming they are build with a clean working tree.

jp7677:
@doitsujin I've changed the commit to use + as dirty indicator.

PR #683, opened 2018-10-03
[build] Use -mtune=core2 to generate better assembly

ryao:
This gives me roughly a 10% improvement in FPS on the `Rise of Nations: Extended Edition` title screen according to the DXVK HUD. This game is CPU bound when vsync is off with a single thread using 100% of the CPU. It also is a 32-bit title and uses Direct3D 11. I tested on a Xeon E5-1650 v2, which is an Ivy Bridge processor. Having the compiler tune for the core2 should give a generic performance improvement on modern processors.

### DIFF
--- a/build-win32.txt
+++ b/build-win32.txt
@@ -6,8 +6,8 @@ strip = 'i686-w64-mingw32-strip'
 exe_wrapper = 'wine'
 
 [properties]
-c_args=['-msse', '-msse2']
-cpp_args=['-msse', '-msse2']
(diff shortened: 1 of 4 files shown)


ryao:
@lieff You are thinking about `-march=core2`. `-mtune=core2` just tries to do things like more intelligent instruction scheduling.

Anyway, I tested this in Unigine Valley and had only a 0.1fps improvement, so there isn't much point to making this change. I am going to withdraw it.

PR #684, opened 2018-10-05
winelib fixes

dhewg:
fix one gcc8 compiler warning
improve `setup_dxvk.sh` so it works for e.g. `meson -Dbindir=bin32 -Dlibdir=lib32`


ssorgatem:
Aren't these 2 unrelated changes?

dhewg:
To some degree, it's just more likely to run into that compiler warning with a winelib build. I didn't think a 2nd PR was justified just for that.

dhewg:
Sounds good to me. I dropped the major/minor patch here and created #689 to drop the unused version info.

dhewg:
Anything I should improve to get this small patch in?

PR #702, opened 2018-10-15
[util] Spoof AMD GPU for SAO FB

zaps166:
Improves performance by not trying to load nvapi.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -81,6 +81,11 @@ namespace dxvk {
       { "dxgi.customVendorId",              "1002" },
       { "dxgi.customDeviceId",              "e366" },
     }} },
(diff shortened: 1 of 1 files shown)


doitsujin:
Github seems to be having some issues right now, I merged this manually.

Anyway, since we're getting more and more games that are problematic due to nvapi, it's probably time to look for a better solution than just adding more and more overrides.

doitsujin:
Github seems to be having some issues right now, I merged this manually.

Anyway, since we're getting more and more games that are problematic due to nvapi, it's probably time to look for a better solution than just adding more and more overrides.

doitsujin:
Merged manually due to github issues. Although we really need to find a proper solution to the nvapi issue.

PR #715, opened 2018-10-19
setup_dxvk: don't install by default

aviau:
Hello!

Not sure if you are willing to break the API here, but allow me to make my case.

I'd like to install this script to the user's path.

Not defaulting to "install" is a great way to educate the user that "reset" is available.

What do you think?

I also think that reset could be renamed to "remove" or "uninstall". What would you think about that? (this is much less important, however).

Cheers,

### DIFF
--- a/utils/setup_dxvk.sh.in
+++ b/utils/setup_dxvk.sh.in
@@ -133,15 +133,15 @@ function createOverride {
 }
 
 case "$1" in
-reset)
+uninstall)
     fun=removeOverride
     ;;
-'')
+install)
     fun=createOverride
     ;;
 *)
     echo "Unrecognized option: $1"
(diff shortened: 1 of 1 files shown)


doitsujin:
Breaking the script should be fine since most people won't ever use it anyway (the release package doesn't include it, Proton doesn't need it, Lutris has its own solution, etc.), but I'm not sure if this particular change is actually going to change the DXVK experience dramatically for the better, there might be slightly more concerning issues to deal with.

Honestly though, I don't really care, while you're at it you might as well rename the `reset` option to something more meaningful and I might just accept this PR.

aviau:
> while you're at it you might as well rename the reset option to something more meaningful

Done! I renamed it to uninstall.

doitsujin:
Merged manually due to github issues.

PR #716, opened 2018-10-19
setup_dxvk.sh: support WINE env variable

aviau:
WINEPREFIX is in caps, it might be a little bit more user friendly to support WINE too for users that want to override it.

### DIFF
--- a/utils/setup_dxvk.sh.in
+++ b/utils/setup_dxvk.sh.in
@@ -20,7 +20,9 @@ if [ ! -f "$dlls_dir/d3d11.$dll_ext" ] || [ ! -f "$dlls_dir/dxgi.$dll_ext" ]; th
 fi
 
 if [ -z "$wine" ]; then
-    if [ $build_arch == "x86_64" ]; then
+    if [ -n "$WINE" ]; then
+        wine="$WINE"
+    elif [ $build_arch == "x86_64" ]; then
         wine="wine64"
     else
         wine="wine"


shmerl:
In such case, you better use standard Wine variable, such as `WINEVERPATH`.

When I need to use a custom Wine with setup_dxvk.sh, I do it as follows:

```
export WINEVERPATH=${WINEVERPATH:-"/opt/mywine"}
export PATH=${WINEVERPATH}/bin:$PATH 
export WINESERVER=${WINEVERPATH}/bin/wineserver
export WINELOADER=${WINEVERPATH}/bin/${wine_bin}
export WINEDLLPATH=${WINEVERPATH}/lib/wine/fakedlls
export LD_LIBRARY_PATH="${WINEVERPATH}/lib:${LD_LIBRARY_PATH}"
````

dhewg:
If that is indented for your debian packages (and to work around the weird suffixes a la `wine-development` it comes with), there's an easy way out without this patch: `export PATH=/usr/lib/wine-development:$PATH`
The standard `wine`, `wine64` etc binaries are installed there, so you can target specific packages without propagating non-standard suffix hacks to other projects ;)

Thaodan:
But $WINE is a standard env var at least.

shmerl:
Is it? Not according to [this](https://wiki.winehq.org/FAQ#Can_I_install_more_than_one_Wine_version_on_my_system.3F) at least. It specifies `$WINEVERPATH` and `$WINELOADER` as key variables (with a few more supporting ones).

Thaodan:
Not according to that but major 
tools like winetricks support 
$WINE.

shmerl:
Sure, I mean that's internal winetricks thing, not standard Wine variable. That was my point.

doitsujin:
Agree with shmerl on this one, there are ways to handle this already without messing with the script.

PR #720, opened 2018-10-21
Fix readme

nextgenthemes:
update-alternatives commands I ironically found in the proton readme. Was pointed out in various issues b4 that the 2 commands where not enough, this even skips the need for further user interaction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ For the current status of the project, please refer to the [project wiki](https:
 
 
 ## How to use
-In order to install DXVK run the following commands. `~/.wine` stands for the default wine prefix in your home folder you may change that to another path if you know what you are doing. Also be sure the version is up to date with version on top of the [release page](https://github.com/doitsujin/dxvk/releases). If its not please do a pull request for this readme to change it.
+In order to install latest realease of DXVK run the following commands. `~/.wine` stands for the default wine prefix in your home folder you may change that to another path if you know what you are doing.
 ```
 mkdir dxvk-download
 cd dxvk-download
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ For the current status of the project, please refer to the [project wiki](https:
 
 
 ## How to use
-In order to install latest realease of DXVK run the following commands. `~/.wine` stands for the default wine prefix in your home folder you may change that to another path if you know what you are doing.
+In order to install the latest realease of DXVK run the following commands. `~/.wine` stands for the default wine prefix in your home folder you may change that to another path if you know what you are doing.
 ```
 mkdir dxvk-download
 cd dxvk-download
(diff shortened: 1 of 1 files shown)


doitsujin:
I don't understand your problem. The README is supposed to tell people how to set up DXVK, not how to use `tar`.

doitsujin:
There are good reasons to keep the files around. If you have multiple wine prefixes or need to set up a new one, there's no reason to re-download the files every single time, and you most certainly need to adjust your `WINEPREFIX` rather than having it set to `~/.wine`. This is actually a very common use case that your instructions neglect.

nextgenthemes:
Well I am not sure how common that really is putting in the default does not mean its unchangeable. I actually can still think like a noob. If someone maybe new to linux/wine just wants to get his games going they do not want to find out what a wineprefix even is. I never used another one then the default. I guess with DXVK it would make sense create a new one without DXVK if something fails but I will still use the default for DXVK.

The actual important part is actually be bottom part if the readme that you completely ignored. instead you claim I would have a 'problem' with the top.

The top is just me obsessing about ease of use and automation. I actually have figured out how to install the latest without the need to change anything now. The extra folder may be removed but then people will end up with x32 and x64 folders in their home or wherever they are without knowing where they belong to.

The /tmp version would be this, I like it more:

```
WINEPREFIX=~/.wine
DXVK_LATEST_TAR=$(curl --silent "https://api.github.com/repos/doitsujin/dxvk/releases" | grep '"browser_download_url":' | head -1 | sed -E 's/.*"([^"]+)".*/\1/')
wget -qO- $DXVK_LATEST_TAR | tar xvz --transform 's/^dbt2-0.37.50.3/dbt2/' --strip-components=1 -C /tmp
winetricks --force /tmp/setup_dxvk.verb
```

ssorgatem:
Well, IMHO, a noob should either try to understand what they are doing (and those commands don't exactly help, rather the opposite; they can be rather obscure even for experienced users) or... you know, use some of the solutions aimed at them, like Lutris,  PlayOnLinux or Steam Play.

So if there's a noob that doesn't want to understand how to setup DXVK or how wine prefixes work (and it's not that hard to grasp; I was playing with wineprefixes long before I even knew what was "wget" or a pipe or git).... what exactly are they doing here? 

That aside, there is no good reason to install DXVK to /tmp (and regardless it's a bad practice hardcoding that to a fixed path). If a noob can't figure out how to take advantadge of that to set up DXVK on several wineprefixes... really, they should be using some of the helpers previously mentioned instead. That's exactly what they are for.

You are not sure how common it is to have several wineprefixes...? Well, it's the recommended practice by Wine: one prefix for every app you want to install. It has many advantadges, and I'd say it's the most common use case by far. That's also what the helpers do.
If you try to install everything on one wineprefix, things will start conflicting and fail.
So that's what you'd keep DXVK around and not just the dlls copied into a wineprefix.
Having to download everything again just to test enabling/disabling DXVK is just silly.

PR #729, opened 2018-10-24
[Untested] [Feedback Wanted] Add option on whether or not to advertise command list support

Guy1524:
In the VKx discord, I heard that DXVK has to emulate deferred contexts, and that games which use them are slower on DXVK.  Because of this, I thought not advertising support for a hardware implementation may trigger some applications to use an alternative codepath that doesn't involve deferred contexts.  I haven't tested this PR with any applications yet, but I would like to know whether I'm doing anything wrong before I proceed.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -1022,10 +1022,44 @@ namespace dxvk {
           UINT                              CopyFlags) {
     if (!pDstResource)
       return;
-    
+
     // We need a different code path for buffers
     D3D11_RESOURCE_DIMENSION resourceType;
     pDstResource->GetType(&resourceType);
+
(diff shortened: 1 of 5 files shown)


K0bin:
I also implemented this some time ago as an experiment.

https://github.com/k0bin/dxvk/tree/no-commandlists

I only tested it with Assassin's Creed Origins but that seemed to work.

Guy1524:
Interesting, did you notice any performance improvements?

K0bin:
Not at all. I don't think there's many games that have different render paths based on that option.

doitsujin:
I don't *really* like the idea because I doubt that any sizable number of games actually resort to using just the immediate context when command lists are not supported, and enabling the option will break pretty much everything that continues to rely on deferred contexts due to D3D11 runtime bugs.

Guy1524:
That is understandable, and I agree that the workaround is really hacky.  Closing.

PR #746, opened 2018-11-03
cross build cleanup

dhewg:
Move non-generic stuff out of the meson cross files.

This allows reusing private cross files without dxvk specific switches.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -37,6 +37,7 @@ if dxvk_winelib
   lib_dxgi    = declare_dependency(link_args: [ '-ldxgi' ])
   lib_d3dcompiler_43 = declare_dependency(link_args: [ '-L'+dxvk_library_path, '-ld3dcompiler_43' ])
   lib_d3dcompiler_47 = declare_dependency(link_args: [ '-ld3dcompiler' ])
+  lib_dl = declare_dependency(link_args: [ '-ldl' ])
   exe_ext = '.exe.so'
   dll_ext = '.dll'
   def_spec_ext = '.spec'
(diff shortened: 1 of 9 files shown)


meson.build
```diff
@@ -25,12 +25,19 @@ else
   dxvk_library_path = meson.source_root() + '/lib32'
 endif
 
-if meson.get_cross_property('winelib', false)
+code = '''#ifndef __WINE__
+#error 1
+#endif'''
```
doitsujin: I'm sure this works, but looks a bit weird, might be better to keep the `winelib` cross-property around.

dhewg: That would defeat the purpose of these changes ;)
But yeah, looks a bit weird. I still think it's better to detect features, maybe we can improve the code chunk instead?

dhewg:
Of course, mingw as well as winelib still build warning free for 32bit and 64bit. Winelib runtime tests still work fine as they did before

meson.build
doitsujin: The reason I think it's awkward is because we explicitly have to enable winelib builds using the cross files anyway, so we might as well tell the build system what we want instead of working around it using some awkward piece of code.

This isn't quite the same as the other changes which mostly avoid setting compiler and linker options in multiple different places.

dhewg: Just to clarify: The motivation is to get rid of all the custom stuff to be able to reuse general cross files.
Like my `.local/share/meson/cross/wine64`, which I can just use with any project using `meson --cross-file=wine64`.

Why, and not just use your bundled version? Because meson doesn't accept `CFLAGS=foo meson` for cross compilation. I have to edit cross files to change any flags...

But okay, let me drop that patch then. I can add `winelib=true' to my file instead, but it'll be project specific and not generic - and everyone has add it.

doitsujin: I wasn't aware of this use case (tbh I *do* find it rather odd and probably no one else is ever going to do it, but whatever), it's probably okay to keep it then.

dhewg: Hehe, ok, I'll add the commit back, thanks!

PR #748, opened 2018-11-04
Fixed for a05c93dd178de722c4e1d257a138fc94c4b877d6

SveSop:
package-release.sh script fails in that it tries to remove non-existing wine folder (since this is no longer creater after a05c93dd178de722c4e1d257a138fc94c4b877d6)


doitsujin:
Good catch, thanks.

PR #751, opened 2018-11-05
[dxvk] Use more threads for pipeline compilation

pchome:
This will report 2 worker threads for 3 core CPUs.

Also, changes behaviour for 4, 6 and 8 cores/threads, and for cases when
non-power-of-2 concurrent threads reported (`taskset`,`cgroups`, ... ?).

### DIFF
--- a/src/dxvk/dxvk_state_cache.cpp
+++ b/src/dxvk/dxvk_state_cache.cpp
@@ -57,11 +57,9 @@ namespace dxvk {
         writeCacheEntry(file, e);
     }
 
-    // Use half the available CPU cores for pipeline compilation
+    // Use 3/4 the available CPU cores for pipeline compilation
     uint32_t numCpuCores = dxvk::thread::hardware_concurrency();
-    uint32_t numWorkers  = numCpuCores > 8
-      ? numCpuCores * 3 / 4
-      : numCpuCores * 1 / 2;
+    uint32_t numWorkers  = numCpuCores * 3 / 4;
 
     if (numWorkers <  1) numWorkers =  1;
     if (numWorkers > 16) numWorkers = 16;


kakra:
Maybe it would be better if DXVK supported a taskset configuration variable to offload shader compilation to CPU cores not used by the rendering engine?

Also, how does it affect perceived game performance for different CPU scenarios? I'm having a 8-thread CPU which would make 6 compiler threads instead of 4. This would put more load during initial shader compilation (which would probably be fine) but also during running the game, stealing resources from other game threads (which would not block on shader compilation). Would it be a gain or loss in this case?

doitsujin:
There's a good reason why the number of threads used for compilation on low core count CPUs is low. When compiling shaders, they can be loaded 100% for extended periods of time, which can lead to severe stuttering and low FPS when it occupies too many cores. Even on my 16-thread CPU the drop is quite noticeable in games like Shadow of the Tomb Raider.

As for taskset-like functionality, I don't like the idea, games can by default use all available cores so the only thing DXVK could do would to restrict its own threads to specific cores, which makes no sense. Also, wine AFAIK doesn't currently support the APIs required for that, and the same goes for thread priority support (lowering thread priority would be possible, but as of right now it just isn't going to work).

I can add a config option for the thread count, but I believe the default behaviour should remain as it is.

kakra:
Actually, wineserver supports setting thread affinity. But games only use "SetThreadIdealProcessor" and that is a no-op in current wine. I partly implemented that by letting SetThreadIdealProcessor use CPU affinity, some game engines explicitly restrict some threads to one specific core. But I think that implementation is somewhat incorrect as Windows only uses it as a hint while setting affinity is a policy to follow. But it seems to do its job. The Linux scheduler should automatically prefer other cores for the remaining processes.

If DXVK supported anything like that, it should probably not do so by default but only by explicit request from the user. That way, the user could run the game on one set of cores and DXVK would compile on another set. This may make sense for AMD dual-die CPUs where it is known to get better performance when restricting a game to one CPU die. In that case, DXVK could stop calculating the number of threads and just use what the user provided with the configuration - accepting that a user shoots its own foot.

kakra:
@doitsujin BTW:
https://github.com/wine-mirror/wine/blob/b78de971f674f7be8b078f0e7f14112a9a892f33/dlls/kernel32/thread.c#L477

It successfully goes through wineserver why submitting a proper request (using `NtSetInformationThread()`), that finally results in setting the affinity via a Linux syscall:
https://github.com/wine-mirror/wine/blob/b78de971f674f7be8b078f0e7f14112a9a892f33/server/thread.c#L463

pchome:
@doitsujin 
Yes, please add a config option.

kakra:
@doitsujin Here's a simple and hacky proof-of-concept:
https://github.com/kakra/dxvk/commit/c08b9da839b7d4b2bdae436ccfae2292bb4e4922

It doesn't really improve anything in-game as far as I could test yet. But I no longer have audio drop-outs during loading screens which is a somewhat nice side-effect. It's only compatible with winelib builds because wine still misses the background priority API for threads. If it had, we could exploit that. SCHED_BATCH seems to most closely match what Windows apps would expect from background scheduling.

kakra:
Here's a test build including the SCHED_BATCH hack if anyone cares to test:
https://keybase.pub/hurikhan77/wine-proton-unofficial/dxvk-dist-wine-3.19-580-g723d54bf655.tar.xz

You need to set `"PROTON_USE_WINED3D": "1"` because DXVK is included as a winelib-builtin. Extract this to your Proton 3.16 folder (do not clear the folder, just extract into the existing folder, you may want to make a backup first).

Issue #706, opened 2018-10-16
[winelib] Building dxvk like a native library?


shmerl:
Proton changelog mentions:

> DXVK is now built as a native Linux library, which may give a small performance boost, and should make debugging easier for DXVK and driver developers.

May be this should be added as an option to build and installation instructions?

pchome:
1. Use corresponding cross-file
2. Done

shmerl:
> corresponding cross-file

Is it `build-wine64.txt`?

pchome:
And `build-wine32.txt`, yes.

shmerl:
Either way, adding it to explicit instructions would make it more clear. And how exactly is it used in such case if it's not a dll

Guy1524:
@shmerl Proton probably has a custom way of installing it, DXVK's winetricks verb only works with the mingw build, so you'll have to put the dll.so in either your wine library folder, your prefix's system library folder, or your game's exe folder.  Then set d3d11 and dxgi to builtin.

shmerl:
OK. I'll give it a try. That's why instructions for this can be useful. May be another section on the Wiki. I can make something once I'll make it work.

pchome:
Proton way:
```sh
env["WINEDLLPATH"] = lib64dir + "/wine:" + libdir + "/wine"
```
and
```sh
if "wined3d11" in config_opts:
    env["WINEDLLPATH"] = lib64dir + "/wine/wined3d:" + libdir + "/wine/wined3d:" + env["WINEDLLPATH"]
else:
    env["WINEDLLPATH"] = lib64dir + "/wine/dxvk:" + libdir + "/wine/dxvk:" + env["WINEDLLPATH"]
```

shmerl:
So simply adding a path where that .so resides to `WINEDLLPATH` should work?

pchome:
https://github.com/doitsujin/dxvk/pull/607#issuecomment-417456171
I personally prefer hack+overrides. 

For `WINEDLLPATH` to work you probably should split WINE built-in *.so the same way Proton does this.
For hack -- add all DXVK's `d*` libs to `WINEDLLOVERRIDES` and set them to either `b` - to use wine's libs, or `n` - to use DXVK's ones.

shmerl:
Ah, good to know that setup_dxvk.sh already handles that, I missed it since I've never tired building it that way yet.

SveSop:
[...]

PR #769, opened 2018-11-16
Add script to automate build winelib due to Ubuntu 18.04/18.10 missing mingw-w64 6.0

SveSop:
Snagged the script from [https://github.com/Tk-Glitch/PKGBUILDS/blob/master/wine-tkg-git/wine-tkg-patches/dxvk-winelib.patch](https://github.com/Tk-Glitch/PKGBUILDS/blob/master/wine-tkg-git/wine-tkg-patches/dxvk-winelib.patch) and modified it slightly to:
1. Rename the .dll.so's to simple dll's
2. added --libdir=lib to the buildoption, cos Ubuntu automagically puts them in x86_64-linux-gnu when not specified.

Not sure if there is any "problem" to using winelib dll's renamed like this, other than it is NOT working for World of Warcraft for some reason. The error from the WoW client is "could not loadlibrary d3d11.dll". Other games MAY also suffer from this, but easy enough for an optional script for those not able to build with mingw-w64 6.0. (Like Ubuntu 18.04/18.10)

PS. World of Warcraft DO work with this if you create symlinks as usual - remove the dll-redirects in winecfg AND copy the the dll's as dll.so's overwriting the original wine d3d1*.dll.so's (and dxgi.dll.so).

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -1671,26 +1671,27 @@ namespace dxvk {
 
 
 
-  WineDXGISwapChainFactory::WineDXGISwapChainFactory(IUnknown* pContainer)
-  : m_container(pContainer) {
+  WineDXGISwapChainFactory::WineDXGISwapChainFactory(
+          IDXGIVkPresentDevice*   pDevice)
+  : m_device(pDevice) {
     
   }
   
   
   ULONG STDMETHODCALLTYPE WineDXGISwapChainFactory::AddRef() {
-    return m_container->AddRef();
+    return m_device->AddRef();
   }
   
   
   ULONG STDMETHODCALLTYPE WineDXGISwapChainFactory::Release() {
-    return m_container->Release();
+    return m_device->Release();
   }
   
   
   HRESULT STDMETHODCALLTYPE WineDXGISwapChainFactory::QueryInterface(
           REFIID                  riid,
           void**                  ppvObject) {
-    return m_container->QueryInterface(riid, ppvObject);
+    return m_device->QueryInterface(riid, ppvObject);
   }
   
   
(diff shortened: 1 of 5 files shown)


ssorgatem:
Why rename the winelib build to .dll?
It's a .so, not a .dll.
You should only rename it to .dll if copying it directly inside a wineprefix; otherwise just symlinking as a .dll works just fine.

SveSop:
It has been debated a wee bit here and there, and the main reason was to do a easy in-place replacement for already created prefixes - Like the steam prefixes where the symlinks points to *.dll files.

In other words, if you were to compile as winelib (for those with ubuntu 18.04/18.10 that does not use a mingw-w64 PPA), you COULD use that to replace steam's proton dxvk's, as im not 100% sure the proton wineprefix symlinks won't be updated by a script if "not correct"? (Probably not tho..)

I still have issues with winelib build and World of Warcraft where it will bug out with a "loadlibrary" error, if not BOTH symlinked AND hard-replaced wine's own dll.so files for some reason. (Suspect missing GNU symbols or something.. dunno)
See also line 4 in src/dxgi/dxgi_include.h
```
//for some reason we need to specify __declspec(dllexport) for MinGW
#if defined(_MSC_VER) || defined(__WINE__)
  #define DLLEXPORT
#else
  #define DLLEXPORT __declspec(dllexport)
#endif
```
Maybe d3d11.dll needs this (dllexport) function for WoW to work? This function in itself is afaik "not supported" in wine : 

> Bugs
> The dllimport/dllexport attributes are not supported at the moment, due to lack of support for these features in the ELF version of gcc.

From [https://linux.die.net/man/1/winegcc](https://linux.die.net/man/1/winegcc)

Not sure why wined3d works, and a mingw build of dxvk works as its supposed.. but nevertheless.

[https://github.com/doitsujin/dxvk/issues/706](https://github.com/doitsujin/dxvk/issues/706)

PR #797, opened 2018-12-03
[dxgi] Check fullscreen state before using m_monitor

zzhiyi:
m_monitor is nullptr when it's not in fullscreen state.

Signed-off-by: Zhiyi Zhang <zzhang@codeweavers.com>

### DIFF
--- a/src/dxgi/dxgi_swapchain.cpp
+++ b/src/dxgi/dxgi_swapchain.cpp
@@ -46,7 +46,7 @@ namespace dxvk {
   DxgiSwapChain::~DxgiSwapChain() {
     Com<IDXGIOutput> output;
     
-    if (SUCCEEDED(m_adapter->GetOutputFromMonitor(m_monitor, &output)))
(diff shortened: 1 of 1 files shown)


doitsujin:
What is this supposed to fix? `GetOutputFromMonitor` fails with an error code if the monitor handle is invalid, so this check is redundant.

zzhiyi:
In GetOutputData you used std::unordered_map::find( const Key& key ), if key is nullptr, then it will crash. I have a demo that ran into this.

doitsujin:
Do you have a code sample to reproduce that? Passing a null pointer as a key to a hash map is legal and works fine in my tests, so I'd be interested to see what is actually causing that crash.

zzhiyi:
I was using this https://www.dropbox.com/s/stl4eraf268b2ln/directx_demo.zip?dl=0

doitsujin:
How do you reproduce the crash with this demo? Just closing it works fine for me.

PR #833, opened 2018-12-28
Lock the d3d10 device during presentation

K0bin:
The D3D10Device uses a mutex to avoid accessing the D3D11ImmediateContext from multiple threads at the same time. The new D3D11Swapchain needs to be locked as well because with D3D10 it's not guaranteed that the game only accesses the ImmediateContext from the render thread.

This fixes Just Cause 2 for me. I haven't done that much testing, just 5 minutes of running around an shooting things.

In theory D3D11VkInterop has the same issue because it also accesses the ImmediateContext.
But that is only used with VR and thus D3D11, right?

### DIFF
--- a/src/d3d10/d3d10_device.cpp
+++ b/src/d3d10/d3d10_device.cpp
@@ -8,9 +8,7 @@ namespace dxvk {
   D3D10Device::D3D10Device(
           D3D11Device*                      pDevice,
           D3D11ImmediateContext*            pContext)
-  : m_device(pDevice), m_context(pContext),
-    m_multithread(this, !(pDevice->GetCreationFlags() & D3D10_CREATE_DEVICE_SINGLETHREADED)) {
-    
+  : m_device(pDevice), m_context(pContext) {
   }
 
   
@@ -716,8 +714,6 @@ namespace dxvk {
   void STDMETHODCALLTYPE D3D10Device::ClearRenderTargetView(
           ID3D10RenderTargetView*           pRenderTargetView,
     const FLOAT                             ColorRGBA[4]) {
-    D3D10DeviceLock lock = LockDevice();
-
     D3D10RenderTargetView* d3d10View = static_cast<D3D10RenderTargetView*>(pRenderTargetView);
     D3D11RenderTargetView* d3d11View = d3d10View ? d3d10View->GetD3D11Iface() : nullptr;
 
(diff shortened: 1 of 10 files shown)


doitsujin:
Interesting, although I'm not sure if this is really the best way to fix this since this thread safety issue is not necessarily exclusive to D3D10. Can you test if this patch works as well?

[dxvk-sync-cs-thread.patch.txt](https://github.com/doitsujin/dxvk/files/2714788/dxvk-sync-cs-thread.patch.txt)

K0bin:
I changed the PR so D3D10Device just enables Multithreaded protection on the contexts ID3D11Multithread.

src/d3d10/d3d10_device.cpp
```diff
@@ -8,9 +8,8 @@ namespace dxvk {
   D3D10Device::D3D10Device(
           D3D11Device*                      pDevice,
           D3D11ImmediateContext*            pContext)
-  : m_device(pDevice), m_context(pContext),
-    m_multithread(this, !(pDevice->GetCreationFlags() & D3D10_CREATE_DEVICE_SINGLETHREADED)) {
-    
+  : m_device(pDevice), m_context(pContext) {
+    m_context->GetMultithread()->SetMultithreadProtected(!(pDevice->GetCreationFlags() & D3D10_CREATE_DEVICE_SINGLETHREADED));
```
doitsujin: This will enable multithreaded protection and thus cause a ~20% performance hit for **all** D3D11 apps as well, since the d3d10 device is always created. The correct fix would be to do this in `D3D10CoreCreateDevice`.

src/d3d10/d3d10_main.cpp
```diff
@@ -41,6 +41,10 @@ extern "C" {
     hr = D3D11CoreCreateDevice(pFactory, pAdapter,
       Flags, &FeatureLevel, 1, &d3d11Device);
     
+    Com<ID3D10Multithread> multithread;
+    d3d11Device->QueryInterface(__uuidof(ID3D10Multithread), reinterpret_cast<void**>(&multithread));
+    multithread->SetMultithreadProtected(!(Flags & D3D10_CREATE_DEVICE_SINGLETHREADED));
+
     if (FAILED(hr))
       return hr;
```
doitsujin: The `if (FAILED(hr))` check needs to be moved up before setting up the MT protection, otherwise things will crash if device creation fails (and it can legitimately fail).

PR #863, opened 2019-01-15
[d3d11] Fix bad check in DrawInstancedIndirect

michalsrb:
The code in DrawInstancedIndirect would reuse the cmdData if it was of type
DrawIndirectIndexed, which does not look correct. It looks like copy-paste
mistake - the same code is in DrawIndexedInstancedIndirect.


doitsujin:
Good catch, thanks.

PR #878, opened 2019-01-21
[d3d11] In D3D11CreateDevice, create DXGI factory with IDXGIFactory1 support

aeikum:
Fixes running the Resident Evil 2 demo against Wine's DXGI.

Tests are here: https://source.winehq.org/git/wine.git/blob/b353d7c9148414be7c928e3925f4075b1578c0a8:/dlls/d3d11/tests/d3d11.c#l2019

### DIFF
--- a/src/d3d11/d3d11_main.cpp
+++ b/src/d3d11/d3d11_main.cpp
@@ -118,7 +118,7 @@ extern "C" {
         Logger::warn("D3D11CreateDevice: Unsupported driver type");
       
       // We'll use the first adapter returned by a DXGI factory
-      if (FAILED(CreateDXGIFactory(__uuidof(IDXGIFactory), reinterpret_cast<void**>(&dxgiFactory)))) {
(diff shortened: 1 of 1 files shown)


aeikum:
Re-reading the docs at https://docs.microsoft.com/en-us/windows/desktop/api/dxgi/nf-dxgi-createdxgifactory1 , perhaps we should change all instances of IDXGIFactory in d3d11/ to IDXGIFactory1.

Joshua-Ashton:
I'm assuming the crash is related to this behavior?:
https://gist.github.com/Joshua-Ashton/39da9a52e1a1f807750afbdcf284fc37

(there are some weird quirks like querying Adapter1 from an Adapter0 from a Factory0 working though :man_shrugging:)

This would make kinda sense would make sense as CreateDXGIFactory1 is backwards compat but CreateDXGIFactory is not forwards compat (via. QueryInterface).

I am curious as to why it is failing though in that game -- could you go into more details for me as to why it does and what it's calling, more-so for my curiosity than anything as I don't think the game could get its hand on an auto generated adapter?

Issue #820, opened 2018-12-18
Better deployment of dll for setup_dxvk.sh


shmerl:
To make setup_dxvk.sh more unified (i.e. one script for all scenarios), it can be useful to rearrange dlls deployment a bit. For example:

```
./bin
./bin/setup_dxvk.sh
./lib
./lib/win64/dxgi.dll
./lib/win64/d3d11.dll
./lib/win64/d3d10core.dll
./lib/win64/d3d10.dll
./lib/win64/d3d10_1.dll
```

This will allow cleaner placement of 32-bit dlls. While .so ones can be deployed cleanly now in the same parent location, dlls are not since they are placed in bin.

I.e. adding this would make it work:

```
./lib/win32/dxgi.dll
./lib/win32/d3d11.dll
./lib/win32/d3d10core.dll
./lib/win32/d3d10.dll
./lib/win32/d3d10_1.dll
```
If you think it's a good idea, I can update my PR to do that.

doitsujin:
Why would you want to rearrange the files, and what's the fuss about something that's meant for **development** purposes? Remember that every change to this also changes my workflow.

doitsujin:
If the debian package actually ships the .sh script like that then someone should tell them that this is a bad idea. If they want DXVK to be set up via symlinks, which is of course a good thing, they should probably put a small script in themselves that points to the correct locations, but this is distro-specific.

I can also see why the current script is less than ideal, but it's not an easy fix since it is being generated during the build process, i.e. once for the 64-bit build and once for the 32-bit build.

shmerl:
> I can also see why the current script is less than ideal, but it's not an easy fix since it is being generated during the build process, i.e. once for the 64-bit build and once for the 32-bit build.

That can be changed by making the script static, and more generic without templatizing it. Kind of the idea I was going in the PR with external variables controlling the changing parts.

But sure. May be simply writing an alternative script will work as well. I can propose it to Debian maintainers or write one myself first.

doitsujin:
> That can be changed by making the script static (not generated), and more generic without templatizing it in the build.

Yes, but then you basically can't do `ninja install` and have a working script either.

shmerl:
By the way, why after building 32-bit / winelib variant, ninja install produces `.a` files in 64-bit subdirectory?

```
./lib/x86_64-linux-gnu
./lib/x86_64-linux-gnu/dxgi.dll.a
./lib/x86_64-linux-gnu/d3d11.dll.a
[...]

PR #892, opened 2019-01-28
Add support for custom dlldir option

shmerl:
Based on suggestions in https://github.com/doitsujin/dxvk/issues/820

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,1 +1,2 @@
 option('enable_tests', type : 'boolean', value : false)
+option('utils_dir', type : 'string', value : '')
--- a/utils/meson.build
+++ b/utils/meson.build
@@ -3,9 +3,15 @@ conf.set('bindir', get_option('bindir'))
 conf.set('libdir', get_option('libdir'))
 conf.set('arch', target_machine.cpu_family())
 conf.set('winelib', dxvk_winelib)
+
+utils_dir = get_option('utils_dir')
+if utils_dir == ''
+  utils_dir = get_option('bindir')
+endif
+
 configure_file(
   configuration : conf,
   input         : 'setup_dxvk.sh.in',
   output        : 'setup_dxvk.sh',
-  install_dir   : get_option('bindir')
+  install_dir   : utils_dir
 )


doitsujin:
Again, *why*? Why do we need to clobber the build system with useless options? What is even a use case for this?

Thaodan:
Maybe they shouldn't look for them in bin then? changing something so that it works doesn't always make it right.
For example you can set $libdir to $bindir at build time.

shmerl:
> Maybe they shouldn't look for them in bin then? changing something so that it works doesn't always make it right.

Well, that's exactly the point. Meson treats dlls as something that goes to bin by default. It's just neater to place them in libdir, at least optionally. Since some current tools rely on placing them in bin, the above change keeps current behavior as is, just provides an option to change it.

SveSop:
> Again, _why_? Why do we need to clobber the build system with useless options? What is even a use case for this?

The use case is obvious tho - Having a flag to set the install dir.

Why have the option to set this? Would it not be a better solution for the package-release.sh script to NOT have to use `cp` to copy files from the `install.$1` but rather just `ninja install` them in the right place to begin with?
Currently:
 ```
meson --cross-file "$DXVK_SRC_DIR/build-win$1.txt"  \
        --buildtype "release"                         \
        --prefix "$DXVK_BUILD_DIR/install.$1"         \
        --strip                                       \
        -Denable_tests=false                          \
        "$DXVK_BUILD_DIR/build.$1"

  cd "$DXVK_BUILD_DIR/build.$1"
  ninja install

  mkdir "$DXVK_BUILD_DIR/x$1"

  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10.dll" "$DXVK_BUILD_DIR/x$1/d3d10.dll"
  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10_1.dll" "$DXVK_BUILD_DIR/x$1/d3d10_1.dll"
  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10core.dll" "$DXVK_BUILD_DIR/x$1/d3d10core.dll"
  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d11.dll" "$DXVK_BUILD_DIR/x$1/d3d11.dll"
  cp "$DXVK_BUILD_DIR/install.$1/bin/dxgi.dll" "$DXVK_BUILD_DIR/x$1/dxgi.dll"
  
  rm -R "$DXVK_BUILD_DIR/build.$1"
  rm -R "$DXVK_BUILD_DIR/install.$1"
```
vs.
```
meson --cross-file "$DXVK_SRC_DIR/build-win$1.txt"  \
        --buildtype "release"                         \
        --prefix "$DXVK_BUILD_DIR"	              \
        -Dlib_dir="x$1"			      \
	--strip                                       \
        -Denable_tests=false                          \
        "$DXVK_BUILD_DIR/build.$1"

  cd "$DXVK_BUILD_DIR/build.$1"
  ninja install

rm -R "$DXVK_BUILD_DIR/build.$1"
```

**Comment by pchome** [comment]
vs
```bash
cp "$DXVK_BUILD_DIR/install.$1/bin/"*.dll "$DXVK_BUILD_DIR/x$1/" \
&& rm -R "$DXVK_BUILD_DIR/build.$1" "$DXVK_BUILD_DIR/install.$1"
```

shmerl:
How to do it externally is really not that important vs not doing it all if build can handle it.

SveSop:
Nah, i much more prefer doing 
`cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10.dll" "$DXVK_BUILD_DIR/x$1/d3d10.dll" && cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10_1.dll" "$DXVK_BUILD_DIR/x$1/d3d10_1.dll" && cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10core.dll" "$DXVK_BUILD_DIR/x$1/d3d10core.dll" && cp "$DXVK_BUILD_DIR/install.$1/bin/d3d11.dll" "$DXVK_BUILD_DIR/x$1/d3d11.dll" && cp "$DXVK_BUILD_DIR/install.$1/bin/dxgi.dll" "$DXVK_BUILD_DIR/x$1/dxgi.dll" && rm -R "$DXVK_BUILD_DIR/build.$1" && rm -R "$DXVK_BUILD_DIR/install.$1" && echo "----" && echo "DONE" && echo "----"`
by hand every time... manually.. dont even use a script! Am i ready for Gentoo soon?

Jokes aside: This is not about "changing the script", but to provide a option that makes it easier for those that uses custom scripts. And the argument is "why".

pchome:
@shmerl 
> How to do it externally is really not that important vs not doing it all if build can handle it.

I cant disagree, see #618.
But in real world we should do 
```
patch -p1 < dxvk-fix-build.patch
meson --cross-file ...
```

shmerl:
This improves flexibility, and is backwards compatible. So I see a point. That answers the question "what's the point".

SveSop:
> @shmerl
> 
> > How to do it externally is really not that important vs not doing it all if build can handle it.
> 
> I cant disagree, see #618.
> But in real world we **could** do
> 
> ```
> patch -p1 < dxvk-fix-build.patch
> meson --cross-file ...
> ```

Did i fix it?

I do `../build_fix.sh` and all patches i use are applied and ready to build. Kinda besides the point tho :)

doitsujin:
> Why have the option to set this? Would it not be a better solution for the package-release.sh script to NOT have to use `cp` to copy files from the `install.$1` but rather just `ninja install` them in the right place to begin with?

This is actually the first somewhat valid reason for making changes to the build system, even though it's purely cosmetic, but I don't see myself accepting it in its current form. I **especially** don't see any reason to accept anything that touches *every single* `meson.build` file in the project just to add *one single* option, this is ludicrous.

If someone comes up with a clean solution that a) doesn't break everyone's workflow and b) doesn't make it impossible to maintain the buildsystem, then I'm fine with such a change.

@shmerl
> Having something like `get_option_default('libdir')` could simplify this

That command does not exist, what is it supposed to do?

SveSop:
> > Why have the option to set this? Would it not be a better solution for the package-release.sh script to NOT have to use `cp` to copy files from the `install.$1` but rather just `ninja install` them in the right place to begin with?
> 
> This is actually the first somewhat valid reason for making changes to the build system, even though it's purely cosmetic, but I don't see myself accepting it in its current form. I **especially** don't see any reason to accept anything that touches _every single_ `meson.build` file in the project just to add _one single_ option, this is ludicrous.
> 
> If someone comes up with a clean solution that a) doesn't break everyone's workflow and b) doesn't make it impossible to maintain the build system, then I'm fine with such a change.

For pure "cosmetic fixup" of the build-script, i guess one could do:
```
diff --git a/package-release.sh b/package-release.sh
index 832f553..3b18354 100755
--- a/package-release.sh
+++ b/package-release.sh
@@ -25,24 +25,17 @@ function build_arch {
 
   meson --cross-file "$DXVK_SRC_DIR/build-win$1.txt"  \
         --buildtype "release"                         \
-        --prefix "$DXVK_BUILD_DIR/install.$1"         \
-        --strip                                       \
+        --prefix "$DXVK_BUILD_DIR"                    \
+        --bindir "x$1"                                \
+       --strip                                       \
         -Denable_tests=false                          \
         "$DXVK_BUILD_DIR/build.$1"
 
   cd "$DXVK_BUILD_DIR/build.$1"
   ninja install
 
-  mkdir "$DXVK_BUILD_DIR/x$1"
-
-  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10.dll" "$DXVK_BUILD_DIR/x$1/d3d10.dll"
-  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10_1.dll" "$DXVK_BUILD_DIR/x$1/d3d10_1.dll"
-  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d10core.dll" "$DXVK_BUILD_DIR/x$1/d3d10core.dll"
-  cp "$DXVK_BUILD_DIR/install.$1/bin/d3d11.dll" "$DXVK_BUILD_DIR/x$1/d3d11.dll"
-  cp "$DXVK_BUILD_DIR/install.$1/bin/dxgi.dll" "$DXVK_BUILD_DIR/x$1/dxgi.dll"
-  
   rm -R "$DXVK_BUILD_DIR/build.$1"
-  rm -R "$DXVK_BUILD_DIR/install.$1"
+  rm -R "$DXVK_BUILD_DIR/lib"
 }
 
 function build_verb {
```
Dll's generated directly in the x32/x64 folder + generate the `setup_dxvk.sh` script with "arch" specific `build_arch='@arch@'` statement. Probably the original idea of generating the `setup_dxvk.sh` script in the first place? Dunno. Should perhaps do a double-check that the scripts actually do work first, as i am not sure about the reasons for not generating it? (For all i know the `package_release.sh` script could have been made the way it is today cos the `setup_dxvk.sh` did more harm than good after the winetricks verb was put in).

SveSop:
Not entirely loving the default setup_dxvk.sh.in script as it is tho, even tho it seems to work. Probably should be tweaked somewhat, since it seems to check for a winelib build aswell, and possibly is made for those cases where you manually compile using meson and ninja install. (And thus have the lib/x86_64-linux-gnu line in it, which according to pchome does not happen on Gentoo).

Perhaps it was not included in the `package_release.sh` script cos its intended for manual builds only?
Could ofc just delete it in the `package_release.sh` script, but then its almost as messy as it was and the whole thing is just pointless.

Nah.. i give up. Ill stop commenting, and wish whatever solution good luck :)

pchome:
> And thus have the lib/x86_64-linux-gnu line in it, which according to pchome does not happen on Gentoo

Gentoo build system passes all required params explicitly, e.g.
```java
meson \
  --buildtype plain \
  --libdir lib \
  --localstatedir /var/lib \
  --prefix /usr \
  --sysconfdir /etc \
  --wrap-mode nodownload \
  --cross-file /var/tmp/portage/games-util/dxvk-tests-0.96/temp/meson.i686-pc-linux-gnu.x86 \
  -Denable_tests=true \
  -Denable_utils=false \
  --unity=on \
  --bindir=lib/dxvk-tests \
  /var/tmp/portage/games-util/dxvk-tests-0.96/work/dxvk-0.96 \
  /var/tmp/portage/games-util/dxvk-tests-0.96/work/dxvk-0.96-abi_x86_32.x86
```
where
```java
  -Denable_tests=true \
  -Denable_utils=false \
  --unity=on \
  --bindir=lib/dxvk-tests \
```
the only options I 'v changed manually, see https://github.com/pchome/dxvk-gentoo-overlay/blob/master/games-util/dxvk-tests/dxvk-tests-0.96.ebuild#L64

Also those options are for patched DXVK's meson.build :wink: 
See https://github.com/pchome/dxvk-gentoo-overlay/blob/master/games-util/dxvk-tests/dxvk-tests-0.96.ebuild#L41
https://github.com/pchome/dxvk-gentoo-overlay/tree/master/games-util/dxvk-tests/files

So, lets say I'm using advanced custom build script. 

_P.S. This particular example is for `games-util/dxvk-tests` package, where tests are build separately.
For `games-util/dxvk` there are different logic/patches, e.g. custom `setup_dxvk_winelib.verb` generation, etc._

shmerl:
I guess this is obsolete now, since setup_dxvk.sh is now a static script, so it doesn't depend on Meson build. Placing dlls in libdir is easy by simply overrding bindir.

doitsujin:
Yeah, that's the idea. `package-release.sh` also uses this to install stuff into the appropriate directories directly, as originally suggested by @SveSop.

If you feel that this is no longer needed, please close the PR.

PR #896, opened 2019-01-30
Constant Buffer Bounds check option

K0bin:
This brings back 621aed5fdbaf92764944be7b3a27cbb3df63ba94 but disables it for everything other than Dark Souls Remastered.

### DIFF
--- a/src/d3d11/d3d11_options.cpp
+++ b/src/d3d11/d3d11_options.cpp
@@ -10,6 +10,7 @@ namespace dxvk {
     this->strictDivision          = config.getOption<bool>("d3d11.strictDivision", false);
     this->zeroInitWorkgroupMemory = config.getOption<bool>("d3d11.zeroInitWorkgroupMemory", false);
     this->relaxedBarriers       = config.getOption<bool>("d3d11.relaxedBarriers", false);
+    this->checkConstantBufferBounds = config.getOption<bool>("d3d11.checkConstantBufferBounds", false);
     this->maxTessFactor         = config.getOption<int32_t>("d3d11.maxTessFactor", 0);
     this->samplerAnisotropy     = config.getOption<int32_t>("d3d11.samplerAnisotropy", -1);
     this->deferSurfaceCreation  = config.getOption<bool>("dxgi.deferSurfaceCreation", false);
--- a/src/d3d11/d3d11_options.h
+++ b/src/d3d11/d3d11_options.h
@@ -7,11 +7,11 @@
 #include "d3d11_include.h"
 
 namespace dxvk {
-  
+
   struct D3D11Options {
     D3D11Options(const Config& config);
     /// Handle D3D11_MAP_FLAG_DO_NOT_WAIT properly.
-    /// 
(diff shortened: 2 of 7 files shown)


doitsujin:
This conflicts with e5a06d3f4a103a54cd4eb51970fedee405d1d698, so a new implementation would be necessary.

And I can't really say I'm a fan of messing around with basic shader code like this right now.

GloriousEggroll:
This also fixes the flicker in Grim Dawn on some nvidia cards

GloriousEggroll:
I think this should be fixed now as of commit: https://github.com/doitsujin/dxvk/commit/8f580efa25054ea46994c63fce6d2b50ebcbde0c

ZeroFault verified on discord that it does

Tk-Glitch:
And I verified that it doesn't :smile:

GloriousEggroll:
> And I verified that it doesn't smile

yep deleted my comment. /facepalm

lieff:
Hmm, may be it's last hotfix2 patch? I verify it again later.

GloriousEggroll:
latest upstream with this patch causes game to crash at crate logo on AMD, here's a patch so that it only checks for the option on dxbc if the user is on nvidia
`
diff --git a/src/dxbc/dxbc_options.cpp b/src/dxbc/dxbc_options.cpp
index cce1391..5bb08aa 100644
--- a/src/dxbc/dxbc_options.cpp
+++ b/src/dxbc/dxbc_options.cpp
@@ -27,10 +27,12 @@ namespace dxvk {
       = (devInfo.core.properties.limits.minStorageBufferOffsetAlignment <= sizeof(uint32_t));
     useSdivForBufferIndex
       = adapter->matchesDriver(DxvkGpuVendor::Nvidia, VK_DRIVER_ID_NVIDIA_PROPRIETARY_KHR, 0, 0);
-    
+
     strictDivision          = options.strictDivision;
     zeroInitWorkgroupMemory = options.zeroInitWorkgroupMemory;
-    
+    if (adapter->matchesDriver(DxvkGpuVendor::Nvidia, VK_DRIVER_ID_NVIDIA_PROPRIETARY_KHR, 0, 0))
+      checkConstantBufferBounds = options.checkConstantBufferBounds;
+
     // Disable early discard on RADV due to GPU hangs
     // Disable early discard on Nvidia because it may hurt performance
     if (adapter->matchesDriver(DxvkGpuVendor::Amd,    VK_DRIVER_ID_MESA_RADV_KHR,          0, 0)

`

K0bin:
That's weird. I don't see why it would crash.

All the patch does is add index < cbuffer.Size ? cBuffer[index] : vec4(0,0,0,0].

Maybe I messed up the rebase.

doitsujin:
I guess the problem is that you're mixing a vec4 with a smaller vector now. Constant buffer loads don't always load a vec4 anymore.

I promise to revisit this issue before tagging the 1.1 release. :frog:

K0bin:
I'll just close this and leave it to you then.

PR #908, opened 2019-02-09
Create separate log files for each d3d10 module

jp7677:
Hi there,

While playing a little bit with a d3d10 application I stumbled over some inconsistencies between the content of the log files and the log output on the console for d3d10 log statements. When digging into it I recognized that every d3d10 module (d3d10, d3d10_1, d3d10core) creates a logger for d3d10.log. These loggers can then overwrite each others content since every logger create its own stream to the log file.

This PR aims to create a separate logger for each d3d10 module by passing the file name for the logger as compiler argument into the compilation step. I did the same for the d3d11 and dxgi module for consistency.

Let me know what you think and where I could improve, _if_ you like this approach. I could actually go even further and determine the file name solemnly within the constructor the the logger. On the other hand I could also revert the changes for d3d11 and dxgi to handle only the problematic d3d10 modules with this way. I haven't touched the logger initialization of the test executable's, e.g. dxbc-compiler (yet).

PS: This is actually my third attempt to set the logger file name for each module sepately. I naively tried to use a static map of streams per logger file name, but that obviously didn't work because the logger sources are compiled into each module separately. (This took me quite some time to figure this out ;)) My second attempt was to get the module name at runtime using 'dllmain' and 'getmodulefilename'. That actually worked but felt a little bit tricky since the logger _must_ be initialized after 'dllmain' in this case. Using a compiler argument seemed the sanest solution imho.

### DIFF
--- a/src/d3d10/d3d10_main.cpp
+++ b/src/d3d10/d3d10_main.cpp
@@ -5,8 +5,10 @@
 
 #include "../dxgi/dxgi_adapter.h"
 
+#include "../util/util_str.h"
+
 namespace dxvk {
-  Logger Logger::s_instance("d3d10.log");
+  Logger Logger::s_instance(xstr(LOGGER_FILENAME));
 }
 
 extern "C" {
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -13,6 +13,7 @@ d3d10_core_dll = shared_library('d3d10core'+dll_ext, d3d10_main_src,
   install             : true,
   objects             : not dxvk_msvc ? 'd3d10core'+def_spec_ext : [],
   vs_module_defs      : 'd3d10core'+def_spec_ext,
+  cpp_args            : '-DLOGGER_FILENAME=d3d10core.log',
   override_options    : ['cpp_std='+dxvk_cpp_std])
 
 d3d10_dll = shared_library('d3d10'+dll_ext, d3d10_main_src,
@@ -22,6 +23,7 @@ d3d10_dll = shared_library('d3d10'+dll_ext, d3d10_main_src,
   install             : true,
   objects             : not dxvk_msvc ? 'd3d10'+def_spec_ext : [],
   vs_module_defs      : 'd3d10'+def_spec_ext,
+  cpp_args            : '-DLOGGER_FILENAME=d3d10.log',
   override_options    : ['cpp_std='+dxvk_cpp_std])
 
 d3d10_1_dll = shared_library('d3d10_1'+dll_ext, d3d10_main_src,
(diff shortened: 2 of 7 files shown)


jp7677:
Resources I used for stringification since I didn't know how to handle this:
https://gcc.gnu.org/onlinedocs/gcc-4.8.5/cpp/Stringification.html
https://en.wikipedia.org/wiki/C_preprocessor#Token_stringification

jp7677:
PPS: The override statement in second paragraph here https://github.com/doitsujin/dxvk/wiki/Common-issues#crashes should also include all d3d10 dll's.

jp7677:
After sleeping a night, a fourth approach would be by introducing a separate main for d3d10core and d3d10_1 something like this: https://github.com/doitsujin/dxvk/compare/master...jp7677:d3d10-logger2?expand=1
(only executed for d3d10core). This is architecture-wise probably the best approach but requires quite some extra lines of code.

jp7677:
Last option is of course to leave the logging behavior for D3D10 as it currently works, considering there are not much game out there that would be affected and that chances are pretty low that someone will really miss some information due to some missing log lines. Doing nothing would be absolutely fine with me.

doitsujin:
I'm actually debating whether it would make more sense to just put all the DXVK code into one single big library, and only implement the entry points in the actual DLLs. But I'm not sure if this would work with games that inspect loaded DLLs and the code that's in them.

Still, you're pointing out something that has been annoying me for quite some time, but I haven't found a good solution to that yet. Ideally we'd have **one** log file for everything, but since it can be initialized from pretty much anywhere and both DXGI and D3D11 contain entry points that may spew log messages, it's not as simple as just deferring it to Vulkan instance creation, for example.

jp7677:
Yeah, while playing with it I figured that too. All the solutions that came to my mind are still far from ideal and the number of log files would indeed further blow up. Something like out-of-process logging could avoid some issues but would introduce lots of other issues.

I guess it is best to close this PR to avoid spoiling your project with open PR's. Feel free to contact me if you want me to work in a certain direction, I would be happy to help.

pchome:
An alternative solution could be to use **stdout** and/or `OutputDebugString()` (or whatever is used for `WINEDEBUG=+debugstr`).
_I'm not sure about usage of WINE's debugging channels (FIXME(), ERR(),...), probably not an option due to MinGW._

This could be logged/collected in similar way as WINE/user does:
```
dxvk:info:d3d11: Message
dxvk:info:d3d11: Message 2
```
No custom log files - no problem.

<hr>

Another one solution is to actually **split d3d10 code** into separate ` d3d10`, `d3d10_1` and `d3d10core` source files. I use such solution in my [patches](https://github.com/pchome/dxvk-gentoo-overlay/blob/master/games-util/dxvk/files/dxvk-0.93-whole-program-support.patch#L107) for a long time w/o any problems (winelib build).
There is usually nothing valuable in d3d10 log files.

##### Notes about the patch:
* please ignore all `whole_program` parts
* the solution optimized for patching, so `#ifdef`'s are used to minimize the patch and the difference w/ upstream code

FYI

doitsujin:
> No custom log files - no problem.

This is not true unfortunately. When built as a Windows DLL, applications can (and some do) capture or redirect stdout/stderr, which would mean no log at all. Unigine Valley and Heaven are examples of this.

PR #909, opened 2019-02-10
[d3d11, d3d10, dxgi] Handle null ppvObject in QueryInterface.

Joshua-Ashton:
When a null ppvObject is passed into a QueryInterface on any IUnknown, a E_POINTER should be returned as the result (and it should not crash.)

This matches native d3d11/d3d10/dxgi behaviour and the documentation found here https://docs.microsoft.com/en-us/windows/desktop/api/unknwn/nf-unknwn-iunknown-queryinterface(q_) for IUnknown.


Joshua-Ashton:
A few d3d9 games rely on this, so I figured I'd check if the same was true for d3d11 and indeed it was.
May or may not fix anything but it's probably a good idea to match native behaviour (and also not crash :stuck_out_tongue:) -- and yes, it returns E_POINTER even if it doesn't have the interface. :frog:

PR #911, opened 2019-02-12
[d3d11] Ignore query interface for IDirectBee.

Joshua-Ashton:
IDirectBee is provided by this GUID in the CryEngine SDK and appears to be an interface of some form of internal CryEngine/CryTek d3d debug wrapper.
This ignores it being listed as an "unknown interface" in the logs to avoid misleading users as to the cause of any issues.

The interface is not supported or provided in native D3D11.


Joshua-Ashton:
or native D3D10 for that matter.

doitsujin:
Since this is specific to some cryengine versions, and therefore rare enough to not cause too much log spam, I'd like to ignore this particular interface.

Joshua-Ashton:
🐝

PR #857, opened 2019-01-13
Changed 2018 -> 2019

Faalagorn:
We are in 2019 now!

PR #919, opened 2019-02-14
fix old year (2018) in file (now 2019)

marcel-klomsbeck:
hello mr. doitsujin

this makes it so the file setup_dxvk.verb doesn't have the old year (2018) but replaces it with the new year (2019)

the commit description is: update the year to the current year (2019)

thank you for your time and open code! 😄


K0bin:
Duplicate of #857

PR #921, opened 2019-02-17
Update DXVK_FILTER_DEVICE_NAME to support matching substrings

thirdeyefunction:
Why? RADV device names include the LLVM version, which means that on every LLVM upgrade, the device name needs to be changed to match. This change allows setting something like "AMD RADV VEGA10" which should remain persistent over time.


doitsujin:
No objecttions.

PR #945, opened 2019-03-01
[build] Mark tests as gui apps in Meson

Joshua-Ashton:
Fixes building the tests on Windows and uses the right entrypoint and linker subsystem.


K0bin:
This has no effect on MinGW or Winelib builds. Those still work fine.

PR #960, opened 2019-03-11
 dxbc_compiler: use arithmetic shift to calculate offset 

libcg:
there's nothing preventing the offset from being negative. according to the SPIR-V spec logical shift will fill MSB with zeros which is invalid here.

### DIFF
--- a/src/dxbc/dxbc_compiler.cpp
+++ b/src/dxbc/dxbc_compiler.cpp
@@ -5122,13 +5122,9 @@ namespace dxvk {
     
     const uint32_t typeId = getVectorTypeId(result.type);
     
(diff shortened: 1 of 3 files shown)


libcg:
This has already been tried... Closing

libcg:
According to TKG this was working after deleting the cache, and using opUDiv instead of opSDiv would bring back the original issue. Reopening just in case

doitsujin:
This doesn't make any sense since the offsets aren't supposed to be negative in the first place. From a D3D perspective, the unsigned version is correct.

Also, why would the cache influence this? The shader code is different, this should trigger a cache miss.

PR #961, opened 2019-03-11
Made shell scripts conform to posix shell syntax.

openglfreak:
Because why not.

### DIFF
--- a/package-release.sh
+++ b/package-release.sh
@@ -1,8 +1,7 @@
-#!/bin/bash
+#!/bin/sh
 
 set -e
 
-shopt -s extglob
 
 if [ -z "$1" ] || [ -z "$2" ]; then
   echo "Usage: $0 version destdir [--no-package] [--dev-build] [--winelib]"
@@ -10,9 +9,9 @@ if [ -z "$1" ] || [ -z "$2" ]; then
 fi
 
 DXVK_VERSION="$1"
-DXVK_SRC_DIR=`dirname $(readlink -f $0)`
-DXVK_BUILD_DIR=$(realpath "$2")"/dxvk-$DXVK_VERSION"
-DXVK_ARCHIVE_PATH=$(realpath "$2")"/dxvk-$DXVK_VERSION.tar.gz"
+DXVK_SRC_DIR="$(readlink -f "$0")"; DXVK_SRC_DIR="${DXVK_SRC_DIR%/*}"
+DXVK_BUILD_DIR="$(realpath "$2")/dxvk-$DXVK_VERSION"
+DXVK_ARCHIVE_PATH="${DXVK_BUILD_DIR}.tar.gz"
 
 if [ -e "$DXVK_BUILD_DIR" ]; then
   echo "Build directory $DXVK_BUILD_DIR already exists"
@@ -47,7 +46,7 @@ while [ $# -gt 0 ]; do
   shift
 done
 
-function build_arch {
+build_arch() {
   export WINEARCH="win$1"
   export WINEPREFIX="$DXVK_BUILD_DIR/wine.$1"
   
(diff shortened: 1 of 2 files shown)


edmondo:
I skipped through this PR and I have following starting observations:
- I can see some good ideas I would agree with and others I would not recommend. Probably some changeset splitting of the ideas would also help the review.
- "for arg; do" : this seems buggy and needs to be fixed (and tested).
- you are changing (removing and sometimes adding) the curly brackets variable references. both of them are POSIX conform, so out of scope for this PR (or should be splitted in an own changeset). I also recommend to be consistent (always with or always without).
- There is one "find" command (replacing the "rm") with many -name options before the -exec. I didn't deeply check, but it rings me a bell: do we need some brackets?

Joshua-Ashton:
Why? Everybody has bash installed -- even if its not their preferred shell.

openglfreak:
@edmondo
1. Splitting the commit is probably a good idea.
2. `for arg; do` is a shorthand for `for arg in "$@"; do`
3. I omitted the braces only at the and of the string and before spaces, everywhere else I made sure t hey are there. I could add them to every variable if you think that's better.
4. No that command should work. Even the `-a`s aren't strictly necessary, as they are added implicitly if missing.

@Joshua-Ashton
Maybe someone doesn't want bash installed?
Idk, I don't like bash-specific scripts.

Joshua-Ashton:
> Maybe someone doesn't want bash installed?
Maybe they don't want to use 50% of Linux scripts 🤷‍♂️

openglfreak:
@doitsujin I have not tested it. I do not even know what the scripts do tbh.

But shellcheck approves, so it's probably fine, right?

openglfreak:
Okay I split the commit and added curly braces to all variable references.

edmondo:
@openglfreak Thanks

> 2. `for arg; do` is a shorthand for `for arg in "$@"; do`

You are right.

> 4. No that command should work. Even the `-a`s aren't strictly necessary, as they are added implicitly if missing.

Yes, it should be OK. Anyways I would make sure we just remove regular files not starting with . and not ending with .dll and probably write it like this:
`find "${DXVK_BUILD_DIR}/x$1" -type f ! \( -name '.*' -o -name '*.dll' \) -exec rm {} \;`

On a more personal note I actually prefer a bash script to use /bin/bash. It's getting a bit theoretical now, but I'm not sure if it's somewhere defined that /bin/sh must be a POSIX conform shell, so maybe it's just easier to let it be a bash script.

And yes, please test both scripts.

package-release.sh
```diff
@@ -27,68 +29,68 @@ opt_winelib=0
...
-    opt_devbuild=1
-    ;;
-  "--winelib")
-    opt_winelib=1
-    crossfile="build-wine"
-    ;;
-  *)
-    echo "Unrecognized option: $1" >&2
-    exit 1
+for arg; do
```
orbea: The while loop is perfectly portable and potentially more robust. I think its also more portable to do `for arg do` instead of `for arg; do`.

Joshua-Ashton:
I still don't see the point -- it's not like anyone doesn't have bash (pretty much every script requires it) and it just makes the code harder to read and more convoluted which overall seems less maintainable and more annoying imo.

openglfreak:
This is still open?
Wow, I completely forgot about this pr.

orbea:
> I still don't see the point -- it's not like anyone doesn't have bash (pretty much every script requires it) 

There are a lot of great reasons to avoid bash including infinite loops that are treated as "Won't fix" because the example scripts are disagreeable...

> it just makes the code harder to read and more convoluted which overall seems less maintainable and more annoying imo.

I won't deny this PR has issues and still required work, but its its very possible to write portable and readable shell. If there is interest I could be convinced to redo this PR.

PR #974, opened 2019-03-18
[d3d11] Handle invalid texture extents on creation

Joshua-Ashton:
d3d11 has a max 2D texture width/height of 16384 and and 3D of 2048 w/h/d.

If the dimensions exceed these values, on native, E_INVALIDARG is returned.

### DIFF
--- a/src/d3d11/d3d11_texture.cpp
+++ b/src/d3d11/d3d11_texture.cpp
@@ -267,6 +267,22 @@ namespace dxvk {
   HRESULT D3D11CommonTexture::NormalizeTextureProperties(D3D11_COMMON_TEXTURE_DESC* pDesc) {
     if (pDesc->Width == 0 || pDesc->Height == 0 || pDesc->Depth == 0)
       return E_INVALIDARG;
+
+    // The rules for invalid 2D textures are the same as 1D textures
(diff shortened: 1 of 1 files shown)


doitsujin:
We should rather look at hardware limitations (in the backend) rather than these cosntants. They should match on most hw, but d3d11 could support larger textures in the future.

Joshua-Ashton:
According to the documentation and debug layers, these values are hardcoded and are constant for all hardware:

https://docs.microsoft.com/en-us/windows/desktop/api/d3d11/ns-d3d11-d3d11_texture2d_desc

Joshua-Ashton:
I also understand that this will probably get caught by the DXVK backend also as it would fail to create the image -- but I have in the past had a GPU hang when a ridiculous size buffer when I was testing D9VK. The same may not apply for textures, but I figured it may be a better solution to early out here to avoid anything like that.

I should probably log it also, stating the Width/Height/Depth.

doitsujin:
Actually I should learn to read my own code, we already have a check:
https://github.com/doitsujin/dxvk/blob/90c7878a53651fad3206dd42ffac4790afee5ae5/src/d3d11/d3d11_texture.cpp#L301-L306

Since hardware limits pretty much exactly match D3D11 requirements these days, this should work, and I don't see any reason to disallow bigger textures in the future, should hardware allow it.

PR #991, opened 2019-04-01
[vulkan] Recreate the surface on surface loss.

cdavis5e:
According to the Vulkan spec:

> Several WSI functions return `VK_ERROR_SURFACE_LOST_KHR` if the
> surface becomes no longer available. After such an error, the surface
> (and any child swapchain, if one exists) **should** be destroyed, as
> there is no way to restore them to a not-lost state.

So if we get this error, we need to recreate the surface in addition to
the swapchain.

### DIFF
--- a/src/vulkan/vulkan_presenter.cpp
+++ b/src/vulkan/vulkan_presenter.cpp
@@ -106,8 +106,8 @@ namespace dxvk::vk {
     
     if ((status = m_vki->vkGetPhysicalDeviceSurfaceCapabilitiesKHR(
         m_device.adapter, m_surface, &caps)) != VK_SUCCESS) {
-      while (status == VK_ERROR_SURFACE_LOST_KHR) {
-        // Recreate the surface and try again.
+      for (uint32_t i = 0; i < 5 && status == VK_ERROR_SURFACE_LOST_KHR; i++) {
+        // Recreate the surface and try again. Give up after 5 tries.
         if (m_surface)
           destroySurface();
         if ((status = createSurface()) != VK_SUCCESS)
--- a/src/vulkan/vulkan_presenter.cpp
+++ b/src/vulkan/vulkan_presenter.cpp
@@ -10,8 +10,8 @@ namespace dxvk::vk {
     const Rc<DeviceFn>&   vkd,
           PresenterDevice device,
     const PresenterDesc&  desc)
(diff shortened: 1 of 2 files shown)


doitsujin:
Does this actually fix anything? Asking because this should *really* not happen in any application that isn't broken on Windows.

cdavis5e:
> Does this actually fix anything? Asking because this should _really_ not happen in any application that isn't broken on Windows.

On Mac, the system will sometimes recreate any `CAMetalLayer` backing a view. This happens, for example, switching between windowed and fullscreen in Final Fantasy XIV--the window style is altered, and the system recreates all the backing stores, including the `CAMetalLayer`, in response. I have [a change out on MoltenVK](/KhronosGroup/MoltenVK/pull/557) to make it generate `VK_ERROR_SURFACE_LOST_KHR` when this happens. Without both changes, the window changes to black and stays that way, because the surface we were using got pulled out from under us. With only the change to MoltenVK, FFXIV exits on switching from windowed to fullscreen, because the surface is now lost. With both these changes, I can switch between windowed and fullscreen at will.

src/vulkan/vulkan_presenter.cpp
```diff
@@ -105,8 +105,20 @@ namespace dxvk::vk {
     VkResult status;
     
     if ((status = m_vki->vkGetPhysicalDeviceSurfaceCapabilitiesKHR(
-        m_device.adapter, m_surface, &caps)) != VK_SUCCESS)
-      return status;
+        m_device.adapter, m_surface, &caps)) != VK_SUCCESS) {
+      for (uint32_t i = 0; i < 5 && status == VK_ERROR_SURFACE_LOST_KHR; i++) {
```
doitsujin: Is there a valid scenario where this can fail an arbitrary number of times in a row? I'd prefer a solution with no more than *one* attempt if that is reasonable, since any functions that affect the window should have completed at this point.

Joshua-Ashton:
> I have a change out on MoltenVK to make it generate VK_ERROR_SURFACE_LOST_KHR when this happens. 

Slightly unrelated note, and I might be wrong here but: I have a feeling this will probably break lots of things.
I would be very skeptical of games/apps actually handling this correctly, (as in, remaining working) if at all. 

If typical games treat that in the same way they treat d3d device loss, or lack thereof, then I'd think twice about not doing that transparently.

PR #1002, opened 2019-04-06
[build] Fix version info generation for winelib builds

pchome:
`*.res` extensions used for winelib, in case there is "magic" in `winegcc`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -25,8 +25,6 @@ else
   dxvk_library_path = meson.source_root() + '/lib32'
 endif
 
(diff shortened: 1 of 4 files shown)


pchome:
Not tested on MinGW builds.

doitsujin:
Does this even do anything without fake dlls?

pchome:
> Does this even do anything without fake dlls?

WINE compiles version information with both `.dll.so` and `.fakedll`, so I suppose reading such data using standard method may work. Reading data directly from "DLL" may, obviously, not work.

At least this shouldn't harm, and should preserve build system consistency.

> Wouldn't mind if you could do that.

I have no MinGW installed

PR #1004, opened 2019-04-07
[build] Meson kwargs support (>=0.49)

pchome:
http://mesonbuild.com/Release-notes-for-0-49-0.html#can-specify-keyword-arguments-with-a-dictionary

Changes nothing, but keyword arguments became more maintainable.

e.g. #945 could be something like
```patch
--- a/meson.build
+++ b/meson.build
@@ -69,6 +69,10 @@ common_exe_opts    = common_opts + { 'install': true }
 common_static_opts = common_opts + { 'include_directories': dxvk_include_path }
 common_shared_opts = common_opts + { 'include_directories': dxvk_include_path, 'name_prefix': '', 'install': true }
 
+if dxvk_compiler.get_id() == 'msvc'
+  common_exe_opts  += { 'gui_app' : true }
+endif
+
 glsl_compiler = find_program('glslangValidator')
 glsl_generator = generator(glsl_compiler,
   output    : [ '@BASENAME@.h' ],
```

_Started this just for fun, to check how those `kwargs` works. I'm sharing this, since patch already exists.
Feel free to reject this PR, and use it's parts as a guide when/if DXVK will support Meson 0.49._

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('dxvk', ['c', 'cpp'], version : 'v1.0.1', meson_version : '>= 0.43')
+project('dxvk', ['c', 'cpp'], version : 'v1.0.1', meson_version : '>= 0.49')
 
 cpu_family = target_machine.cpu_family()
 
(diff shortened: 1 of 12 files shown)


doitsujin:
I think it's far too early to **require** 0.49. We have to account for build systems being at least one year old.

doitsujin:
Don't get me wrong though, I don't mind these changes - I'll keep this one open and probably merge them at a later time, when pretty much everyone will have meson 0.49 available.

PR #1036, opened 2019-04-29
[util] Enable constant buffer range check for NieR:Automata

Bownairo:
This fixes a graphical corruption issue where Operator 6O's portrait
displays as a large flashing circle due to uninitialized buffer reads.

Fixes issue: https://github.com/ValveSoftware/Proton/issues/1543

Reviewed-by: Liam Middlebrook <lmiddlebrook@nvidia.com>

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -138,6 +138,10 @@ namespace dxvk {
     { "Grim Dawn.exe", {{
       { "d3d11.constantBufferRangeCheck",   "True" },
     }} },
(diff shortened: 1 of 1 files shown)


doitsujin:
Is this another case of a shader reading constant buffer data outside of the *bound* buffer range?

I'm willing to accept this as a workaround for now if it doesn't regress anything (this workaround doesn't match D3D11 behaviour), but I feel like we're going to need a real solution eventually, i.e. some way to specify that out-of-bound-range reads return 0 rather than undefined values.

PR #1038, opened 2019-04-30
[d3d11] Check if uav's counter slice is defined in CopyStructureCount

werman:
This fixes one of the crashes when working with RenderDoc and D3D11 under wine.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -662,9 +662,13 @@ namespace dxvk {
     if (!buf || !uav)
       return;
 
+    auto counterSlice = uav->GetCounterSlice();
+    if (!counterSlice.defined())
+      return;
+
     EmitCs([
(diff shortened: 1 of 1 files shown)


src/d3d11/d3d11_context.cpp
```diff
@@ -662,9 +662,13 @@ namespace dxvk {
...
       return;
 
+    auto counterSlice = uav->GetCounterSlice();
+    if (!counterSlice.defined())
+      return;
+
     EmitCs([
       cDstSlice = buf->GetBufferSlice(DstAlignedByteOffset),
-      cSrcSlice = uav->GetCounterSlice()
+      cSrcSlice = counterSlice
```
doitsujin: Can we get an `std::move` here? Saves a few CPU cycles due to atomic ref counts not getting changed redundantly, although to be fair, this function is hardly performance-critical.

werman: Yes, done.

PR #1058, opened 2019-05-15
[build] Add options to disable dxgi/d3d* build

### LINKED ISSUE #1040: Feature Request: options to disable dxgi/d3d*
Example: [0001-build-Support-separate-d3d-dxgi-builds.patch.txt](https://github.com/doitsujin/dxvk/files/3138703/0001-build-Support-separate-d3d-dxgi-builds.patch.txt) (based on d9vk)

Such changes are questionable for DXVK, so not a PR.

Motivation:
* can be useful for package maintainers, e.g. to build separate
  - dxvk-dxgi (or skip, as it's  optional for recent WINE versions)
  - dxvk-d3d10
  - dxvk-d3d11
* can drastically reduce compile times for d9vk git users/testers/developers, who currently compiles "whole DXVK/DX9VK" but uses only `d3d9.dll`.

Well, we can go deeper, and extend this changes to produce `dxvk-libs`, and then optionally reuse them when found in systems library paths.

pchome:
Supported following combinations:
* d3d11
* d3d11, dxgi
* d3d11, d3d10
* d3d11, d3d10, dxgi
* tests

Closes #1040

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,1 +1,4 @@
 option('enable_tests', type : 'boolean', value : false)
+option('enable_dxgi',  type : 'boolean', value : true, description: 'Build DXGI')
+option('enable_d3d10', type : 'boolean', value : true, description: 'Build D3D10')
+option('enable_d3d11', type : 'boolean', value : true, description: 'Build D3D11')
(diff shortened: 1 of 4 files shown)


Joshua-Ashton:
I think the DXGI Present shaders should be moved to D3D11.

pchome:
Yes, it was started such way, after the first PR in your repo.
But then I realized there is again too many changes in code (see patch in mentioned issue). So I decided to redo both PRs.

Anyway, shaders could be moved by developer later.

PR #1064, opened 2019-05-17
[build] Configure setup script to install only the enabled dlls

loathingKernel:
Configures setup_dxvk.sh based on what is enabled in meson. This way it won't try to install non-existent files and fail because of it. Mostly needed for d9vk since it makes the two projects install under the same prefix cleanly with the provided script if the user selects which dlls to build. There should be a similar PR for d9vk but since that project pulls from dxvk regularly, I think this PR should start from here.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -107,7 +107,7 @@ setup_dxvk_sh_conf = configuration_data()
 if get_option('enable_dxgi')
   setup_dxvk_sh_conf.set('action_dxgi', '$action dxgi')
 else
-  setup_dxvk_sh_conf.set('action_dxgi', '')
+  setup_dxvk_sh_conf.set('action_dxgi', 'true')
 endif
 
 if get_option('enable_d3d10')
--- a/meson.build
+++ b/meson.build
@@ -101,3 +101,31 @@ enable_tests = get_option('enable_tests')
 if enable_tests
   subdir('tests')
 endif
+
+setup_dxvk_sh_conf = configuration_data()
+
+if get_option('enable_dxgi')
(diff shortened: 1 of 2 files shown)


pchome:
One problem here: meson will generate two different scripts (for x32 and x64).

I tried to do something similar, basically to set `@libdir@` and `@prefix@`, but :man_shrugging: 

Also, empty
```bash
if [ ... ]; then

fi
```
will fail.

loathingKernel:
> One problem here: meson will generate two different scripts (for x32 and x64).
Yeah because meson is run twice. This could be fixed if the build system supported building both x64 and x32 architectures simultaneously but that is a different and larger change. The files should be equal though if the arguments are the same.

> Also, empty if will fail
Yeap, good catch. I assumed it would fall through but I was wrong. I think replacing it with the `true` utility is ok-ish since it just returns 0 and is part of the coreutils and quite possibly present in every distribution out there.

doitsujin:
@loathingKernel Please don't tie the setup script to meson config options again. I rewrote the script explicitly to avoid that, and to avoid having to ship **two** scripts for each separate architecture.

The best way to fix this would probably be to just check whether the file exists in the script.

@pchome I think this also serves as a good example why I'm not a huge fan of pointless changes to the build system.

loathingKernel:
I see, that is reasonable. I will close this and create a new one that handles it in the script.

pchome:
@doitsujin 
> I think this also serves as a good example why I'm not a huge fan of more or less pointless changes to the build system.

_The script is the root of most problems in DXVK ;)_
(_seriously_) Yes, I understand.

PR #1065, opened 2019-05-18
[utils] Don't try to install or remove missing dlls

loathingKernel:
Same goal as https://github.com/doitsujin/dxvk/pull/1064 but touching only setup_dxvk.sh.

I can think of one instance this is going to silently fail. If the user already has dxvk installed in a prefix, with more dlls than their new build includes, it won't remove the missing ones. It is a side-effect of trying to make partially-built dxvk and d9vk co-exist under the same prefix without trying to remove each other. Is it something that should be handled?

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -50,6 +50,7 @@ if [ -z "$wine" ]; then
 fi
 
 wine64="${wine}64"
+wineboot="${wine}boot"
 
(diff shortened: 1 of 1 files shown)


doitsujin:
> Is it something that should be handled?

I don't think that's relevant. Most users are going to install the whole thing anyway, and d9vk should be merged eventually.

PR #1071, opened 2019-05-21
[meta] Adding `compiler` to `DXVK_HUD`

concatime:
https://github.com/doitsujin/dxvk/commit/7dc449a

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ The `DXVK_HUD` environment variable controls a HUD which can display the framera
 - `memory`: Shows the amount of device memory allocated and used.
 - `version`: Shows DXVK version.
 - `api`: Shows the D3D feature level used by the application. Does not work correctly for D3D10 at the moment.
+- `compiler`: Shows shader compiler activity
 
 Additionally, `DXVK_HUD=1` has the same effect as `DXVK_HUD=devinfo,fps`, and `DXVK_HUD=full` enables all available HUD elements.
 


doitsujin:
That's a bit misleading since it does **not** show anything when the shaders are not being compiled from the state cache.

For that reason I'm not sure if it's a great idea to document this at all, it's not a useful HUD indicator and should probably be removed again.

concatime:
It's not that useless, because it's used by some install script from lutris: https://lutris.net/games/install/10154/view (`DXVK_HUD: compiler`).

doitsujin:
Yes, I'm aware of that; that's why it was added in the first place. I still don't want the broader public to think it's doing something that it isn't actually capable of doing.

h1z1:
Just to confirm are you saying when that shows on the HUD it's _only_ for new shaders?   To wit the lack of such output means either no compilation or it's coming from cache?

If that is the case Valve's precaching isn't doing much because I see that all the time.

K0bin:
> Just to confirm are you saying when that shows on the HUD it's only for new shaders? To wit the lack of such output means either no compilation or it's coming from cache?

It's the exact opposite. It's only showing when DXVK compiles shaders it already has encountered in an earlier run. It's impossible to show the message for new shaders.

h1z1:
Interesting.  That still doesn't match up with what I've seen which is most of the time that message conincides with studdering.   Granted it never occurred to me such distinction were present nor to check the code I'm quite curious.  Can't imagine I'd be the only one.

Kreyren:
i think that more options for info in `DXVK_HUD` is always good thing

pchome:
@doitsujin 
I have a question: is it possible to make counters visible outside DXVK?
To use them in e.g. [Conky](https://github.com/brndnmtthws/conky). 

Why? Well, even not 100% accurate it should be useful while playing a game in window, where the counters data would be represented **near** a game window, not inside.
(thinking about sysfs/procfs/debugfs, but not sure it's possible)

odysseywestra:
What about just changing the wording on the HUD to `Compiling Pre-Cached Shaders`? That should help with some of the confusion since it's already know that DXVK uses shader caches. Then that would make sense on the documentation end.

PR #1089, opened 2019-06-10
[dxgi] Return non-exact matches from FindClosestMatchingMode1

aeikum:
Team Sonic Racing requests a 60Hz mode. This method would fail if the hardware didn't supply a 60Hz mode, and the game would quit. Instead we should return the closest mode.

This change allows DXVK's DXGI to pass Wine's `test_find_closest_matching_mode` test.

### DIFF
--- a/src/dxgi/dxgi_output.cpp
+++ b/src/dxgi/dxgi_output.cpp
@@ -105,6 +105,95 @@ namespace dxvk {
   }
   
   
+  static void filterModesByDesc(std::vector<DXGI_MODE_DESC1> &modes, const DXGI_MODE_DESC1 &targetMode)
+  {
+    std::vector<DXGI_MODE_DESC1> matched;
+
+    if (targetMode.ScanlineOrdering != DXGI_MODE_SCANLINE_ORDER_UNSPECIFIED) {
+      for (auto mode : modes) {
+        if (mode.ScanlineOrdering == targetMode.ScanlineOrdering)
+          matched.push_back(mode);
+      }
+      if (matched.size() > 0) {
+        modes = matched;
+        matched.clear();
+      }
+    }
+
+    if (targetMode.Scaling != DXGI_MODE_SCALING_UNSPECIFIED) {
+      for (auto mode : modes) {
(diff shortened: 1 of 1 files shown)


doitsujin:
Thanks. I made some changes in order to better utilize C++ constructs (also, note that the `variable = struct { .member = value };` syntax isn't allowed) and merged that in cd6e3ffe758fe1a60c5f6d0d7aed0d43024ef466.

aeikum:
Thanks for the cleanup. I assumed the `.member` thing was in C++ these days, but I guess it's still a GNU extension or something. Bummer, I like that feature :)

The commit you made is significantly different from what I submitted. While it still fixes Team Sonic Racing, it now fails the Wine tests again. Your change requires the returned mode to match ScanlineOrdering, Scaling, and Format exactly, while the function should return the "closest" mode, even if it doesn't match those exactly.

Also you have a typo when checking the Format in FilterModesByDesc.

(On another note, it's kind of awkward to have my authorship on a commit that you basically authored. The bugs above are now attributed to me =/ )

doitsujin:
I think I'm having some trouble understanding what we're actually supposed to do (given that the behaviour of this function is undocumented), what your code does, and what the wine test wants us to do. Are we supposed to ignore Scaling etc. if no exact match exists?

aeikum:
That's my read on the documentation, yes. https://docs.microsoft.com/en-us/windows/desktop/api/dxgi1_2/nf-dxgi1_2-idxgioutput1-findclosestmatchingmode1

Not "ignore" exactly, but "closest", whatever that can mean for these enum values. In my patch, and in Wine's current implementation, it was ignored, yes.

I added some output to Wine's DXGI tests and I found that Win10 gave me S_OK with `DXGI_MODE_SCANLINE_ORDER_UPPER_FIELD_FIRST` when I asked for `LOWER_FIELD_FIRST`, which is more evidence for this behavior.

aeikum:
Confirmed current master (acab2bd8ce13510dc8feeeb48839b8c302136b80) passes the tests. Thanks!

Issue #816, opened 2018-12-17
Monster Hunter World randomly freezes


buscher:
Monster Hunter World (with proton) randomly freezes.
This usually happens in between after 10min to 4hours, so a long random time period.

As the DXVK_HUD (with memory) was enabled, at the time of the freeze, around ~3.9gb (assuming this is vram) of 6gb were used.

Most noticeable the dmesg output:
```
NVRM: Xid (PCI:0000:09:00): 31, Ch 0000004b, intr 10000000. MMU Fault: ENGINE GRAPHICS GPCCLIENT_T1_4 faulted @ 0x0_00000000. Fault is of type FAULT_PDE ACCESS_TYPE_READ
```
Xid 31, the addr 0x0_00000000, intr 10000000 and ACCESS_TYPE_READ are always constant.

To me, it looks like a simple nullptr access, as it is always the 0x0 addr, but I don't know how to investigate this problem further. I can not let the game run with
`VK_INSTANCE_LAYERS=VK_LAYER_LUNARG_standard_validation` or apitrace for hours, as this makes it very unplayable.

`PROTON_USE_WINED3D=1` just results in a black screen.
Allow flipping (in nvidia-setting) on/off does not change anything.

Please let me know how to make this report more useful, I am out of ideas.

### Software information
- Monster Hunter World
- vsync: off
- 30fps lock (getting weird input lag otherwise sometimes)
- Steam / Proton 3.16-beta5

### System information
- GPU: Nvidia Geforce 1060gtx 6gb
- Driver: nvidia-drivers-415.23
- Wine version: Proton 3.16-beta5 (???)
- DXVK version: Proton 3.16-beta5 (dxvk 0.93)
- Kernel: 4.19.10
- Ram: 16gb
- CPU: Ryzen 2700X

### Log files
(with DXVK_LOG_LEVEL=debug and DXVK_HUD=devinfo,fps,memory)

- d3d11.log: [MonsterHunterWorld_d3d11.log](https://github.com/doitsujin/dxvk/files/2687470/MonsterHunterWorld_d3d11.log)
- dxgi.log: [MonsterHunterWorld_dxgi.log](https://github.com/doitsujin/dxvk/files/2687471/MonsterHunterWorld_dxgi.log)

EDIT:
The game overall runs pretty well, just the random freezes are a pretty frustrating problem.

EDIT2:
The screen freezes but the game background music is still running.

doitsujin:
[...]

PR #1108, opened 2019-06-29
Enable deferSurfaceCreation for ELEX

DadSchoorse:
Elex crashes when creating some particles e.g. when using the Jetpack.
This problem only seems to affect nvidia.
deferSurfaceCreation fixes this behavior.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -153,6 +153,10 @@ namespace dxvk {
     /* The Surge                                  */
     { "TheSurge.exe", {{
       { "d3d11.allowMapFlagNoWait",         "True" },
+    }} },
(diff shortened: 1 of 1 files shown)


doitsujin:
This is a known issue, see https://github.com/doitsujin/dxvk/issues/816.

`deferSurfaceCreation` only affects the creation of a Vulkan surface for the application window when starting the game, and should only be used when the game creates swap chains for different APIs on the same window. This is **not** the case for ELEX.

DadSchoorse:
I don't know why deferSurfaceCreation helps but it does make the game playable.

doitsujin:
This doesn't fix the root cause of the problem however.

PR #1112, opened 2019-07-02
Print VK_EXT_memory_budget stats on allocation failure

liam-middlebrook:
This should help with logging efforts for #1100 .


liam-middlebrook:
Let me know if you'd like me to `fixup` the second into the first.

doitsujin:
Thanks; I added a cleaner way to check whether the DxvkDevice supports `VK_EXT_memory_budget` and added the logging stuff based on that.

Issue #1138, opened 2019-07-21
Modded Final Fantasy XIV corrupted rendering with DXVK 1.3.1


fosspill:
After switching from DXVK 1.3 to 1.3.1 FFXIV has started crashing when mods are installed.

Note: Unmodded still works fine.

I have been unable so far to track down specific types of mods that cause issues. Everything I've tried seems to be working as excepted with 1.3.1 as long as you have just a few mods installed.

I've also been unable to create an apitrace. Likely due to the launcher being 32bit and the game itself being 64 bit. I'll keep trying when I have time.

A picture of the "crash" / "corruption": https://i.imgur.com/nZniTwj.png - The DXVK HUD is still updating, but nothing functions.
### Workaround
Switch back to DXVK 1.3

### Software information
Final Fantasy XIV, default settings, modded.

### System information
- GPU: Vega 56
- Driver: Mesa-Aco 19.2.0 and Mesa (normal) 19.1.2
- Wine version: Wine-staging 4.12.1 and Wine-staging 4.10
- DXVK version: **1.3.1**

### Apitrace file(s)
- Unsure about how to go about this as the game crashes instantly upon starting with apitrace.

### Log files
- d3d11.log: https://gist.github.com/shinnova/dc573350adc8d181907eb0cb22db650f/raw/81801b3dc6b07974d88135ff4df346930bf8c6c4/ffxiv_dx11_d3d11.log
- dxgi.log: https://gist.github.com/shinnova/dc573350adc8d181907eb0cb22db650f/raw/81801b3dc6b07974d88135ff4df346930bf8c6c4/ffxiv_dx11_dxgi.log
- wine console output: https://gist.githubusercontent.com/shinnova/dc573350adc8d181907eb0cb22db650f/raw/86af91df05de15e36ed3dbcc7f7a7844fe89f59c/wine%2520console%2520log

doitsujin:
Game works fine on my RX 480, but I'm also not using any mods. Also, please test **without** ACO, since older ACO builds are known to break the game (it has been fixed upstream, but I have no idea where you're getting your build from).

Does this build work? It disables the optimization you mentioned, but I don't see why it would cause the game to crash.
[dxvk-ffxiv-test.tar.gz](https://github.com/doitsujin/dxvk/files/3414780/dxvk-ffxiv-test.tar.gz)

fosspill:
[...]

PR #1141, opened 2019-07-21
[dxvk] Handle divide by zero when determining slice count

Joshua-Ashton:
Fixes crash when making a zero-sized buffer.

May help #1138


doitsujin:
As discussed on Discord, this is bollocks - creating buffers of a size of zero is illegal and should be fixed properly.

PR #1150, opened 2019-08-03
[hud] Fix font rendering

pchome:
Some characters (like `i`,`l`, `|`) should be close to the "middle", not to prev. character.

dxvk-hud-default-ttf

Bonus: Terminus (TTF)
dxvk-hud-terminus-ttf
The "Advance" value should be less than '20', but was set to `20` because of "font shadows".
[0001-hud-Use-Terminus-TTF-font.patch.txt](https://github.com/doitsujin/dxvk/files/3464287/0001-hud-Use-Terminus-TTF-font.patch.txt)

### DIFF
--- a/src/dxvk/hud/dxvk_hud_renderer.cpp
+++ b/src/dxvk/hud/dxvk_hud_renderer.cpp
@@ -70,7 +70,7 @@ namespace dxvk::hud {
         sizeFactor * static_cast<float>(glyph.h) };
       
       const HudPos origin = {
-        pos.x + sizeFactor * static_cast<float>(glyph.originX),
+        pos.x - sizeFactor * static_cast<float>(glyph.originX),
         pos.y - sizeFactor * static_cast<float>(glyph.originY) };
       
       const HudPos posTl = { origin.x,          origin.y          };


doitsujin:
This doesn't really look like a correct fix to me to be perfectly honest. Inverting `x` would mean that the horizontal origin is on the right side, which isn't the case.

pchome:
Generated by `font-texture-generator` for Terminus (TTF): `{'|', 36, 80, 6, 25, -5, 23},` , originX mostly have negative values for mentioned characters.

doitsujin:
Hm, looks like you're right and it does indeed generate negative lower numbers for those, but this is counter-intuitive and goes against what e.g. Freetype does. I think we should fix the data instead.

pchome:
There is the PR in original repo to " Export advance parameter in C metadata format", maybe this value should be used.  Fixed repo : https://kevinmoran.github.io/font-texture-generator/

pchome:
Maybe next time I'll try to use different tool to generate glyphs and data. Currently I have no idea how to fix this properly. Inverting the sign of `originX` value, even if looks better, may still be wrong.

PR #1166, opened 2019-08-15
[util] Implement parent dir. based app configs

Joshua-Ashton:
## What?

This helps with identifying games where the executable name is variable or non-unique/generic.

To use a directory check in an app config, prepend it with a \\\\ [escaped backslash \\] in the name.

This is not an issue as \ is an illegal character for filenames on Windows.

## Why?

Sonic Adventure 2 uses the name "launcher.exe" for it's filename, but the parent directory is `Sonic Adventure 2`

## Example

```cpp
    /* Sonic Adventure 2                          */
    { "\\Sonic Adventure 2", {{
      { "d3d9.floatEmulation",              "False" },
    }} }
```

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -324,11 +324,22 @@ namespace dxvk {
   }
 
 
-  Config Config::getAppConfig(const std::string& appName) {
+  Config Config::getAppConfig(const std::string& appName, const std::string& parentDir) {
+    // Find configs by executable name
     auto appConfig = g_appDefaults.find(appName);
+
+    // Didn't find anything?
+    // Find configs by parent directory
+    if (appConfig == g_appDefaults.end()) {
+      appConfig = std::find_if(g_appDefaults.begin(), g_appDefaults.end(),
+        [&](const auto& appDir) -> bool {
+          return parentDir.ends_with(appDir.first);
+      });
+    }
+
     if (appConfig != g_appDefaults.end()) {
       // Inform the user that we loaded a default config
-      Logger::info(str::format("Found built-in config: ", appName));
+      Logger::info(str::format("Found built-in config: ", appConfig->first));
 
       return appConfig->second;
     }
--- a/src/util/util_env.h
+++ b/src/util/util_env.h
@@ -16,6 +16,16 @@ namespace dxvk::env {
   std::string getEnvVar(const char* name);
   
   /**
+   * \brief Gets the executable path
+   * 
+   * Returns the the full path of the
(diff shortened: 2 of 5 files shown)


Joshua-Ashton:
Sounds like a good idea. I will implement this tomorrow.

src/util/util_env.h
```diff
@@ -24,6 +24,16 @@ namespace dxvk::env {
    * \returns Executable name
    */
   std::string getExeName();
```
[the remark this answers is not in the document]
Joshua-Ashton: Can do

Joshua-Ashton:
https://github.com/Joshua-Ashton/dxvk/commit/c73d43653051aa9a889e7f3e2b8276d8347a4811

It now matches against the end of the string, so we can support `Crysis\Bin32`, etc

I also added `getExePath` to the header

Joshua-Ashton:
`Config Config::getAppConfig(const std::string & appName, const std::string & parentDir)` oops looks like some random formatting got in there somehow, lemme fix that

Joshua-Ashton:
Fixed that!

https://github.com/Joshua-Ashton/dxvk/commit/0d6542827d8ef01c045717f9258ee3aa6a8e14fe

src/util/config/config.cpp
```diff
@@ -324,11 +324,22 @@ namespace dxvk {
...
+  Config Config::getAppConfig(const std::string& appName, const std::string& parentDir) {
+    // Find configs by executable name
     auto appConfig = g_appDefaults.find(appName);
+
+    // Didn't find anything?
+    // Find configs by parent directory
+    if (appConfig == g_appDefaults.end()) {
+      appConfig = std::find_if(g_appDefaults.begin(), g_appDefaults.end(),
+        [&](const auto& appDir) -> bool {
+          return parentDir.ends_with(appDir.first);
```
liam-middlebrook: It looks like `::ends_with()` was added in C++20 [[1]](https://en.cppreference.com/w/cpp/string/basic_string/ends_with). In `meson.build` we set MSVC to use the latest standard available, but set `c++1z` for other compilers. It might be better to implement `::ends_with()` within `util_string` depending on what minimum compilers @doitsujin wants DXVK to support building with.

Joshua-Ashton: Ah I was unaware of that.

On Discord we discussed moving to a regex approach on the full path

doitsujin:
Here's a regex-based implementation that we were discussing:
https://github.com/doitsujin/dxvk/tree/dxvk-config-regex

PR #1173, opened 2019-08-26
Llvm mingw

jacekcw:
Fixes for compilation with llvm-mingw:
https://github.com/mstorsjo/llvm-mingw

### DIFF
--- a/build-win32.txt
+++ b/build-win32.txt
@@ -8,7 +8,7 @@ strip = 'i686-w64-mingw32-strip'
 c_args=['-msse', '-msse2']
 cpp_args=['-msse', '-msse2']
 c_link_args = ['-static', '-static-libgcc']
-cpp_link_args = ['-static', '-static-libgcc', '-static-libstdc++', '-Wl,--add-stdcall-alias,--enable-stdcall-fixup']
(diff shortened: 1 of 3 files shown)


pchome:
Maybe better use separate cross-files? More freedom to unconditionally set compiler binary names (`cpp = 'i686-w64-mingw32-clang++'` ?) and compiler flags.

jacekcw:
I don't see how it would be better. In my opinion, build system should be able to figure it out instead of hardcoding every possible configuration.

pchome:
Sure, it completely on developers how to configure their project's build system, but in theory in such cross-files we configuring build environment only once. Then we should be able to build every similar project using the same cross-file, w/o any modification to build system. Those files even not needed to be a part of the project (meson can find them in `~/.local/share/meson/cross/` or `/usr/share/meson/cross/`).

In real world we have multiple similar projects with copy-pasted cross-files and meson.build. Many of them (projects) took DXVK's build system as template, and now they may feel like they should adjust their build systems in the similar way for llvm-mingw, but they shouldn't (IMHO).

I personally would like to see this cross-files as a hint, to be moved in separate directory and extended with "debug" variants, even if I use completely different variant locally.

Anyway, not a big deal.

PR #1180, opened 2019-09-02
[util] Define Windows API version

Joshua-Ashton:
Exposes some more API features that are locked away behind ifdefs...

Some D3D9 stuff is like this for some reason in the MinGW headers I build with at least -- this patch would allow me to remove my hardcoded ``ifdef undef define endif`` for the proper solution as per https://docs.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=vs-2019

This also fixes the fact that previously the WIN32_LEAN_AND_MEAN define was doing nothing as there were other includes that included ``windows.h`` before that definition.

### DIFF
--- a/src/d3d10/d3d10_include.h
+++ b/src/d3d10/d3d10_include.h
@@ -3,5 +3,6 @@
 #include "../dxgi/dxgi_include.h"
 #include "../util/sync/sync_spinlock.h"
 
+#include "../util/util_targetver.h"
 #include <d3d10_1.h>
 #include <d3d11_1.h>
--- a/src/d3d10/d3d10_reflection.h
+++ b/src/d3d10/d3d10_reflection.h
@@ -4,6 +4,7 @@
 
 #include "d3d10_include.h"
 
+#include "../util/util_targetver.h"
 #include <d3d10shader.h>
 #include <d3d11shader.h>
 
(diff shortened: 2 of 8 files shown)


doitsujin:
The question is, as always, *why*?

For all I care we **don't** want to target Windows 10 exclusively.

Joshua-Ashton:
Well it just controls what is exposed in the headers and what isn't.
When I build on MSVC it automatically sets this for me.

For some reason MinGW defines Windows XP by default so some things I use don't get included.

doitsujin:
Such as?

I **really** don't think it's a good idea to use anything that might be unsupported by Windows 7/8 or Wine, for obvious reasons.

bno1:
You generally want to set those headers for the minimum windows version that you want to support (e.g. Windows 7). If you hit compilation problems it means that you're using a feature not supported by the windows version set in those headers and you either have to find a workaround or bump up those headers. AFAIK there is no benefit from setting those headers higher than you actually need to, but if you do you might unknowingly use a functions that's only available on Windows 10 and the library will fail to load in Windows 7.

Joshua-Ashton:
We can change it to 7... As I said it defaults to XP on MinGW but to 10 on MSVC. I doubt we'd do that unknowingly but I can change it to 7

doitsujin:
Yeah I'm fine with Windows 7, **as long as it does not break any system that dxvk can reasonably be used or compiled on**. This does include winelib builds and clang, please test that.

Also, please fix the blank lines between the includes. It's super ugly to have blank lines between two includes from `utils` but not between your `targetver.h` and system headers.

Issue #1182, opened 2019-09-04
Build fails with winegcc built by clang.


orbea:
When building dxvk with a copy of `winegcc` which has been built by `clang` and using the `build-wine32.txt` or `build-wine64.txt` cross files the build will fail because only `gcc` supports `--no-gnu-unique`.

I am not sure if this flag is important to keep or how to work around this using meson...

I am building `dxvk` with this command line.
```
meson \
    --prefix=/usr \
    --libdir=share/dxvk/x32 \
    --cross-file=build-wine32.txt \
    -Dstrip=true \
    -Db_ndebug=true \
    -Dbuildtype=plain \
    build32
```

### System information

- Wine version: `4.15`
- DXVK version: `1.3.3`
- meson version: `0.51.2`
- clang version: `8.0.1`

### Log files
```
The Meson build system
Version: 0.51.2
Source dir: /tmp/SBo/dxvk-1.3.3
Build dir: /tmp/SBo/dxvk-1.3.3/build32
Build type: cross build
Project name: dxvk
Project version: v1.3.3
Appending CFLAGS from environment: '-O2 -fPIC'
C compiler for the build machine: clang (clang 8.0.1 "clang version 8.0.1 (tags/RELEASE_801/final)")
Appending CXXFLAGS from environment: '-O2 -fPIC'
C++ compiler for the build machine: clang++ (clang 8.0.1 "clang version 8.0.1 (tags/RELEASE_801/final)")
C compiler for the host machine: winegcc (clang 8.0.1 "clang version 8.0.1 (tags/RELEASE_801/final)")

meson.build:1:0: ERROR: Compiler wineg++ can not compile programs.

A full log can be found at /tmp/SBo/dxvk-1.3.3/build32/meson-logs/meson-log.txt
```
- meson-log.txt: [meson-log.txt](https://github.com/doitsujin/dxvk/files/3572226/meson-log.txt)
```
clang-8: error: unsupported option '--no-gnu-unique'
winegcc: clang++ failed
```

orbea:
If the `--no-gnu-unique` flag is removed the build will fail later.
```
[12/207] Compiling C++ object 'src/dxbc/5c53ffa@@dxbc@sta/dxbc_header.cpp.o'.
In file included from ../src/dxbc/dxbc_reader.cpp:3:
In file included from ../src/dxbc/dxbc_reader.h:6:
In file included from ../src/dxbc/dxbc_tag.h:3:
[...]

PR #1184, opened 2019-09-05
Workaround build failure with winegcc/clang.

orbea:
Fixes https://github.com/doitsujin/dxvk/issues/1182

I briefly tested this with one game (`Grim Dawn`) without issues. I also tested that `--no-gnu-unique` is still used with `winegcc` as built by `gcc`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -33,6 +33,9 @@ dxvk_winelib = dxvk_compiler.compiles(code, name: 'winelib check')
 dxvk_extradep = [ ]
 
 if dxvk_winelib
+  if dxvk_compiler.has_argument('--no-gnu-unique')
+    add_project_arguments('--no-gnu-unique', language : ['cpp'])
+  endif
   wrc = find_program('wrc')
   lib_vulkan  = declare_dependency(link_args: [ '-lwinevulkan' ])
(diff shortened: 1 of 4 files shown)


meson.build
```diff
@@ -32,7 +32,12 @@ code = '''#ifndef __WINE__
 dxvk_winelib = dxvk_compiler.compiles(code, name: 'winelib check')
 dxvk_extradep = [ ]
 
+cpp_args = [ ]
+
 if dxvk_winelib
+  if dxvk_compiler.has_argument('--no-gnu-unique')
+    cpp_args += '--no-gnu-unique'
```
doitsujin: Please just use `add_project_arguments` directly like in all the other places instead of introducing random new variables.

orbea:
> Does this actually produce working binaries or is it "just" to fix the build process?

Yes and no, as explained in the OP dxvk built by winegcc/clang does work with the game `Grim Dawn`, but winecfg itself still has problems so I could only test against wine built with gcc. The remaining issues should be handled in wine, but it will be more annoying to debug this if dxvk doesn't even build.

meson.build
orbea: Done.

This was done following the example in mesa's build and is more robust, but okay...

PR #1185, opened 2019-09-05
Clean up redundant compiler and linker arguments.

orbea:
I'm not sure if you want this, but this is some trivial clean up following the examples in mesa's meson.build so I am submitting it for consideration. There should be no change in behavior, only in maintainability.

The basic goal is to reduce redundancy in meson.build and the cross files. I find this helps in the following ways.

* Its easier to add any compiler or linker arguments to meson.build.
* The cross files are more simple and have less redundancy.
* Its no longer needed to read 3 different files to find all of the compiler and linker arguments and only meson.build needs to be read. (I didn't touch the win32/win64 cross files since I am not testing them)

### DIFF
--- a/build-wine32.txt
+++ b/build-wine32.txt
@@ -7,13 +7,8 @@ strip = 'strip'
 [properties]
 needs_exe_wrapper = true
 
-c_args=['-m32', '-msse', '-msse2', '-fvisibility=hidden']
-cpp_args=['-m32', '-msse', '-msse2', '-fvisibility=hidden', '-fvisibility-inlines-hidden', '-D__WIDL_objidl_generated_name_0000000C=']
-cpp_link_args=['-m32', '-mwindows']
-
 [host_machine]
 system = 'linux'
 cpu_family = 'x86'
 cpu = 'i686'
 endian = 'little'
-
--- a/build-wine64.txt
+++ b/build-wine64.txt
@@ -7,13 +7,8 @@ strip = 'strip'
 [properties]
 needs_exe_wrapper = true
 
-c_args=['-m64', '-fvisibility=hidden']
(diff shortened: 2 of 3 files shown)


pchome:
Well,

https://mesonbuild.com/Style-guide.html#cross-compilation-arguments
> Try to keep cross compilation arguments away from your build files as much as possible. Keep them in the cross file instead. This adds portability, since all changes needed to compile to a different platform are isolated in one place.

Also, Meson 0.51.0 supports multiple cross files, so having multiple cross-files like `--cross-file  build-wine32.txt --cross-file clang-debug.txt` probably makes sense, but atm (min. required version is `meson_version : '>= 0.46'`) better to keep things as-is (and/or duplicate data in cross-files).

https://mesonbuild.com/Release-notes-for-0-51-0.html#multipe-cross-files-can-be-specified
> `--cross-file` can be passed multiple times, with the configuration files overlaying the same way as `--native-file`.

See also https://mesonbuild.com/Cross-compilation.html#cross-file-locations

I'm not sure, but you probably could use a list in the `[binaries]` section too, like `c = ['winegcc', '-m32']`. So, keeping only flags in that cross-files, required by compiler to build project(s). Other flags, like `-msse` and `-fvisibility=hidden` are optional, and completely on user to use them or not. Maybe only required preprocessor directives should be defined in the `meson.build` file. IMHO.

orbea:
Honestly I am having a hard time understanding. I thought cross files were meant for cross compiling, not enabling debug builds? Personally I use build scripts where it would be easy to add a `DEBUG` conditional.

In practice I am not sure these flags are perceived as optional so much as required or recommended defaults and at least in other projects defining such flags in the main configure file is normal. I'm obviously not familiar with what flags are really needed, but my goal was not to change behavior here.

As for the scope of these changes, I realize more can probably be done, but I decided to start small and see what other people think.

PR #1197, opened 2019-09-20
Update comment for FIFA 19+ workaround

m4sk1n:


doitsujin:
I mean you're not wrong, but... really?

PR #1227, opened 2019-10-16
[dxvk] Apply offset to render area in blitImageFb

Joshua-Ashton:
Fixes weird brown-ness in Halo 2


doitsujin:
As discussed on Discord, the correct fix is to cover the entire image in the render area. Good spot though.

PR #1238, opened 2019-11-05
Build fixes + Platform Compartmentalization w/ Ext Providers

Joshua-Ashton:
Enables building when no frontend/tests are selected (bugging me while I have been working on this)

Introduces a new, single system for providing extra instance/device extensions.

This consolidates platform-specific and vr-related to its' own thing for potential future cross-platform/native Linux work.

I've got the actual native stuff on my end but I'm going to wait until I've worked out all the kinks in the frontend before PRing that to avoid any things I've overlooked if it would be accepted ever :smile:

Cheers!

### DIFF
--- a/src/dxvk/meson.build
+++ b/src/dxvk/meson.build
@@ -98,6 +98,9 @@ dxvk_src = files([
   'dxvk_stats.cpp',
   'dxvk_unbound.cpp',
   'dxvk_util.cpp',
+  'dxvk_extension_provider.cpp',
+
+  'platform/dxvk_win32_exts.cpp',
   
   'hud/dxvk_hud.cpp',
   'hud/dxvk_hud_config.cpp',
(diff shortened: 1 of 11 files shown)


src/dxvk/meson.build
```diff
@@ -98,6 +98,9 @@ dxvk_src = files([
   'dxvk_stats.cpp',
   'dxvk_unbound.cpp',
   'dxvk_util.cpp',
+  'dxvk_extension_provider.cpp',
+
+  'platform/dxvk_win32_exts.cpp',
```
doitsujin: I'd rather have this file inside a `dxvk/platform` subdirectory rather than a new top-level one.

doitsujin: Actually, forget about that if there's more plaform stuff to come in the future.

Joshua-Ashton: Extension providers can be anything that could potentially influence the extensions we need. The change also made OpenVR an extension provider. In my mind it's not *just* a platform thing.

PR #1240, opened 2019-11-07
[dxvk] Hold a reference to the DxvkAdapter's DxvkInstance

aeikum:
This fixes a crash when playing Beat Saber with Wine's DXGI.

### DIFF
--- a/src/dxvk/dxvk_adapter.h
+++ b/src/dxvk/dxvk_adapter.h
@@ -255,7 +255,7 @@ namespace dxvk {
     
   private:
     
-    DxvkInstance*       m_instance;
+    Rc<DxvkInstance>    m_instance;
     Rc<vk::InstanceFn>  m_vki;
     VkPhysicalDevice    m_handle;
 


aeikum:
To elaborate a bit more.

- When using Wine's DXGI, D3D11CoreCreateDevice creates a DxvkInstance and holds a ref to it.
- DxvkInstance::findAdapterByLuid returns a DxvkAdapter which holds a pointer to that DxvkInstance.
- The DxvkAdapter is passed to D3D11DXGIDevice and persists.
- At the end of the function, the original ref is dropped, so the DxvkInstance is destroyed.
- The DxvkAdapter now holds an invalid pointer.

doitsujin:
I see the problem and wonder why we haven't run into this earlier since it's a really dumb bug; unfortunately this solution creates a circular reference since `DxvkAdapter` is owned by `DxvkInstance`, so neither object will ever be destroyed.

I'll work out why there's an adapter->instance relation in the first place, it probably shouldn't exist.

Joshua-Ashton:
Could solve this the d3d9 surface/texture way and make that child object inherit the refcount of its parent

doitsujin:
Doesn't really work that way since this isn't COM ref counting but the internal DXVK stuff.

aeikum:
> I see the problem and wonder why we haven't run into this earlier since it's a really dumb bug; 

FWIW the crash I saw happens when Proton's `vrclient.ivrcompositor_submit_dxvk` calls `IDXGIVkInteropDevice::GetVulkanHandles`. It crashed in that function once, but much more often it returns a bogus VkInstance, which we then crash on later.

doitsujin:
Beat Saber works with the above commit on my end; a proper solution will require significantly more work.

aeikum:
Thanks. I picked that commit on top of 1.4.4 and confirmed it's working here. Are you OK with releasing like that in the next Proton?

doitsujin:
Yes; I don't know how long it'll take to rework the whole adapter thing.

PR #1248, opened 2019-11-16
[util] Wide character conversion changes

Joshua-Ashton:
Replaces tows with an easier helper that fits in nicer with fixed-size arrays in DXGI, etc.

Make setThreadName take a wide char name by default given we know this entirely deterministically. Avoids needing to convert this and alloc a std::vector every time we call this.

Prefer UTF8 in tows/fromws.

### DIFF
--- a/src/dxgi/dxgi_adapter.cpp
+++ b/src/dxgi/dxgi_adapter.cpp
@@ -223,8 +223,7 @@ namespace dxvk {
     
     // Convert device name
     std::memset(pDesc->Description, 0, sizeof(pDesc->Description));
-    ::MultiByteToWideChar(CP_UTF8, 0, deviceProp.deviceName, -1,
-        pDesc->Description, sizeof(pDesc->Description) / sizeof(*pDesc->Description));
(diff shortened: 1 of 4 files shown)


doitsujin:
> Make setThreadName take a wide char name by default given we know this entirely deterministically.

I deliberately changed `setThreadName` to **not** take wide string arguments because I don't think we should pollute the backend with win32 specifics. This would need a ws->mbs conversion with your dxvk-native work *anyway* should you ever implement thread names.

Also, for future notice, please split separate changes into separate commits.

Joshua-Ashton:
Would a fixed-size limit on the thread name be acceptable? I kinda want to remove the alloc in setThreadName too...

Could constexpr magic the length with a string_view, but that'd mean the impl needs to be in a header...
If you're OK with an ifdef for that in the header platform-specific stuff for native, then I can do that.

Joshua-Ashton:
I've force pushed a new version.

Joshua-Ashton:
I've force pushed a new version, removing template stuff as per discussion on Discord.

PR #1251, opened 2019-11-23
stress free compilation on ubuntu

ViNi-Arco:
I left everything chewed for those who want to compile on ubuntu. ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ export WINEPREFIX=/path/to/.wine-prefix
 ## Commands to install as needed to compile on ubuntu:
 - sudo apt install mingw-w64
 - sudo apt install meson
-- sudo snap install cmake --candidate --classic
 - sudo apt install --no-install-recommends dh-autoreconf
 - sudo apt install glslang-tools
 
--- a/README.md
+++ b/README.md
@@ -35,6 +35,13 @@ export WINEPREFIX=/path/to/.wine-prefix
 - [MinGW64](http://mingw-w64.org/) 6.0 compiler and headers
 - [glslang](https://github.com/KhronosGroup/glslang) compiler
 
+## Commands to install as needed to compile on ubuntu:
+- sudo apt install mingw-w64
+- sudo apt install meson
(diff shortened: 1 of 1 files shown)


ViNi-Arco:
I chose it because it is more current and gets fixes faster, but I decided to take it off

Joshua-Ashton:
That doesn't really make sense given DXVK doesn't use `cmake` nor `dh-autoreconf`, whatever that is.

Specifying ``sudo`` in instructions is incorrect, not everybody uses ``sudo``.

I don't think we need instructions on how to install dependencies for every distro, we already list what you need, and people can easily find that themselves.

ZachBacon:
Meson only uses cmake if pkg-config isn't found. But I don't think really dxvk makes use of that since everything is self-contained, iirc you'd only need meson and the mingw compiler.

ViNi-Arco:
dh-autoreconf was needed to install here on ubuntu as it asks for autoconf and dh-autoreconf offers it, sudo is required to install them.

ViNi-Arco:
I did more testing and it really is necessary to compile on ubuntu, I did it to help those who want to try the latest dxvk news, because I saw some people with problems to compile on ubuntu

ViNi-Arco:
cmake não é necessário se não usar --no-install-recommends no MinGW64 ao instalalo, erro meu por querer muito minimalismo :)

Joshua-Ashton:
Getting some CI like on D9VK may be a good idea too 🐸

ViNi-Arco:
Do you speak from here? https://git.froggi.es/joshua/d9vk/pipelines; I downloaded a lot of their compilations but the last ones were giving problems in the GTA Vice City cutsenes and I had to compile mine, which is even less megabytes, around 14.1mb and your pass of 20mb, I don't understand rsrsr

doitsujin:
In general I'd like to avoid distro-specific instructions as much as possible; in particular, I would assume that users know how to install packages on their respective distro.

I'm fine with the `update-alternatives` changes, but the build dependencies are already documented.

PR #1260, opened 2019-11-28
[dxvk] Fix multiple inclusion of dxvk_platform_exts.h

lacc97:
This fixes build failures when using `--unity on` meson parameter


doitsujin:
Nice catch.

PR #1270, opened 2019-12-11
D9VK Shared Changes

Joshua-Ashton:
Upstreaming some shared-code changes/additions in D9VK.

### DIFF
--- a/src/spirv/spirv_module.cpp
+++ b/src/spirv/spirv_module.cpp
@@ -278,6 +278,23 @@ namespace dxvk {
     
     return this->constComposite(vectorTypeId, args.size(), args.data());
   }
+
+
+  uint32_t SpirvModule::constvec4b32(
+          bool                    x,
+          bool                    y,
+          bool                    z,
+          bool                    w) {
+    std::array<uint32_t, 4> args = {{
+      this->constBool(x), this->constBool(y),
+      this->constBool(z), this->constBool(w),
+    }};
+    
+    uint32_t scalarTypeId = this->defBoolType();
+    uint32_t vectorTypeId = this->defVectorType(scalarTypeId, 4);
+    
+    return this->constComposite(vectorTypeId, args.size(), args.data());
+  }
   
   
   uint32_t SpirvModule::constvec4u32(
--- a/src/spirv/spirv_module.cpp
+++ b/src/spirv/spirv_module.cpp
@@ -2613,6 +2613,21 @@ namespace dxvk {
     m_code.putWord(operand);
     return resultId;
   }
+
+
+  uint32_t SpirvModule::opExp(
+          uint32_t                resultType,
+          uint32_t                operand) {
+    uint32_t resultId = this->allocateId();
+    
+    m_code.putIns (spv::OpExtInst, 6);
+    m_code.putWord(resultType);
+    m_code.putWord(resultId);
+    m_code.putWord(m_instExtGlsl450);
+    m_code.putWord(spv::GLSLstd450Exp);
+    m_code.putWord(operand);
+    return resultId;
+  }
   
   
   uint32_t SpirvModule::opLog2(
--- a/src/spirv/spirv_module.cpp
+++ b/src/spirv/spirv_module.cpp
@@ -342,6 +342,24 @@ namespace dxvk {
     
     return this->constComposite(vectorTypeId, args.size(), args.data());
   }
+
+
+  uint32_t SpirvModule::constfReplicant(
+          float                   replicant,
+          uint32_t                count) {
+    uint32_t value = this->constf32(replicant);
+
+    std::array<uint32_t, 4> args = { value, value, value, value };
+
+    // Can't make a scalar composite.
+    if (count == 1)
+      return args[0];
+    
+    uint32_t scalarTypeId = this->defFloatType(32);
+    uint32_t vectorTypeId = this->defVectorType(scalarTypeId, count);
+    
+    return this->constComposite(vectorTypeId, count, args.data());
+  }
   
   
   uint32_t SpirvModule::constComposite(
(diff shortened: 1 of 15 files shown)


src/spirv/spirv_module.cpp
```diff
@@ -342,6 +359,44 @@ namespace dxvk {
...
     return this->constComposite(vectorTypeId, args.size(), args.data());
   }
+
+
+  uint32_t SpirvModule::constfReplicant(
+          float                   replicant,
+          uint32_t                count) {
+    std::array<uint32_t, 4> args = {{
+      this->constf32(replicant), this->constf32(replicant),
+      this->constf32(replicant), this->constf32(replicant),
```
doitsujin: This is a bit silly; storing the result of `constf32(...)` in a variable and then referencing that would be better.

src/spirv/spirv_module.cpp
```diff
@@ -342,6 +359,44 @@ namespace dxvk {
...
+    return this->constComposite(vectorTypeId, count, args.data());
+  }
+
+
+  uint32_t SpirvModule::constbReplicant(
+          bool                    replicant,
+          uint32_t                count) {
+    std::array<uint32_t, 4> args = {{
+      this->constBool(replicant), this->constBool(replicant),
+      this->constBool(replicant), this->constBool(replicant),
```
doitsujin: Same here.

Joshua-Ashton:
Force pushed to fix lzcnt on Clang. (forgot to git add during rebase edit)

src/util/com/com_object.h
```diff
@@ -67,13 +67,48 @@ namespace dxvk {
...
+  *
+  * This version of ComObject ensures that the reference
+  * count does not wrap around if a release happens at zero.
+  * eg. [m_refCount = 0]
+  *     Release()
+  *     [m_refCount = 0]
+  * This is a notable quirk of D3D9's COM implementation
+  * and is relied upon by some games.
+  */
+  template<typename... Base>
```
doitsujin: Whitespace, please make it like this:
```
/**
 *
```

src/util/com/com_object.h
```diff
@@ -67,13 +67,48 @@ namespace dxvk {
...
+  class ComObjectClamp : public ComObject<Base...> {
```
doitsujin: We no longer allow multiple inheritance, so the `...` should be removed.

src/util/config/config.h
```diff
@@ -120,6 +120,10 @@ namespace dxvk {
     static bool parseOptionValue(
       const std::string&  value,
             int32_t&      result);
+
+    static bool parseOptionValue(
+      const std::string&  value,
+            uint32_t&     result);
```
doitsujin: Any particular reason why we need unsigned integers as well? Currently we're using `int32_t` for all the options, if the user enters a negative number it's their own fault.

src/util/util_bit.h
```diff
@@ -14,8 +14,20 @@
 #endif
 
 #include "util_likely.h"
```
doitsujin: newline

src/util/util_math.h
```diff
@@ -1,19 +1,34 @@
...
+  constexpr T align(T what, U to) {
     return (what + to - 1) & ~(to - 1);
   }
+
+  template<typename T, typename U = T>
+  constexpr T alignDown(T what, U to) {
+    return (what / to) * to;
+  }
+
+  // Equivelant of std::clamp for use with floating point numbers
```
doitsujin: "Equivelant"

src/util/util_vector.cpp
```diff
@@ -0,0 +1,9 @@
+#include "util_vector.h"
```
doitsujin: If this file is empty anyway, please remove it.

src/dxvk/dxvk_device.h
```diff
@@ -472,6 +472,8 @@ namespace dxvk {
     DxvkStatCounters            m_statCounters;
     
     DxvkDeviceQueueSet          m_queues;
+
+    std::atomic<uint32_t>       m_numSamplers = { 0 };
```
doitsujin: I still don't like this at all, and would like to ask to drop the sampler HUD thing for now.

Instead of tracking sampler count in the backend (which is *completely* pointless for D3D11 and just confusing, and also incredibly ugly code-wise), we should probably work on making the HUD a bit more flexible to allow this to be done on the D3D9 side.

src/util/util_vector.cpp
Joshua-Ashton: I rebased this commit to do that before and forgot to actually make the change to the commit 🐸 I'll fix this, yea

src/util/config/config.h
Joshua-Ashton: A uint32_t works out nicely here because the max value of GetAvailableTextureMem is UINT32_MAX (the counting is done from min(vramAmount, UINT32_MAX))

We can then limit the amount of VRAM exposed to the game/tracking via the option. 

https://github.com/Joshua-Ashton/d9vk/blob/master/src/d3d9/d3d9_options.cpp#L52

src/dxvk/dxvk_device.h
Joshua-Ashton: Alright, I've dropped the commit

Joshua-Ashton:
Pushed new revisions.

src/util/config/config.h
doitsujin: Wouldn't it be more useful to specify this in Megabytes rather than bytes, for user convenience?

Joshua-Ashton: Some Sims 2 people are using their own dxvk.confs right now and I am wary of breaking that and getting 100 of them on my tracker 🐸 

I'll make that change anyway though.

PR #1283, opened 2019-12-17
Report gpu vendor as Nvidia for Star Wars Battlefront II

gort818:
Star Wars Battlefront II requires requires AGS library.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -157,6 +157,10 @@ namespace dxvk {
     { R"(\\Atelier_Ryza\.exe$)", {{
       { "d3d9.deferSurfaceCreation",        "True" },
     }} },
+    /* Star Wars Battlefront II: amdags issues    */
+    { R"(\\starwarsbattlefrontii\.exe$)", {{
+      { "dxgi.customVendorId",              "10de" },
+    }} },
 
     /**********************************************/
     /* D3D9 GAMES                                 */


doitsujin:
What hardware was this tested on?

Mostly asking because Frostbite Engine relies on the reported vendor being correct, since it applies some quirks depending on it. It might be a better idea to figure out which AGS version the game uses and to use [dxvk-ags](https://github.com/doitsujin/dxvk-ags) instead of changing the vendor id.

gort818:
Vega 56, without changing the vendor id they game will not launch complaining about an out of date driver.

Ok tips on  how to find the AGS version?

I could just built all versions of dxvk-ags and see which one works.

doitsujin:
> I could just built all versions of dxvk-ags and see which one works.

that's probably your best bet, the only way i've found so far is dumping the exports and trying to figure it out that way. The joy of libraries with zero backwards compatibility...

gort818:
I tried version 5.0-5.3 of dxvk-ags and replaced the original and unfortunately no luck. The same message appears.
amdags

Reporting as Nvidia works great for me, game is running super smooth.

Ahmed-E-86:
Now, Star Wars Battlefront II Celebration Edition complains about outdated Nvidia driver

Screenshot from 2021-08-14 20-07-39

I use an AMD gpu, not Nvidia gpu.

Ahmed-E-86:
These settings work for me:

```
dxgi.customDeviceId = 10de
dxgi.customVendorId = 1c06
```
`dxgi.customDeviceId = 10de` does not work alone.

PR #1310, opened 2019-12-31
Add installation support for pure 64-bit Wine (non WoW64)

shmerl:
This makes setup script work correctly with pure 64-bit Wine (non WoW64) since currently the script makes an implicit assumption that every 64-bit Wine is WoW64 one. Custom built Wine is a lot easier to build when it's pure 64-bit, so the change makes it skip 32-bit related steps in such case.

It also adds some minor clean-ups.

The general logic of using `$wine` to determine location is also replaced with Wine's recommended implicit logic of `$PATH`. I.e. if the user wants to use custom Wine location, it's enough to override `$PATH` and `$LD_LIBRARY_PATH` and usage of `wine`, `wine64`, `wineboot` and so on will implicitly pick up that change.

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+# default directories
+dxvk_lib32=${dxvk_lib32:-"x32"}
+dxvk_lib64=${dxvk_lib64:-"x64"}
+
 # figure out where we are
 basedir=`dirname "$(readlink -f $0)"`
 
@@ -45,12 +49,20 @@ fi
 # find wine executable
 export WINEDEBUG=-all
 
-if [ -z "$wine" ]; then
-  wine="wine"
-fi
+wine="wine"
+wine64="wine64"
+wineboot="wineboot"
 
-wine64="${wine}64"
-wineboot="${wine}boot"
+# $PATH is the way for user to control where wine is located (including custom Wine versions).
+# Pure 64-bit Wine (non Wow64) requries skipping 32-bit steps.
+# In such case, wine64 and winebooot will be present, but wine binary will be missing,
+# however it can be present in other PATHs, so it shouldn't be used, to avoid versions mixing.
+wine_path=$(dirname "$(which $wineboot)")
+wow64=true
+if ! [ -f "$wine_path/$wine" ]; then
+   wine=$wine64
+   wow64=false
+fi
 
 # resolve 32-bit and 64-bit system32 path
 winever=`$wine --version | grep wine`
@@ -63,8 +75,10 @@ fi
 # if they are missing
 $wineboot -u
 
-win32_sys_path=$($wine winepath -u 'C:\windows\system32' 2> /dev/null)
 win64_sys_path=$($wine64 winepath -u 'C:\windows\system32' 2> /dev/null)
+if $wow64; then
+  win32_sys_path=$($wine winepath -u 'C:\windows\system32' 2> /dev/null)
+fi
 
 if [ -z "$win32_sys_path" ] && [ -z "$win64_sys_path" ]; then
   echo 'Failed to resolve C:\windows\system32.' >&2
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
What's with the mix and match of `-v` being added to `mv`/`rm` and not?

shmerl:
I tried to add -v to every operation to make it output a nice list of changes, did I skip something?

Joshua-Ashton:
https://github.com/doitsujin/dxvk/pull/1310/files#diff-52f09cbfd5cc4ba6f672196484b85e34R122

setup_dxvk.sh
```diff
@@ -107,7 +121,7 @@ installFile() {
       if ! [ -f "${dstfile}.old" ]; then
         mv "${dstfile}" "${dstfile}.old"
```
Joshua-Ashton: Missing -v

shmerl: Fixed

doitsujin:
> It also adds some minor clean-ups.

If you feel those are necessary for whatever reason, please make them a separate commit. It's hard to reason about changes when there are 10 different things batched together.

That said, what exactly is the use case for a pure 64-bit wine install? There is no such Windows environment.

telans:
> That said, what exactly is the use case for a pure 64-bit wine install? There is no such Windows environment.

I think the BSD's do 32 or 64bit only? Not sure on that

shmerl:
I can split it into several commits for clarity.

> That said, what exactly is the use case for a pure 64-bit wine install? There is no such Windows environment.

One use case I often have is building Wine + esync patches. Building pure 64-bit Wine is a simple process - just check out the source, apply patches and build. The result has only `wine64` binary. It's quite enough to run any modern game all of which are 64-bit. Building WoW64 Wine on the other hand is a complicated process, requiring having some VM or container, or less likely cross compilation (that Wine project doesn't recommend due to high complexity of setting it up).

So I rarely go through that when I want to have customized Wine, since I usually customize it for newer games.

pchome:
@shmerl 
> The result has only wine64 binary.

Create a symlink.

shmerl:
> Create a symlink.

That doesn't solve the problem. Note how the script tries to figure out location of 32-bit and 64-bit prefix locations. With symlink it's going to be the same (system32 basically) and it messes up the whole sequence.

aqxa1:
>Building WoW64 Wine on the other hand is a complicated process, requiring having some VM or container, or less likely cross compilation (the last one Wine project doesn't recommend due to high complexity of setting it up).

That's odd. It's normally enough to configure the wine32 build using "--with-wine64=<64bit build>". No need for a cross compiler, or any complicated container setup. 

I'm not saying that pure 64-bit support is a bad idea, though.

shmerl:
According to Wine wiki, you need a complicated setup: https://wiki.winehq.org/Building_Wine#Shared_WoW64

How exactly are you building it?

loathingKernel:
```
cd ~/wine-dirs/wine64-build/
../wine-source/configure --enable-win64
make

cd ~/wine-dirs/wine32-build/
PKG_CONFIG_PATH=/path/to/pkgconfig ../wine-source/configure --with-wine64=../wine64-build
make

```

In what way is this complicated?

shmerl:
That only prepares it for wow64, you still need to do the 32-bit build. How exactly are you doing it?

doitsujin:
@shmerl are you still going to clean this up as requested in https://github.com/doitsujin/dxvk/pull/1310#issuecomment-569898610? I won't consider merging this until that is done.

shmerl:
Yes, I'll split it.

loathingKernel:
> It's far from trivial, unless your distro has full multiarch support.

Yes, you need 32bit libraries for the required things too. At the same time you can disable whatever features are using libraries your distribution doesn't provide for its multilib setup. You will need to disable those features both in 64 and 32 bit wine configure steps. Take a look at the output of ``./configure --help``

shmerl:
Let's not go into off-topic much. Enough to say, that when I tried, Wine with multilib / multiarch was a huge mess.

pchome:
@shmerl 
>> Create a symlink.

> That doesn't solve the problem. Note how the script tries to figure out paths for 32-bit and 64-bit prefix locations. With symlink it's going to be the same (system32 basically) and it messes up the whole sequence.

I mean do this on your side, you seem good in scripting, so e.g. an **alias** for `wine winepath` should help you even with other scripts calling `wine`.  Return the *nix path to your Trash folder, or TMP.
IDK, there are a lot of possibilities and variations how it can be handled on Linux. Unless this is hardcoded in an installation script.
Just a thoughts.

PR #1311, opened 2020-01-01
[meta] Happy New Year 🎉💃🎊

Joshua-Ashton:
It's now 2020!


libcg:
Not quite...

telans:
> Not quite...

I wonder if other countries use different timezones 🤔

doitsujin:
Nah, obviously everything uses the same single time zone as the US.

Oh. :frog:

PR #1317, opened 2020-01-04
Fixes refcounting to match native d3d11 behavior

kingscallop:
These bugs were found when trying to run bgfx on top of dxvk.
The dxvk refcount on COM objects doesn't match the refcount of native d3d11.
Bgfx expects the refcount values to be as on windows.

### DIFF
--- a/src/d3d11/d3d11_context_imm.cpp
+++ b/src/d3d11/d3d11_context_imm.cpp
@@ -35,22 +35,6 @@ namespace dxvk {
   }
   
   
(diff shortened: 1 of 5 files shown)


doitsujin:
I am aware of this, unfortunately fixing reference counting causes crashes in some games.

kingscallop:
I understand your perspective.
If you ever decide to follow more closely the ref counting of d3d11 I'll help you debug it.

PR #1344, opened 2020-01-16
Add support for exporting ELF information alongside the DLLS for Linux tools like perf.

Guy1524:
Based off Remi's wine patch.  I tested this with d3d11-triangle, and it seems to work, however, lot's of functions in the various dxvk dlls are actually part of mingw's libc++, so they don't show up in perf top.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -53,6 +53,15 @@ else
     wrc = find_program('rc')
   else
     wrc = cpu_family == 'x86_64' ? find_program('x86_64-w64-mingw32-windres') : find_program('i686-w64-mingw32-windres')
+    if get_option('dump_debug')
+      if get_option('buildtype') == 'release'
+        error('Option dump_debug specified in a release build')
+      endif
+      add_global_link_arguments('-Wl,--file-alignment,4096', language : ['cpp'])
+      objcopy = cpu_family == 'x86_64' ? find_program('x86_64-w64-mingw32-objcopy') : find_program('i686-w64-mingw32-objcopy')
+      bfd_target = cpu_family == 'x86_64' ? 'elf64-x86-64' : 'elf32-i386'
+      dump_debug = [objcopy, '-O', bfd_target, '--only-keep-debug', '--file-alignment=4096', '@INPUT@', '@OUTPUT@']
+    endif
   endif
   if cpu_family == 'x86'
     if dxvk_compiler.has_link_argument('-Wl,--add-stdcall-alias')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -1,4 +1,5 @@
 option('enable_tests', type : 'boolean', value : false)
+option('dump_debug',   type : 'boolean', value : false, description: 'Copy debug symbols alongside DLLs for linux debugging tools')
 option('enable_dxgi',  type : 'boolean', value : true, description: 'Build DXGI')
 option('enable_d3d9',  type : 'boolean', value : true, description: 'Build D3D9')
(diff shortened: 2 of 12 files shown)


doitsujin:
Thanks; I popped this into a local branch for now since it can be useful, but not necessarily something to upstream (might still do that eventually after some cleanup).

pchome:
FYI
Some thoughts:
* if you want ask users to run debug tools, then better send them a debug build, or ask them to build DXVK with debug options, or better create official debug build ([just a random repo release page as example](https://github.com/KhronosGroup/glslang/releases))

* "splitdebug" and "compressdebug" (--compress-debug-sections) it's more for distribution part, this two options BTW are Gentoo built-in features, other distributions should have similar things too. I mean no need to reinvent this in DXVK, better fix this in distributions if broken and if possible. I'm not sure what the "Remi's wine patch" is, but I built Wine many times like this, w/o any Wine patches (I guess).

* if you do debug work by yourself, and not for mingw debugging, then why not to use winelib builds. Hard to handle this libs? Try the "WINEMOD" patch I PR'ed  for Protons winelib once. Just patch your Wine, add path to your debug/release/etc. DXVK build and now you running Wine with desired build. That's it, no installations/overrides/copying/etc. Very useful thing sometimes, I doing patchworks on my live Wine installation with this, w/o modifying Wine itself and reinstallations. TBH I spent zero time to create it, it work for me and I use it on per-day basis. If it will became useful for you but is broken, then just fix and upstream.

TL;DR: I think this feature is useless as built-in for DXVK (or any other project). No offence.

doitsujin:
Well the idea is to make DXVK show up in `perf top` on my end when I need it. This is by no means intended for end users and the reason why it exists is because I explicitly asked Derek for a way to do this in DXVK.

pchome:
So, you just don't want meson `--strip` option for `--dev-build` mode in the ` package-release.sh`, right? Doing objcopy/strip manually after installation should be possible then, if needed.

doitsujin:
Again, this is **not intended for end users** and **not intended to be merged**, this exists **BECAUSE I WANTED TO KNOW HOW IT WORKS**. Jesus.

pchome:
Don't get me wrong, I just sharing my knowledge how it works (or should work), and I will appreciate if you correct me when I'm wrong.

pchome:
Sorry, I can't live this as is, I wanted a reference to a valid solution for future me, in case I want to debug mingw builds.

So, if you have to generate separate debug info, then this should work before or after `ninja install` for `--dev-build` option (I mean do objdump and strip separately for x86 and x86_64). Moreover, new --debug-build mode can be introduced (likely), which will do the same, but with `--buildtype "plain"` and custom c_args/cpp_args to set more detailed debug info (-g[1-3] levels or so).

If this is the correct solution, then it can be safely merged into master branch (package-release.sh), and we'll have valid working solution and reference how to cook debug symbols for winelib/mingw builds.

I understand you have more important tasks to do, but as I said before I will appreciate for the info why it should be hardcoded into the build system scripts, and I'll gladly help to test with non-mingw builds otherwise (or I already did, I don't remember).

**TL;DR**: I just need to know if `x86_64-w64-mingw32-objcopy -O elf64-x86-64 --only-keep-debug --file-alignment=4096 in_file_name out_file_name` and then `x86_64-w64-mingw32-strip out_file_name` works, w/o *.build files being modified.
_But I'm too lazy to install mingw and check by myself, so I disturbing developers, which don't even care about this part of code/script._

PR #1350, opened 2020-01-17
[d3d9] Create front buffer

doitsujin:
Makes GetFrontBufferData use the correct image.


Joshua-Ashton:
Change to m_backbuffers
```cpp
    for (uint32_t i = 0; i < m_backBuffers.size(); i++)
      m_backBuffers[i] = new D3D9Surface(m_parent, &desc);```

Joshua-Ashton:
Should be fine now :)

PR #1351, opened 2020-01-17
[d3d9] Fix broken image view creation

### LINKED ISSUE #1349: D3D9 games do not start
GW2 does not start with b954ab886f2eaa76393b6f52b921bcb0df0ca308

### System information
- GPU: nvidia 970
- Driver: 440.48.02
- Wine version: 5.0 rcX
- DXVK version: b954ab886f2eaa76393b6f52b921bcb0df0ca308

### Log files
[Gw2-64_d3d9.log](https://github.com/doitsujin/dxvk/files/4075331/Gw2-64_d3d9.log)

doitsujin:
This addresses a regression introduced with 7c53a997effc64adbe52a8f67ee6bb6b5f2fe580 that would lead to us creating views with `VK_IMAGE_FORMAT_UNDEFINED`.

The first page refactors the `D3D9CommonTexture` constructor to not take a `D3D9_VK_FORMAT_MAPPING`. This seemed like bad design to begin with since it expects the mapping to match the image format anyway, and might have been causing issues for images with `Unknown` format.

The second patch fixes sRGB lookup for formats that have no corresponding sRGB format, which seems ot be responsible for most of the current issues.

Fixes #1349.

### DIFF
--- a/src/d3d9/d3d9_common_texture.cpp
+++ b/src/d3d9/d3d9_common_texture.cpp
@@ -455,7 +455,9 @@ namespace dxvk {
       return;
 
     m_sampleView.Color = CreateView(AllLayers, Lod, VK_IMAGE_USAGE_SAMPLED_BIT, false);
-    m_sampleView.Srgb  = CreateView(AllLayers, Lod, VK_IMAGE_USAGE_SAMPLED_BIT, true);
+
+    if (IsSrgbCompatible())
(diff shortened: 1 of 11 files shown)


Joshua-Ashton:
Let's not duplicate the view -- I think we should make the linear-space one if we have no valid SRGB format and then make the SRGB Rc<Doobeedoo> point to that one

PR #1366, opened 2020-01-23
[d3d9] Rotate swap chain back buffers

doitsujin:
Restores functionality removed in 81c3daa3d068f6a1fd2e8960efa0a109a61ebe78.

The implementation assumes that it is **not** possible for the game or DXVK itself to create additional views for swap chain back buffers (looking at the API, there doesn't seem to be a way to do this), and that the DXVK views are not queried from the D3D9 surface by the CS thread.

The framebuffer on the D3D9 device gets invalidated so that we render to the correct back buffer in the next frame. Textures are not invalidated since we can't seem to bind a `Direct3DSurface9` as a texture.

This time we also shift correctly, turns out the old implementation was wrong after all.

### DIFF
--- a/src/d3d9/d3d9_subresource.h
+++ b/src/d3d9/d3d9_subresource.h
@@ -111,6 +111,14 @@ namespace dxvk {
       return m_container;
     }
 
+    void Swap(D3D9Subresource* Other) {
+      // Only used for swap chain back buffers that don't
+      // have a container and all have identical properties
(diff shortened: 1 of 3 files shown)


Joshua-Ashton:
Tested on some demos that need frontbuffer (ATi Toyshop) and some games/traces of that need multiple backbuffers (Altairiari Sophie and One Piece Pirate Warriors 3)

PR #1382, opened 2020-01-28
Move creation of Client API HUD item to D3D9 and D3D11

doitsujin:
Since the HUD item is created after the device now, we can **finally** display D3D10 apps correctly.


src/d3d9/d3d9_swapchain.cpp
```diff
@@ -1383,4 +1385,17 @@ namespace dxvk {
...
   }
 
+
+  std::string D3D9SwapChainEx::GetApiName() {
+    Com<IDirect3D9> d3d9;
+    Com<IDirect3D9Ex> d3d9Ex;
+
+    this->GetParent()->GetDirect3D(&d3d9);
+
+    HRESULT hr = d3d9->QueryInterface(__uuidof(IDirect3D9Ex),
```
Joshua-Ashton: this->GetParent()->IsExtended()

doitsujin: Fixed.

jqadev:
@doitsujin should below description from README.md file be updated according to this PR?:
`api`: Shows the D3D feature level used by the application. Does not work correctly for D3D10 at the moment.

PR #1390, opened 2020-01-31
[GCC10] Include ostream in src/util/rc/util_rc_ptr.h

frantisekz:
Apparently, GCC 10 requires this for successful compilation[0].

[0] https://kojipkgs.fedoraproject.org//work/tasks/8740/41288740/build.log (error near the end of the file)

### DIFF
--- a/src/util/rc/util_rc_ptr.h
+++ b/src/util/rc/util_rc_ptr.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <functional>
+#include <ostream>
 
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
I would call this commit `[util] Include ostream in util_rc_ptr.h` if we want to keep things consistent.

frantisekz:
> I would call this commit `[util] Include ostream in util_rc_ptr.h` if we want to keep things consistent.

Done.

PR #1405, opened 2020-02-05
[d3d9] Copy shader metadata to shader constant structs

doitsujin:
This makes primarly `SetShaderConstants` a bit cheaper by avoiding pointer chasing, and also cleans up the code a bit. The metadata struct is small so the tradeoff should be well worth it.

### DIFF
--- a/src/d3d9/d3d9_constant_set.h
+++ b/src/d3d9/d3d9_constant_set.h
@@ -40,7 +40,7 @@ namespace dxvk {
 
   struct D3D9ConstantSets {
     Rc<DxvkBuffer>            buffer;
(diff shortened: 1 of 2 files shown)


Joshua-Ashton:
You should also change this to be used in `PrepareDraw` where it calls `UpdateBoolSpecConstantVertex` , I think:

```cpp
      if (likely(!CanSWVP())) {
        UpdateBoolSpecConstantVertex(
          m_state.vsConsts.bConsts[0] &
          GetCommonShader(m_state.vertexShader)->GetMeta().boolConstantMask);
```

PR #1431, opened 2020-02-09
[d3d9] Defer managed texture uploads until PrepareDraw and when needed

Joshua-Ashton:
Fixes perf problems in games that lock D3DPOOL_MANAGED textures multiple times in a frame without doing any draws with those textures inbetween... (ie. uploading characters to a font image 1 character at a time [Silent Hill 2], uploading lightmap faces 1 at a time [VTMB])

### DIFF
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -357,6 +357,15 @@ namespace dxvk {
             UINT                   Lod,
             VkImageUsageFlags      UsageFlags,
             bool                   Srgb);
+    D3D9SubresourceBitset& GetUploadBitmask() { return m_needsUpload; }
+
+    void SetUploading(UINT Subresource, bool uploading) { m_uploading.set(Subresource, uploading); }
+    void ClearUploading() { m_uploading.clearAll(); }
+    bool GetUploading(UINT Subresource) const { return m_uploading.get(Subresource); }
+
+    void SetNeedsUpload(UINT Subresource, bool upload) { m_needsUpload.set(Subresource, upload); }
+    bool NeedsAnyUpload() { return m_needsUpload.any(); }
+    void ClearNeedsUpload() { return m_needsUpload.clearAll();  }
 
   private:
 
@@ -391,6 +400,9 @@ namespace dxvk {
     D3D9SubresourceBitset         m_readOnly = { };
 
     D3D9SubresourceBitset         m_dirty = { };
+
+    D3D9SubresourceBitset         m_uploading = { };
+    D3D9SubresourceBitset         m_needsUpload = { };
 
     /**
      * \brief Mip level
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -6,6 +6,8 @@
 
 #include "../dxvk/dxvk_device.h"
 
+#include "../util/util_bit.h"
+
 namespace dxvk {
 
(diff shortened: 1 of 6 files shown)


src/d3d9/d3d9_common_texture.h
```diff
@@ -394,11 +395,15 @@ namespace dxvk {
...
     D3D9ColorView                 m_sampleView;
 
-    D3D9SubresourceArray<
-      bool>                       m_locked = { };
+    D3D9SubresourceBitset         m_locked = { };
 
-    D3D9SubresourceArray<
-      bool>                       m_dirty = { };
+    D3D9SubresourceBitset         m_readOnly = { };
+
```
doitsujin: Why so many newlines? :frog:

src/d3d9/d3d9_device.cpp
```diff
@@ -4672,7 +4693,28 @@ namespace dxvk {
...
+
+    UpdateManagedTextureUploadsUsed();
+  }
+
+
+  inline void D3D9DeviceEx::UpdateManagedTextureUploadsUsed() {
+    auto masks = GetShaderMasks();
+
+    // Update the bitmask for those uploads we need to do this draw.
+    m_activeTexturesToUploadUsed = m_activeTexturesToUpload & masks.samplerMask;
```
doitsujin: `&=`

src/d3d9/d3d9_device.cpp
```diff
@@ -4681,6 +4723,7 @@ namespace dxvk {
     if (tex != nullptr && tex->IsRenderTarget())
       m_activeRTTextures |= bit;
 
+    UpdateManagedTextureUploadIndex(index);
```
doitsujin: Most of the code in `UpdateManagedTextureUploadIndex` is the same as in this function (computin the index and fetching the texture pointer etc). Calling `UpdateManagedTextureUploadsUsed` for every single texture also seems unnecessary, so it would seem useful to rewrite this as
```c++
if (tex != nullptr) {
  if (unlikely(tex->IsRenderTarget())
    m_activeRTTextures |= bit;
  if (unlikely(tex->NeedsAnyUpload())
    m_activeTexturesToUpload |= bit;
}
```

Then maybe do `UpdateManagedTextureUploadsUsed` during the upload.

src/d3d9/d3d9_device.cpp
```diff
@@ -4724,6 +4767,24 @@ namespace dxvk {
   }
 
 
+  void D3D9DeviceEx::UploadManagedTextures() {
```
doitsujin: Call `UpdateManagedTextureUploadsUsed` here instead.

src/d3d9/d3d9_device.cpp
```diff
@@ -5379,6 +5440,13 @@ namespace dxvk {
         FlushBuffer(vbo);
     }
 
+    if (unlikely(m_activeTexturesToUploadUsed != 0)) {
+      UploadManagedTextures();
+
+      m_activeTexturesToUpload     = 0;
+      m_activeTexturesToUploadUsed = 0;
```
doitsujin: The `= 0` should go to `UploadManagedTextures` imho.

src/d3d9/d3d9_device.cpp
Joshua-Ashton: It's not the same var.

Joshua-Ashton:
Updated the PR.

PR #1437, opened 2020-02-11
[RFC] [d3d9] Add option to disable the explicit frontbuffer

### LINKED ISSUE #1368: [d3d9] Front/Backbuffer flipping leads to flickering in games that don't respect DISCARD SwapEffect
Thanks for dxvk, it's great :-) It looks like my FPS in a benchmark situation in ZUSI 3 are now better than on Windows (20 vs. 18, wined3d gets 14), but I need to double check.

The recent introduction of a front buffer in the swap chain causes some issues though. After looking into it, I believe the game actually doesn't redraw all pixels for every frame even though it uses the DISCARD SwapEffect. It apparently still expects the "current" backbuffer to be unchanged after Present(), even though that is not guaranteed on any system unless the COPY SwapEffect is used. Apparently it still works under most circumstances when setting the presentation params to windowed. (Exclusive fullscreen mode is available in the game settings, but heavily discouraged.)

Or maybe I'm reading the apitrace wrong / missing something and it actually DOES redraw everything, and dxvk just fails to do that successfully.

This results in strange flickering effects when using dxvk, but not wined3d and most Windows setups in windowed mode. Thankfully, someone posted a vid when he had the issue on Windows: https://www.youtube.com/watch?v=bm7Lk2O0Ink That issue eventually solved itself without changing anything in the game, probably an automatic Windows or driver update.

I could understand if this is a WONTFIX if the bug is in ZUSI, but since most Windows drivers seem to handle this pretty gracefully, maybe it would be possible to fix it in DXVK. I'm not really sure how that would work though. If it comes at a performance cost, then it shouldn't be a default behavior to imitate COPY SwapEffect behavior even when the application requests DISCARD.

I have verified that simply commenting out the "Rotate swap chain buffers" loop fixes the flickering. Also, released versions including v1.5.1 obviously don't flicker so you could call this a regression.

### Software information
ZUSI 3 - Aerosoft Edition https://store.steampowered.com/app/1040730/ZUSI_3__Aerosoft_Edition/
Pretty much default settings, but in the wine prefix / proton "dist" directory I use dsound.dll from wine-staging, a hacked iphlpapi.dll, winetricks gdiplus and some winetricks dotnet versions and fonts to fix various game issues.

### System information
- GPU: HD 7870 (GCN 1.0)
- Driver: amdgpu from XanMod kernel + radv from Ubuntu 19.10 mesa (19.2.1)
- Wine version: Proton 4.11-12, but with modified wine prefix (see last section)
- DXVK version: git 67767650 with added FlushImplicit(FALSE) in FlushBuffer() to make MANAGED vertex/index buffers load correctly in ZUSI (sync/driver bug?), that's why my version info in the log is different

### Apitrace file(s)
https://drive.google.com/file/d/1aJEv2aArCWT-1cStf5h1p3j7bmJOvqwz/view?usp=sharing . Flickering starts at frame ~370 when using dxvk.

### Log files
- d3d9.log: [ZusiSim_d3d9.log](https://github.com/doitsujin/dxvk/files/4108651/ZusiSim_d3d9.log)

w-flo:
I'm not too happy about this, especially the wording. "Explicit frontbuffer" was the best term I was able to come up with. Maybe you can suggest a better name if the patch is acceptable in general.

Maybe there are better approaches to fix the issue, but this one seemed the easiest to implement so I gave it a try, and it fixes #1368 for me.

I reproduced the same ZUSI 3 bevhior on Windows by setting the game to "DirectX fullscreen" (aka exclusive fullscreen). So I'm almost convinced now that Zusi 3 does not redraw the full screen in every frame, only regions that have changed vs. the previous frame. So there is probably nothing other than a workaround similar to this one to fix it in dxvk. Exclusive fullscreen is marked as "special use only" in the game so I guess there are no plans to fix the game itself.


Joshua-Ashton:
I would like to see if we can *actually* fix this issue without an app profile. I think it is related to the swap effect that the game is using -- currently we are entirely disregarding that and I know there are some quirks wrt. buffer count/front buffer with those (probably the same in D3D11 too @doitsujin)

w-flo:
As far as I can tell (I don't have a lot of apitrace experience though, so you may want to double check - trace is in the bug report), the game uses the DISCARD swap effect to be able to use MSAA. MSAA is only allowed when using DISCARD.

However, DISCARD requires the game to always draw complete frames: "An application that uses this swap effect should update an entire back buffer before invoking a Present operation that displays it.". So it's perfectly legal for dxvk to flip buffers, or even throw away old buffers, or fill them with random noise, when a game uses DISCARD.

ZUSI 3 apparently ignores that part and gets away with it because most Windows drivers implement "COPY" swapeffect semantics ("the [Present] operation leaves the content of the back buffer unchanged") when requesting DISCARD in windowed mode, because that is apparently the most efficient thing to do when working with the Windows compositor (?).

Initially I also assumed that the bug is caused by unimplemented "COPY" swapeffect in dxvk, but it looks like ZUSI 3 really requests DISCARD and then fails to follow DISCARD rules. This is also why the same bug occurs on Windows when setting ZUSI 3 to use exclusive fullscreen.

Joshua-Ashton:
Merged as ba41a5219ae69753b4c7106bed99d310a5b28e01

PR #1442, opened 2020-02-14
[util] Force 16:9 aspect ratio for Hyperdimension Neptunia U: Action Unleashed

ryester27:
Fixes letterboxed fullscreen when in-game.

Ran into this exact same issue in Joshua-Ashton/d9vk#460. Setting the d3d9.forceAspectRatio option also fixes this instance.

Before
Screenshot_20200213_192600

After
Screenshot_20200213_192746

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -314,6 +314,11 @@ namespace dxvk {
     { R"(\\rthdribl\.exe$)", {{
       { "d3d9.allowDoNotWait",              "False" },
     }} },
+    /* Hyperdimension Neptunia U: Action Unleashed */
+    { R"(\\Neptunia\.exe$)", {{
+      { "d3d9.forceAspectRatio",            "16:9" },
+    }} },
+
   }};
 
 


Joshua-Ashton:
Cool! Merged it in.

Cheers. :)

ryester27:
Thanks. While I have you here, I have a quick question about this source file. Let's say that there's another Neptinia.exe in the future that needs to be added. How would DXVK know which options segment to use? Since I presume that it's detecting the process name. Would something need to be done with the regex?

Joshua-Ashton:
Well that will trigger for anything named Neptunia.exe, if there is another one then we could go onto check the parent dir etc

turol:
You could also add file size or checksum checking.

ryester27:
> You could also add file size or checksum checking.

Then you would need to check and make sure that the checksum never changed due to a patch. Don't really need it for Steam releases because the game directory name in the steamapps folder never changes. Checksum might be useful for other storefronts, but we're also talking about a situation here that might never be run into.

PR #1454, opened 2020-02-16
[d3d10] Implement D3D10[Core]GetVersion and D3D10[Core]RegisterLayers

Joshua-Ashton:
I spent some time testing these functions' behaviour to make some decent stub implementations that satisfy enough for Crysis x64 to run with our D3D10 implementation.

See the test commits/files for how the signatures and return values were derived.

It was a fun little side-project anyway 🐸!

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -65,6 +65,7 @@ else
 
   lib_vulkan  = dxvk_compiler.find_library('vulkan-1', dirs : dxvk_library_path)
   lib_d3d9    = dxvk_compiler.find_library('d3d9')
+  lib_d3d10   = dxvk_compiler.find_library('d3d10')
   lib_d3d11   = dxvk_compiler.find_library('d3d11')
   lib_dxgi    = dxvk_compiler.find_library('dxgi')
   lib_d3dcompiler_43 = dxvk_compiler.find_library('d3dcompiler_43', dirs : dxvk_library_path)
--- a/src/d3d10/d3d10.def
+++ b/src/d3d10/d3d10.def
@@ -25,3 +25,4 @@ EXPORTS
     D3D10StateBlockMaskGetSetting
     D3D10StateBlockMaskIntersect
     D3D10StateBlockMaskUnion
+    D3D10GetVersion
--- a/src/d3d10/d3d10.def
+++ b/src/d3d10/d3d10.def
@@ -26,3 +26,4 @@ EXPORTS
     D3D10StateBlockMaskIntersect
     D3D10StateBlockMaskUnion
     D3D10GetVersion
+    D3D10RegisterLayers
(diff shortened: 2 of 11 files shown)


doitsujin:
Why do we need this? Crysis already runs fine.

pchome:
Meanwhile
https://github.com/doitsujin/dxvk/blob/50cf3a465b65ea40deeee8e80ecc8d5018cf70fb/src/d3d9/d3d9_adapter.cpp#L80
:wink: 

P.S. I remembered because I tried to fix an "Outdated driver version" message in ... IIRC **Underlords** under Proton. W/o success, so I gave up. This numbers looks pretty similar to driver version numbers, e.g. 10.1.4.6000 (windows10, api version 1, subversion 4, build 6000),  but who knows. 

[Ref](https://source.winehq.org/git/wine.git/blob/da816802e022e8aa8eb197de94e18305a92766c1:/dlls/wined3d/directx.c#l193)
Patch, just in case: [d3d9-adapter-report-more-real-version.patch.gz](https://github.com/doitsujin/dxvk/files/4209613/d3d9-adapter-report-more-real-version.patch.gz)

Joshua-Ashton:
@pchome Can you make a separate issue about this if it exists in `master` still? It's not really relevant to this PR 🐸

doitsujin:
I don't really see a way to accurately emulate the Windows driver versions each and every games expects to run on, especially since they should be vendor-specific.

Joshua-Ashton:
@pchome I don't think this is our issue/the issue is elsewhere for Dota Underlords given it works just fine on Windows w/ DXVK.

D3D11:
image

The slightly broken D3D9 renderer in the game because it doesnt have hero shaders compiled for D3D9:
image

Joshua-Ashton:
Dota Underlords also works fine for me in Proton with DXVK 1.5.4.

Not sure what driver version issue you are going on about.

pchome:
Oh, no issue there, just a message in the log file, so I wondered why big numbers reported as a driver version and the game still report it's too old drivers for it.
I don't think there is a problem, but if you want to check: In the very beginning of `Underlords/game/dac/console.log`, GPU profiles in `Underlords/game/bin/dxsupport.cfg`.

I mentioned "versions" and the game only because of those hex-numbers, I didn't mean any real problems.

Joshua-Ashton:
It doesn't complain it's too old, all it says is:

`02/16 22:43:26 [RenderSystem] Determined driver version for graphics adapter 0 [output 0] 'AMD Radeon VII': 32767.65535.65535.65535`

pchome:
Nvidia, I can't find old logs, they were rotated and replaced by ones from current Vulkan version installed. Not sure it worth to investigate, but if you want to, I'll launch the game via Proton.

ptrb78:
Could anyone tell how to compile dxvk with this PR?

doitsujin:
See 7a134e49be309fe8eabfe336473c60621e829ba6.

PR #1469, opened 2020-02-22
[meta] use c++latest only for old versions of MSVC

nopjmp:
c++latest is only needed for c++17 features when the compiler is before MSVC 15.3

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,16 +5,12 @@ cpu_family = target_machine.cpu_family()
 add_project_arguments('-DNOMINMAX', language : 'cpp')
 
 dxvk_compiler = meson.get_compiler('cpp')
-if dxvk_compiler.get_id() == 'msvc'
+dxvk_msvc=dxvk_compiler.get_id() == 'msvc'
+# c++17 was added in 15.3, older version needs c++latest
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Why do we want/need to support older MSVC?

nopjmp:
We don't need to, I have a diff where I set the default options in the project rather than using a variable and setting them in the individual subprojects. Would that be preferred?

Joshua-Ashton:
So you aren't getting build problems?

Then why do we need to change this, I'm confused.

nopjmp:
I'm not getting build problems because MSVC added back c++latest for C++20 support. It was breaking in older versions.

pchome:
@nopjmp 
> I have a diff where I set the default options in the project rather than using a variable and setting them in the individual subprojects. Would that be preferred?

like it was before, see https://github.com/doitsujin/dxvk/commit/9deb73a2a7007c20d306f2e88d783458b4935825#diff-969b60ad3d206fd45c208e266ccfed38

I guess developers can safely switch to c++2a and start using features available in GCC 8 (https://en.cppreference.com/w/cpp/compiler_support), such move will break MSVC support again with your PR.

P.S. AFAIK GCC 8 used by Proton in Debian Buster build environment, so I said GCC 8. Even so, this should be valid for GCC 9 and maybe GCC 10 too.

PR #1489, opened 2020-03-01
Skip display mode setting if already set

zvova7890:
That mostly taken from wine's code. Before setting new display mode, we are checking if our current display mode isn't equal new required mode, if new mode is same as current, do not call ChangeDisplaySettingsExW. Excludes display flickering if the application several times setup same display mode

### DIFF
--- a/src/dxgi/dxgi_monitor.cpp
+++ b/src/dxgi/dxgi_monitor.cpp
@@ -129,6 +129,9 @@ namespace dxvk {
       return E_FAIL;
     }
     
+    DEVMODEW currentMode = { };
+    currentMode.dmSize   = sizeof(currentMode);
+    
     DEVMODEW devMode = { };
     devMode.dmSize       = sizeof(devMode);
     devMode.dmFields     = DM_PELSWIDTH | DM_PELSHEIGHT | DM_BITSPERPEL;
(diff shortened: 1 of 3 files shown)


doitsujin:
I'm fine with the functionality change, but this introduces a non-trivial amount of code-duplication. @Joshua-Ashton what exactly was the reason again for having all the DXGI monitor code duplicated in D3D9, save struct conversion?

Also, there's an unrelated commit that's already being tracked in #1436.

src/dxgi/dxgi_monitor.cpp
```diff
@@ -142,6 +145,26 @@ namespace dxvk {
...
+    if (EnumDisplaySettingsW(monInfo.szDevice, ENUM_CURRENT_SETTINGS, &currentMode)) {
+      if (currentMode.dmPelsWidth == devMode.dmPelsWidth &&
+        currentMode.dmPelsHeight == devMode.dmPelsHeight &&
+        currentMode.dmBitsPerPel == devMode.dmBitsPerPel &&
+        (currentMode.dmDisplayFrequency == devMode.dmDisplayFrequency ||
+        !(devMode.dmFields & DM_DISPLAYFREQUENCY)) &&
+        (currentMode.dmDisplayFlags == devMode.dmDisplayFlags ||
+        !(devMode.dmFields & DM_DISPLAYFLAGS)))
+      {
+        Logger::info(str::format("DXGI: Skipping redundant mode setting call: ",
```
doitsujin: I highly doubt we need a log message to inform the user that nothing happened.

src/dxgi/dxgi_monitor.cpp
```diff
@@ -142,6 +145,26 @@ namespace dxvk {
...
     
+    /* Compare new mode with current and change if necessary */
+    if (EnumDisplaySettingsW(monInfo.szDevice, ENUM_CURRENT_SETTINGS, &currentMode)) {
+      if (currentMode.dmPelsWidth == devMode.dmPelsWidth &&
+        currentMode.dmPelsHeight == devMode.dmPelsHeight &&
+        currentMode.dmBitsPerPel == devMode.dmBitsPerPel &&
+        (currentMode.dmDisplayFrequency == devMode.dmDisplayFrequency ||
+        !(devMode.dmFields & DM_DISPLAYFREQUENCY)) &&
+        (currentMode.dmDisplayFlags == devMode.dmDisplayFlags ||
+        !(devMode.dmFields & DM_DISPLAYFLAGS)))
```
doitsujin: `(devMode.dmFields & DM_DISPLAYFLAGS)` is always `false`, so this check should probably be removed.

Joshua-Ashton:
The code duplication really only exists because I wasn't exactly sure what I needed when I implemented, and at that time the D3D9 frontend was still a fork so I couldn't just move stuff around willy nilly there. 😛

Ideally this is the sort of thing that the wsi abstraction in the native port solves...

src/dxgi/dxgi_monitor.cpp
zvova7890: Agree, but if this will have some problems, log message will help with debugging. If it's too much, I will remove

doitsujin:
Well I guess factoring out the monitor code is an option. I kind of don't want to deal with the native stuff right now.

zvova7890:
How about merge this, and then thinking about code unify? :) Code duplication is already present and will not get worse after merge

doitsujin:
> and then thinking about code unify?

We all know that that's not going to happen and instead we'll end up with bit rot.

I've been busy working on vkd3d lately so not much time to work on this.

doitsujin:
Superseded by #1494. This reworks the monitor stuff to remove code duplication, and includes the feature discussed in this PR based on your work (I added you as co-author to give credit, not sure if there's any better way to do this).

zvova7890:
Thank you, i had test you branch, looks fine

PR #1491, opened 2020-03-02
[d3d9] Clean up subresource handling 

Joshua-Ashton:
This PR cleans up the whacky undocumented allocation system for subresources

### DIFF
--- a/src/d3d9/d3d9_texture.h
+++ b/src/d3d9/d3d9_texture.h
@@ -5,11 +5,12 @@
 #include "d3d9_volume.h"
 #include "d3d9_util.h"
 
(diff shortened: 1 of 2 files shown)


doitsujin:
Not sure if this is any cleaner than the current code to be honest, since now we have something in "common" code that is *only* ever going to be useful in D3D9 texture management code and has to be used in a very specific way to even work.

I'd say the best thing to do would be to move all the weird things accessing the array into their own private methods rather than introducing the helper, and writing a few comments to explain why a regular vector isn't viable..

Joshua-Ashton:
I need to use an array like this in swapchain code too for surfaces to fix refcounting for container stuff

PR #1496, opened 2020-03-05
[d3d9] Don't use m_monitor in GetDisplayModeEx

### LINKED ISSUE #1495: Monitor rework breaks Guild Wars 2
With the monitor rework commits Guild Wars 2 crashes when the gamelauncher tries to start the client    
```
info:  D3D9DeviceEx::ResetSwapChain:
info:    Requested Presentation Parameters
info:      - Width:              3832
info:      - Height:             2133
info:      - Format:             D3D9Format::X8R8G8B8
info:      - Auto Depth Stencil: true
info:                  ^ Format: D3D9Format::D24S8
info:      - Windowed:           true
info:  Presenter: Actual swap chain properties:
info:    Format:       VK_FORMAT_B8G8R8A8_UNORM
info:    Present mode: VK_PRESENT_MODE_IMMEDIATE_KHR
info:    Buffer size:  3832x2133
info:    Image count:  3
info:    Exclusive FS: 0
err:   Failed to query monitor info
err:   D3D9SwapChainEx::GetDisplayModeEx: Failed to enum display settings
```

doitsujin:
Should fix #1495.

This looks like a bit of a mess though. Previously this would use `m_monInfo`, which itself was initialized to contain monitor info of the *default* monitor. Meanwhile, `m_monitor` is `nullptr` unless in fullscreen mode, otherwise the default monitor. Why is this so weird?

### DIFF
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -435,7 +435,7 @@ namespace dxvk {
       DEVMODEW devMode = DEVMODEW();
       devMode.dmSize = sizeof(devMode);
 
-      if (!GetMonitorDisplayMode(m_monitor, ENUM_CURRENT_SETTINGS, &devMode)) {
+      if (!GetMonitorDisplayMode(GetDefaultMonitor(), ENUM_CURRENT_SETTINGS, &devMode)) {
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
I was using it to keep track of what monitor we set display mode on etc
I mean sure that will be GetDefaultMonitor now, but... down the line

This should be OK for now, just gotta pick the right monitor there when we actually have multimonitor

Issue #1358, opened 2020-01-18
[d3d9] Dragon's Dogma: screen tearing with vsync and enabled and FIFO present mode in use


Oschowa:
Dragon's Dogma shows screen tearing despite vsync being enabled in the game settings and FIFO present mode being used according to the log file. This doesn't happen with wined3d.
d3d9.presentInterval = 1 also does not help.

### Software information
Dragon's Dogma

### System information
- GPU: RX580
- Driver: RADV/ACO 19.3.2
- Wine version: Proton 4.11-12
- DXVK version: 35a9934

### Apitrace file(s)
https://drive.google.com/open?id=11YHQBLHzjDpdxlO5Z82Llwhw0VZ7hkGN

### Log files
- d3d9.log:
[DDDA_d3d9.log](https://github.com/doitsujin/dxvk/files/4082045/DDDA_d3d9.log)

doitsujin:
There's absolutely nothing we can do to fix tearing with FIFO, since FIFO is **guaranteed** to be vsynced.

aufkrawall:
Does vsync cap the frame rate? Is this with recent xorg 1.20.7 and xf86-video-amdgpu 19.1?

Oschowa:
Yes, yes and yes. Also happens with amdvlk-pro 19.50 with `UseFlipHint,1` although the tearing seems more 'subtle'.

aufkrawall:
Ok, that's weird. The game's window is fullscreen and its size exactly matches the display's native resolution? Then I'm really out of ideas.

Oschowa:
The window definitely appears to be proper fullscreen, but the log has this:
```
info:  D3D9: Setting display mode: 1920x1080@60
info:  Presenter: Actual swap chain properties:
info:    Format:       VK_FORMAT_B8G8R8A8_UNORM
info:    Present mode: VK_PRESENT_MODE_FIFO_KHR
info:    Buffer size:  1918x1078
```
so maybe the buffer size not being equal with the window size causes problems? FWIW, windowed mode with compositor enabled works around the problem. This is also the only out of many games i tested with DXVK/D9VK which has this problem.

dlshinobi:
Hmm, for me it works as intended (nVidia + KDE compositor enabled + fullsceen mode):
```
info:  D3D9: Setting display mode: 1920x1080@60
warn:  D3D9DeviceEx::SetRenderState: Unhandled render state D3DRS_MULTISAMPLEANTIALIAS
[...]

PR #1502, opened 2020-03-09
[d3d9] Hook WM_NCCALCSIZE to get rid of fullscreen non-client areas.

rbernon:
I believe this should fix https://github.com/doitsujin/dxvk/issues/1358.

The issue is that the game restores a WS_BORDER window style after D9VK has cleared them. In theory D3D should not change window styles -or hook this WM_NCCALCSIZE fwiw- and some other internal mechanism would change full screen non-client areas size computations instead. Hooking this message seems to produce the same results.

Note that this can also be reproduced using D9VK to run the game on Windows (tested with Windows 10), and this PR also fixes that.

### DIFF
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -9,6 +9,45 @@
 
 namespace dxvk {
 
+
+  struct D3D9WindowData {
+    bool unicode;
+    WNDPROC proc;
+  };
+
+
+  static std::recursive_mutex windowProcMapMutex;
+  static std::unordered_map<HWND, D3D9WindowData> windowProcMap;
+
+
+  static D3D9WindowData GetD3D9WindowData(HWND window) {
+    std::lock_guard<std::recursive_mutex> lock(windowProcMapMutex);
+
+    auto it = windowProcMap.find(window);
+    if (it == windowProcMap.end())
+      return {};
+
+    return it->second;
+  }
+
+
+  static LRESULT CALLBACK D3D9WindowProc(HWND window, UINT message, WPARAM wparam, LPARAM lparam)
+  {
+      if (message == WM_NCCALCSIZE && wparam == TRUE)
+        return 0;
+
+      auto windowData = GetD3D9WindowData(window);
+      if (!windowData.proc)
+        return IsWindowUnicode(window)
+          ? DefWindowProcW(window, message, wparam, lparam)
+          : DefWindowProcA(window, message, wparam, lparam);
+
+      return windowData.unicode
+        ? CallWindowProcW(windowData.proc, window, message, wparam, lparam)
+        : CallWindowProcA(windowData.proc, window, message, wparam, lparam);
+  }
+
+
   static uint16_t MapGammaControlPoint(float x) {
     if (x < 0.0f) x = 0.0f;
     if (x > 1.0f) x = 1.0f;
@@ -55,6 +94,7 @@ namespace dxvk {
 
 
   D3D9SwapChainEx::~D3D9SwapChainEx() {
+    ResetWindowProc();
     RestoreDisplayMode(m_monitor);
 
     m_device->waitForSubmission(&m_presentStatus);
@@ -1251,6 +1291,14 @@ namespace dxvk {
       Logger::err("D3D9: EnterFullscreenMode: Failed to change display mode");
       return D3DERR_INVALIDCALL;
     }
+
+    // Testing shows we shouldn't hook WM_NCCALCSIZE but we shouldn't change
+    // windows style either.
+    //
+    // Some games restore window styles after we have changed it, so hooking is
+    // also required. Doing it will allow us to create fullscreen windows
+    // regardless of their style and it also appears to work on Windows.
+    HookWindowProc();
     
     // Change the window flags to remove the decoration etc.
     LONG style   = ::GetWindowLongW(m_window, GWL_STYLE);
@@ -1286,6 +1334,8 @@ namespace dxvk {
       Logger::warn("D3D9: LeaveFullscreenMode: Failed to restore display mode");
     
     m_monitor = nullptr;
+
+    ResetWindowProc();
     
     // Only restore the window style if the application hasn't
     // changed them. This is in line with what native D3D9 does.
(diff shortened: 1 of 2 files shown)


Joshua-Ashton:
Hi, this will need either a global mutex (ew.) or the unordered_map to be moved into the device (with ``auto lock = m_parent->LockDevice();`` around it.

rbernon:
Sure about the mutex. The issue of putting the map in the device is that we need to access the original window proc from within the hook, hence the global map.

Joshua-Ashton:
Hm, yeah. Might need some global mutex then which is kinda ew, but it seems like that's the only real solution. (Which is why I never touched the window proc at all so far)

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -9,6 +9,25 @@
...
+      return CallWindowProcA(windowProcMap[window], window, message, wparam, lparam);
+  }
+
+
+  static LRESULT CALLBACK WindowProcW(HWND window, UINT message, WPARAM wparam, LPARAM lparam)
+  {
+      if (message == WM_NCCALCSIZE && wparam == TRUE)
+        return 0;
+      return CallWindowProcW(windowProcMap[window], window, message, wparam, lparam);
+  }
```
Joshua-Ashton: Can we merge this? This needs some form of mutex anyway, so I am sure we can shove in some extra data long with WNDPROC to make this

```suggestion
  static LRESULT CALLBACK WindowProc(HWND window, UINT message, WPARAM wparam, LPARAM lparam) {
    auto windowData = windowMap[window];
    if (message == WM_NCCALCSIZE && wparam == TRUE)
      return 0;
    return windowData.unicode
      ? CallWindowProcW(windowData.proc, window, message, wparam, lparam)
      : CallWindowProcA(windowData.proc, window, message, wparam, lparam);
  }
```

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -1405,4 +1435,26 @@ namespace dxvk {
...
   }
 
+
+  void D3D9SwapChainEx::hookWindowProc() {
+    resetWindowProc();
+    if (IsWindowUnicode(m_window)) {
+      windowProcMap[m_window] = (WNDPROC)SetWindowLongPtrW(m_window, GWLP_WNDPROC, (LONG_PTR)WindowProcW);
+    } else {
+      windowProcMap[m_window] = (WNDPROC)SetWindowLongPtrA(m_window, GWLP_WNDPROC, (LONG_PTR)WindowProcA);
+    }
```
Joshua-Ashton:
```suggestion
    windowProcMap[m_window] = IsWindowUnicode(m_window)
      ? (WNDPROC)SetWindowLongPtrW(m_window, GWLP_WNDPROC, (LONG_PTR)WindowProc)
      : (WNDPROC)SetWindowLongPtrA(m_window, GWLP_WNDPROC, (LONG_PTR)WindowProc);
```

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -1405,4 +1435,26 @@ namespace dxvk {
...
+  void D3D9SwapChainEx::resetWindowProc() {
+    if (IsWindowUnicode(m_window)) {
+      if ((WNDPROC)GetWindowLongPtrW(m_window, GWLP_WNDPROC) == WindowProcW)
+        SetWindowLongPtrW(m_window, GWLP_WNDPROC, (LONG_PTR)windowProcMap[m_window]);
+      windowProcMap.erase(m_window);
+    } else {
+      if ((WNDPROC)GetWindowLongPtrA(m_window, GWLP_WNDPROC) == WindowProcA)
+        SetWindowLongPtrA(m_window, GWLP_WNDPROC, (LONG_PTR)windowProcMap[m_window]);
+      windowProcMap.erase(m_window);
+    }
```
Joshua-Ashton: Similar here...

src/d3d9/d3d9_swapchain.h
```diff
@@ -221,6 +221,10 @@ namespace dxvk {
 
     std::string GetApiName();
 
+    void hookWindowProc();
+
+    void resetWindowProc();
+
```
Joshua-Ashton:
```suggestion
    void HookWindowProc();

    void ResetWindowProc();

```

Joshua-Ashton:
I was more of a fan of the VK_EXT_full_screen_exclusive idea -- did the native behaviour fall through there?

rbernon:
From my investigation the VK extension is not enough. Even with it fully plugged and with vkAcquireFullScreenExclusiveModeEXT calls (as in https://github.com/rbernon/dxvk/tree/bug/17639/v1, which implements it in addition to these hooks) the 1px border is still there with Dragon's Dogma.

It seems that something in native D3D triggers an internal mechanism that makes DefWindowProc handle this WM_NCCALCSIZE message differently. I'm still investigating it to see if we can manage to trigger it ourselves, but hooking the message works in the meantime.

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -9,6 +9,43 @@
...
+
+
+  static LRESULT CALLBACK D3D9WindowProc(HWND window, UINT message, WPARAM wparam, LPARAM lparam)
+  {
+      if (message == WM_NCCALCSIZE && wparam == TRUE)
+        return 0;
+
+      auto windowData = GetD3D9WindowData(window);
+      if (!windowData.proc)
+        return 0;
```
Joshua-Ashton: ``return DefWindowProcA(...)``?

rbernon: ofc

rbernon: Although we have lost the unicode information as well...

rbernon: I used IsWindowUnicode here as well, I'm not sure why it was avoided in wined3d, maybe it's expensive. In any case this case should hopefully not happen.

rbernon:
> It seems that something in native D3D triggers an internal mechanism that makes DefWindowProc handle this WM_NCCALCSIZE message differently. I'm still investigating it to see if we can manage to trigger it ourselves, but hooking the message works in the meantime.

I was actually mixing up things. Windows 10's D3D9 changes the window styles as well, and that's what changing DefWindowProc behavior. However, restoring the styles after the fullscreen window has been created, as Dragon's Dogma does, still behaves correctly there and the rendering is done over the whole screen. There's still a internal hidden mechanism involved.

For Wine, we could mimic that in the VK extension implementation, and this message hooking would be optional, but that would make the extension implementation differ from what happens on Windows -- where this hooking is _not_ optional.

Oschowa:
I can confirm this fixes the issue with Dragon's Dogma, thanks!

doitsujin:
@Joshua-Ashton feel free to merge.

Issue #1511, opened 2020-03-15
apitrace's d3dretrace.exe crashes at d3d11.dll when trying to take a snapshot most of the time


tanty:
Replaying a captured D3D11 apitrace works well but, when trying to take a snapshot, wine crashes with a "exception 0xc0000005" most of the times (not always).

### Software information
WickedEngine Tests for x64:
https://wickedengine.net/downloads/

### System information
- Intel Kabylake, and AMD Radeon (POLARIS10)
- anv and radv
- Reproduced with 4.0-2, 4.2-2 (development), 5.0.0~buster (stable), and 5.3-buster (staging). All in Debian Buster.
- Reproduced with all the versions from 1.4.6 to 1.5.5 and master.

### Apitrace file(s)
- https://people.igalia.com/agomez/files/wicked-engine-Tests.trace

### Log files
- [d3dretrace_dxgi.log](https://github.com/doitsujin/dxvk/files/4335023/d3dretrace_dxgi.log)
- [d3dretrace_d3d11.log](https://github.com/doitsujin/dxvk/files/4335024/d3dretrace_d3d11.log)
### Wine debug log
- [wine-debug.log](https://github.com/doitsujin/dxvk/files/4335022/wine-debug.log)

doitsujin:
What exactly do you mean by "snapshot"?

tanty:
The equivalent to "apitrace dump-images". Creating a capture from a frame call.

tanty:
I should have placed the command line call directly in the description. It is in the "wine-debug.log" file:

`wine d3dretrace -S 368720 -s wicked-engine-Tests.trace- wicked-engine-Tests.trace`

doitsujin:
This appears to be a reference counting bug. Unfortunately, the last time I tried to fix that, it broke a very large number of games, despite passing most wine tests, so I'm not sure if this can be fixed without a rewrite.

Why is this important anyway? There are way better debuggers out there than apitrace.

tanty:
Just checking it for Mesa's CI in order to do pre-merge trace checking.

If I cannot use apitrace I may check to use renderdoc.

In the worst case I can just replay with apitrace and capture with LunarG's screenshot Vulkan Layer. I've checked and it seems to work, although it often skips the last frame :man_shrugging: 

Anyway, just checking if the bug was really in DXVK and was easily fixale to ponder the alternatives.

doitsujin:
Hm, I see. I guess it would be nice if that worked indeed.

[...]

PR #1512, opened 2020-03-16
[d3d11] Init D3D11ContextState in D3D11DeviceContextState's constructor

werman:
SwapDeviceContextState could be called with freshly created
D3D11DeviceContextState and RestoreState would crash if internal
D3D11ContextState is not initialized.

Closes: #1511

@tanty Does this fix the issue for you? It works for me with your trace.

### DIFF
--- a/src/d3d11/d3d11_state_object.cpp
+++ b/src/d3d11/d3d11_state_object.cpp
@@ -4,7 +4,7 @@ namespace dxvk {
 
   D3D11DeviceContextState::D3D11DeviceContextState(
           ID3D11Device*         pDevice)
-  : m_device(pDevice) {
+  : m_device(pDevice), m_state() {
 
   }
 


doitsujin:
Why exactly is this supposed to fix anything? All members of `D3D11DeviceContextState` *do* get initialized.

werman:
I cannot step into DXVK code in gdb at the moment since symbols are lost somewhere. But from what I see in the code I don't see `m_state` being initialized and initializing it prevent the crash. Where the initialization suppose to happen?

doitsujin:
The struct definition itself initializes (almost) everything.

werman:
I agree... Let me step back and look again, because it does help.

doitsujin:
I suppose there *was* a genuine bug there, but I don't think this fully solves #1511. We're now relying on use-after-free behaviour.

PR #1535, opened 2020-03-28
[documentation] improve README.md with regards to mingw64

stefson:
this makes things a bit easier to understand if you're not on a debian based distro


doitsujin:
Fair enough.

Which distro are we talking about here specifically? On Arch I never had issues with mingw's threading model.

PR #1536, opened 2020-03-28
[util] Spoof AMD GPU for GTA IV

doitsujin:
See https://github.com/ValveSoftware/Proton/issues/350#issuecomment-605420176.


mrpippy:
This allows the game to launch and run for me.
(There's frequent stuttering when VSync is enabled though, not sure if that's a Wine or DXVK problem though)

PR #1577, opened 2020-04-18
[d3d9] Fix swap chain surface ref counting

### LINKED ISSUE #1571: [d3d9] Final Fantasy 9 crashes at launch with DXVK 1253+
Title.

It works with 1252 and less.

Precision: I don't launch the game from the launcher, directly with FF9.exe (x64). The issue doesn't come from the launcher.

This is the error message:

Screenshot_20200415_142220

Crash reports from the game itself:

[error.log](https://github.com/doitsujin/dxvk/files/4481131/error.log)
[output_log.txt](https://github.com/doitsujin/dxvk/files/4481132/output_log.txt)

Option in Steam to launch the game (from protondb):
`set -- %command%; "${@:1:2}" x64/FF9.exe "${@:4}"`

### System information
- GPU: Nvidia Geforce 730M
- Driver: nvidia-vulkan 440.66.08
- Wine version: Proton 5.0-5
- DXVK version: 1253 and 1254

### Apitrace file(s)
https://mega.nz/file/v3hUAITT#23Fe7RRU5WfyxNz3wfM4R0YwB-guLiOwPXdQiH0oCG4

### Log files
[steam-377840-dxvk.log](https://github.com/doitsujin/dxvk/files/4481180/steam-377840-dxvk.log)

doitsujin:
Does away with the `unique_ptr` and `aligned_storage` for surface objects. The former is an issue for swap chains because a call to `Reset` can destroy the back buffers while the app still holds a reference to them or they are bound as a render target.

In order to avoid circular references, the container which owns the object in question will increase **only** the surface ref count, not the container's ref count.

Should fix #1571.

### DIFF
--- a/src/d3d9/d3d9_surface.cpp
+++ b/src/d3d9/d3d9_surface.cpp
@@ -31,34 +31,36 @@ namespace dxvk {
         pBaseTexture) { }
 
   void D3D9Surface::AddRefPrivate(bool RefContainer) {
-    if (m_baseTexture) {
-      D3DRESOURCETYPE type = m_baseTexture->GetType();
-      if (type == D3DRTYPE_TEXTURE)
-        static_cast<D3D9Texture2D*>(m_baseTexture)->AddRefPrivate();
-      else //if (type == D3DRTYPE_CUBETEXTURE)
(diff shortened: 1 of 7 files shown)


Joshua-Ashton:
Pushed a fix that worked with some more games with funnier refcounting

PR #1588, opened 2020-04-22
[meta] Restore .a files removal on non-devbuilds.

Tk-Glitch:
This was -supposedly mistakenly- removed with 436357e28096f5e1e25aa8b72fceb77123ea8404


doitsujin:
Derp, you're right.

PR #1635, opened 2020-05-20
[d3d9] Only allow ColorFill on surfaces in D3DPOOL_DEFAULT

CME42:
Both nine and native nvidia d3d9 seems to do this,
this fixes a crash in spacemail by xplsv.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1035,6 +1035,9 @@ namespace dxvk {
 
     D3D9CommonTexture* dstTextureInfo = dst->GetCommonTexture();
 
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -1035,6 +1035,9 @@ namespace dxvk {
 
     D3D9CommonTexture* dstTextureInfo = dst->GetCommonTexture();
 
+    if(unlikely(dstTextureInfo->Desc()->Pool != D3DPOOL_DEFAULT))
```
libcg, line 1038: should be `if (`

PR #1643, opened 2020-05-27
[dxso] Handle half texel offset for FETCH4

Joshua-Ashton:


doitsujin:
I don't like the `const*Replicant` helpers. `constuReplicant` is unused, `constfReplicant` is used literally *once*, the name sounds stupid, and we're probably not going to be using those *anywhere* else.

Joshua-Ashton:
> constfReplicant is used literally once

????

image

image

doitsujin:
sorry, i meant consti*.

Ah, well, whatever, if we have `f` anyway... just saying I don't like it very much.

Joshua-Ashton:
May as well complete the family :frog:

PR #1654, opened 2020-06-07
[vulkan] Don't use more images than necessary

haasn:
The +1 doesn't make sense for FIFO swapchains, because they behave
similarly to IMMEDIATE in that only one backbuffer is required. The
extra image only really makes sense for MAILBOX swapchains, because they
need an extra backbuffer to support flipping back and forth.

This improves latency by one frame when vsync is enabled.

### DIFF
--- a/src/vulkan/vulkan_presenter.cpp
+++ b/src/vulkan/vulkan_presenter.cpp
@@ -430,7 +430,7 @@ namespace dxvk::vk {
           uint32_t                  desired) {
     uint32_t count = caps.minImageCount;
     
-    if (presentMode != VK_PRESENT_MODE_IMMEDIATE_KHR)
+    if (presentMode == VK_PRESENT_MODE_MAILBOX_KHR)
       count = caps.minImageCount + 1;
     
     if (count < desired)


Joshua-Ashton:
Hm, do we also need to consider DXGI/D3D9 swap effect here too?

doitsujin:
I don't think this is the right thing to do since this may force double-buffering on some platforms and we **really** don't want that, especially on Linux where vsync doesn't actually synchronize the application to the display.

In fact, we're adding the extra image because not doing so made some games unplayable in the past due to exceptionally bad frame pacing.

Joshua-Ashton:
> In fact, we're adding the extra image because not doing so made some games unplayable in the past due to exceptionally bad frame pacing.

Don't we also do that in D3D11/DXGI for the front buffer though?

doitsujin:
Yes, but that was a separate problem. We're allocaing an extra Vulkan swap chain image so that `vkAcquireNextImageKHR` can return early - it's a blocking call on various drivers.

haasn:
> it's a blocking call on Mesa drivers.

Looking at the mesa WSI source code, they already add the extra buffer on their side, to the minimum swapchain size, to prevent `vkAcquireNextImageKHR` from stalling the application under 'normal' circumstances. (They set the minimum to 3 and 4 respectively for X11 and Wayland) So adding an extra buffer in dxvk adds one more than is actually needed to avoid bad stalling.

That being said, keep in mind that `vkAcquireNextImageKHR` will effectively *always* block, on essentially any driver, no matter how many images your swapchain contains, *unless* you have another mechanism of metering rendering speed. That would, however, require waiting on a fence of some sort (which I'm not sure I see your code doing?). I'm also not familiar enough with how D3D9/D3D11 swapchains work to know whether there's another mechanism in place somewhere to accomplish this.

In theory we could request as many images as we want without affecting application latency as long as we make sure we never rely on the blocking behaviour of vkAcquireNextImage.

Is there a way to debug this in practice, perhaps by viewing the swapchain state with some sort of debug overlay?

doitsujin:
> That being said, keep in mind that vkAcquireNextImageKHR will effectively always block

Thing is, it blocks the CPU, even though the API is laid out to only stall future command buffer submissions with a semaphore (unless you explicitly specify a fence and wait for that, which we're not doing).

> They set the minimum to 3 and 4 respectively for X11 and Wayland

That wasn't the case when the code was first written. Not sure what Nvidia does though, or proprietary AMD. Nvidia in particular was already *very* sensitive to an off-by-one issue in the DXGI swap chain code where we wouldn't account for the implicit front buffer.

Thing is, I **really** don't want to risk reintroducing terrible frame times for some users for no particular reason, and the behaviour we're talking about is that, when vsynced to 60 FPS, you'd get alternating present intervals of something like 30ms and 3ms on the CPU timeline. I don't think I need to explain why that makes vsync completely unusable when it happens.

haasn:
Yeah, it's understandable. I guess at the end of the day, to make a change like this we'd need a better methodology of analyzing the gains and drawbacks across a wide variety of game, including subtle factors like frame pacing.

And for the latency argument, I've come to believe that the better way to solve it is not by limiting the maximum swapchain size but rather by introducing a different mechanism to enforce waiting on the completion of previous commands before queueing new ones, so this needs to be a fundamentally different patch anyway.

> Why is this one extra frame of latency important anyway? I'd argue that if your game is really latency-sensitive, you probably shouldn't be using vsync in the first place.

I mean the ideal is to have low latency *and* no tearing. That compromise is rooted on either getting low latency, or no tearing, but never both.

aufkrawall:
I get a pleasantly low input lag when configuring  `dxgi.numBackBuffers = 2` and then force mailbox vsync via MangoHud. Though this gives me regular spikes when frame rate is dropping below refresh rate. On the other hand, that's totally broken with Nvidia anyway.

I wonder if the effect with AMD driver would be the same regarding input lag?

aufkrawall:
> I wonder if the effect with AMD driver would be the same regarding input lag?

Nope, it seems `numBackBuffers` doesn't matter much (if at all?) with AMD drivers, mailbox mode is always uncapped (as it certainly is intended).

PR #1658, opened 2020-06-09
[util] Set enableRtOutputNaNFixup for Stars End

hakzsam:


doitsujin:
I'm starting to think that it might be better to make NaN texture reads defined behaviour in Vulkan rather than piling up workarounds.

PR #1661, opened 2020-06-10
[d3d9] Implement AL16 & R16 formats

K0bin:
Apparently that weird weeb game someone complained about on Discord needs those two.


K0bin:
I originally got those from Josh (https://github.com/Joshua-Ashton/d9vk/pull/155#issuecomment-500135920) but seems like it's actually just RRRR for both.

PR #1662, opened 2020-06-10
[d3d9] Implicitly discard managed buffers to avoid stalls

K0bin:
I did some further investigating about the GH3 performance problem and tried the following:
- relying on the game to call PreLoad()
- flushing all managed textures & buffers on present
- flushing all managed textures & buffers before every draw call (Nine works that way)

Neither of those fixed the problem. Then I tried discarding managed buffers when they are currently in use and that did the trick.

Looks like Nine does something similar.
When flushing managed textures it calls `NineSurface9_UploadSelf` (https://github.com/anholt/mesa/blob/master/src/gallium/state_trackers/nine/basetexture9.c#L271).
This uses `nine_context_box_upload` (https://github.com/mesa3d/mesa/blob/master/src/gallium/frontends/nine/surface9.c#L774) which basically does a map with discard. (https://github.com/mesa3d/mesa/blob/a887ad7c84e14fdad7907037a39e9fee9d504bf3/src/gallium/frontends/nine/nine_state.c#L2531)

The changes are based on Joshs `PreLoad` implementation. (I made some changes to the last commit to IMO clean it up a little.)

Memory usage might be a problem with this approach. That's why I check if the mapped buffer is getting used atm before doing it. For most games this should hopefully result in the exact same behavior as before.
We could however restrict this to small mapping buffers but GH3 would need a limit of at least 61KB.

Risen for example (hi @DadSchoorse) doesn't discard any mapping buffers at all with those changes.
A Hat in Time discarded two buffers (in like 10 minutes of play time) with both only going up to 2 slices. One was 6.4kb, the other one 23kb.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3988,16 +3988,13 @@ namespace dxvk {
       // calling app promises not to overwrite data that is in use
       // or is reading. Remember! This will only trigger for MANAGED resources
       // that cannot get affected by GPU, therefore readonly is A-OK for NOT waiting.
-      const bool uploading = pResource->GetUploading(Subresource);
-      const bool skipWait = (managed && !uploading) || (readOnly && managed) || scratch || (readOnly && systemmem && !dirty);
+      const bool skipWait = (readOnly && managed) || scratch || (readOnly && systemmem && !dirty);
 
       if (alloced)
         std::memset(physSlice.mapPtr, 0, physSlice.length);
       else if (!skipWait) {
         if (!WaitForResource(mappedBuffer, Flags))
           return D3DERR_WASSTILLDRAWING;
-
-        pResource->ClearUploading();
       }
     }
     else {
@@ -4188,8 +4185,6 @@ namespace dxvk {
       subresource.arrayLayer, 1 };
 
     auto convertFormat = pResource->GetFormatMapping().ConversionFormatInfo;
-
-    pResource->SetUploading(Subresource, true);
 
     if (likely(convertFormat.FormatType == D3D9ConversionFormat_None)) {
       EmitCs([
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3971,7 +3971,7 @@ namespace dxvk {
       physSlice = pResource->DiscardMapSlice(Subresource);
 
       EmitCs([
-        cImageBuffer = mappedBuffer,
+        cImageBuffer = std::move(mappedBuffer),
         cBufferSlice = physSlice
       ] (DxvkContext* ctx) {
         ctx->invalidateBuffer(cImageBuffer, cBufferSlice);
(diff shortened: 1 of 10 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -4314,15 +4328,30 @@ namespace dxvk {
       const bool skipWait = (Flags & D3DLOCK_NOOVERWRITE) ||
                             quickRead                     ||
                             (boundsCheck && !pResource->DirtyRange().Overlaps(pResource->LockRange()));
-
       if (!skipWait) {
-        if (!(Flags & D3DLOCK_DONOTWAIT)) {
-          pResource->SetReadLocked(false);
-          pResource->DirtyRange().Clear();
-        }
+        if (desc.Pool & D3DPOOL_MANAGED) {
```
Joshua-Ashton, line 4332: ``IsPoolManaged(desc.Pool)``

src/d3d9/d3d9_device.cpp
```diff
@@ -4314,15 +4328,30 @@ namespace dxvk {
...
+          if (!WaitForResource(mappingBuffer, D3DLOCK_DONOTWAIT)) {
+            // if the mapped buffer is currently being copied to image and it's small enough
```
Joshua-Ashton, line 4334: not an image this time

src/d3d9/d3d9_device.cpp
```diff
@@ -3992,7 +3992,21 @@ namespace dxvk {
 
       if (alloced)
         std::memset(physSlice.mapPtr, 0, physSlice.length);
-      else if (!skipWait) {
+      else if (managed && !skipWait) {
+        if (!WaitForResource(mappedBuffer, D3DLOCK_DONOTWAIT)) {
```
Joshua-Ashton, line 3996: maybe we should take lock flags out of WaitForResource and just make it a BOOL

src/d3d9/d3d9_device.cpp
```diff
@@ -4314,15 +4328,30 @@ namespace dxvk {
...
+            EmitCs([
+              cBuffer = std::move(mappingBuffer),
+              cBufferSlice = physSlice
+            ] (DxvkContext* ctx) {
+              ctx->invalidateBuffer(cBuffer, cBufferSlice);
+            });
+          }
+        } else {
+          if (!(Flags & D3DLOCK_DONOTWAIT)) {
+            pResource->SetReadLocked(false);
```
Joshua-Ashton, line 4348: This needs a change but I can do that after the fact...

K0bin: What do you mean?

src/d3d9/d3d9_device.cpp
K0bin: Would probably make sense to rename the function as well if we do that, given that it doesn't really wait every time.

src/d3d9/d3d9_device.cpp
Joshua-Ashton: this logic needs to be changed slightly, it can be made better + i think it should trigger in the other path?

src/d3d9/d3d9_device.cpp
Joshua-Ashton: ye but we can do that after

src/d3d9/d3d9_device.cpp
Joshua-Ashton: nvm ill fix this when it gets merged in

Joshua-Ashton: actually nah add this to the other path please for now.

PR #1688, opened 2020-07-14
[dxgi] Add dxgi.customDeviceDesc option

FLGX06:
I thought also thought of setting RX 480 for NvAPI workaround but because that's enabled by default some people would be confused by games showing RX 480 on NVIDIA hardware

### DIFF
--- a/src/dxgi/dxgi_adapter.cpp
+++ b/src/dxgi/dxgi_adapter.cpp
@@ -242,13 +242,23 @@ namespace dxvk {
     auto memoryProp = m_adapter->memoryProperties();
     auto deviceId   = m_adapter->devicePropertiesExt().coreDeviceId;
     
-    // Custom Vendor / Device ID
+    // Custom Vendor / Device ID / Description
     if (options->customVendorId >= 0)
       deviceProp.vendorID = options->customVendorId;
     
     if (options->customDeviceId >= 0)
       deviceProp.deviceID = options->customDeviceId;
     
+    if (!options->customDeviceDesc.empty())
+      #ifdef __GNUC__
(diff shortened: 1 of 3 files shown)


src/dxgi/dxgi_adapter.cpp
```diff
@@ -242,13 +242,23 @@ namespace dxvk {
...
     
     if (options->customDeviceId >= 0)
       deviceProp.deviceID = options->customDeviceId;
     
+    if (!options->customDeviceDesc.empty())
+      #ifdef __GNUC__
+      #pragma GCC diagnostic push
+      #pragma GCC diagnostic ignored "-Wstringop-truncation"
+      #endif // __GNUC__
+      std::strncpy(deviceProp.deviceName, options->customDeviceDesc.c_str(), sizeof(deviceProp.deviceName));
```
libcg, line 257: `sizeof(deviceProp.deviceName) - 1` should allow removing the ifdefs

FLGX06: Done

doitsujin:
Why? Is this required for any particular game?

FLGX06:
> Why? Is this required for any particular game?
> 
> I don't really want to add useless options that no one is going to use anyway "just because". Games will already only show RX 480 if they have an internal list based on PCI IDs.

It is not needed for any game I know about, just saw that d3d9 has this option. So feel free to close this PR

K0bin:
D3D9 has this option because some games (namely Sims 2) rely on the device name for some stupid reason.

FLGX06:
Yep, I saw that

PR #1702, opened 2020-07-25
[build] Add ccache support

gasinvein:
Starting with meson 0.51 or so, it supports multiple `--cross-file` args and list values for `c` and `cpp` values in cross-files, so we can add supplementary cross-files that would override `c` and `cpp`, prepending `ccache` binary to actual compiler.
This adds a new option `--ccache` for `package-release.sh` to enable ccache.

### DIFF
--- a/build-win32-ccache.txt
+++ b/build-win32-ccache.txt
@@ -1,0 +1,3 @@
+[binaries]
+c = ['ccache', 'i686-w64-mingw32-gcc']
+cpp = ['ccache', 'i686-w64-mingw32-g++']
--- a/build-win64-ccache.txt
+++ b/build-win64-ccache.txt
@@ -1,0 +1,3 @@
+[binaries]
+c = ['ccache', 'x86_64-w64-mingw32-gcc']
+cpp = ['ccache', 'x86_64-w64-mingw32-g++']
--- a/package-release.sh
+++ b/package-release.sh
@@ -23,6 +23,7 @@ shift 2
 
 opt_nopackage=0
 opt_devbuild=0
+opt_ccache=0
 
 crossfile="build-win"
 
@@ -34,6 +35,9 @@ while [ $# -gt 0 ]; do
   "--dev-build")
     opt_nopackage=1
     opt_devbuild=1
+    ;;
+  "--ccache")
+    opt_ccache=1
     ;;
   *)
     echo "Unrecognized option: $1" >&2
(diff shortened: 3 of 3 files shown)


Joshua-Ashton:
Doesn't Meson automatically use ccache if it's available?

gasinvein:
It does, but not for cross-compilers.

Thaodan:
Why just don't set CXX=cacche/bin/c++ when using ccache? Looks like this adds unnecessary complications to me.

gasinvein:
@Thaodan Because it won't work for cross-compiler, see mesonbuild/meson#7147

doitsujin:
Is this really important enough to clutter the repo with *even more* files? DXVK compiles in less than two minutes here.

gasinvein:
@doitsujin Not really, I just thought it might be useful for someone.

kassindornelles:
i think it would be useful for me because i have a core2quad CPU, but i dont really think theres much people out there compiling DXVK or anything else in such old CPU's in these days

Endle:
Is it necessary to change project config?
On my computer I have
```
$ which i686-w64-mingw32-gcc
/usr/lib64/ccache/i686-w64-mingw32-gcc
```

`ccache` is triggered when compileing dxvk

orbea:
> Because it won't work for cross-compiler

This seems entirely false. I have ccache enabled.
```
$ which x86_64-w64-mingw32-g++ i686-w64-mingw32-g++
/usr/local/bin/x86_64-w64-mingw32-g++
/usr/local/bin/i686-w64-mingw32-g++

$ ls -l /usr/local/bin/x86_64-w64-mingw32-g++ /usr/local/bin/i686-w64-mingw32-g++ 
lrwxrwxrwx 1 root root 6 Nov 10 10:31 /usr/local/bin/i686-w64-mingw32-g++ -> ccache
lrwxrwxrwx 1 root root 6 Nov 10 10:31 /usr/local/bin/x86_64-w64-mingw32-g++ -> ccache

$ ls -l /usr/local/bin/ccache 
lrwxrwxrwx 1 root root 15 Dec  3  2017 /usr/local/bin/ccache -> /usr/bin/ccache

Then I compile dxvk 3 times in a row (Includes time spent extracting the tarball, configuring meson, compiling both 32 and 64-bit dxvk and creating a package).
```
real 71.67
user 72.08
sys 27.33

real 32.72
user 49.57
sys 24.12


real 32.13
user 49.15
sys 24.11
```
And once more with `export CCACHE_DISABLE=1`:
```
real 363.05
user 1234.38
sys 147.83
```
Seems my ccache was already somewhat used in the first build...

gasinvein:
@orbea My comment was about using env vars to enable ccache, like `CXX=ccache i686-...`. If you've added a wrapper for `i686-w64-mingw32-g++` command to your `PATH`, surely it would work since meson just executes whatever it points to. But such wrappers are not always there for mingw.

orbea:
> But such wrappers are not always there for mingw.

Its trivial to add them to the user's `$PATH`.

gasinvein:
Fine, it's easy to workaround indeed, and hopefully cross-ccache handling will be fixed on meson side.

Issue #1419, opened 2020-02-07
GTA IV Frame skipping / jumping lag


NikoBellicRU:
I dont know if you can see this in the video but you can clearly see it in the frametime graph.
https://www.youtube.com/watch?v=clxQTkD14y0&feature=youtu.be

Also for some reason there's a bug with gta4Browser.exe that its constantly using 25% of the cpu and really drops performance without stoping it or kill it soon as i enter the game, not sure if its wine/proton problem tho.

- GPU: GTX 670
- Driver: 440.59
- Wine version: proton 4.11-12
- DXVK version: 1.5.4

[steam-12210.log](https://github.com/doitsujin/dxvk/files/4173034/steam-12210.log)
[GTAIV_d3d9.log](https://github.com/doitsujin/dxvk/files/4173036/GTAIV_d3d9.log)
[LaunchGTAIV_d3d9.log](https://github.com/doitsujin/dxvk/files/4173038/LaunchGTAIV_d3d9.log)

doitsujin:
Low performance is expected on weak hardware, especially if your CPU is weak.

NikoBellicRU:
Where is that file and whats the name? Also i played this game with d9vk before it was merged with dxvk a while ago and it waas working great i played with all very high settings and now with medium settings that wierd lag happends

telans:
I encountered this a few days ago, but didn't think this was DXVK's issue as I also have this issue in Modern Warfare Remastered.

I don't think specs are the issue here, I have an i7-4770 & 980 Ti (not the best, but good enough for GTA IV)

Iglu47:
@NikoBellicRU use this instruction for dx9 games https://github.com/Joshua-Ashton/d9vk/wiki/Making-a-Trace#on-linux 

instead editing setting via `winecfg` try use environment variable through `set launch option...` for steam games, example 
`WINEDLLOVERRIDES="d3d9,d3d11,d3d10,d3d10core,d3d10_1,dxgi=n,b" DXVK_HUD=full %command%`

NikoBellicRU:
https://drive.google.com/file/d/1RDlcqGlgY2yCfETqUI6YoFlpTJDFd9ML/view?usp=sharing
https://drive.google.com/file/d/17h4ugjTxXDnDvhBDWCmlA1-aHKDnIivO/view?usp=sharing

ViNi-Arco:
[...]

PR #1713, opened 2020-08-09
[d3d9] Do Discard in UpdateSurface/-Texture when there were stalls for the source texture

K0bin:
Ref #1419

GTA IV locks the same D3DPOOL_SYSTEMMEMORY texture multiple times per frame without DISCARD or NOOVERWRITE and uses that to update multiple D3DPOOL_DEFAULT textures with UpdateTexture.
That means that we have to sync with the GPU and wait for the previous copies to finish on subsequent Lock calls.

The way to work around this is to make another copy of the data in UpdateTexture so we're free to overwrite anything in the source texture. I implemented this by doing a simple Lock(Discard), memcpy, unlock. That would however mean more CPU work which obviously isn't great for D3D9 games.

So we could:
* Just always do it like this.
* Add another config option and enable it specifically for GTA IV.
* Detect stalls on system memory textures and turn it on for those

I implemented option 3 as a POC here. Adding even more complexity to both locking and the texture class is pretty awful but so would be adding another config option. The actual implementation pretty much copies what you've done with queries except that I put the threshold much lower because it didnt trigger otherwise in apitrace.

My PR does a Discard Lock on the destination texture in UpdateTexture. Alternatively you could allocate a new slice for the source buffer in UpdateTexture or do the same thing in Lock similar to what we're doing for managed textures.

### DIFF
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -366,6 +366,19 @@ namespace dxvk {
     void PreLoadAll();
     void PreLoadSubresource(UINT Subresource);
 
+    bool IsStalling() const {
+      return m_stallFlag;
+    }
+
+    void NotifyLocked() {
+      m_stallMask <<= 1;
+    }
+
+    void NotifyStall() {
+      m_stallMask |= 1;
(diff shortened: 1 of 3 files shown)


K0bin:
I have a better idea on how to fix this.

PR #1733, opened 2020-08-16
[d3d9] Respect D3DLOCK_NOOVERWRITE for pools other than DEFAULT

K0bin:
Should hopefully improve #1730 

Konosuba uses a big buffer and locks part of that with NOOVERWRITE before almost every draw call.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4276,11 +4276,10 @@ namespace dxvk {
     if (unlikely((Flags & (D3DLOCK_DISCARD | D3DLOCK_NOOVERWRITE)) == (D3DLOCK_DISCARD | D3DLOCK_NOOVERWRITE)))
       Flags &= ~D3DLOCK_DISCARD;
 
-    // Ignore DISCARD and NOOVERWRITE if the buffer is not DEFAULT pool (tests + Halo 2)
     // The docs say DISCARD and NOOVERWRITE are ignored if the buffer is not DYNAMIC
     // but tests say otherwise!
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -4280,7 +4280,7 @@ namespace dxvk {
     // The docs say DISCARD and NOOVERWRITE are ignored if the buffer is not DYNAMIC
```
414n, line 4280: Should probably change the previous comment line (not in the patch context) accordingly too

K0bin: Not really. We never respected the docs here and what the docs say hasn't changed.

414n: Sorry, I think I was ambiguous. What I was referring to was line 4279, where there is a description of what's going to happen practically.

K0bin: Ah, I see what you mean. You're right, Im gonna change that.

Joshua-Ashton:
This breaks Halo CE (flickering).

I can't reproduce the performance problems on either Nvidia or AMD in the Konosuba game.

Joshua-Ashton:
(I checked HaloCE and Halo 2 and all their D3DPOOL_SYSTEMMEM buffers are D3DUSAGE_WRITEONLY | D3DUSAGE_DYNAMIC)

I'll need to go through and check if renaming the exe/changing the hash breaks this on native. Maybe they have a hack for it.

K0bin:
Maybe it's based on the usage instead of the pool?

PR #1736, opened 2020-08-19
[dxvk] Add (debug) suffix to debug builds in HUD

colonal-cpu:
I think it looks cool


doitsujin:
"looks cool" isn't really a good justification for code bloat to be honest. Is people accidentally using debug builds even an issue in practice, given that neither Lutris nor Proton ship them?

PR #1750, opened 2020-08-26
Fix a crash when calling Create*Shader with limited stack space

Joshua-Ashton:
Fixes a crash with a weird nGlide demo.


doitsujin:
Do we really need to work around a dumb app being compiled with questionable settings?

Joshua-Ashton:
Turns out it wasnt compiled with anything questionable, its just very limited by the time it calls us.

It works with WineD3D and Native so I feel we should possibly also have it work here given its only minimal changes.

doitsujin:
Feel free to merge then. I'm just not a huge fan of having crap code to work around weird app behaviour.

PR #1753, opened 2020-09-02
d3d9: Send necessary messages for full-screen windows during WM_ACTIVATEAPP

g-insn:
Some fullscreen games such as `Heroes of Might and Magic V' have a black screen when restored from being minimized (out of focus). The patch has the same behavior that wined3d does here. With this patch the game restores properly even with DXVK, not just wined3d.

### DIFF
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -14,6 +14,7 @@ namespace dxvk {
     bool unicode;
     bool filter;
     WNDPROC proc;
+    D3D9SwapChainEx* swapchain;
   };
 
 
@@ -82,7 +83,7 @@ namespace dxvk {
   }
 
 
-  void HookWindowProc(HWND window) {
+  void HookWindowProc(HWND window, D3D9SwapChainEx* swapchain) {
     std::lock_guard lock(g_windowProcMapMutex);
 
     ResetWindowProc(window);
@@ -94,6 +95,7 @@ namespace dxvk {
       CallCharsetFunction(
       SetWindowLongPtrW, SetWindowLongPtrA, windowData.unicode,
         window, GWLP_WNDPROC, reinterpret_cast<LONG_PTR>(D3D9WindowProc)));
+    windowData.swapchain = swapchain;
 
     g_windowProcMap[window] = std::move(windowData);
   }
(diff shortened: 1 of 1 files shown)


jqadev:
@Joshua-Ashton is there a chance to merge this PR before the next release?

jqadev:
@doitsujin is there a problem with these changes that they weren't merged yet?

aeikum:
Our QA found that this commit breaks alt-tabbing out of A Hat In Time. Quote from QA:

Found a windowing regression with a hat in time. The game now minimizes if you alt+tab in fullscreen mode (it used to stay open and other windows would go on top when they gained focus) but when you alt+tab again to regain focus, the screen in blown up and only a small portion is visible. Using alt+enter twice will restore the screen to look normal. Able to reproduce on ubuntu+gnome, and on kde as well.

Reverting the commit restores the old behavior.

g-insn:
What happens if you remove the else {} block with ShowWindow(window, SW_MINIMIZE), does it solve the issue?

That's technically not needed for Heroes of Might and Magic V to work, either—I just took it from wined3d, assuming it's correct. But if it causes a regression it should be left out.

jqadev:
@aeikum IMHO it is good idea to report an issue, since you already reversed this commit in Valve's fork https://github.com/ValveSoftware/dxvk/commit/85c70ad5db0863ffbc4afee2ca57a6e6e92e8ef6

jqadev:
@doitsujin are you going to revert this commit as well?

g-insn:
Again, what happens if you remove just the else {} block that minimizes the window, does it solve the problem? Instead of reverting the whole patch.

Joshua-Ashton:
@aeikum Alt-tabbing in Hat in Time works fine for me on KDE. I've tried at multiple resolutions.

Alt-tabbing does make the game minimise when fullscreen now, but I don't see why that's an issue. That behaviour happens on Windows too with the game.

Joshua-Ashton:
However, I think I understand why. Please try with latest master.

alasky17:
@Joshua-Ashton I did some testing, and here are the results.  For reference, the bug with the "send necessary messages" commit was that it would regain focus in a super zoomed in way (see screenshot, I'll refer to this as "zoom bug").  You are correct that minimizing on focus loss in fullscreen, non-native mode is the Windows behavior and that part was an improvement :)

1) Proton 5.13-5 with "send necessary messages" commit still has "zoom bug" - game update did not change behavior.
2) Proton 5.13-6 with "send necessary messages" commit does NOT have "zoom bug".  minimizes on focus loss and regains focus correctly
3) Proton 5.13-6 checked out to dxvk master tip (with "send necessary messages" commit and "[d3d9] Don't minimise in WM_ACTIVATEAP" commit) does NOT have "zoom bug".  also does NOT minimize on focus loss (just like all past behavior from 5.13-4 and earlier)
4) Proton 5.13-6 checked out to dxvk master tip with "[d3d9] Don't minimise in WM_ACTIVATEAP" commit reverted (but *with* "send necessary messages" commit) does NOT have "zoom bug" and does minimize correctly on focus loss.
5) Proton 5.13-6 with new fullscreen commits from 5.13-6 reverted and "send necessary messages commit" applied - has "zoom bug" -- proves those commits fixed "zoom bug" in Proton 5.13-6

In summary, the problem with the original "send necessary messages" commit with Proton 5.13-5 was inadvertently resolved with some fullscreen hack fixes for DOOM Eternal in 5.13-6 (specifically commit 16de1f8d19d9fe92a02e5d5f6bdf611e69e6999f and e839e7f6b632217e76215e4b9ad23a27f92afa0d).  My opinion would be to use the dxvk master without "[d3d9] Don't minimise in WM_ACTIVATEAP" during the next Proton release testing and confirm that the commit is not needed.  If you think that "[d3d9] Don't minimise in WM_ACTIVATEAP" is actually correct, then I believe it would not regress things further to keep both commits and just revert the revert of the "send necessary messages" commit in the Proton dxvk branch.  
hat-in-time-zoom-bug

PR #1759, opened 2020-09-07
[d3d9] Track dirty regions for UpdateTexture

### LINKED ISSUE #1757: [d3d9] Battle Engine Aquila → broken textures
WineD3D:
battle-engine-aquila-wined3d

DXVK:
battle-engine-aquila-dxvk

### System information
- GPU: Nvidia Geforce 730M
- Driver: 450.66
- Wine version: Proton 5.0-9 
- DXVK version: 1347

### Apitrace file(s)
https://mega.nz/file/j6JwmIra#MSxWfQyvzIKfcA0cMBkTk4YUnZS4ZEEq6x-wht7VHko

### Log files
[steam-1346400-wined3d.log](https://github.com/doitsujin/dxvk/files/5180089/steam-1346400-wined3d.log)
[steam-1346400-dxvk.log](https://github.com/doitsujin/dxvk/files/5180088/steam-1346400-dxvk.log)

K0bin:
Fixes #1757

Battle Engine Aquila relies on us not copying anything outside of the dirty regions that were set by `LockRect` and `AddDirtyRect`.

This could technically either be considered a game bug if the documentation is to be believed:
> If the source texture has dirty regions, the copy can be optimized by restricting the copy to only those regions. It is not guaranteed that only those bytes marked dirty will be copied.

This has serious potential for regressions and I have not yet tested it with a variety of games. I'd like to get some feedback on the implementation first.

### DIFF
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -366,6 +366,40 @@ namespace dxvk {
     void PreLoadAll();
     void PreLoadSubresource(UINT Subresource);
 
+    void AddUpdateDirtyBox(CONST D3DBOX* pDirtyBox, uint32_t layer) {
+      if (pDirtyBox) {
+        D3DBOX box = *pDirtyBox;
+        if (box.Right <= box.Left
+          || box.Bottom <= box.Top
+          || box.Back <= box.Front)
+          return;
+
+        D3DBOX& updateBox = m_updateDirtyBoxes[layer];
+        if (updateBox.Left == updateBox.Right) {
+          updateBox = box;
+        } else {
+          updateBox.Left    = std::min(updateBox.Left,   box.Left);
+          updateBox.Right   = std::max(updateBox.Right,  box.Right);
+          updateBox.Top     = std::min(updateBox.Top,    box.Top);
+          updateBox.Bottom  = std::max(updateBox.Bottom, box.Bottom);
+          updateBox.Front   = std::min(updateBox.Front,  box.Front);
+          updateBox.Back    = std::max(updateBox.Back,   box.Back);
+        }
+      } else {
+        m_updateDirtyBoxes[layer] = { 0, 0, m_desc.Width, m_desc.Height, 0, m_desc.Depth };
+      }
+    }
+
+    void ClearUpdateDirtyBoxes() {
+      for (uint32_t i = 0; i < m_updateDirtyBoxes.size(); i++) {
+        m_updateDirtyBoxes[i] = { 0, 0, 0, 0, 0, 0 };
+      }
+    }
+
+    const D3DBOX& GetUpdateDirtyBox(uint32_t layer) const {
+      return m_updateDirtyBoxes[layer];
+    }
+
   private:
 
     D3D9DeviceEx*                 m_device;
@@ -405,6 +439,8 @@ namespace dxvk {
     bool                          m_needsMipGen = false;
 
     D3DTEXTUREFILTERTYPE          m_mipFilter = D3DTEXF_LINEAR;
+
+    std::array<D3DBOX, 6>         m_updateDirtyBoxes;
 
     /**
      * \brief Mip level
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5016,6 +5016,7 @@ namespace dxvk {
   
   void D3D9DeviceEx::MarkTextureMipsDirty(D3D9CommonTexture* pResource) {
     pResource->SetNeedsMipGen(true);
+    pResource->MarkAllDirty();
 
     for (uint32_t tex = m_activeTextures; tex; tex &= tex - 1) {
       // Guaranteed to not be nullptr...
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -646,6 +646,19 @@ namespace dxvk {
     if (unlikely(srcTextureInfo->Desc()->Format != dstTextureInfo->Desc()->Format))
       return D3DERR_INVALIDCALL;
 
+    if (dst->GetMipLevel() == 0) {
+      if (pSourceRect) {
+        D3DBOX updateDirtyRect = { UINT(pSourceRect->left), UINT(pSourceRect->top), UINT(pSourceRect->right), UINT(pSourceRect->bottom), 0, 1 };
+        if (pDestinationSurface) {
+          updateDirtyRect.Left = pDestPoint->x;
+          updateDirtyRect.Top = pDestPoint->y;
+        }
+        dstTextureInfo->AddUpdateDirtyBox(&updateDirtyRect, dst->GetFace());
+      } else {
+        dstTextureInfo->AddUpdateDirtyBox(nullptr, dst->GetFace());
+      }
+    }
+
     const DxvkFormatInfo* formatInfo = imageFormatInfo(dstTextureInfo->GetFormatMapping().FormatColor);
 
     VkOffset3D srcBlockOffset = { 0u, 0u, 0u };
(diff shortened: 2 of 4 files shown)


Joshua-Ashton (changes requested):
Also needs to handle:
 - Autogen mipmaps (should probably dirty everything?)
 - Render target/depth stencil
 - pInitialData

Does native actually hold a list of boxes to update? Or does it just grow a single box?

K0bin:
pInitialData is handled (notice I set the whole texture dirty in the constructor of CommonTexture).

Aside from that, I've implemented it the way the docs describe it:
> You can dirty a portion of a texture by locking it, or by calling one of the following methods.
>    IDirect3DCubeTexture9::AddDirtyRect
>    IDirect3DTexture9::AddDirtyRect
>    IDirect3DVolumeTexture9::AddDirtyBox
>    IDirect3DDevice9::UpdateSurface

https://docs.microsoft.com/en-us/windows/win32/api/d3d9helper/nf-d3d9helper-idirect3dtexture9-adddirtyrect

Why does auto mip gen matter? We only track mip 0.

Joshua-Ashton:
I am mainly thinking of the case where someone calls UpdateTexture after mips have been generated.

K0bin:
>If pSourceTexture is a non-autogenerated mipmap and pDestinationTexture is an autogenerated mipmap, only the topmost matching level is updated, and the destination sublevels are regenerated. All other source sublevels are ignored.
>If both pSourceTexture and pDestinationTexture are autogenerated mipmaps, only the topmost matching level is updated. The sublevels from the source are ignored and the destination sublevels are regenerated.
>If pSourceTexture is an autogenerated mipmap and pDestinationTexture a non-autogenerated mipmap, UpdateTexture will fail.

K0bin:
Adjusted how auto mip gen textures are handled in UpdateTexture according to the docs above.

src/d3d9/d3d9_common_texture.h
```diff
@@ -366,6 +366,42 @@ namespace dxvk {
...
+        } else {
+          updateBox.Left    = std::min(updateBox.Left,   box.Left);
+          updateBox.Right   = std::max(updateBox.Right,  box.Right);
+          updateBox.Top     = std::min(updateBox.Top,    box.Top);
+          updateBox.Bottom  = std::max(updateBox.Bottom, box.Bottom);
+          updateBox.Front   = std::min(updateBox.Front,  box.Front);
+          updateBox.Back    = std::max(updateBox.Back,   box.Back);
+        }
+      } else {
+        m_updateDirtyBoxes[layer] = { 0, 0, m_desc.Width, m_desc.Height };
```
Joshua-Ashton, line 391: should be 0, 0, width, height, 0, depth?

based on:

typedef struct D3DBOX {
  UINT Left;
  UINT Top;
  UINT Right;
  UINT Bottom;
  UINT Front;
  UINT Back;
} D3DBOX, *LPD3DBOX;

src/d3d9/d3d9_common_texture.h
```diff
@@ -366,6 +366,42 @@ namespace dxvk {
...
+    void AddUpdateDirtyBox(CONST D3DBOX* pDirtyBox, uint32_t layer) {
+      if (pDirtyBox) {
+        D3DBOX box = *pDirtyBox;
+        if (box.Right <= box.Left
+          || box.Bottom <= box.Top
+          || box.Back <= box.Front)
+          return;
+
+        D3DBOX& updateBox = m_updateDirtyBoxes[layer];
+        if (updateBox.Left == 0 && updateBox.Right == 0
```
Joshua-Ashton, line 378: This degenerate test could be simpler.

We could just do Left == Right probably

src/d3d9/d3d9_device.cpp
```diff
@@ -723,39 +736,67 @@ namespace dxvk {
...
+        VkExtent3D levelExtent = dstImage->mipLevelExtent(m);
+        VkExtent3D blockCount  = util::computeBlockCount(levelExtent, formatInfo->blockSize);
+        VkOffset2D srcByteOffset = {
+          offset.x / int32_t(formatInfo->blockSize.width),
+          offset.y / int32_t(formatInfo->blockSize.height)
+        };
+        VkDeviceSize srcOffset = srcByteOffset.y * formatInfo->elementSize * blockCount.width
+                                  + srcByteOffset.x * formatInfo->elementSize;
+        VkExtent2D srcExtent = VkExtent2D{ blockCount.width  * formatInfo->blockSize.width,
+                                              blockCount.height * formatInfo->blockSize.height };
```
Joshua-Ashton, line 775: spacing

src/d3d9/d3d9_device.cpp
```diff
@@ -723,39 +736,67 @@ namespace dxvk {
...
+          alignDown(uint32_t((box.Back - box.Front) >> m), formatInfo->blockSize.depth),
+        };
+
+        VkExtent3D levelExtent = dstImage->mipLevelExtent(m);
+        VkExtent3D blockCount  = util::computeBlockCount(levelExtent, formatInfo->blockSize);
+        VkOffset2D srcByteOffset = {
+          offset.x / int32_t(formatInfo->blockSize.width),
+          offset.y / int32_t(formatInfo->blockSize.height)
+        };
+        VkDeviceSize srcOffset = srcByteOffset.y * formatInfo->elementSize * blockCount.width
```
Joshua-Ashton, line 772: Do we need to do anything different here for 3D textures?

src/d3d9/d3d9_common_texture.h
Joshua-Ashton: Doesn't this also need to account for mip size?

src/d3d9/d3d9_device.cpp
```diff
@@ -723,39 +736,67 @@ namespace dxvk {
       return D3DERR_INVALIDCALL;
 
     const Rc<DxvkImage> dstImage  = dstTexInfo->GetImage();
-      
+    const DxvkFormatInfo* formatInfo = imageFormatInfo(dstTexInfo->GetFormatMapping().FormatColor);
     uint32_t mipLevels   = std::min(srcTexInfo->Desc()->MipLevels, dstTexInfo->Desc()->MipLevels);
     uint32_t arraySlices = std::min(srcTexInfo->Desc()->ArraySize, dstTexInfo->Desc()->ArraySize);
+
+    if (unlikely(srcTexInfo->IsAutomaticMip() && !dstTexInfo->IsAutomaticMip()))
```
Joshua-Ashton, line 743: I'm scared of this

Joshua-Ashton (changes requested):

src/d3d9/d3d9_device.cpp
Joshua-Ashton: Docs seem to say this is correct so I guess we can go with it for now...

Docs have lied in the past however :frog:

src/d3d9/d3d9_device.cpp
Joshua-Ashton: Do we need anything more for srcOffset on 3D images?

src/d3d9/d3d9_common_texture.h
K0bin: > Doesn't this also need to account for mip size?

Tracking only happens for mip 0 (according to the Microsoft documentation)

Joshua-Ashton: Oh yeah, I was mistaking layer for mip and not face.

I think we should still do the 0, depth thing though.

src/d3d9/d3d9_device.cpp
doitsujin: Write a test then.

src/d3d9/d3d9_common_texture.h
K0bin: This is not a degenerate test. It tests whether to use the entire box passed in.

src/d3d9/d3d9_device.cpp
K0bin: FWIW, Nine does the same thing.
https://gitlab.freedesktop.org/mesa/mesa/-/blob/master/src/gallium/frontends/nine/device9.c#L1460

Joshua-Ashton:
Going to merge this so it can get some testing.

PR #1761, opened 2020-09-10
Properly handle non-ASCII characters in paths

ishitatsuyuki:
Fixes the mojibake in log/cache files when running executables with CJK characters in their name. Tested on Windows 10.

### DIFF
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -5,10 +5,13 @@
 namespace dxvk::env {
 
   std::string getEnvVar(const char* name) {
-    char* result = std::getenv(name);
-    return (result)
-      ? result
-      : "";
+    std::vector<WCHAR> result;
+    result.resize(MAX_PATH + 1);
+
+    DWORD len = ::GetEnvironmentVariableW(str::tows(name).c_str(), result.data(), MAX_PATH);
+    result.resize(len);
+
+    return str::fromws(result.data());
   }
   
   
--- a/src/d3d11/d3d11_shader.cpp
+++ b/src/d3d11/d3d11_shader.cpp
@@ -27,7 +27,7 @@ namespace dxvk {
     const std::string dumpPath = env::getEnvVar("DXVK_SHADER_DUMP_PATH");
     
     if (dumpPath.size() != 0) {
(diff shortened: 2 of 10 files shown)


Joshua-Ashton:
I have a feeling wchar nonsense may get interesting on Wine and I'm not a big fan of the ``tows``/``fromws`` spam.

Might it be easier to use something like ``wformat``?

Converting to/from all the time is usually a bad idea in general.
Maybe make getenv just return wstring instead of converting it back and forth to UTF-8? Still kinda 😬 though.

Bleh, why can't everything be UTF-8.

ishitatsuyuki:
> I have a feeling wchar nonsense may get interesting on Wine and I'm not a big fan of the tows/fromws spam.

What do you mean by this?

I agree that `tows` got a little bit too verbose in this, but in this case conversions are fine as they are not on hot path.

Making `getenv` to return wstring seemed to require much more refactors which I didn't want to do. After all wchar is mostly useless on platforms other than Windows, given that some implementations use 4 bytes for wchar, so I thought using UTF-8 strings would be better for readability.

ishitatsuyuki:
I'm using MinGW cross-compile from a Linux box as described in the README. I think the C++ standard library calling the ANSI versions of API is perfectly intended; "char" on Windows means ANSI, and UTF-8 is not used in any of the native Windows APIs (except the multibyte conversion routines). In C++20 there's a new type called u8char_t that actually makes things UTF-8, and the reason it's a distinct type is presumably due to Windows.

Anyway, are you OK with this approach of keeping most things in UTF-8, even if that means a conversion? In addition to that, do you have other complaints regarding the code style (mainly repetitive `tows`, `c_str()`)?

doitsujin:
I guess we can merge this as-is, since there doesn't really seem to be a better way - it's just a stupid problem, and it makes me angry that we are forced to write shit code to work around issues on a shit platform.

src/util/util_env.cpp
```diff
@@ -5,10 +5,13 @@
 namespace dxvk::env {
 
   std::string getEnvVar(const char* name) {
-    char* result = std::getenv(name);
-    return (result)
-      ? result
-      : "";
+  std::vector<WCHAR> result;
```
doitsujin, line 8: indentation here is wrong.

Issue #1553, opened 2020-04-06
[d3d11] Just Cause 3 flickering terrain


BlazeKl:
Just Cause 3 has some flickering terrain, changing settings in game makes no difference. This also happen on both LLVM and ACO compilers

The trace (recorded on WIN10) replays with some HUD errors for me but it reproduces the terrain bug

### Software information
Just Cause 3, max settings

### System information
- GPU: AMD RX 470 4GB
- Driver: Mesa-tkg 20.1.0-devel (git-8970b7839a)
- Wine version: Proton 5.0-5
- DXVK version: 1.6-18-g00d371d7

### Apitrace file(s)
- https://drive.google.com/open?id=1HaHi6CG4V3rPMo2_imprxXKEtlfyeaYZ

### Log files (using Wine's DXGI)
- d3d11.log: [JustCause3_d3d11.log](https://github.com/doitsujin/dxvk/files/4435300/JustCause3_d3d11.log)

doitsujin:
See #613 .

doitsujin:
Actually, nevermind, this seems to be a separate issue. Looks like a RADV bug to me since it doesn't happen with AMDVLK.

BlazeKl:
I have only noticed HUD problems on the trace, everything else is rendered properly, even the terrain bug on RADV, I used apitrace's dlls directly from its github page, it's there any fork that works better for this case?

doitsujin:
Not really; sometimes it just bugs out and you can't do much about it.

Can you test whether amdvlk actually fixes the bug in-game on your end?

BlazeKl:
it seems that it only happens on RADV, AMDVLK and -PRO don't have the bug, where should I report bugs to mesa?

aufkrawall:
It's already reported:
https://gitlab.freedesktop.org/mesa/mesa/issues/851
Another bump likely wouldn't hurt though.

doitsujin:
Wasn't that *yet another* problem? What I see when replaying the apitrace is that the terrain geometry is glitching randomly.

PR #1762, opened 2020-09-11
Workaround flickering terrain for Just Cause 3 on RADV

pendingchaos:
A compute shader seems to expect out-of-bounds temporary array writes to be dropped. Out-of-bounds access seems to be undefined with D3D11 and SPIR-V, so this PR adds and uses an option to add bounds-checking to the writes

For small arrays, Mesa clamps the index, which causes flickering terrain in Just Cause 3: https://gitlab.freedesktop.org/mesa/mesa/-/issues/851

### DIFF
--- a/src/dxbc/dxbc_compiler.cpp
+++ b/src/dxbc/dxbc_compiler.cpp
@@ -387,6 +387,7 @@ namespace dxvk {
       m_xRegs.resize(regId + 1);
     
     m_xRegs.at(regId).ccount = info.type.ccount;
+    m_xRegs.at(regId).alength = info.type.alength;
     m_xRegs.at(regId).varId  = emitNewVariable(info);
     
     m_module.setDebugName(m_xRegs.at(regId).varId,
@@ -4724,28 +4725,7 @@ namespace dxvk {
   
   DxbcRegisterPointer DxbcCompiler::emitGetIndexableTempPtr(
     const DxbcRegister&           operand) {
-    // x# regs are indexed as follows:
-    //    (0) register index (immediate)
-    //    (1) element index (relative)
-    const uint32_t regId = operand.idx[0].offset;
-    
-    const DxbcRegisterValue vectorId
-      = emitIndexLoad(operand.idx[1]);
-    
-    DxbcRegisterInfo info;
-    info.type.ctype   = DxbcScalarType::Float32;
-    info.type.ccount  = m_xRegs[regId].ccount;
-    info.type.alength = 0;
-    info.sclass       = spv::StorageClassPrivate;
-    
-    DxbcRegisterPointer result;
-    result.type.ctype  = info.type.ctype;
-    result.type.ccount = info.type.ccount;
-    result.id = m_module.opAccessChain(
-      getPointerTypeId(info),
-      m_xRegs.at(regId).varId,
-      1, &vectorId.id);
-    return result;
+    return getIndexableTempPtr(operand, emitIndexLoad(operand.idx[1]));
   }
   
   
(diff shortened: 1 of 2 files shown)


pendingchaos:
Should close https://github.com/doitsujin/dxvk/issues/1553

Joshua-Ashton (approved):
LGTM

aufkrawall:
Can confirm this fixes the issue on Polaris. Thx @pendingchaos !

src/dxbc/dxbc_compiler.h
```diff
@@ -79,12 +79,14 @@ namespace dxvk {
   struct DxbcRegisterPointer {
     DxbcVectorType    type;
     uint32_t          id;
+    uint32_t          boundsCheck;
```
doitsujin, line 82: I'm really not a big fan of adding a member that's only ever going to be used by one thing.

src/dxbc/dxbc_compiler.cpp
```diff
@@ -4745,6 +4747,13 @@ namespace dxvk {
       getPointerTypeId(info),
       m_xRegs.at(regId).varId,
       1, &vectorId.id);
+
+    if (boundsCheck) {
+      result.boundsCheck = m_module.opULessThan(
+        m_module.defBoolType(), vectorId.id,
+        m_module.constu32(m_xRegs.at(regId).alength));
+    }
```
doitsujin, line 4755: Probably better to move this into emitRegisterStore instead so that we don't need the new struct member.

src/dxbc/dxbc_options.cpp
```diff
@@ -41,6 +41,7 @@ namespace dxvk {
     enableRtOutputNanFixup   = options.enableRtOutputNanFixup;
     zeroInitWorkgroupMemory  = options.zeroInitWorkgroupMemory;
     forceTgsmBarriers        = options.forceTgsmBarriers;
+    boundsCheckTempStores    = options.boundsCheckTempStores;
```
doitsujin, line 44: We should enable this unconditionally since it clearly works in D3D11, even though this is an application bug.

src/dxbc/dxbc_compiler.cpp
pendingchaos: How should I do that?

Should I add a `emitIndexLoad(operand.idx[1])` in `emitRegisterStore`, create a `getIndexableTempPtr(const DxbcRegister& operand, DxbcRegisterValue vectorId)` helper or add a output `vectorId` parameter to `emitGetIndexableTempPtr`?

doitsujin: Actually, don't we have to consume the OpAccessChain in the same block that contains the instruction for the SPIR-V to be valid?

pendingchaos: I don't see any requirement in the SPIR-V spec. There is a requirement that OpSampledImage instructions are in the same block which they are consumed though

doitsujin: Right, that shouldn't be much of a concern then (at least in theory, in practice I wouldn't want to rely on it since drivers tend to be buggy with code patterns that aren't generated by glslang)

> Should I add a emitIndexLoad(operand.idx[1]) in emitRegisterStore, create a getIndexableTempPtr(const DxbcRegister& operand, DxbcRegisterValue vectorId) helper

I suppose that's the better option.

src/dxbc/dxbc_options.cpp
pendingchaos: Done

doitsujin:
Looks good, and thanks for figuring out the cause of this issue.

Issue #1764, opened 2020-09-11
Need for Speed Heat missing ground texture


Linubuser:
NFS HEAT missing ground texture with DXVK 1.7.1 as per attached pictures.

What I tried:

1. Proton 5.0-9 with DXVK 1.7.1 = missing texture
2. Windows 10 with DXV 1.7.1 = missing texture
3. Windows 10 with D3D11 = no problem,

I created the API trace with the 3 files provided with pierre-loup. Hopefully this will be useful. 
Thanks for the great work!!

### Software information
NFS Heat, any settings

### System information
- GPU: Nvidia GTX 970
- Driver: 450.66
- Wine version: Proton 5.0-9
- DXVK version: 1.7.1 

### Apitrace file(s)
[https://drive.google.com/file/d/1ru2Cq-xLxPre0ghI8ncR9EnNnKdelOaZ/view?usp=sharing](url)
[https://drive.google.com/file/d/1zm7GmifS9ByPy6UZXHTH8z1WySfe_0vb/view?usp=sharing](url)
[https://drive.google.com/file/d/1Y-CveCEb-VvLZr_e4rtXjDJIvOw8f6G8/view?usp=sharing](url)

### Log files
[https://drive.google.com/file/d/1J9z2Gy9ZONaLZbaGnbT5yIK6okvS8uoA/view?usp=sharing](url)
[https://drive.google.com/file/d/15ycs4RkzA6pWIfQWqBqSILrWsO_rNHlx/view?usp=sharing](url)
[https://drive.google.com/file/d/1VYXqqZEA33zipNqiKJwmpHflHsweeLVV/view?usp=sharing](url)

DXVK
<img width="960" alt="DXVK 1 7 1" src="https://user-images.githubusercontent.com/38225213/92981095-741b2900-f466-11ea-86b6-a7dfff2fb8a5.png">

D3D11
<img width="960" alt="DX11" src="https://user-images.githubusercontent.com/38225213/92981108-84cb9f00-f466-11ea-910e-f9d02da08b98.png">

Joshua-Ashton:
I'm confused as to what these log files are because they certainly aren't ours.

I've never seen ``HackerSwapchain`` before but Google tells me it might be related to Nvidia 3D Vision or something called Helix Mod.

Please disable any mods/graphics API hooks you might be using.

Linubuser:
Humm I don't have any mods installed. 

[...]

PR #1765, opened 2020-09-14
[d3d11] Clamp bound constant buffer range

K0bin:
NFS Heat binds a 1048576 byte constant buffer.
So as discussed on Discord, we just bind the range of it thats legal.

```
VUID-VkWriteDescriptorSet-descriptorType-00332(ERROR / SPEC): msgNum: -1974202905 - Validation Error: [ VUID-VkWriteDescriptorSet-descriptorType-00332 ] Object 0: handle = 0x31ad1d0000107dcb, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x8a540de7 | vkUpdateDescriptorSetWithTemplate() failed write update validation for VkDescriptorSet 0x31ad1d0000107dcb[] with error: Write update to VkDescriptorSet VkDescriptorSet 0x31ad1d0000107dcb[] allocated with VkDescriptorSetLayout VkDescriptorSetLayout 0xcadcd00000000364[] binding #0 failed with error message: Attempted write update to buffer descriptor failed due to: For buffer VkBuffer 0x9fae5200000008cf[] VkDescriptorBufferInfo range is 1048576 which is greater than this device's maxUniformBufferRange (65536). The Vulkan spec states: If descriptorType is VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER or VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC, the range member of each element of pBufferInfo, or the effective range if range is VK_WHOLE_SIZE, must be less than or equal to VkPhysicalDeviceLimits::maxUniformBufferRange (https://vulkan.lunarg.com/doc/view/1.2.148.0/windows/1.2-extensions/vkspec.html#VUID-VkWriteDescriptorSet-descriptorType-00332)
    Objects: 1
        [0] 0x31ad1d0000107dcb, type: 23, name: NULL
```

Unfortunately this does not fix the issue in #1764.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -3498,18 +3498,6 @@ namespace dxvk {
 
   void D3D11DeviceContext::BindConstantBuffer(
           UINT                              Slot,
-          D3D11Buffer*                      pBuffer) {
-    EmitCs([
-      cSlotId      = Slot,
-      cBufferSlice = pBuffer ? pBuffer->GetBufferSlice() : DxvkBufferSlice()
-    ] (DxvkContext* ctx) {
-      ctx->bindResourceBuffer(cSlotId, cBufferSlice);
-    });
-  }
-  
-  
-  void D3D11DeviceContext::BindConstantBuffer1(
-          UINT                              Slot,
           D3D11Buffer*                      pBuffer,
           UINT                              Offset,
           UINT                              Length) {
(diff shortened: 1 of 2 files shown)


src/d3d11/d3d11_context.cpp
```diff
@@ -3624,28 +3624,34 @@ namespace dxvk {
...
+       || Bindings[StartSlot + i].constantCount  != constantCount) {
         Bindings[StartSlot + i].buffer         = newBuffer;
         Bindings[StartSlot + i].constantOffset = 0;
-        Bindings[StartSlot + i].constantCount  = constantBound;
-        Bindings[StartSlot + i].constantBound  = constantBound;
-        
-        BindConstantBuffer(slotId + i, newBuffer);
+        Bindings[StartSlot + i].constantCount  = constantCount;
+
+        if (likely(constantCount <= UINT(D3D11_REQ_CONSTANT_BUFFER_ELEMENT_COUNT))) {
```
doitsujin, line 3642: Why do we need this `if`? We should be able to just set `constantCount` to the proper value in line 3634.

K0bin: We need the if because otherwise it calls BindConstantBuffer (without the 1) which binds the entire slice.

doitsujin: Just always call 1 instead?

Anyway, gonna do some testing since I want to make sure this is even the right thing to do.

K0bin: Updated the PR to just use BindConstantBuffer1.

doitsujin:
I did some more testing and found that:
- SetConstantBuffers will only bind the first 64 KiB, and GetConstantBuffers1 will return 4096 constants as bound.
- SetConstantBuffers1 will return rather than clamp if an invalid buffer range is explicitly provided, and behave the same as SetConstantBuffers if no range is provided.

Fixed in 4801fbe0980bb7e062b3223a10e57d56ed5c0aab.

PR #1781, opened 2020-09-28
[util]: Set enableRtOutputNaNFixup for Art of Rally

pendingchaos:
Fixes the (Windows) Art of Rally demo.

See https://gitlab.freedesktop.org/mesa/mesa/-/issues/3562

I got the filename for the full version from here: https://steamdb.info/app/550320/config/ . I'll ask on the GitLab whether the workaround works for the full version.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -254,6 +254,10 @@ namespace dxvk {
     { R"(\\The Convenience Store Yakin Jiken\.exe$)", {{
       { "d3d11.enableRtOutputNanFixup",     "True" },
     }} },
+    /* Art of Rally                               */
+    { R"(\\artofrally(_demo)?\.exe$)", {{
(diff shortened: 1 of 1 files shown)


src/util/config/config.cpp
```diff
@@ -254,6 +254,13 @@ namespace dxvk {
     { R"(\\The Convenience Store Yakin Jiken\.exe$)", {{
       { "d3d11.enableRtOutputNanFixup",     "True" },
     }} },
+    /* Art of Rally                               */
+    { R"(\\artofrally_demo\.exe$)", {{
```
DadSchoorse, line 258: `{ R"(\\artofrally(_demo)?\.exe$)", {{` should make this one if.

pendingchaos: Done

Issue #1845, opened 2020-12-06
Performance tanks in Mass Effect when looking at occluded part of the level


mirh:
So, I like to benchmark the game in the comm room on the Normandy, since I found out (once the door closes) that I can get the highest framerates of all the game there. 
I just walk against the FTL comm link at the bottom of it, and then stare at the empty wall. 
In D3D9 I get around 420 fps, while dxvk gives me around 350, not that bad I thought. 

I just figured out though, that if I just rotate the camera to "face the rest of the ship", while D3D9 only slows down to about 360fps, dxvk gets almost *3* times slower at 130. 

### Software information
Mass Effect 1, max in-game settings, no mods. 

### System information
- GPU: Radeon R9 290
- Driver: 20.11.2
- Wine version: Windows 7 SP1
- DXVK version: 1.7.3

### Apitrace file(s)
- [Apitrace](https://mega.nz/file/BN4zVYQI#t720f9B95CccQKNstiqRRlM8CfGdHCdWrNWIFsZ1yyk)

### Log files
- [d3d9.log](https://github.com/doitsujin/dxvk/files/5649612/MassEffect_d3d9.log)

doitsujin:
Are you CPU-bound or GPU-bound? That kind of info is **extremely** important when dealing with performance issues.

Also, be aware that DXVK is known not to perform well on first- and second-gen GCN cards due to some limitations around how they handle depth buffers.

mirh:
I think it's actually both in the apitrace you can find. 
I lowered the resolution to 640x480, and D3D9 wasn't actually affected at all (even though I could have swore gpu usage at 97% meant it was gpu-bound). 

But then this had a big effect on DXVK. When I'm staring at the wall I gain those 70fps that I was missing earlier. 
While I'm staring behind my shoulders, I'm still getting the same ~130 fps.

K0bin:
I can confirm that it's slower when looking at the wall that has geometry behind it and the hit is much larger than with Windows D3D9. Windows D3D9 pretty much runs the same and DXVK drops from 550 fps to 430 for me.

Fully CPU bound of course. I dont really know why though. There's nothing that obviously sticks out.

K0bin:
Can you try this?

[d3d9.zip](https://github.com/doitsujin/dxvk/files/5691568/d3d9.zip)

mirh:
This is enormously better. Performance hit is around 20% now, which is *more or less* the same ballpark of the one you get with native D3D9.

[...]

PR #1857, opened 2020-12-15
D3D9: Query stuff for Mass Effect 1

K0bin:
Mass Effect 1 issues occlusion queries early in the frame and reads those in the next one.  So I extended the stall tracking to other query types but only do an implicit flush to make sure those queries are ready.
In the apitrace provided in #1845 you can see how the game is waiting for the first query of the batch to finish.

It also retrieves the data of its queries multiple times without issuing them inbetween, so I cache those results to avoid having to go through multiple Vulkan queries and the Vulkan event again.

And for the last commit, I moved the stalling query flush after the call to End (the way it is for D3D11 queries) so the query is actually done by the time we flush. :frog:

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -6008,11 +6008,13 @@ namespace dxvk {
   void D3D9DeviceEx::End(D3D9Query* pQuery) {
     D3D9DeviceLock lock = LockDevice();
 
+    pQuery->NotifyEnd();
     if (unlikely(pQuery->IsEvent())) {
-      pQuery->NotifyEnd();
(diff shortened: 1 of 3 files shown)


Joshua-Ashton:
Some questions:

- Does the game really hit on the cached path?
- I'm interested in seeing perf effects in a few other games.
  - The flipping of End makes perfect sense, but I want to make sure we don't regress perf in other games with the stall check; some games I know spam queries are AHiT and Halo CE.

K0bin:
I removed the strong hint from the FlushImplicit call.

K0bin:
Okay a simple FlushImplicit should be fairly safe and apparently still does the trick.

PR #1886, opened 2021-01-05
Using wineopenxr to support OpenXR

aeikum:
Two commits here to support OpenXR with DXVK.

The first commit extends the interop interfaces to provide a couple more methods that our OpenXR wrapper needs. We need the queue index, not the queue handle, for [XrGraphicsBindingVulkanKHR](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#_xrgraphicsbindingvulkankhr3), so I added a new method to retrieve that. Also, [OpenXR returns VkImages](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#xrEnumerateSwapchainImages) which the application should render to. We need a method to wrap that in an ID3D11Texture to return to the application in [XrSwapchainImageD3D11KHR](https://www.khronos.org/registry/OpenXR/specs/1.0/man/html/openxr.html#XrSwapchainImageD3D11KHR).

_(Review note: Is this the correct way to extend the interop interfaces? How should the new IID be generated?)_

The second commit adds a new extension provider for OpenXR. Unfortunately SteamVR's OpenXR implementation is very picky. Notably, you cannot create two XrInstances simultaneously. That means we can't use the method we use for OpenVR where we create and destroy an instance, as the application may have already created an instance. For that reason, I added two methods that are specific to wineopenxr to retrieve the required Vulkan instance and device extensions. That means this support is limited to wineopenxr and won't work with other OpenXR implementations.

Are you interested in merging this, or something like it? If not, it could instead live in the Proton DXVK fork.

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -2531,7 +2531,8 @@ namespace dxvk {
       return S_OK;
     }
     
-    if (riid == __uuidof(IDXGIVkInteropDevice)) {
+    if (riid == __uuidof(IDXGIVkInteropDevice)
+     || riid == __uuidof(IDXGIVkInteropDevice1)) {
       *ppvObject = ref(&m_d3d11Interop);
       return S_OK;
     }
--- a/src/d3d11/d3d11_interop.cpp
+++ b/src/d3d11/d3d11_interop.cpp
@@ -104,4 +104,63 @@ namespace dxvk {
     m_device->GetDXVKDevice()->unlockSubmission();
   }
   
+  
+  void STDMETHODCALLTYPE D3D11VkInterop::GetSubmissionQueue1(
+          VkQueue*              pQueue,
+          uint32_t*             pQueueIndex,
+          uint32_t*             pQueueFamilyIndex) {
+    auto device = static_cast<D3D11Device*>(m_device)->GetDXVKDevice();
+    DxvkDeviceQueue queue = device->queues().graphics;
+    
+    if (pQueue != nullptr)
+      *pQueue = queue.queueHandle;
+    
+    if (pQueueIndex != nullptr)
+      *pQueueIndex = queue.queueIndex;
+    
+    if (pQueueFamilyIndex != nullptr)
+      *pQueueFamilyIndex = queue.queueFamily;
+  }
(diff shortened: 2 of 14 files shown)


src/d3d11/d3d11_device.cpp
```diff
@@ -2531,7 +2531,8 @@ namespace dxvk {
       return S_OK;
     }
     
-    if (riid == __uuidof(IDXGIVkInteropDevice)) {
+    if (riid == __uuidof(IDXGIVkInteropDevice)
+     || riid == __uuidof(IDXGIVkInteropDevice2)) {
```
doitsujin, line 2535: To keep this consistent with Microsoft's weird D3D naming schemes, this should be `IDXGIVkInteropDevice1`, and `GetSubmissionQueue1`, respectively.

doitsujin (commented):
Looks good overall besides one naming nit (see below), and I think we can merge this just like the OpenVR code.

Are there any (demo) apps using OpenXR to test this with?

> (Review note: Is this the correct way to extend the interop interfaces? How should the new IID be generated?)

Yes, it is correct. I think most of our GUIDs are the result of typing "random guid" into DuckDuckGo, so anything that doesn't conflict with existing D3D interfaces is fine really.

aeikum:
Thanks. I will change the iface and method names. I'll leave the IID as it is unless you think it should change.

[hello_xr](https://github.com/KhronosGroup/OpenXR-SDK-Source/tree/master/src/tests/hello_xr) is the best sample app right now, and Microsoft Flight Simulator is the only released game that I know of that uses OpenXR. However, you'll need wineopenxr and also some changes to winevulkan, which I haven't made public anywhere yet. I'm planning to make an Experimental build very soon (today?) that will include it.

aeikum:
I've updated that branch with those changes. The build with OpenXR support is now available on the Proton experimental branch, including source in the usual places.

ChristophHaag:
Very nice work. A few more open source d3d11 OpenXR apps/samples:

https://github.com/microsoft/OpenXR-MixedReality
https://github.com/maluoi/OpenXRSamples
https://github.com/maluoi/StereoKit

PR #1887, opened 2021-01-06
[d3d11] Fix device ref counting from queries.

gofman:
Fixes AO Tennis 2 crash on exit.

The crash is due to DxvkGpuQueryAllocator::freeQuery is called after the DxvkGpuQueryAllocator has been already freed, which is probably due to missing reference counting from d3d11 query to d3d11 device (device gets destroyed while there are still queries created from it). The similar crash is reproducible on Windows with dxvk.

### DIFF
--- a/src/d3d11/d3d11_query.cpp
+++ b/src/d3d11/d3d11_query.cpp
@@ -126,7 +126,7 @@ namespace dxvk {
   
   
   void STDMETHODCALLTYPE D3D11Query::GetDevice(ID3D11Device **ppDevice) {
-    *ppDevice = ref(m_device);
+    *ppDevice = m_device.ref();
   }
   
   
--- a/src/d3d11/d3d11_query.h
+++ b/src/d3d11/d3d11_query.h
@@ -96,7 +96,7 @@ namespace dxvk {
     
   private:
     
-    D3D11Device* const m_device;
+    const Com<D3D11Device> m_device;
     D3D11_QUERY_DESC1  m_desc;
 
     D3D11_VK_QUERY_STATE m_state;


src/d3d11/d3d11_query.cpp
```diff
@@ -126,7 +126,7 @@ namespace dxvk {
   
   
   void STDMETHODCALLTYPE D3D11Query::GetDevice(ID3D11Device **ppDevice) {
-    *ppDevice = ref(m_device);
```
Joshua-Ashton, line 129: I don't get this change.

src/d3d11/d3d11_query.h
```diff
@@ -96,7 +96,7 @@ namespace dxvk {
     
   private:
     
-    D3D11Device* const m_device;
+    const Com<D3D11Device> m_device;
```
Joshua-Ashton, line 99: Are we sure this should be ``Com<D3D11Device>`` and not ``Com<D3D11DeviceChild, false>`` for a private ref?

src/d3d11/d3d11_query.cpp
gofman: ref(m_device) won't compile after m_device is Com<D3D11Device>. Or am I missing something obvious?

src/d3d11/d3d11_query.h
gofman: > Are we sure this should be `Com<D3D11Device>` and not `Com<D3D11DeviceChild, false>` for a private ref?

I suggested it similar to how it looks like for, e. g., D3D11Buffer, any reason it should be different? I might have missed something here though, can't say I really went deep into dxvk ref counting while thought this is rather obvious case. If this is actually more involved maybe worth dropping this request in favour of a better solution.

src/d3d11/d3d11_query.cpp
Joshua-Ashton: Nope, I just forgot about that quirk.

src/d3d11/d3d11_query.h
Joshua-Ashton: I'd imagine D3D11Buffer using public refs is probably wrong too. I will test this later today.

K0bin: D3D11Buffer inherits from ID3D11DeviceChild. Would be very weird if that didnt hold a public reference imo.

gofman: I've just added the test to Wine tests for CreateBuffer specifically and tested on Windows, looks like the buffer holds a visible reference to device: https://gist.github.com/gofman/18b57f2e5907fc04cee60ff8e9462626

Joshua-Ashton: Alright then, this lgtm!

Joshua-Ashton (approved):

PR #1901, opened 2021-01-20
[d3d9] Early updates for textures

K0bin:
Mass Effect locks 2 textures every frame and those are always dirty so the game has to queue a copy and synchronize the gpu immediately afterwards.

So I propose copying the UpdateMappedBufferEarly approach from D3D11 and updating the mapping buffer after most writing operations (except if it's used as an RT). I put it behind a super simple heurstic though..

@mirh Please test if this PR improves performance for you.

### DIFF
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -300,9 +300,19 @@ namespace dxvk {
       return m_type;
     }
 
+    bool ShouldUpdateMappedBufferEarly() const {
+      // The texture has been locked very frequently.
+      // This indicates that it gets locked every frame
+      // and we should try our hardest to avoid the late copy to the mapping buffer.
+      return m_totalLockCount > 256;
+    }
+
     const D3D9_VK_FORMAT_MAPPING& GetMapping() { return m_mapping; }
 
-    void SetLocked(UINT Subresource, bool value) { m_locked.set(Subresource, value); }
+    void SetLocked(UINT Subresource, bool value) {
+      m_totalLockCount += 1;
+      m_locked.set(Subresource, value);
+    }
 
     bool GetLocked(UINT Subresource) const { return m_locked.get(Subresource); }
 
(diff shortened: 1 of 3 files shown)


mirh:
I'm not really in the bestest of situations to compile myself the thing.

Joshua-Ashton:
I only implemented deferred texture uploads because some games depend on it because they keep writing after unlock or use the image while locked.

Might be easier just to disable that behaviour for this game?

mirh:
It doesn't make the slightest difference.

K0bin:
And I still can't reproduce that.

At 4k, the game runs less than 20% slower with DXVK.

pchome:
@K0bin 
> Mass Effect locks 2 textures every frame and those are always dirty so the game has to queue a copy and synchronize the gpu immediately afterwards.

Is this behavior the game or engine specific?

I downloaded a random UE3 benchmark (PLA Benchmark) and it shows ~1%  avg. FPS improvement in d3d9 mode with this patch. So should be bigger improvement in some specific scenes, I guess. I don't know where to look at.

mirh:
What if you try 8K with vsr?

K0bin:
4k (4x) is the highest I can get with VSR.

mirh:
Try SSAA then I guess (hoping driver can force it in vulkan too)?
And can't you use some older version of Nsight, or renderdoc?

K0bin:
> Try SSAA then I guess (hoping driver can force it in vulkan too)?

No it can't.

>  And can't you use some older version of Nsight, or renderdoc?

No version ever supported Vulkan GPU profiling on anything older than Turing.

Nothing I can do here. Just buy new hardware.

mirh:
[Mhh I see](https://forums.developer.nvidia.com/t/profile-vulkan-on-non-turing-class-gpu/115527), and anyhow even their [raw api](https://github.com/baldurk/renderdoc/pull/930) doesn't seem to expose vulkan.... but do you really need hardware counters here? Tracy, optick or renderdoc may be already good enough here perhaps. 

As for DSR, it's your [lucky](https://www.forum-3dcenter.org/vbulletin/showthread.php?t=593534) [day](https://framedsc.github.io/GeneralGuides/custom_dsr_resolutions.htm).

K0bin:
> but do you really need hardware counters here? Tracy, optick or renderdoc may be already good enough here perhaps.

Why don't you just go ahead and optimize it then. I'm not gonna waste anymore time on a game that already runs at 500 fps on my system.

mirh:
[Well, I tried](https://github.com/baldurk/renderdoc/issues/2157)

PR #1907, opened 2021-01-22
Workaround F1 2020 artifact bug.

giomasce:
This PR is an attempt to fix the artifacts in F1 2020. While developing it I was not aware of the TGSM barrier workaround, and I basically discovered the same trick independently. However, the TGSM barrier workaround as it is currently implemented in DXVK does not completely fix the issue, although it goes really close, so I would like to propose my solution as well. My current patch is definitely not ready for inclusion, so I am writing to know from DXVK developers their opinion on accepting a more complete solution for the F1 shader bugs.

### Context: what is the problem with the buggy shaders
I did this analysis on F1 2020; I guess the same is happening for earlier versions of F1, but I am not sure. A compute shader used to preprocess lighting divides the scene into a number of 16x16 tiles, and would like to compute for each tile the minumum (and maximum, but let us focus on the minimum) of an array of 64 values (which happen to be depth values for that tile). A thread group is spawned for each tile, and 64 threads are spawned in each thread group. In the interesting section, the 64 threads are executing along the lines of this pseudo-code:
```c
float x[64];    // the array for which we want to compute the minimum, which is in TGSM
if (thread_id < 32) {
    x[thread_id] = min(x[thread_id], x[thread_id+32]);
    x[thread_id] = min(x[thread_id], x[thread_id+16]);
    x[thread_id] = min(x[thread_id], x[thread_id+8]);
    x[thread_id] = min(x[thread_id], x[thread_id+4]);
    x[thread_id] = min(x[thread_id], x[thread_id+2]);
    x[thread_id] = min(x[thread_id], x[thread_id+1]);
}
float minimum = x[0];    // this is the result
```
This code is correct as long as the seven lines inside the `if` block are executed in lockstep and all writes are immediately visible to other threads, which is not enforced by the code. I don't know why it works with D3D11, but clearly it does not with Vulkan.

### Why the current solution does not completely fix the problem
The current solution simply causes a memory barrier to be inserted between the seven lines inside the `if` block in my pseudo-code, generating this pseudo-code:
```c
float x[64];    // the array for which we want to compute the minimum, which is in TGSM
if (thread_id < 32) {
    x[thread_id] = min(x[thread_id], x[thread_id+32]);
    memory_barrier();
    x[thread_id] = min(x[thread_id], x[thread_id+16]);
    memory_barrier();
    x[thread_id] = min(x[thread_id], x[thread_id+8]);
    memory_barrier();
    x[thread_id] = min(x[thread_id], x[thread_id+4]);
    memory_barrier();
    x[thread_id] = min(x[thread_id], x[thread_id+2]);
    memory_barrier();
    x[thread_id] = min(x[thread_id], x[thread_id+1]);
    memory_barrier();
}
float minimum = x[0];    // this is the result
```
This clearly helps a lot, but doesn't completely fix the problem, because, since they are not control barriers, a thread might be reading for the following line while another hasn't even issued to write for the previous line. So, even if writes are assumed to be immediately visible, the more advanced thread will miss the read. What we really need is a memory and control barrier between each line in the `if` block.

Unfortunately we cannot just change the memory barriers in control barriers, because control barriers cannot be inserted inside dynamic control flow.

The problem here is not just theoretical: I can still see some artifacts even with the TGSM barrier workaround enable. Sure, they are much much fewer and much much smaller, but still there they are (on my Intel iGPU I can see a few pixels worth of artifact every ~10 seconds).

### My solution
My solutions attempts to completely tackle the problem (and I couldn't observe any artifacts with it) by putting actual control barriers in the code and modifying control flow so that the resulting code is legal. In practice, the code above is changed to this:
```c
float x[64];    // the array for which we want to compute the minimum, which is in TGSM
int condition = thread_id < 32;
if (condition) {
    x[thread_id] = min(x[thread_id], x[thread_id+32]);
}
memory_and_control_barrier();
if (condition) {
    x[thread_id] = min(x[thread_id], x[thread_id+16]);
}
memory_and_control_barrier();
if (condition) {
    x[thread_id] = min(x[thread_id], x[thread_id+8]);
}
memory_and_control_barrier();
if (condition) {
    x[thread_id] = min(x[thread_id], x[thread_id+4]);
}
memory_and_control_barrier();
if (condition) {
    x[thread_id] = min(x[thread_id], x[thread_id+2]);
}
memory_and_control_barrier();
if (condition) {
    x[thread_id] = min(x[thread_id], x[thread_id+1]);
}
memory_and_control_barrier();
float minimum = x[0];    // this is the result
```
This should be correct both theoretically and practically. It is also very complicated, unfortunately, because it has to track at least to some degree the flow of the shader code. Also, for the moment it works by patching the DXBC code, not the SPIR-V code, hijacking the DXBC decoder in a very unclean way.

So the final question to DXVK authors is this: could you give me some suggestion on how to rework this patch in a way that can be accepted in DXVK? The hard part IMHO is how to manage the control flow: does DXVK have any facility to help with that?

### DIFF
--- a/src/d3d11/d3d11_shader.cpp
+++ b/src/d3d11/d3d11_shader.cpp
@@ -20,7 +20,7 @@ namespace dxvk {
       reinterpret_cast<const char*>(pShaderBytecode),
       BytecodeLength);
     
-    DxbcModule module(reader);
+    DxbcModule module(reader, name);
     
     // If requested by the user, dump both the raw DXBC
     // shader and the compiled SPIR-V module to a file.
--- a/src/dxbc/dxbc_decoder.cpp
+++ b/src/dxbc/dxbc_decoder.cpp
@@ -35,7 +35,12 @@ namespace dxvk {
       throw DxvkError("DxbcCodeSlice: End of stream");
     return DxbcCodeSlice(m_ptr + n, m_end);
   }
-  
+
+
+  static bool check_f1_2020() {
+      const char *sgi = getenv("SteamGameId");
+      return sgi && strcmp(sgi, "1080110") == 0;
+  }
   
   
   void DxbcDecodeContext::decodeInstruction(DxbcCodeSlice& code) {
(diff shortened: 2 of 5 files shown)


doitsujin:
Which GPU is this a problem on? All current Nvidia and AMD GPUs have a subgroup size of *at least* 32, so control barriers should not be needed.

Ultimately this is still a game bug though and I'd rather not introduce any of this disgusting control flow hackery. DXVK does not have any way to deal with this properly.

giomasce:
Yes, I totally agree that they should fix their shaders, but I have no leverage for that. Though I also understand you're not willing to include such a patch, my only hope was that there was some better way to write this I was not aware of, and apparently there is not. Fortunately the current TGSM barrier workaround is close enough.

doitsujin:
Can you test [this](https://github.com/doitsujin/dxvk/tree/subgroup-size-control) branch with the game and see if that resolves the artifacts?

If not, then in addition to forcing the subgroup size, we may need to introduce control barriers that operate on the Subgroup scope, which should be allowed in non-uniform control flow (rather than Workgroup).

giomasce:
Unfortunately with that branch the game works correctly until I reach the screen when the bug is visible, but blocks at that moment and after a few dozen seconds it crashes. For the record, the HUD reports device Intel(R) UHD Graphics 630 (CFL GT2), driver version 20.3.3 and Vulkan version 1.2.145.

doitsujin:
Got any log files? It's important to know whether this is a GPU hang or something else going wrong.

Anyway, if this doesn't work, I'd rather just leave this broken.

giomasce:
BTW, given that the TGSM barrier workaround is good enough and Intel cards are not generally advised for gaming, I am not interested in this issue any more. Since I guess upstream isn't too, I am closing.

PR #1930, opened 2021-02-17
[build] Add optional build-id

ludocode:
Build-ids are required for doing certain kinds of performance tracing.
For example build-ids let Linux perf find the correct objects and
transfer them to other machines via "perf archive".

This adds an option to add -Wl,--build-id to the linker. It can be
enabled by passing -Dbuild_id=true to meson or --build-id to
package-release.sh.

Signed-off-by: Nicholas Fraser <nfraser@codeweavers.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -26,6 +26,12 @@ if dxvk_compiler.get_id() == 'clang'
   endif
 endif
 
+if get_option('build_id') and dxvk_compiler.get_id() != 'msvc'
+  if dxvk_compiler.has_link_argument('-Wl,--build-id')
+    add_global_link_arguments('-Wl,--build-id', language: 'cpp')
+  endif
+endif
+
 dxvk_include_path = include_directories('./include')
 
 if (cpu_family == 'x86_64')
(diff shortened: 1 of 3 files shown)


doitsujin:
Would there be any sort of downside to always enabling it? Could get rid of the config option if not, but if this blows up binary size or creates extra files or whatever, we should keep it.

Joshua-Ashton:
I don't think so, apparently this just adds something to the ELF header?

How does this work given we do PE builds? Does it replace the PDB dir or something?

ludocode:
It does the same thing for PE builds, just adds a 53 byte .buildid section. The reason I didn't turn it on by default is because it's a random GUID so it could interfere with reproducible builds. (This isn't the case for ELF build-ids because they are (by default) an SHA-1 hash of the .text section so they are reproducible.) I'm not sure if you or Valve are doing reproducible builds or planning on it and I figure most people doing performance tracing are going to be building from master anyway so they can turn this on themselves.

pchome:
@doitsujin 
> Would there be any sort of downside to always enabling it?

It potentially can brake some debug tools which rely on CRC. As opposite it will unbrake those tools which rely on build-id only (AutoFDO ?). See also: https://sourceware.org/gdb/current/onlinedocs/gdb/Separate-Debug-Files.html

doitsujin:
Fair enough, gonna keep it as an option then.

PR #1931, opened 2021-02-20
Fix some crashes with a user after free with barriers

Joshua-Ashton:
imgslices stores a raw pointer to a DxvkImage
- operator == in Rc with a raw pointer will cast the raw to a Rc
 - Because it was a raw pointer, and not ref-tracked it may have already been freed
 - Comparing with a raw pointer directly would be bad as we could get another image allocated at the same address so we must ref track for barrier lists

Also fixes a global out of bounds read with the HUD texture.


doitsujin:
This is not a correct fix. Images in the barrier list *cannot* be freed because they must be tracked by the current command buffer, which will keep them alive until the GPU finishes processing the command list. If there's a problem, we're missing a trackResource somewhere, which means that even with this patch, we may destroy the image and free its backing storage before too soon.

doitsujin:
FWIW I changed the barrier list to store VkImages instead, since we a) we don't dereference the pointers anyway and b) we really shouldn't be accidentally refcounting anything for a comparison, but while this will probably also resolve this particular crash, the underlying issue remains.

PR #1934, opened 2021-02-20
[dxvk] Fix resource tracking in attachment transitions

Joshua-Ashton:
Fixes a crash in Portal 2 on DXVK native in which an old depth stencil is used after free after a device reset.

### DIFF
--- a/src/dxvk/dxvk_context.cpp
+++ b/src/dxvk/dxvk_context.cpp
@@ -4063,6 +4063,8 @@ namespace dxvk {
         attachment.view->imageInfo().layout,
         attachment.view->imageInfo().stages,
         attachment.view->imageInfo().access);
+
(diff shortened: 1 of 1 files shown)


src/dxvk/dxvk_context.cpp
```diff
@@ -4063,6 +4063,9 @@ namespace dxvk {
         attachment.view->imageInfo().layout,
         attachment.view->imageInfo().stages,
         attachment.view->imageInfo().access);
+
+      m_cmd->trackResource<DxvkAccess::None>(attachment.view);
```
doitsujin, line 4067: We don't need to track the views here since the GPU won't use them, just the image.

Joshua-Ashton: Resolved.

PR #1938, opened 2021-02-23
[vr] Use registry key to get OpenVR extensions list if available.

gofman:
The current way of querying Vulkan extensions required for OpenVR is problematic:

- The current version can work only if the app didn't initialize VR before creating d3d11 device. Otherwise we will teardown the VR connections as the implemented way to check if VR is already initialized doesn't work, duplicated inits always succeed;
- Extra tinkering of VR client connection on and off inside the game process is very fragile, we see that native VR client depends on process state a lot. The native VR objects are per process singletons, and currently there is no known way to make sure that the initialized state is completely cleaned up. There is also seems to be some cached state in the vrserver. This problem is not fully solved by removing the extra OpenVR initialization / tear down as the game can do it itself. E. g., this is done by Until You Fall, and it still doesn't work (hangs the whole VR on attempt to render again after second initialization) on AMD (surprisingly, works fine on Nvidia). I have a standalone native OpenVR / Vulkan sample which reproduces this issue without Proton involved. Still, removing the extra initialization induced by us helps some games (at least Left Hand Path), and hopefully makes the thing more reliable. Besides, it shaves off a some time from the games' startup when Steam VR is running.

The idea is to query the extensions on Proton prefix startup from outside of the game process and store the required information in registry. Currently I used Proton's steam.exe helper for that purpose. The keys are initialized asynchronously (to avoid adding potential delays to game process startup) and it may happen the initialization is not completed yet by the time DXVK tries to create device. There is a registry value holding the OpenVR status, the value of 0 means it is not initialized yet, ~0 means VR is not available and 1 means VR is up and running and keys with extensions are valid. The registry key is volatile, so there is no leftover state from previous runs. I am attaching the Proton patch which sets this registry (as it is not in any release yet), and also a patch for wineopenxr which also going to use this (just for reference).

[patches.zip](https://github.com/doitsujin/dxvk/files/6028212/patches.zip)

### DIFF
--- a/src/dxvk/dxvk_openvr.cpp
+++ b/src/dxvk/dxvk_openvr.cpp
@@ -34,7 +34,9 @@ namespace dxvk {
   VrFunctions g_vrFunctions;
   VrInstance VrInstance::s_instance;
 
-  VrInstance:: VrInstance() { }
+  VrInstance:: VrInstance() {
+    m_no_vr = env::getEnvVar("DXVK_NO_VR") == "1";
+  }
   VrInstance::~VrInstance() { }
 
 
@@ -62,10 +64,21 @@ namespace dxvk {
   void VrInstance::initInstanceExtensions() {
     std::lock_guard<std::mutex> lock(m_mutex);
 
-    if (!m_compositor)
+    if (m_no_vr || m_initializedDevExt)
+        return;
+
+    if (!m_vr_key)
+    {
+        LSTATUS status;
+
+        if ((status = RegOpenKeyExA(HKEY_CURRENT_USER, "Software\\Wine\\VR", 0, KEY_READ, &m_vr_key)))
+            Logger::info(str::format("OpenVR: could not open registry key, status ", status));
+    }
+
+    if (!m_vr_key && !m_compositor)
       m_compositor = this->getCompositor();
 
-    if (!m_compositor || m_initializedInsExt)
+    if (!m_vr_key && !m_compositor)
       return;
     
     m_insExtensions = this->queryInstanceExtensions();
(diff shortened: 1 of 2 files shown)


src/dxvk/dxvk_openvr.cpp
```diff
@@ -62,10 +64,21 @@ namespace dxvk {
...
-    if (!m_compositor)
+    if (m_no_vr || m_initializedDevExt)
+        return;
+
+    if (!m_vr_key)
+    {
+        LSTATUS status;
+
+        if ((status = RegOpenKeyExA(HKEY_CURRENT_USER, "Software\\Wine\\VR", 0, KEY_READ, &m_vr_key)))
+            Logger::info(str::format("OpenVR: could not open registry key, status", status));
```
doitsujin, line 75: `status "` <- there should be a space here

src/dxvk/dxvk_openvr.cpp
```diff
@@ -76,31 +89,137 @@ namespace dxvk {
...
+        }
+        size = sizeof(value);
+        if ((status = RegQueryValueExA(m_vr_key, "state", NULL, &type, (BYTE *)&value, &size)))
+        {
+            Logger::err(str::format("OpenVR: could not query value, status ", status));
+            goto done;
+        }
+        if (value)
+            break;
+        while ((wait_status = WaitForSingleObject(event, 1000)) == WAIT_TIMEOUT)
```
doitsujin, line 140: Do we have any sort of guarantee that setting the registry key can never go wrong and that we'll never have stale entries in the registry even when something crashed during a previous launch? If not, we really shouldn't be creating an infinite loop here.

doitsujin (commented):
Why does the current code not work outside of some potential race conditions if the app initializes both at the same time? We shouldn't touch initialization unless `getGenericInterface` fails to return a compositor.

I'm not really a big solution either since it heavily depends on external factors working correctly, which also seems more than a little bit fragile.

src/dxvk/dxvk_openvr.cpp
gofman: > Why does the current code not work outside of some potential race conditions if the app initializes both at the same time? We shouldn't touch initialization unless `getGenericInterface` fails to return a compositor.

Yes, you are right, assuming DXVK gets the same higher level openvr_api.dll loader as the game is using (which stores the global vrclient's interface pointer during initialization and will succeed the interface query if the initialization has already been performed through VR_InitInternal[2] from this dll). That must be usually the case in practice, the other cases are more of theoretical concern and I probably shouldn't have mentioned this part.

The real issue which I currently saw affecting games is the extra initialization and teardown of vrclient done in the game process.

> 
> I'm not really a big fan of this solution either since it heavily depends on external factors working correctly, which also seems more than a little bit fragile.

Yes, doing the same VR client init / tear down outside of game requests is still not ideal (as vrclient links some things to steam game id also), but looks less fragile than doing the same in the game process.

> Do we have any sort of guarantee that setting the registry key can never go wrong and that we'll never have stale entries in the registry even when something crashed during a previous launch? If not, we really shouldn't be creating an infinite loop here.

The guarantee from any leftovers from the previous launch is using the volatile keys, they are not stored anywhere persistently and do not exist after Proton prefix shutdown, so upon new start the HKCU\\Software\\Wine\\VR key never exists until created by Proton's steam.exe. Unless someone sets it manually of course, but I believe we already have enough ways of breaking things manually like that anyway.

The more concerning part is probably the extension query process with VR init in initialize_vr_data() in steam.exe (in the attached patch). But I believe it is effectively equivalent to what DXVK currently does, so if something is going to hang there and leave status '0' the same seems to have equal chances to happen on the same route DXVK. If SteamVR is not running or libraries are not available we should be setting the ~0u status quickly, and it is rather hard to guess what can go wrong on this path unless some fundamental things (like setting registry key) are not working in Proton prefix.

I thought of adding some sort of timeout, but that timeout must be rather big. FWIW, with the way currently in DXVK the initialization takes ~2 sec in many cases when SteamVR is running with Steam's vrclient log complaining about timeout waiting for valid appkey after 2 sec. I am avoiding that now in steam.exe VR init, but not entirely sure such thing cannot happen under some different conditions. So my reasoning was that introducing the timeout will be only hiding some problems if they are going to happen. If you think it worth it though I don't mind adding such.

doitsujin:
Just changed some C stuff to the C++ equivalents (casts, `NULL`->`nullptr`) etc, fine otherwise.

pchome:
> Just changed some C stuff to the C++ equivalents (casts, `NUL`->`nullptr`) etc, fine otherwise.

formatting inconsistency,  `goto done;` could be replaced with just `break;` here, `sprintf` ...

I like this PR in many senses.

PR #1951, opened 2021-02-27
Presenter stuff

doitsujin:
Less radical than the previous version.


src/dxvk/dxvk_swapchain_blitter.cpp
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    m_gammaDirty = false;
+  }
+
+
+  void DxvkSwapchainBlitter::createSampler() {
+    DxvkSamplerCreateInfo samplerInfo;
+    samplerInfo.magFilter       = VK_FILTER_LINEAR;
+    samplerInfo.minFilter       = VK_FILTER_LINEAR;
```
Joshua-Ashton, line 269: We used to use _NEAREST when the extents matched, do we still want to do that maybe?

I know we are using pixelCoord, but...

doitsujin: We're using texelFetch in that.

Joshua-Ashton: Got that after looking again, and realised that was only on the blit path LGTM then!

PR #1958, opened 2021-03-03
[d3d9] Convert window position relative to its parent.

rbernon:
When using child windows, GetWindowRect is in screen coordinates while MoveWindow expects parent-relative position.


Joshua-Ashton (approved):
LGTM

PR #1959, opened 2021-03-03
[build] Use --file-alignment=4096 with MinGW

Joshua-Ashton:
Avoids a copy in the Wine loader as well as enables debug symbols to work in perf.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,15 +5,16 @@ cpu_family = target_machine.cpu_family()
 add_project_arguments('-DNOMINMAX', language : 'cpp')
 
 dxvk_compiler = meson.get_compiler('cpp')
-dxvk_msvc=dxvk_compiler.get_id() == 'msvc'
+dxvk_is_msvc = dxvk_compiler.get_id() == 'msvc'
+
 # c++17 was added in 15.3, older version needs c++latest
-if dxvk_compiler.get_id() == 'msvc' and dxvk_compiler.version().version_compare('<15.3')
+if dxvk_is_msvc and dxvk_compiler.version().version_compare('<15.3')
   dxvk_cpp_std='c++latest'
 else
   dxvk_cpp_std='c++17'
 endif
 
-if dxvk_compiler.get_id() == 'msvc'
+if dxvk_is_msvc
   add_project_arguments('/std:' + dxvk_cpp_std, language : 'cpp')
 endif
 
@@ -26,7 +27,7 @@ if dxvk_compiler.get_id() == 'clang'
   endif
 endif
 
-if dxvk_compiler.get_id() != 'msvc'
+if not dxvk_is_msvc
   if get_option('build_id') and dxvk_compiler.has_link_argument('-Wl,--build-id')
     add_global_link_arguments('-Wl,--build-id', language: 'cpp')
   endif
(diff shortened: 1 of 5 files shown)


meson.build
```diff
@@ -5,15 +5,16 @@ cpu_family = target_machine.cpu_family()
 add_project_arguments('-DNOMINMAX', language : 'cpp')
 
 dxvk_compiler = meson.get_compiler('cpp')
-dxvk_msvc=dxvk_compiler.get_id() == 'msvc'
+dxvk_msvc = dxvk_compiler.get_id() == 'msvc'
```
doitsujin, line 8: can we name this `is_msvc` or something?

Joshua-Ashton: Done.

Joshua-Ashton: wait

Joshua-Ashton: Forgot to add a few files to the commit, fixed.

pchome:
Is `objects             : not dxvk_is_msvc ? 'd3d10'+def_spec_ext : [],` still required by anything? It was a hack for winelib to add *.spec file name to link process, IIRC. I guess I about to add additional headache for me and other winelib users, but better remove this if unused.

Joshua-Ashton:
Its needed for MSVC to compile

pchome:
Likely `vs_module_defs      : 'd3d10'+def_spec_ext,`  needed for MSVC to compile. The question is this also used for MinGW, or mingw use `objects`.

Joshua-Ashton:
MinGW takes in it's def as an object, MSVC takes it in another way, hence why this is needed.

pchome:
Still doubt, but okay I won't insist.

DXVK was fine w/o this.
- then they added msvc support: https://github.com/doitsujin/dxvk/commit/9deb73a2a7007c20d306f2e88d783458b4935825#diff-02d517bb3be49b91036dde4a613d0b4b5793091470cb8e3f51f01ee1ce7f0218
- then winelib support was added and broke msvc: https://github.com/doitsujin/dxvk/commit/06511aa72c05538884b937227ecfa0ab041fe854#diff-02d517bb3be49b91036dde4a613d0b4b5793091470cb8e3f51f01ee1ce7f0218
(based on my PR https://github.com/doitsujin/dxvk/pull/268/files#diff-02d517bb3be49b91036dde4a613d0b4b5793091470cb8e3f51f01ee1ce7f0218, so it was literally me who introduced this shit (`object:`))
- then msvc was fixed: https://github.com/doitsujin/dxvk/commit/f38ee85a396bbc2a0ad2dba8f14f5a34d61457ee#diff-02d517bb3be49b91036dde4a613d0b4b5793091470cb8e3f51f01ee1ce7f0218

That's why I doubt. I don't have all compilers to check, but I bet you can remove this (`object:`) to clean up a bit.

Joshua-Ashton:
MinGW does not respect vs_module_defs, MSVC fails if you have the .def in the object array.

I don't get what you don't get.

pchome:
I don't get why you need `objects` for MSVC if it is empty `[]` for MSVC. And if you do not need `objects` not for MSVC nor for MinGW then remove it, because it is not intended to be used for def/spec anyway.

Joshua-Ashton:
This is where you put .def on MinGW. Unless something has changed, it does not respect vs_module_defs as I mentioned above.

PR #1966, opened 2021-03-07
[build] Meson kwargs support

pchome:
Supersedes: #1004

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,4 +1,4 @@
-project('dxvk', ['c', 'cpp'], version : 'v1.8.1', meson_version : '>= 0.46')
+project('dxvk', ['c', 'cpp'], version : 'v1.8.1', meson_version : '>= 0.49')
 
 cpu_family = target_machine.cpu_family()
 
@@ -119,6 +119,13 @@ dxvk_version = vcs_tag(
   command: ['git', 'describe', '--dirty=+'],
   input:  'version.h.in',
   output: 'version.h')
+
+# Common kwargs for targets
+common_opts = { 'override_options': ['cpp_std=' + dxvk_cpp_std] }
+
+common_exe_opts    = common_opts + { 'gui_app' : true, 'install': true }
+common_static_opts = common_opts + { 'include_directories': dxvk_include_path }
+common_shared_opts = common_opts + { 'include_directories': dxvk_include_path, 'name_prefix': '', 'install': true }
 
 subdir('src')
 
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -12,37 +12,36 @@ d3d10_main_src = [
   'd3d10_state_block.cpp',
 ]
 
-d3d10_core_dll = shared_library('d3d10core'+dll_ext, d3d10_core_src, d3d10_core_res,
-  name_prefix         : '',
-  dependencies        : [ d3d11_dep ],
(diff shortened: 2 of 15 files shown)


meson.build
```diff
@@ -120,6 +120,13 @@ dxvk_version = vcs_tag(
   input:  'version.h.in',
   output: 'version.h')
 
+# Common kwargs for targets
+common_opts = { 'override_options': ['cpp_std=' + dxvk_cpp_std], 'gnu_symbol_visibility': 'inlineshidden' }
```
doitsujin, line 124: Where does the gnu_symbol_visibility thingy come from? We're not using it right now and this doesn't look like something that would work with MSVC.

Joshua-Ashton:
I do not know what this is

meson.build
pchome: Added this as a bonus (meson ~v0.48). I guess it should not have any effect on MSVC, same as `vs_module_defs` have no effect on GCC and enabled unconditionally. Instead will be used by builds like winelib or dxvk native and kinds.

Also `'path' / 'to' / 'file'` concatenation was used as one of new features, and `objects` replaced with more (i guess) correct version.

pchome:
This is #1004 updated for current master branch.

doitsujin (changes requested):
There's a bunch of things I don't understand here that really need explanation on why they work, see comments.

Joshua-Ashton:
I'm not really a fan of this whole thing -- I know it's trying to clean up but it seems to just be making things much more of a mess.

meson.build
pchome: You mean separate commit in this PR, or separate PR? Or I should simply remove this and forget?

meson.build
```diff
@@ -120,6 +120,13 @@ dxvk_version = vcs_tag(
...
+
+common_exe_opts    = common_opts + { 'gui_app' : true, 'install': true }
```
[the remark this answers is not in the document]
pchome: Grouped with other similar variables, easy to find and change. Not really meter, they are visible everywhere to `src/`, can be moved to `test/` if you want it to be there.

src/d3d10/meson.build
```diff
@@ -12,37 +12,36 @@ d3d10_main_src = [
...
-  override_options    : ['cpp_std='+dxvk_cpp_std])
+d3d10_core_def = 'd3d10core' + def_spec_ext
+d3d10_def      = 'd3d10'     + def_spec_ext
+d3d10_1_def    = 'd3d10_1'   + def_spec_ext
+
+# d3d10core
+d3d10_core_opts = common_shared_opts + { 'dependencies': [ d3d11_dep ], 'vs_module_defs': d3d10_core_def }
+
+if not dxvk_is_msvc
+  d3d10_core_opts += { 'link_depends': files(d3d10_core_def), 'link_args': meson.current_source_dir() / d3d10_core_def }
```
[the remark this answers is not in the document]
pchome: Yes, `link_args` wants string, unlike generated files this `*.def` files are not copied to the build directory, so linker can't find them if they specified as file name.
`link_depends` just adds "tracking", so meson will trigger re-link in configured build directory if the file will be changed. Maybe `files()` not really needed here for single file name.

pchome:
Use this as example, there was (and will be) too many changes in build system to rebase this PR.

PR #1973, opened 2021-03-10
Switch to the portable shebang to bash

fabiscafe:
Hi.
We should go for the [portable shebang-form](https://en.wikipedia.org/w/index.php?title=Shebang_(Unix)&oldid=878552871#Portability) to get the bash environment. `/bin/bash` is also not a real thing on most distros anymore¹. This will bring compatiblity for distros who don't create the compat-symlink.

¹ Most distros:
```
file /bin
/bin: symbolic link to usr/bin
```

### DIFF
--- a/package-release.sh
+++ b/package-release.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 set -e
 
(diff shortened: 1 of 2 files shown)


jqadev:
@fabiscafe to be consistent you could also add these changes to
https://github.com/HansKristian-Work/vkd3d-proton/blob/master/setup_vkd3d_proton.sh
https://github.com/HansKristian-Work/vkd3d-proton/blob/master/package-release.sh

PR #1975, opened 2021-03-11
[dxvk] Use image memory barriers for readback

Joshua-Ashton:
Targets only the specific subresource of the bound color attachments.

From the Vulkan spec:
Each of the synchronization scopes and access scopes of a vkCmdPipelineBarrier2KHR or vkCmdPipelineBarrier command inside a render pass instance must be a subset of the scopes of one of the self-dependencies for the current subpass.


doitsujin:
What exactly is this supposed to fix? Unspecified memory barriers shoud always be legal.

doitsujin:
As discussed on Discord, not very useful.

Issue #1980, opened 2021-03-17
[D3D9]Dragon Nest Water is not displaying correctly


dnblank123:
### Software information
Dragon Nest

### System information
- GPU: Nvidia GTX 1050ti
- Driver: 456.98
- Windows 10
- DXVK version: 1.8.1

### Apitrace file(s)
- https://drive.google.com/file/d/1daCTTV76LHC_Jxf_47L4iSmj4JAlHJo-/view?usp=sharing

### Log files
- d3d9.log: [DragonNest_x64_d3d9.log](https://github.com/doitsujin/dxvk/files/6153977/DragonNest_x64_d3d9.log)

[without DXVK]
![DN 2021-03-17 12-35-12 Wed](https://user-images.githubusercontent.com/67945097/111421215-951b4a00-8727-11eb-9e53-e319ccbf552b.png)

[with DXVK]
![DN 2021-03-17 13-07-09 Wed](https://user-images.githubusercontent.com/67945097/111421270-af552800-8727-11eb-9e55-e5b014cf4cbc.png)

K0bin:
Can you try turning putting `d3d9.allowLockFlagReadonly = False` into your DXVK config (dxvk.conf in your user directory or right next to the games exe) and report back whether that fixes it?

dnblank123:
> Can you try turning putting `d3d9.allowLockFlagReadonly = False` into your DXVK config (dxvk.conf in your user directory or right next to the games exe) and report back whether that fixes it?

It did fixed, Thank you!

K0bin:
Reopening this so we can make it work out of the box.

K0bin:
Should be fixed out of the box now. Pls re-open if the issue comes back.

PR #1982, opened 2021-03-17
D3D9 readonly locking cleanup

K0bin:
This still needs a lot of testing and some cleanup (especially the commits) so it's a draft.
I've tested Risen, Burnout Paradise and Nostale. Those three games needed the hack before.

It does a couple of things to clean up locking:

## Buffers

- Give `dirtyRange`, `lockRange` and `readLocked` names that better explain their purpose.
- Remove `needsUpload` because the same thing is being tracked by `dirtyRange` (or formally `lockRange`) anyway.
- Mark non-default buffers as dirty by default. This combined with the upload on the first draw is what allows us to get rid of the hack to disable `D3DLOCK_READONLY`.

## Textures

- Mark the texture as `needsUpload` by default. We still need that one here because the dirty box only for mip 0.
- Fix a couple of issues with how we handle the dirty box.
- Rename `updateDirtyBox` to just `dirtyBox` as it's no longer only being used for `UpdateTexture`/`UpdateSurface`

## General

- Remove the hack that disables `D3DLOCK_READONLY`

After that we should try to just generally clean up the logic in the Lock* functions and also only copy dirty parts of managed resources.

### DIFF
--- a/src/d3d9/d3d9_common_buffer.h
+++ b/src/d3d9/d3d9_common_buffer.h
@@ -168,9 +168,10 @@ namespace dxvk {
     }
     uint32_t GetLockCount() const { return m_lockCount; }
 
-    void MarkUploaded()      { m_needsUpload = false; }
-    void MarkNeedsUpload()   { m_needsUpload = true; }
-    bool NeedsUpload() const { return m_needsUpload; }
+    /**
+     * \brief Whether or not the staging buffer needs to be copied to the actual buffer
+     */
+    bool NeedsUpload() { return m_desc.Pool != D3DPOOL_DEFAULT && !m_dirtyRange.IsDegenerate(); }
 
     void PreLoad();
 
@@ -205,8 +206,6 @@ namespace dxvk {
     D3D9Range                   m_gpuReadingRange;
 
     uint32_t                    m_lockCount = 0;
-
-    bool                        m_needsUpload = false;
 
   };
 
--- a/src/d3d9/d3d9_common_buffer.h
+++ b/src/d3d9/d3d9_common_buffer.h
@@ -139,11 +139,25 @@ namespace dxvk {
 
     static HRESULT ValidateBufferProperties(const D3D9_BUFFER_DESC* pDesc);
 
-    D3D9Range& LockRange()  { return m_lockRange; }
-    D3D9Range& DirtyRange() { return m_dirtyRange; }
+    /**
+     * \brief The range of the buffer that was changed using Lock calls
+     */
+    D3D9Range& DirtyRange()  { return m_dirtyRange; }
 
-    bool GetReadLocked() const     { return m_readLocked; }
-    void SetReadLocked(bool state) { m_readLocked = state; }
+    /**
+     * \brief The range of the buffer that might currently be read by the GPU
+     */
+    D3D9Range& GPUReadingRange() { return m_gpuReadingRange; }
+
+    /**
+    * \brief Whether or not the buffer was written to by the GPU (in IDirect3DDevice9::ProcessVertices)
+    */
+    bool WasWrittenByGPU() const     { return m_wasWrittenByGPU; }
+
+    /**
+    * \brief Sets whether or not the buffer was written to by the GPU
+    */
+    void SetWrittenByGPU(bool state) { m_wasWrittenByGPU = state; }
 
     uint32_t IncrementLockCount() { return ++m_lockCount; }
     uint32_t DecrementLockCount() {
@@ -157,12 +171,6 @@ namespace dxvk {
     void MarkUploaded()      { m_needsUpload = false; }
     void MarkNeedsUpload()   { m_needsUpload = true; }
     bool NeedsUpload() const { return m_needsUpload; }
-
-    bool MarkLocked() {
-      bool locked = m_readLocked;
-      m_readLocked = true;
-      return locked;
-    }
 
     void PreLoad();
 
@@ -186,15 +194,15 @@ namespace dxvk {
     D3D9DeviceEx*               m_parent;
     const D3D9_BUFFER_DESC      m_desc;
     DWORD                       m_mapFlags;
-    bool                        m_readLocked = false;
+    bool                        m_wasWrittenByGPU = false;
 
     Rc<DxvkBuffer>              m_buffer;
     Rc<DxvkBuffer>              m_stagingBuffer;
 
     DxvkBufferSliceHandle       m_sliceHandle;
 
-    D3D9Range                   m_lockRange;
     D3D9Range                   m_dirtyRange;
+    D3D9Range                   m_gpuReadingRange;
 
     uint32_t                    m_lockCount = 0;
 
--- a/src/d3d9/d3d9_common_texture.cpp
+++ b/src/d3d9/d3d9_common_texture.cpp
@@ -21,6 +21,13 @@ namespace dxvk {
       AddDirtyBox(nullptr, i);
     }
 
+    if (m_desc.Pool != D3DPOOL_DEFAULT) {
+      const uint32_t subresources = CountSubresources();
+      for (uint32_t i = 0; i < subresources; i++) {
+        SetNeedsUpload(i, true);
+      }
+    }
+
     m_mapping = pDevice->LookupFormat(m_desc.Format);
 
     m_mapMode = DetermineMapMode();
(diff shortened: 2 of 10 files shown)


K0bin:
Got rid of the commits that switched it to use track the dirty box per subresource and use that to figure out if something needs to be uploaded.

src/d3d9/d3d9_device.cpp
```diff
@@ -5771,7 +5766,7 @@ namespace dxvk {
       GenerateTextureMips(texturesToGen);
 
     auto* ibo = GetCommonBuffer(m_state.indices);
-    if (ibo != nullptr && ibo->NeedsUpload())
+    if (ibo != nullptr && ibo->Desc()->Pool != D3DPOOL_DEFAULT  && !ibo->DirtyRange().IsDegenerate())
```
Joshua-Ashton, line 5747: Maybe add a NeedsUpload helper method here. Feel free to pick a better name. Maybe NeedsDeferredUpload

src/d3d9/d3d9_device.cpp
```diff
@@ -3980,7 +3980,18 @@ namespace dxvk {
...
 
     if (!(Flags & D3DLOCK_NO_DIRTY_UPDATE) && !(Flags & D3DLOCK_READONLY)) {
+      if (pBox && MipLevel != 0) {
+        D3DBOX scaledBox = *pBox;
+        scaledBox.Left   <<= MipLevel;
+        scaledBox.Right  <<= MipLevel;
+        scaledBox.Top    <<= MipLevel;
+        scaledBox.Bottom <<= MipLevel;
+        scaledBox.Back   <<= MipLevel;
+        scaledBox.Front  <<= MipLevel;
```
Joshua-Ashton: Do we need to clamp this? Eg. if there are several mips with a 1xY size. That 1 could get scaled up outside the bounds of the texture.

src/d3d9/d3d9_device.cpp
```diff
@@ -4252,8 +4239,11 @@ namespace dxvk {
          shouldFlush &= !pResource->GetReadOnlyLocked(Subresource);
          shouldFlush &= !pResource->IsManaged() || m_d3d9Options.evictManagedOnUnlock;
 
-    if (shouldFlush)
+    if (shouldFlush) {
         this->FlushImage(pResource, Subresource);
+        if (pResource->IsManaged())
```
Joshua-Ashton, line 4232: != DEFAULT

src/d3d9/d3d9_common_texture.h
```diff
@@ -369,6 +369,9 @@ namespace dxvk {
     void PreLoadSubresource(UINT Subresource);
 
     void AddDirtyBox(CONST D3DBOX* pDirtyBox, uint32_t layer) {
+      if (m_desc.Pool == D3DPOOL_DEFAULT)
```
Joshua-Ashton, line 368: I would pull this out.

Joshua-Ashton (changes requested):
I still need to review `Correct handling of D3DLOCK_READONLY` and take a closer look at things tomorrow.

I go to bed.

src/d3d9/d3d9_device.cpp
```diff
@@ -4059,20 +4040,13 @@ namespace dxvk {
       // No need to wait as its not in use.
       physSlice = pResource->GetMappedSlice(Subresource);
 
-      // We do not need to wait for the resource in the event the
-      // calling app promises not to overwrite data that is in use
-      // or is reading. Remember! This will only trigger for MANAGED resources
-      // that cannot get affected by GPU, therefore readonly is A-OK for NOT waiting.
-      const bool skipWait = (readOnly && managed) || scratch || (readOnly && systemmem && !wasWrittenByGPU);
```
Joshua-Ashton: ?

src/d3d9/d3d9_common_texture.h
```diff
@@ -314,10 +314,6 @@ namespace dxvk {
 
     void MarkAllWrittenByGPU() { m_wasWrittenByGPU.setAll(); }
 
-    void SetReadOnlyLocked(UINT Subresource, bool readOnly) { return m_readOnly.set(Subresource, readOnly); }
```
Joshua-Ashton, line 317: ?

src/d3d9/d3d9_device.cpp
K0bin: Well readonly can write after all so we can't skip the wait here.

src/d3d9/d3d9_device.cpp
```diff
@@ -4461,12 +4446,6 @@ namespace dxvk {
     *ppbData = reinterpret_cast<void*>(data);
 
     DWORD oldFlags = pResource->GetMapFlags();
-
-    // We need to remove the READONLY flags from the map flags
-    // if there was ever a non-readonly upload.
-    if (!(Flags & D3DLOCK_READONLY))
```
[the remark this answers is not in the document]
K0bin: Same thing as the comment above.

src/d3d9/d3d9_common_texture.h
K0bin: Same thing as the comment above.

src/d3d9/d3d9_device.cpp
Joshua-Ashton: Games are using it for writes but as I understood based on our conversation it was in a NO_OVERWRITE fashion which would make skipping waits okay.

The problem was I was aware of was that we weren't initially dirtying stuff.

src/d3d9/d3d9_device.cpp
K0bin: The dirty box is also used for UpdateTexture and UpdateSurface so I thought we cant clear that but that line isnt called for backed system memory textures anyway so your suggestion is fine too.

Joshua-Ashton: Oh right, this is for the funny evictManagedOnUnlock path. Ignore this then.

src/d3d9/d3d9_device.cpp
K0bin: Any idea what to clamp this to?

K0bin: Dont need to clamp the left/top/front because it's all unsigned.

src/d3d9/d3d9_common_buffer.h
```diff
@@ -171,7 +171,7 @@ namespace dxvk {
     /**
      * \brief Whether or not the staging buffer needs to be copied to the actual buffer
      */
-    bool NeedsUpload() { return m_desc.Pool != D3DPOOL_DEFAULT && !m_dirtyRange.IsDegenerate(); }
+    bool NeedsUpload() { return m_desc.Pool != D3DPOOL_DEFAULT && !m_dirtyRange.IsDegenerate() && m_stagingBuffer != nullptr; }
```
Joshua-Ashton, line 174: != DEFAULT always has a staging buffer

src/d3d9/d3d9_common_texture.cpp
```diff
@@ -21,6 +21,13 @@ namespace dxvk {
       AddDirtyBox(nullptr, i);
     }
 
+    if (m_desc.Pool != D3DPOOL_DEFAULT) {
```
Joshua-Ashton, line 24: We have observed initial upload always being true for buffers, is this also true for textures?

K0bin: I've seen games that rely on readonly uploads for textures (#1980 for example) so I think so.

K0bin: I've seen games that rely on readonly uploads for textures (#1980 for example) so I think so.

src/d3d9/d3d9_common_buffer.h
K0bin: Done.

src/d3d9/d3d9_common_texture.h
```diff
@@ -353,7 +353,7 @@ namespace dxvk {
     D3D9SubresourceBitset& GetUploadBitmask() { return m_needsUpload; }
 
     void SetNeedsUpload(UINT Subresource, bool upload) { m_needsUpload.set(Subresource, upload); }
-    bool NeedsUpload(UINT Subresource) const { return m_needsUpload.get(Subresource); }
+    bool NeedsUpload(UINT Subresource) const { return m_needsUpload.get(Subresource) && m_buffers[Subresource] != nullptr; }
```
Joshua-Ashton, line 356: We still want to mark this as not needing upload anymore + clear dirty region regardless of whether we have a staging buffer to do it from or not.

We should factor that outside of the NeedsUpload call into FlushBuffer

Joshua-Ashton (changes requested):

Joshua-Ashton (approved):

PR #1995, opened 2021-03-29
[util] Override default thread stack reserve size.

rbernon:
Some applications (Dragon Quest Builder 2 for instance) use an oversized
default thread stack reserve size (2G in this case), which causes every
DXVK thread to allocate as much memory, quickly exhausting system memory.

Explicitly specifying the desired stack reserve size overrides the default application setting.

### DIFF
--- a/src/util/thread.h
+++ b/src/util/thread.h
@@ -40,8 +40,9 @@ namespace dxvk {
       // Reference for the thread function
       this->incRef();
 
-      m_handle = ::CreateThread(nullptr, 0,
-        ThreadFn::threadProc, this, 0, nullptr);
+      m_handle = ::CreateThread(nullptr, 0x100000,
+        ThreadFn::threadProc, this, STACK_SIZE_PARAM_IS_A_RESERVATION,
+        nullptr);
       
       if (m_handle == nullptr)
         throw DxvkError("Failed to create thread");


rbernon:
The memory exhaustion is probably more an issue with Wine how allocates virtual memory, but I'd still think that controlling the thread stack size is better in any case.

Joshua-Ashton:
> The memory exhaustion is probably more an issue with how Wine allocates virtual memory

VA exhaustion is a pretty big problem for us in D3D9 DXVK.
It's much better under Windows but still a bit problematic there.

Finding where the extraneous VA comes from would be interesting at some point. We tried tracing maps but nothing really stuck out ever.
We've considered unmapping MANAGED and stuff but that doesn't seem to match native behaviour on Windows so... :man_shrugging:.

doitsujin (approved):

rbernon:
The problem here is that all memory was mapped to be backed by physical pages, which quickly exhaust the amount of physical memory. Wine doesn't treat MEM_RESERVE very differently from MEM_COMMIT and I think it could more generally use MAP_NORESERVE to avoid requiring physical memory when it's not needed.

Sadly I don't think this will change anything regarding 32bit address space exhaustion, as the issue seems to more directly related to running out of address space (although I didn't do any kind of debugging on that topic myself).

Issue #1996, opened 2021-03-29
Origin Overlay does not work in Proton (specifically, with d3d11 and dxvk's dxgi) due to IDXGISwapchain methods being not in dxgi.dll


gofman:
The Origin's detour patcher expects the methods to be hotpatched from IDXGISwapchain1 interface to live in dxgi.dll. It checks that the methods lie within the loaded dxgi.dll image range and refuses to hotpatch them otherwise.

That's currently the same for Proton's dxgi as well due to the same reasons. 
The attached PoC hack fixes the issue (attaching just for illustration of what it wants, I would not consider this a fix).

[0001-d3d11-Create-dxgi-swapchain-from-dxgi.dll.txt](https://github.com/doitsujin/dxvk/files/6225012/0001-d3d11-Create-dxgi-swapchain-from-dxgi.dll.txt)

doitsujin:
Well that's a huge pain in the arse. It doesn't look like this is possible to fix without losing compatibility to Wine's DXGI implementation, is it?

gofman:
As a probably saner workaround, adding a proxy interface to dxgi itself without touching d3d11 helps too (that's what I did for Proton's dxgi, see attached). Please note it might happen (especially with a very short function which just jumps to the underlying proxy method) that it still won't be hotpatchable by Origin due to some opcode it won't understand or the function being just too short. In that case __attribute__((__ms_hook_prologue__)) function attribute (DECLSPEC_HOTPATCH in Wine) helps. This is less likely for PE .dll (the Wine's dxgi is currently dll.so), if to go this way we can check that it is actually needed before adding that.

[0001-dxgi-Use-proxy-IDXGISwapChain-interface-when-importi.txt](https://github.com/doitsujin/dxvk/files/6227692/0001-dxgi-Use-proxy-IDXGISwapChain-interface-when-importi.txt)

Joshua-Ashton:
I was planning on just writing a simple swapchain wrapper/dispatcher in our DXGI. I imagine that works too?

I'm not super familiar with `DECLSPEC_HOTPATCH` from your patch there. What does that do to help the situation?

gofman:
Yes, that works, the second patch I linked (for Wine's dxgi) is doing exactly that.

[...]

PR #1997, opened 2021-03-30
[dxgi] Implement swapchain dispatcher

Joshua-Ashton:
Some overlays and hooks check if the vtable funcs reside in the dxgi.dll module.

Fixes: #1996

### DIFF
--- a/src/dxgi/dxgi_factory.cpp
+++ b/src/dxgi/dxgi_factory.cpp
@@ -1,5 +1,6 @@
 #include "dxgi_factory.h"
 #include "dxgi_swapchain.h"
+#include "dxgi_swapchain_dispatcher.h"
 
 namespace dxvk {
 
@@ -130,9 +131,18 @@ namespace dxvk {
     if (SUCCEEDED(pDevice->QueryInterface(
           __uuidof(IWineDXGISwapChainFactory),
           reinterpret_cast<void**>(&wineDevice)))) {
-      return wineDevice->CreateSwapChainForHwnd(
+      IDXGISwapChain4* frontendSwapChain;
+
+      HRESULT hr = wineDevice->CreateSwapChainForHwnd(
         this, hWnd, pDesc, pFullscreenDesc,
-        pRestrictToOutput, ppSwapChain);
+        pRestrictToOutput, reinterpret_cast<IDXGISwapChain1**>(&frontendSwapChain));
+
+      // No ref as that's handled by the object we're wrapping
+      // which was ref'ed on creation.
+      if (SUCCEEDED(hr))
+        *ppSwapChain = new DxgiSwapChainDispatcher(frontendSwapChain);
+
+      return hr;
     }
     
     Logger::err("DXGI: CreateSwapChainForHwnd: Unsupported device type");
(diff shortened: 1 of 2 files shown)


doitsujin:
Can you move this to its own file?

Joshua-Ashton:
Will do

src/dxgi/dxgi_factory.cpp
```diff
@@ -3,6 +3,266 @@
 
 namespace dxvk {
 
+  class DxgiSwapChainDispatcher : public ComObject<IDXGISwapChain4> {
```
doitsujin, line 6: This should probably not be a `ComObject`, instead, just forward AddRef/Release to the underlying swap chain.

src/dxgi/dxgi_factory.cpp
```diff
@@ -3,6 +3,266 @@
...
+    HRESULT STDMETHODCALLTYPE SetHDRMetaData(
+            DXGI_HDR_METADATA_TYPE    Type,
+            UINT                      Size,
+            void*                     pMetaData) final {
+      return m_dispatch->SetHDRMetaData(Type, Size, pMetaData);
+    }
+
+  private:
+
+    Com<IDXGISwapChain4> m_dispatch;
```
doitsujin, line 262: Then this needs to be a raw `IDXGISwapChain4*`.

Joshua-Ashton:
Updated he PR.

src/dxgi/dxgi_swapchain_dispatcher.h
```diff
@@ -0,0 +1,280 @@
...
+    HRESULT STDMETHODCALLTYPE QueryInterface(
+            REFIID                    riid,
+            void**                    ppvObject) final {
+      if (ppvObject == nullptr)
+        return E_POINTER;
+
+      *ppvObject = nullptr;
+
+      if (riid == __uuidof(IUnknown)
+      || riid == __uuidof(IDXGIObject)
```
doitsujin, line 40: weird indentation

src/dxgi/dxgi_swapchain_dispatcher.h
```diff
@@ -0,0 +1,280 @@
...
+
+  class DxgiSwapChainDispatcher : public IDXGISwapChain4 {
+
+  public:
+
+    DxgiSwapChainDispatcher(IDXGISwapChain4* dispatch)
+      : m_dispatch(dispatch) {
+    }
+
+    virtual ~DxgiSwapChainDispatcher() {
```
doitsujin, line 15: This doesn't really need to be virtual since nothing will ever inherit from this class. We can just use the default destructor.

Joshua-Ashton: Gives a warning if I don't do this.

doitsujin: even if you mark the class as `final`?

src/dxgi/dxgi_swapchain_dispatcher.h
Joshua-Ashton: Fixed.

PR #2002, opened 2021-04-01
Implement ID3DUserDefinedAnnotations for D3D11

liam-middlebrook:
I plan to follow-up with the D3D9-era `D3DPERF_*` implementation in a separate PR. That implementation isn't quite a straight-forward given that those functions are global and operate outside of any device/instance context.

### DIFF
--- a/src/d3d11/d3d11_annotation.cpp
+++ b/src/d3d11/d3d11_annotation.cpp
@@ -1,9 +1,12 @@
 #include "d3d11_annotation.h"
+#include "d3d11_context.h"
+#include "d3d11_device.h"
 
 namespace dxvk {
 
-  D3D11UserDefinedAnnotation::D3D11UserDefinedAnnotation(ID3D11DeviceContext* ctx)
-  : m_container(ctx) { }
+  D3D11UserDefinedAnnotation::D3D11UserDefinedAnnotation(D3D11DeviceContext* ctx)
+  : m_container(ctx),
+    m_eventDepth(0) { }
 
 
   D3D11UserDefinedAnnotation::~D3D11UserDefinedAnnotation() {
@@ -30,26 +33,60 @@ namespace dxvk {
 
   INT STDMETHODCALLTYPE D3D11UserDefinedAnnotation::BeginEvent(
           LPCWSTR                 Name) {
-    // Currently not implemented
-    return -1;
+    if (!m_container->IsAnnotationEnabled())
+      return -1;
+
+    m_container->EmitCs([labelName = dxvk::str::fromws(Name)](DxvkContext *ctx) {
+      VkDebugUtilsLabelEXT label;
+      label.sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT;
+      label.pNext = nullptr;
+      label.pLabelName = labelName.c_str();
+      label.color[0] = 1.0f;
+      label.color[1] = 1.0f;
+      label.color[2] = 1.0f;
+      label.color[3] = 1.0f;
+
+      ctx->beginDebugLabel(&label);
+    });
+
+    return m_eventDepth++;
   }
 
 
   INT STDMETHODCALLTYPE D3D11UserDefinedAnnotation::EndEvent() {
-    // Currently not implemented
-    return -1;
+    if (!m_container->IsAnnotationEnabled())
+      return -1;
+
+    m_container->EmitCs([](DxvkContext *ctx) {
+      ctx->endDebugLabel();
+    });
+
+    return m_eventDepth--;
   }
 
 
   void STDMETHODCALLTYPE D3D11UserDefinedAnnotation::SetMarker(
           LPCWSTR                 Name) {
-    // Currently not implemented
+    if (!m_container->IsAnnotationEnabled())
+      return;
+
+    m_container->EmitCs([labelName = dxvk::str::fromws(Name)](DxvkContext *ctx) {
+      VkDebugUtilsLabelEXT label;
+      label.sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT;
+      label.pNext = nullptr;
+      label.pLabelName = labelName.c_str();
+      label.color[0] = 1.0f;
+      label.color[1] = 1.0f;
+      label.color[2] = 1.0f;
+      label.color[3] = 1.0f;
+
+      ctx->insertDebugLabel(&label);
+    });
   }
 
 
   BOOL STDMETHODCALLTYPE D3D11UserDefinedAnnotation::GetStatus() {
-    // Currently not implemented
-    return FALSE;
+    return m_container->IsAnnotationEnabled();
   }
 
 }
--- a/src/d3d11/d3d11_annotation.h
+++ b/src/d3d11/d3d11_annotation.h
@@ -4,11 +4,13 @@
 
 namespace dxvk {
 
+  class D3D11DeviceContext;
+
   class D3D11UserDefinedAnnotation : ID3DUserDefinedAnnotation {
 
   public:
 
-    D3D11UserDefinedAnnotation(ID3D11DeviceContext* ctx);
+    D3D11UserDefinedAnnotation(D3D11DeviceContext* ctx);
     ~D3D11UserDefinedAnnotation();
 
     ULONG STDMETHODCALLTYPE AddRef();
@@ -31,8 +33,10 @@ namespace dxvk {
 
   private:
 
-    ID3D11DeviceContext*  m_container;
+    D3D11DeviceContext*  m_container;
 
+    // Stack depth for non-finalized BeginEvent calls
+    int32_t m_eventDepth;
   };
 
 }
--- a/src/dxvk/dxvk_cmdlist.cpp
+++ b/src/dxvk/dxvk_cmdlist.cpp
@@ -6,6 +6,7 @@ namespace dxvk {
   DxvkCommandList::DxvkCommandList(DxvkDevice* device)
   : m_device        (device),
     m_vkd           (device->vkd()),
+    m_vki           (device->instance()->vki()),
     m_cmdBuffersUsed(0),
     m_descriptorPoolTracker(device) {
     const auto& graphicsQueue = m_device->queues().graphics;
@@ -206,4 +207,15 @@ namespace dxvk {
     return m_vkd->vkQueueSubmit(queue, 1, &submitInfo, fence);
   }
   
+  void DxvkCommandList::cmdBeginDebugUtilsLabel(VkDebugUtilsLabelEXT *pLabelInfo) {
+    m_vki->vkCmdBeginDebugUtilsLabelEXT(m_execBuffer, pLabelInfo);
+  }
+
+  void DxvkCommandList::cmdEndDebugUtilsLabel() {
+    m_vki->vkCmdEndDebugUtilsLabelEXT(m_execBuffer);
+  }
+
+  void DxvkCommandList::cmdInsertDebugUtilsLabel(VkDebugUtilsLabelEXT *pLabelInfo) {
+    m_vki->vkCmdInsertDebugUtilsLabelEXT(m_execBuffer, pLabelInfo);
+  }
 }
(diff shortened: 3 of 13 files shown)


src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
 
   INT STDMETHODCALLTYPE D3D11UserDefinedAnnotation::BeginEvent(
           LPCWSTR                 Name) {
-    // Currently not implemented
-    return -1;
+    D3D11DeviceContext *context = dynamic_cast<D3D11DeviceContext *>(m_container);
```
Joshua-Ashton, line 36: Instead of this, just make `m_container` a `D3D11DeviceContext` (forward declare the class in the header).

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
...
+
+    if (context) {
+      context->EmitCs([labelName = dxvk::str::fromws(Name)](DxvkContext *ctx) {
+        VkDebugUtilsLabelEXT label = { .sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT };
```
Joshua-Ashton, line 40: This is a C++20 feature :(

Sorry

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
...
+        label.pLabelName = labelName.c_str();
+
+        ctx->beginDebugLabel(
```
Joshua-Ashton, line 43: ctx->beginDebugLabel(&label);

Joshua-Ashton (changes requested):

src/dxvk/dxvk_extensions.h
```diff
@@ -292,8 +292,9 @@ namespace dxvk {
    * used by DXVK if supported by the implementation.
    */
   struct DxvkInstanceExtensions {
+    DxvkExt extDebugUtils                   = { VK_EXT_DEBUG_UTILS_EXTENSION_NAME,                      DxvkExtMode::Optional };
```
Joshua-Ashton, line 295: We shouldn't enable this by default. It's not free.

Joshua-Ashton (changes requested):

src/dxvk/dxvk_extensions.h
Joshua-Ashton: Maybe use an env var for this like we do in VKD3D.

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
 
   INT STDMETHODCALLTYPE D3D11UserDefinedAnnotation::BeginEvent(
           LPCWSTR                 Name) {
-    // Currently not implemented
-    return -1;
+    D3D11DeviceContext *context = dynamic_cast<D3D11DeviceContext *>(m_container);
+
+    if (context) {
```
Joshua-Ashton, line 38: We always have a context. We're a QueryInterface of it!

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
...
+    return m_eventDepth++;
   }
 
 
   INT STDMETHODCALLTYPE D3D11UserDefinedAnnotation::EndEvent() {
-    // Currently not implemented
-    return -1;
+    D3D11DeviceContext *context = dynamic_cast<D3D11DeviceContext *>(m_container);
+
+    if (context) {
```
Joshua-Ashton, line 55: We always have a context. We're a QueryInterface of it!

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
...
+    return m_eventDepth--;
   }
 
 
   void STDMETHODCALLTYPE D3D11UserDefinedAnnotation::SetMarker(
           LPCWSTR                 Name) {
-    // Currently not implemented
+    D3D11DeviceContext *context = dynamic_cast<D3D11DeviceContext *>(m_container);
+
+    if (context) {
```
Joshua-Ashton, line 69: We always have a context. We're a QueryInterface of it!

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
...
+      context->EmitCs([labelName = dxvk::str::fromws(Name)](DxvkContext *ctx) {
+        VkDebugUtilsLabelEXT label = { .sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT };
+        label.pLabelName = labelName.c_str();
+
+        ctx->insertDebugLabel(
```
Joshua-Ashton, line 74: ctx->insertDebugLabel(&label);

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,59 @@ namespace dxvk {
...
+      });
+    }
   }
 
 
   BOOL STDMETHODCALLTYPE D3D11UserDefinedAnnotation::GetStatus() {
-    // Currently not implemented
+    D3D11DeviceContext *context = dynamic_cast<D3D11DeviceContext *>(m_container);
+
+    if (context) {
```
Joshua-Ashton, line 84: You get the picture.

Joshua-Ashton (changes requested):

Joshua-Ashton (changes requested):
Another thing: I'd prefer having a !extDebugUtils check on the frontend also, that way, if games end up using this, we don't have a bunch of overhead from allocating a string and copying into that and pushing that on the CS thread for it to never be used.

src/dxvk/dxvk_extensions.h
DadSchoorse: Or a dxvk.conf option.

liam-middlebrook: I'd prefer to go with an environment variable for toggling this on/off, since it's a bit more lightweight to enable while changing between different applications. How does a boolean `DXVK_PERF_EVENTS` sound?

liam-middlebrook:
> Another thing: I'd prefer having a !extDebugUtils check on the frontend also, that way, if games end up using this, we don't have a bunch of overhead from allocating a string and copying into that and pushing that on the CS thread for it to never be used.

Sure I think that makes sense. Especially since applications _can_ still end these perf events even if the mechanism isn't reported as enabled/supported.

src/dxvk/dxvk_extensions.h
Joshua-Ashton: Sounds good.

src/d3d11/d3d11_annotation.cpp
```diff
@@ -1,9 +1,12 @@
 #include "d3d11_annotation.h"
+#include "d3d11_context.h"
+#include "d3d11_device.h"
 
 namespace dxvk {
 
   D3D11UserDefinedAnnotation::D3D11UserDefinedAnnotation(ID3D11DeviceContext* ctx)
-  : m_container(ctx) { }
+  : m_container(ctx),
```
doitsujin, line 8: Now that we actually interact with the implementation, `ctx` should be changed to `D3D11DeviceContext` to avoid the useless casts.

src/d3d11/d3d11_annotation.h
```diff
@@ -33,6 +33,9 @@ namespace dxvk {
 
     ID3D11DeviceContext*  m_container;
 
+    // Stack depth for non-finalized BeginEvent calls
+    int m_eventDepth;
```
doitsujin, line 37: is there a reason why this is signed? (also, prefer `int32_t` or `uint32_t`)

src/dxvk/dxvk_cmdlist.cpp
```diff
@@ -206,4 +206,15 @@ namespace dxvk {
     return m_vkd->vkQueueSubmit(queue, 1, &submitInfo, fence);
   }
   
-}
\ No newline at end of file
+  void DxvkCommandList::cmdBeginDebugUtilsLabel(VkDebugUtilsLabelEXT *pLabelInfo) {
+    m_device->instance()->vki()->vkCmdBeginDebugUtilsLabelEXT(m_execBuffer, pLabelInfo);
```
doitsujin, line 210: we could also store the instance dispatch table in the command list somewhere to make this a bit less silly

doitsujin (changes requested):
See the things Josh pointed out as well.

src/d3d11/d3d11_annotation.h
Joshua-Ashton: I guess because D3D returns this signed

liam-middlebrook: Sure. I'll make this a sized type. Josh is correct, it's signed because that's how the D3D interface wants it. Basically it's a stack-like behavior of events being pushed/popped and if we have a negative number that indicates an error from the application ending more event regions than it began.

src/d3d11/d3d11_annotation.cpp
liam-middlebrook: Sure.

src/dxvk/dxvk_cmdlist.cpp
liam-middlebrook: Sure.

liam-middlebrook:
Okay, I've addressed the feedback from above and added a new environment variable `DXVK_PERF_EVENTS` for enabling this functionality for use with debugging/profiling tools.

src/dxvk/dxvk_instance.cpp
```diff
@@ -105,6 +105,26 @@ namespace dxvk {
...
+      if (!extensionsAvailable.enableExtensions(
+            utilsExtList.size(),
+            utilsExtList.data(),
+            utilsExtensionsEnabled))
+        throw DxvkError("DxvkInstance: Failed to create instance");
+
+      extensionsEnabled.merge(utilsExtensionsEnabled);
+    }
+
+    m_extensions = insExtensions;
```
doitsujin, line 126: I think a much easier way to do this would be to turn `insExtensionList` into an `std::vector` and conditionally push the pointer to it. The whole extension merging thing is meant for extensions that we don't necessarily know about.

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,60 @@ namespace dxvk {
...
           LPCWSTR                 Name) {
-    // Currently not implemented
-    return -1;
+    if (!m_container->IsAnnotationEnabled())
+      return -1;
+
+    m_container->EmitCs([labelName = dxvk::str::fromws(Name)](DxvkContext *ctx) {
+      VkDebugUtilsLabelEXT label;
+      label.sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT;
+      label.pNext = NULL;
```
doitsujin, line 42: `nullptr`

src/d3d11/d3d11_annotation.cpp
```diff
@@ -30,26 +33,60 @@ namespace dxvk {
...
   void STDMETHODCALLTYPE D3D11UserDefinedAnnotation::SetMarker(
           LPCWSTR                 Name) {
-    // Currently not implemented
+    if (!m_container->IsAnnotationEnabled())
+      return;
+
+    m_container->EmitCs([labelName = dxvk::str::fromws(Name)](DxvkContext *ctx) {
+      VkDebugUtilsLabelEXT label;
+      label.sType = VK_STRUCTURE_TYPE_DEBUG_UTILS_LABEL_EXT;
+      label.pNext = NULL;
```
doitsujin, line 76: `nullptr`

doitsujin (changes requested):
Small nits, looks good otherwise.

src/dxvk/dxvk_instance.cpp
liam-middlebrook: Okay that makes sense. So the merging is for the "plugins" like for OpenVR/OpenXR or DXVK Extensions like dxvk-ags or dxvk-nvapi ? Either way, fixed.

Joshua-Ashton (approved):
LGTM

PR #1987, opened 2021-03-22
D3D9: Copy only dirty regions

K0bin:
- Remove implicit discard
- Copy the dirty parts of a resource when flushing it (or when the managed buffer is otherwise used)
- So unless we recently wrote to the managed buffer (GetRenderTargetData), the game can do whatever the hell it wants with that data.

This has a number of advantages over the current approach:
- When a game uses dirty regions, we only need to make a copy of those parts. Previously we had to copy the entire resource.
This is significant in Nostale which keeps reusing a single 512x512 managed texture and writes data into it before every draw call. The data written is usually just ~60x60px so we only have to copy that instead of the whole 512x512 texture over and over again.
- Should result in slightly less GPU load because the GPU also has to do less copying.
- We don't need to synchronize the CS thread at all when a game locks a managed/sysmem resource
- It works with broken games that keep the managed pointer for too long (Warhammer Online)

Nine (or rather Radeonsi) also does it this way. The downside is, that games that only lock and write their managed textures once might need more memory/address space and might load a bit slower. We copy the entire texture for textures that need to be converted in a compute shader because it seemed a bit painful to only copy the dirty parts there. I don't think that's a big deal anyway.

I'm rather scared of this, so it's gonna need a lot of testing.

PR #2006, opened 2021-04-03
[d3d9] Allow aligned Vector4

DadSchoorse:
GCC generates scalar assembly for the current code at `-O2`, so here's my attempt to fix that.

WIP because this requires testing. I only tested one Oblivion apitrace.
It's also my first time working with SSE intrinsics, so :frog:

### DIFF
--- a/src/d3d9/d3d9_device.h
+++ b/src/d3d9/d3d9_device.h
@@ -1152,16 +1152,16 @@ namespace dxvk {
           return D3DERR_INVALIDCALL;
 
         if constexpr (ConstantType == D3D9ConstantType::Float) {
-          auto begin = &set.fConsts[StartRegister];
-          auto end = &begin[Count];
+          const float* source = set.fConsts[StartRegister].data;
+          const size_t size   = Count * sizeof(Vector4);
 
-          std::copy(begin, end, reinterpret_cast<Vector4*>(pConstantData));
+          std::memcpy(pConstantData, source, size);
         }
         else if constexpr (ConstantType == D3D9ConstantType::Int) {
-          auto begin = &set.iConsts[StartRegister];
-          auto end = &begin[Count];
+          const int*  source = set.iConsts[StartRegister].data;
+          const size_t size  = Count * sizeof(Vector4i);
 
-          std::copy(begin, end, reinterpret_cast<Vector4i*>(pConstantData));
+          std::memcpy(pConstantData, source, size);
         }
         else {
           for (uint32_t i = 0; i < Count; i++) {
--- a/src/d3d9/d3d9_state.h
+++ b/src/d3d9/d3d9_state.h
@@ -245,19 +245,21 @@ namespace dxvk {
           bool                 FloatEmu) {
     auto UpdateHelper = [&] (auto& set) {
       if constexpr (ConstantType == D3D9ConstantType::Float) {
-        auto begin = reinterpret_cast<const Vector4*>(pConstantData);
-        auto end   = begin + Count;
 
-        if (!FloatEmu)
-          std::copy(begin, end, &set.fConsts[StartRegister]);
-        else
-          std::transform(begin, end, &set.fConsts[StartRegister], replaceNaN);
+        if (!FloatEmu) {
+          size_t size = Count * sizeof(Vector4);
+
+          std::memcpy(set.fConsts[StartRegister].data, pConstantData, size);
+        }
+        else {
+          for (UINT i = 0; i < Count; i++)
+            set.fConsts[StartRegister + i] = replaceNaN(pConstantData + (i * 4));
+        }
       }
       else if constexpr (ConstantType == D3D9ConstantType::Int) {
-        auto begin = reinterpret_cast<const Vector4i*>(pConstantData);
-        auto end   = begin + Count;
+        size_t size = Count * sizeof(Vector4i);
 
-        std::copy(begin, end, &set.iConsts[StartRegister]);
+        std::memcpy(set.iConsts[StartRegister].data, pConstantData, size);
       }
       else {
         for (uint32_t i = 0; i < Count; i++) {
--- a/src/d3d9/d3d9_util.h
+++ b/src/d3d9/d3d9_util.h
@@ -138,7 +138,7 @@ namespace dxvk {
     if (Matrix == nullptr) // Identity.
       return Matrix4();
 
-    return *(reinterpret_cast<const Matrix4*>(Matrix));
+    return Matrix4(Matrix->m);
   }
 
   uint32_t GetVertexCount(D3DPRIMITIVETYPE type, UINT count);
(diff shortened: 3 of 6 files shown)


DadSchoorse:
Okay, I tested a nostale apitrace + a nsight capture, there's a 1.5% perf increase. (tested against #1987 on top of master).

DadSchoorse:
I'm not quite sure about Vector4i's loads/stores, the integer intrinsics are pretty weird.

K0bin:
Compiles fine on MSVC.

src/d3d9/d3d9_util.h
```diff
@@ -138,7 +138,7 @@ namespace dxvk {
     if (Matrix == nullptr) // Identity.
       return Matrix4();
 
-    return *(reinterpret_cast<const Matrix4*>(Matrix));
+    return Matrix4(reinterpret_cast<const float*>(Matrix));
```
doitsujin, line 141: `Matrix.m`

Absolutely no reason to use `reinterpret_cast` here.

src/d3d9/d3d9_state.h
```diff
@@ -245,19 +245,23 @@ namespace dxvk {
...
-        auto begin = reinterpret_cast<const Vector4*>(pConstantData);
-        auto end   = begin + Count;
 
-        if (!FloatEmu)
-          std::copy(begin, end, &set.fConsts[StartRegister]);
-        else
-          std::transform(begin, end, &set.fConsts[StartRegister], replaceNaN);
+        if (!FloatEmu) {
+          size_t len = Count * sizeof(Vector4);
+          void* dest = reinterpret_cast<void*>(&set.fConsts[StartRegister]);
```
doitsujin, line 251: `reinterpret_cast<void*>` makes no sense

doitsujin (changes requested):
We should split this into two parts, one that allows us to use aligned vector and matrix types and one that actually rewrites the classes. The current implementation looks like it's going to make things worse in some cases because some compilers will not optimize redundant store->load pairs so this needs a lot more work.

src/d3d9/d3d9_state.h
DadSchoorse: it's there to avoid a warning
```
warning: 'void* memcpy(void*, const void*, size_t)' copying an object of non-trivial type 'using Vector4i = struct dxvk::Vector4Base<int>' {aka 'struct dxvk::Vector4Base<int>'} from an array of 'const int' [-Wclass-memaccess]
  251 |         std::memcpy(&set.fConsts[StartRegister], pConstantData, len);
```
open for suggestions what do instead.

DadSchoorse:
I removed the commit that changed Vector4.

src/d3d9/d3d9_state.h
DadSchoorse: ah I guess I can just use 
```c
fConsts[StartRegister].data
```
:frog:

src/util/util_matrix.cpp
```diff
@@ -29,6 +29,13 @@ namespace dxvk {
     data[3] = v3;
   }
 
+  Matrix4::Matrix4(const float matrix[4][4]) {
```
Joshua-Ashton, line 32: Should probably move this and the other constructors to the header

src/d3d9/d3d9_device.h
```diff
@@ -1152,16 +1152,16 @@ namespace dxvk {
           return D3DERR_INVALIDCALL;
 
         if constexpr (ConstantType == D3D9ConstantType::Float) {
-          auto begin = &set.fConsts[StartRegister];
-          auto end = &begin[Count];
+          const float* source = set.fConsts[StartRegister].data;
+          size_t       len    = Count * sizeof(Vector4);
```
Joshua-Ashton, line 1156: `const size_t`, then it will line up :D

Joshua-Ashton: also, call this size, not len. It is not a length.

src/d3d9/d3d9_device.h
```diff
@@ -1152,16 +1152,16 @@ namespace dxvk {
...
 
-          std::copy(begin, end, reinterpret_cast<Vector4*>(pConstantData));
+          std::memcpy(pConstantData, source, len);
         }
         else if constexpr (ConstantType == D3D9ConstantType::Int) {
-          auto begin = &set.iConsts[StartRegister];
-          auto end = &begin[Count];
+          const int* source = set.iConsts[StartRegister].data;
+          size_t     len    = Count * sizeof(Vector4i);
```
Joshua-Ashton, line 1162: size.

src/d3d9/d3d9_device.cpp
```diff
@@ -4755,12 +4755,12 @@ namespace dxvk {
       UploadSoftwareConstantSet(slice.mapPtr, Src, Layout, Shader);
 
     if (constSet.meta.needsConstantCopies) {
-      Vector4* data = reinterpret_cast<Vector4*>(slice.mapPtr);
+      float* data = reinterpret_cast<float*>(slice.mapPtr);
```
Joshua-Ashton, line 4758: The mapped pointer will be page aligned to 4096 so this should be okay to stay as Vector4?

src/d3d9/d3d9_state.h
```diff
@@ -245,19 +245,21 @@ namespace dxvk {
...
       if constexpr (ConstantType == D3D9ConstantType::Float) {
-        auto begin = reinterpret_cast<const Vector4*>(pConstantData);
-        auto end   = begin + Count;
 
-        if (!FloatEmu)
-          std::copy(begin, end, &set.fConsts[StartRegister]);
-        else
-          std::transform(begin, end, &set.fConsts[StartRegister], replaceNaN);
+        if (!FloatEmu) {
+          size_t len = Count * sizeof(Vector4);
```
Joshua-Ashton, line 250: size!

src/d3d9/d3d9_state.h
```diff
@@ -245,19 +245,21 @@ namespace dxvk {
...
+        }
+        else {
+          for (UINT i = 0; i < Count; i++)
+            set.fConsts[StartRegister + i] = replaceNaN(pConstantData + (i * 4));
+        }
       }
       else if constexpr (ConstantType == D3D9ConstantType::Int) {
-        auto begin = reinterpret_cast<const Vector4i*>(pConstantData);
-        auto end   = begin + Count;
+        size_t len = Count * sizeof(Vector4i);
```
Joshua-Ashton, line 260: size

Joshua-Ashton (changes requested):

src/d3d9/d3d9_device.cpp
Joshua-Ashton: Same for other uploads to a buffer ig.

DadSchoorse: Removed the commit

Joshua-Ashton (approved):
LGTM

Issue #1879, opened 2021-01-02
[d3d9] Guild Wars 1 hangs


accela:
Guild Wars 1 at a specific scene hangs and have to kill the game. It is 100% reproducible. 
Disabling dxvk and using wined3d works fine.
Possible point of interest the validation errors in the log below. Thank you.

### Software information
Guild Wars 1

### System information
- GPU: AMD RX 480
- Driver: Mesa git
- Wine version: wine-staging-6.0_rc4
- DXVK version: Latest from master

### Apitrace file(s)
https://drive.google.com/file/d/1XRuh6yCElI-VNbhQWCJ2dhd0sGRgqOwL/view?usp=sharing

### Log files
[guildwars.log](https://github.com/doitsujin/dxvk/files/5760500/guildwars.log)

jrugia:
Can confirm this also happens with an RX 5500XT and Intel HD 5500 (i3 5005U cpu) with Mesa 20.3.2 and dxvk 1.7.3. An ideal spot to reproduce it is by trying to load the "Augury Rock" mission which hangs all the time. Works fine on Nvidia. Weirdly enough it doesn't affect the entire game.

I tried older Mesa builds and all 20.x.x have same issue. Works fine with 19.3.5 though.

@Joshua-Ashton or @doitsujin : This game is an MMO, if you need a test account to work on the issue I'd be happy to lend you one.

doitsujin:
Can you please clarify what "hang" means in this case? GPU hang?

accela:
Graphics freeze, music keep playing, can return to desktop.

jrugia:
I've done a little more research into this. I initially thought Mesa 19.3.5 was the last known version to work with recent dxvk's, I was wrong. The game works fine with Mesa 20.1.xx and dxvk 1.7.3.

If you go as far back as DXVK 1.5.4 though, the game works fine again with newer Mesa versions, in my case I'm running 20.3.2.  Switching to 1.5.5 makes the game unstable again.

@doitsujin: I believe it's the game that hangs, process has to be killed manually. Rest of desktops keeps working fine otherwise.

DXVK 1.5.5: https://user-images.githubusercontent.com/13750621/103485596-be4e5e80-4df7-11eb-907f-513948893c09.mp4

DXVK 1.5.4: https://user-images.githubusercontent.com/13750621/103485617-f655a180-4df7-11eb-96f8-fa064c51f3b1.mp4

Gnatzelle:
[...]

PR #2010, opened 2021-04-04
[d3d9] FF: use correct vector type when dealing with tex coords

K0bin:
Came up in #1879

No idea, if it fixed the issue. The trace doesn't freeze on Nvidia so it might be something else entirely.

Here's the broken FF shaders and their validation errors for reference: [shaders.zip](https://github.com/doitsujin/dxvk/files/6254807/shaders.zip)

### DIFF
--- a/src/d3d9/d3d9_fixed_function.cpp
+++ b/src/d3d9/d3d9_fixed_function.cpp
@@ -1560,7 +1560,7 @@ namespace dxvk {
 
       bool processedTexture = false;
 
-      auto DoBumpmapCoords = [&](uint32_t baseCoords) {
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_fixed_function.cpp
```diff
@@ -1560,7 +1560,7 @@ namespace dxvk {
 
       bool processedTexture = false;
 
-      auto DoBumpmapCoords = [&](uint32_t baseCoords) {
+      auto DoBumpmapCoords = [&](uint32_t baseCoords, uint32_t vectorSpirVType) {
```
Joshua-Ashton, line 1563: uint32_t typeId, uint32_t baseCoords

Joshua-Ashton (changes requested):

PR #2011, opened 2021-04-04
Adjust required features for lavapipe

doitsujin:
This + some lavapipe bug fixes + the pending Vulkan 1.1 MR should allow DXVK to run on that, at least to a degree.


Joshua-Ashton:
First two look fine, I wouldn't bother with the `image_format_list` one given it can essentially be a no-op for everyone who it doesn't affect.

doitsujin:
Fair enough, I'll drop that commit.

PR #2039, opened 2021-04-22
[d3d9] Minor changes to buffer locking

K0bin:
Now that we're uploading using a staging buffer,
we don't need to wait for non default buffers.

We should also respect READONLY for WRITEONLY buffers
(yes, it sounds dumb) because Nostale relies on that.


K0bin:
Changes were merged in #2111

PR #2043, opened 2021-04-26
[build] enable -mincoming-stack-boundary=2 on i386

DadSchoorse:
On 32bit windows the stack only has to be 4 byte aligned, but for some reason gcc expects 16 by default.

Needs some performance testing to make sure there are no big regressions.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -81,6 +81,11 @@ elif cpu_family == 'x86'
   if dxvk_compiler.has_argument('-mfpmath=sse')
     add_project_arguments('-mfpmath=sse', language: ['c', 'cpp'])
   endif
+  # On 32bit windows the stack only has to be 4 byte aligned,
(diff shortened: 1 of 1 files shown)


DadSchoorse:
I think thing the change is correct. But I'm closing this since it's not resulting in issues rn and I want to have a clean PR overview

mirh:
Follow the links here for why this could have been the case. 
https://gitlab.freedesktop.org/mesa/mesa/-/issues/122

DadSchoorse:
What are you trying to say? I know what's the issue and I know how you could trigger it (e.g. by aligning dxvk's Vector4 class to 16 bytes).

mirh:
You said in the first post that you didn't know what reason gcc had to do what they do, and there you have a decade old issue that discuss the matter.

Issue #2046, opened 2021-04-30
[d3d9] Lightning Returns: Final Fantasy XIII


dlshinobi:
Hi, 

The game renders without any graphical issues with DXVK 1.7.0, but going up to 1.8.1+, the 3D elements (world, models) is a mess, 2D elements (GUI) is fine.

If it's DXVK regression then it's between 1.7.0 - 1.7.1 releases.
Can't bisect any further, 'cause don't have DXVK build setup.

If it's nVidia driver problem, then why it's working with 1.7.0 and not 1.7.1 or above?

So to sum up: no matter of the drivers or wine/proton version, it works with 1.7.0, but not with 1.7.1 or above.

`winetricks d3dx9` didn't help.

P.S. Unrelated with the issue, but game spams a lot of this message, so can we just print this once?
https://github.com/doitsujin/dxvk/blob/525fd53bbc6c2c0ff73560bca9f543652035eddc/src/d3d9/d3d9_swapchain.cpp#L658

---

DXVK v.1.7.0:
![1 7 0](https://user-images.githubusercontent.com/7222444/116659699-a7310f00-a99a-11eb-8fb9-bc9804773277.png)

DXVK 1.8.1:
![1 8 1](https://user-images.githubusercontent.com/7222444/116659712-ac8e5980-a99a-11eb-814c-34572d8bf5e7.png)

### Software information
Lightning Returns: Final Fantasy XIII

### System information
- GPU: nVidia GTX 960
- Driver: stable (465.24.02), vulkan-dev (455.26.01, 455.50.14)
- Wine version: proton 5.13, proton 6.3 (experimental)
- DXVK version: 1.7.0 - 1.8.1 + master artifact.

### Apitrace file(s)
- apitrace: https://drive.google.com/file/d/1pNDpECJ_WH6iWzTLPTYCmLOxrXJ5_wOF/view?usp=sharing

### Log files
- d3d9.log: https://drive.google.com/file/d/1_u7Tm9Gs4A3gME8RQqEdW9seroolKmnO/view?usp=sharing

Joshua-Ashton:
Does https://github.com/doitsujin/dxvk/pull/2050 fix the issue?

dlshinobi:
> Does #2050 fix the issue?

Aha, works like a charm! Thank you! :thumbsup:

Issue #1908, opened 2021-01-25
Spec Ops: The Line - God Rays rendering issue


gabriele2000:
### Software information
Spec Ops: The line - DXGI latest release

### System information
- GPU: Nvidia GTX1050TI 4GB (notebook)
- Driver: Nvidia-460
- Wine version: Latest staging (6.0)
- DXVK version: Latest release 1.7.3
### Trace files:
https://mega.nz/file/tp10QCQT#uoaTnly6UzNyR0U-JPQ6PrwrHndprgRFmH0L9C8vC-M

### Log files
- d3d9.log: [SpecOpsTheLine_d3d9.log](https://github.com/doitsujin/dxvk/files/5866512/SpecOpsTheLine_d3d9.log)

![SpecOpsTheLine 2021-01-25 13-47-54](https://user-images.githubusercontent.com/12186181/105708297-875dfb00-5f14-11eb-86e7-15ac73f19e9e.png)
![SpecOpsTheLine 2021-01-25 13-47-49](https://user-images.githubusercontent.com/12186181/105708242-71e8d100-5f14-11eb-8f67-0588dcc85eac.png)

gabriele2000:
Apitrace uploaded

gabriele2000:
The issue seems releated to a particular fog effect too, post-processing stuff I guess, later in that level, that diagonal line was still there, acting as "saturation" thing

doitsujin:
The apitrace renders fine on my Polaris card as far as I can tell. Not sure if the Nvidia driver is screwing up somewhere but it doesn't look like we're doing anything wrong.

gabriele2000:
> 
> 
> The apitrace renders fine on my Polaris card as far as I can tell. Not sure if the Nvidia driver is screwing up somewhere but it doesn't look like we're doing anything wrong.

Today my pop_os updated the nvidia driver to its nvidia driver made for that distro... I'll test again

WinterSnowfall:
Initially this kind of looked like yer' old Nvidia Optimus diagonal screen tearing to me, based on the screenshots screenshots. But I played back the api trace on my (desktop) GTX 1080, and I'm seeing the same thing (with 460.32.03) . So yes, probably an Nvidia issue, since AMD seems fine as per the above.

gabriele2000:
[...]

PR #2050, opened 2021-04-30
[d3d9] Account for vertex declaration size for UP draws

Joshua-Ashton:
The stride may not give us the full picture here as the stride
may not encompass the vertex declaration entirely.

Consider a vertex declaration of size 20, and a stride of 12,
we may not have covered the whole range of space the draw wants with
VertexCount * Stride.

Instead, take (VertexCount - 1) * Stride + VertexDeclSize

Closes: #2046
Closes: #1908

### DIFF
--- a/src/d3d9/d3d9_device.h
+++ b/src/d3d9/d3d9_device.h
@@ -1085,6 +1085,31 @@ namespace dxvk {
       const DWORD*                pShaderBytecode,
       const DxsoModuleInfo*       pModuleInfo);
 
+    inline uint32_t GetUPDataSize(uint32_t vertexCount, uint32_t stride) {
+      return vertexCount * stride;
+    }
+
+    inline uint32_t GetUPBufferSize(uint32_t vertexCount, uint32_t stride) {
+      return (vertexCount - 1) * stride + m_state.vertexDecl->GetSize();
+    }
+
+    inline void FillUPVertexBuffer(void* buffer, const void* userData, uint32_t dataSize, uint32_t bufferSize) {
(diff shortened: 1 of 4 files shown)


Joshua-Ashton:
Draft until I hear back about this fixing the issue

src/d3d9/d3d9_device.h
```diff
@@ -1186,6 +1186,13 @@ namespace dxvk {
         : GetHelper(m_state.psConsts);
     }
 
+    inline uint32_t CalculateVertexSizeUP(uint32_t vertexCount, uint32_t stride) {
+      if (unlikely(!vertexCount))
```
doitsujin, line 1190: can we exit early from `DrawPrimitiveUP` instead if the app is dumb? We shouldn't submit an empty draw to the backend.

doitsujin (commented):
Nit, looks fine otherwise. Dumb games be dumb.

Joshua-Ashton:
I wanna take a second look at this tomorrow... Had some bedtime thoughts about it.

Joshua-Ashton:
Ended up looking into this further and what I did here was slightly incorrect, the size of the user data remains the same.

However we need to pad the buffer to this size with 0s instead.

The game has a float3 texcoord trailing just outside the buffer range which causes the whole vertex element to be 0 on NVIDIA, rather than just the .z oob component.

src/d3d9/d3d9_device.h
Joshua-Ashton: Done

PR #2052, opened 2021-05-07
artifacts.yml: Remove unused ids, make env vars into outputs

Margen67:


Joshua-Ashton:
Can we not change the Ubuntu version? We don't actually use it, we immediately use a docker inside of it that runs Arch and I don't want that to randomly break.

Please don't remove the IDs, I may want them in future when adding matrix builds.

Margen67:
> 
> 
> Can we not change the Ubuntu version? We don't actually use it, we immediately use a docker inside of it that runs Arch and I don't want that to randomly break.
> 
> Please don't remove the IDs, I may want them in future when adding matrix builds.

Reverting the Ubuntu version is fine.

The steps that had their id removed don't have any outputs, so unless you're wanting to use [conclusion/outcome](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#steps-context) I don't see why you'd want them.

PR #2053, opened 2021-05-07
Add workarounds for RE8

Joshua-Ashton:
This PR adds a global DXGI factory for Resident Evil 8 as it spams `CreateDXGIFactory` calls.

This also caches display modes on DXGI outputs as the game also enumerates them every frame which was causing allocations for us due to the way this was handled previously.

Now the game gets put on a fast path where it just ends up returning the cached data constantly.

### DIFF
--- a/src/dxgi/dxgi_adapter.cpp
+++ b/src/dxgi/dxgi_adapter.cpp
@@ -61,7 +61,20 @@ namespace dxvk {
     m_adapter (adapter),
     m_interop (this),
     m_index   (index) {
-    
+    for (uint32_t i = 0; ; i++) {
+      MonitorEnumInfo info;
+      info.iMonitorId = i;
+      info.oMonitor   = nullptr;
+      
+      ::EnumDisplayMonitors(
+        nullptr, nullptr, &MonitorEnumProc,
+        reinterpret_cast<LPARAM>(&info));
+      
+      if (info.oMonitor == nullptr)
+        break;
+      
+      m_outputCache.emplace_back(new DxgiOutput(m_factory, this, info.oMonitor));
+    }
   }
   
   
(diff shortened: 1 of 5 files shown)


doitsujin:
> This PR adds a global DXGI factory for Resident Evil 8 as it spams CreateDXGIFactory calls.

Does this match Windows behaviour?

Joshua-Ashton:
It does not. The pointers are definitely different on Windows, but the overhead is much lower there.

Andon13:
Actually, there's pretty significant performance overhead from this in Windows too.

Special K shows about 3 ms in API shenanigans every time the engine does this stuff.

My solution involves caching the factory, caching the IDXGIAdapter1 acquired using said factory, and then caching the mode list until the factory's `IsCurrent (...)` method fails. At which point, it's necessary to teardown that factory, cache and all.

doitsujin:
caching DXGI factories seems strictly wrong though so we're not going to do that.

Andon13:
Fair enough.

I'm all but certain the reason for this behavior in the engine is because Capcom doesn't know about `WM_DISPLAYCHANGE` notifications and the correct usage of `IDXGIFactory1::IsCurrent (...)`. They are releasing an re-acquiring a new factory every frame because otherwise Windows returns a stale cached mode list.

Their resolution management code in general is terrible.


As long as DXVK is performant while the engine is going about getting resolutions the wrong way, you shouldn't need to cache factories. On Windows, there's substantial overhead caused by releasing these factories, since `GetDisplayModeList (...)` API overhead is much higher immediately after creating a new factory.

Oschowa:
I'm pretty sure this is fixed in the game with the patch released on August 10th. I've tried with an unpatched build of DXVK's DXGI and get 300 fps with CPU bound settings on the title screen which was a slide-show before without this workaround. Also tagging @aeikum for the workaround in wine's DXGI in proton experimental.

PR #2061, opened 2021-05-13
Return custom device description for Crazy Machines 3 

Joshua-Ashton:


Joshua-Ashton:
Turns out this isn't required anymore.

PR #2075, opened 2021-05-31
Low-latency frame latency

Joshua-Ashton:
Move image aquisition to  after presentation rather than before to avoid stalling when the game wants to present when we could be waiting on WSI crap.

### DIFF
--- a/src/d3d11/d3d11_swapchain.cpp
+++ b/src/d3d11/d3d11_swapchain.cpp
@@ -178,9 +178,8 @@ namespace dxvk {
       return DXGI_ERROR_INVALID_CALL;
 
     m_frameLatency = MaxLatency;
-    m_frameLatencySignal->wait(m_frameId - GetActualFrameLatency());
 
-    SignalFrameLatencyEvent();
+    SyncFrameLatency();
     return S_OK;
   }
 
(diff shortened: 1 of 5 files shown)


src/d3d11/d3d11_swapchain.cpp
```diff
@@ -241,59 +241,89 @@ namespace dxvk {
...
     }
 
+    // Signal the frame latency event for *this* frame.
     SignalFrameLatencyEvent();
-    return S_OK;
+
+    // Wait on the frame latency for the subsequent frame
+    // to avoid any latency at the next presentation.
+    // Bump our frame id.
+    WaitForFrameLatency(++m_frameId);
```
doitsujin, line 289: This will immediately synchronize if frame latency is 1, rather than allow for a single frame to be queued.

src/d3d11/d3d11_swapchain.cpp
```diff
@@ -241,59 +240,80 @@ namespace dxvk {
...
+
+    SyncFrameLatency();
+
+    // Acquire the image for the next present.
+    return AcquireNextImage();
+  }
+
+
+  bool D3D11SwapChain::AcquireNextImage() {
+    m_nextImageIndex = UINT32_MAX;
```
doitsujin, line 293: Can we use some properly named constant instead of `UINT32_MAX`? This is some non-standard windows crap that we're not really using anywhere at least in the d3d11/dxgi code.

doitsujin (changes requested):
Nit, looks good otherwise.

src/d3d11/d3d11_swapchain.cpp
Joshua-Ashton: UINT32_MAX is from stdint, it's not non-standard Windows crap.

But yeah, it can have it's own constant

Joshua-Ashton: Added `InvalidImageIndex = ~0u`

doitsujin:
Actually, as discussed on Discord, this currently breaks pretty much everything due to the numerous threads involved. I'll try hacking this into `vk::Presenter` instead, *but* we should keep the frame latency fixes (notably, moving the wait after the present).

Joshua-Ashton:
Updated PR to only include the frame latency changes with the recent stuff in master.

PR #2077, opened 2021-06-01
Add some Proton per-game workarounds

Joshua-Ashton:


dxvk.conf
```diff
@@ -31,6 +31,15 @@
 # d3d9.customVendorId = 0000
 
 
+# Override the reported device description
+#
+# Supported values: Any string.
+
+# dxgi.customDeviceDesc = ""
+
```
doitsujin, line 39: no space here please

src/dxgi/dxgi_adapter.cpp
```diff
@@ -283,6 +283,15 @@ namespace dxvk {
...
         sharedMemory += heap.size;
     }
+
+    // Some games think we are on Intel given a lack of
+    // NVAPI or AGS/atiadlxx support.
+    // Report our device memory as shared memory,
+    // and some small amount for the carveout.
+    if (options->emulateUMA && !m_adapter->isUnifiedMemoryArchitecture()) {
+      deviceMemory = 128 * (1 << 20);
+      sharedMemory = deviceMemory;
```
doitsujin, line 293: surely this should be the other way around? We're now reporting 128MB for *both* shared and device memory.

doitsujin (changes requested):

src/dxgi/dxgi_adapter.cpp
Joshua-Ashton: Yes this should :frog:

src/dxgi/dxgi_adapter.cpp
```diff
@@ -278,6 +283,15 @@ namespace dxvk {
       else
         sharedMemory += heap.size;
     }
+
+    // Some games think we are on Intel given a lack of
+    // NVAPI or AGS/atiadlxx support.
+    // Report our device memory as shared memory,
+    // and some small amount for the carveout.
+    if (options->emulateUMA && !m_adapter->isUnifiedMemoryArchitecture()) {
+      sharedMemory = deviceMemory;
```
DadSchoorse, line 292: Won't this be bad on low VRAM cards, like 2GB?

doitsujin: i mean this is literally a GTA IV hack, whatever makes that game happy should work here. Most games seem to ignore this value anyway.

Issue #2079, opened 2021-06-04
Duplicate dxvk-cache/logs due to case-(in)sensitivity


SirFrags:
GetModuleFileNameW returns a string with the execution path case and not the actual path case.  This causes the file extension detection for dxvk-cache/logs to be unsuccessful.  Example:

Running GAME.EXE as GAME.EXE results in a file prefix of "GAME.EXE" while running it as "GAME.exe" results in a prefix of "GAME" causing two sets of dxvk-cache/logs.

### Software information
Windows (10)

### Possible solution
dxvk_state_cache.cpp
> getCacheFileName()
```
if (extp != std::string::npos && exeName.substr(extp + 1)== "exe")
  exeName.erase(extp);
```

change to:
```
std::string extLower = exeName.substr(extp + 1);
for (auto& c : extLower)
  c = std::tolower(c);

if (extp != std::string::npos && extLower == "exe")
  exeName.erase(extp);
```
log.cpp
> getFileName()

`Same fix`

Joshua-Ashton:
Does https://github.com/doitsujin/dxvk/pull/2080 fix it?

SirFrags:
@Joshua-Ashton the pr fails to build but I made the intended changes locally and it fixes the problem - needs <algorithm> for std::transform instead of <cctype>

Joshua-Ashton:
Well, cctype gives tolower :p 

Ye, I just wrote that before i head out, will fix it up shortly.

PR #2080, opened 2021-06-04
Handle casing correctly in finding cache/log name

Joshua-Ashton:
Otherwise any game that has a .EXE instead of a .exe or some other casing variant will generate a bad state cache name.

Closes: #2079


doitsujin:
This will almost certainly break unicode file names.

Can we just fix the `== "exe"` thing to something that essentially checks for `[eE][xX][eE]`?

Joshua-Ashton:
Ye

Issue #2056, opened 2021-05-10
[d3d9] Second Sight → missing HUD, subtitles, transition effect during cutscenes...


NerosTie:
There are many issues with DXVK and Second Sight.

First, the HUD and the subtitles are missing:

WINED3D:
![Screenshot_20210510_200000](https://user-images.githubusercontent.com/4572452/117716616-1a782380-b1da-11eb-93b1-0b47d5486844.png)

DXVK:
![Screenshot_20210510_195457](https://user-images.githubusercontent.com/4572452/117716628-1e0baa80-b1da-11eb-9888-d6e8e8372da0.png)

Before the main menu, there is a screen with the title of the game, it's missing with DXVK:

WINED3D:
![Screenshot_20210510_201933](https://user-images.githubusercontent.com/4572452/117717762-85762a00-b1db-11eb-8e8e-7c72b480479a.png)

DXVK:
![Screenshot_20210510_215638](https://user-images.githubusercontent.com/4572452/117717708-71cac380-b1db-11eb-9000-ea5a379e7613.png)

During the cutscene of the first level, there are a transition effect which is missing with DXVK. Also notice that the 2 videos begin at the same time, so the black screen with the word "ISOLEMENT" is missing with DXVK.

WINED3D:
https://streamable.com/lsq57d

DXVK:
https://streamable.com/xaoip4

### System information
- GPU: AMD RX 580
- Driver: Mesa 21.1.0
- Wine version: proton-experimental (2021/05/07)
- DXVK version: 1.8.1 (also tested with the last [dev build,](https://github.com/doitsujin/dxvk/actions/runs/821617018) same results)

### Apitrace file(s)
I can't get one... I've tried everything but the wrapper does nothing or it's ignored. I even tried with a GOG version found *somewhere* but with the same result... If someone knows how to get one, please upload it, it would be a great help!

### Log files
[steam-11550_dxvk.log](https://github.com/doitsujin/dxvk/files/6454676/steam-11550_dxvk.log)
[steam-11550_wined3d.log](https://github.com/doitsujin/dxvk/files/6454677/steam-11550_wined3d.log)

Joshua-Ashton:
I cannot do anything without an apitrace I am afraid.

NerosTie:
I tried for 2+ hours to have an apitrace but with no success, I really don't know what's the issue with this game or where I do something wrong :/

DadSchoorse:
[...]

PR #2088, opened 2021-06-06
[d3d9] Don't scale z to [0, 1] for POSITIONT

Joshua-Ashton:
I originally thought the depth clipping region was always [0, 1] when I first implemented this nearly 2 years ago.

The depth clipping region is already in the viewport's depth range, so just don't do anything here if we are z-testing.

( We still need to keep the flattening around for when ztest is disabled though :( )

Fixes: #2056

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -52,6 +52,9 @@ fi
 
 # find wine executable
 export WINEDEBUG=-all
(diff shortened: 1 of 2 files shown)


doitsujin (approved):
LGTM assuming that this has been tested.

PR #2091, opened 2021-06-09
Frame rate limiter

doitsujin:
Title. We can enable this by default for games that break at certain frame rates but are too dumb to realize it.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -24,6 +24,7 @@
 # 
 # Supported values : Any non-negative integer
 
+# dxgi.maxFrameRate = 0
 # d3d9.maxFrameRate = 0
 
 
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -19,6 +19,14 @@
 # d3d9.maxFrameLatency = 0
 
 
+# Enables a frame rate limiter, unless the game is already
+# limited to the same refresh rate by vertical synchronization.
+# 
+# Supported values : Any non-negative integer
+
+# d3d9.maxFrameRate = 0
+
+
 # Override PCI vendor and device IDs reported to the application. Can
 # cause the app to adjust behaviour depending on the selected values.
 #
--- a/src/d3d11/d3d11_options.cpp
+++ b/src/d3d11/d3d11_options.cpp
@@ -19,6 +19,7 @@ namespace dxvk {
     this->deferSurfaceCreation  = config.getOption<bool>("dxgi.deferSurfaceCreation", false);
     this->numBackBuffers        = config.getOption<int32_t>("dxgi.numBackBuffers", 0);
     this->maxFrameLatency       = config.getOption<int32_t>("dxgi.maxFrameLatency", 0);
+    this->maxFrameRate          = config.getOption<int32_t>("dxgi.maxFrameRate", 0);
     this->syncInterval          = config.getOption<int32_t>("dxgi.syncInterval", -1);
     this->tearFree              = config.getOption<Tristate>("dxgi.tearFree", Tristate::Auto);
 
--- a/src/d3d11/d3d11_options.h
+++ b/src/d3d11/d3d11_options.h
@@ -83,6 +83,9 @@ namespace dxvk {
     /// a higher value. May help with frame timing issues.
     int32_t maxFrameLatency;
 
+    /// Limit frame rate
+    int32_t maxFrameRate;
+
     /// Defer surface creation until first present call. This
     /// fixes issues with games that create multiple swap chains
     /// for a single window that may interfere with each other.
--- a/src/d3d11/d3d11_swapchain.cpp
+++ b/src/d3d11/d3d11_swapchain.cpp
@@ -22,7 +22,8 @@ namespace dxvk {
     m_desc      (*pDesc),
     m_device    (pDevice->GetDXVKDevice()),
     m_context   (m_device->createContext()),
-    m_frameLatencyCap(pDevice->GetOptions()->maxFrameLatency) {
+    m_frameLatencyCap(pDevice->GetOptions()->maxFrameLatency),
+    m_fpsLimiter(double(pDevice->GetOptions()->maxFrameRate)) {
     CreateFrameLatencyEvent();
 
     if (!pDevice->GetOptions()->deferSurfaceCreation)
@@ -285,6 +286,8 @@ namespace dxvk {
 
       SubmitPresent(immediateContext, sync, i);
     }
+
+    m_fpsLimiter.delay(SyncInterval);
 
     SyncFrameLatency();
     return S_OK;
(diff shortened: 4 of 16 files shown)


aufkrawall:
Great to have this feature.

Though the limiter of libstrangle shows less variance in Heroes of the Storm, which is also reflected in DXVK's HUD (no vsync and both CPU & GPU forced to maximum clocks):
DXVK:
Screenshot_20210610_141447

libstrangle:
Screenshot_20210610_141122

It doesn't work correctly on Windows. When I set it to 85fps there, it limits at ~65fps with uneven variance:
HeroesOfTheStorm_x64_2021 06 10-14 25


Regarding vsync detection: I wonder if that can interfere with VRR? It often makes sense to combine VRR with vsync + fps limiter, and in such cases it'd likely be desirable to make sure the limiter never suspends. Would be a toggle for this be within scope, just in case?

doitsujin:
Note that this stalls the application thread that presents, while libstrangle stalls a different thread, so naturally the latter will have more consistent present timings on the Vulkan side, but not necessarily the application side.

Not sure what to do about Windows since timers are pretty much broken there.

aufkrawall:
The 1st WIP commit unfortunately doesn't fix Windows. I'm going to try the 2nd.

Regarding fps limiter & VRR: When only vsync is used with VRR without an fps limiter and vsync actually caps the frame rate, I think it presents from the backbuffer queue with additional input/output lag (like without VRR).
The Nvidia D3D11 driver automatically turns on its fps limiter when the "Ultra Low Latency" driver feature is set to "ultra" when VRR is active, allegedly due to this reason. Also, turning off vsync instead isn't an option either if the aim is to guarantee 100% absence of tearing.
So, I think if the fps limiter turns itself off with vsync and VRR, this can only do harm (increase latency) in the best case. Also, fifo vsync is in incredibly bad shape with amdvlk even on Windows and it should be desirable to never let it cap the frame rate with VRR. I've just recently witnessed amdvlk screwing up fifo vsync on Windows in Days Gone when it caps fps, despite of VRR being enabled. :(

aufkrawall:
2nd WIP commit fixes Windows, great! :)

doitsujin:
We already have a mechanism in place for that, and that is setting `maxFrameLatency` to 1. It synchronizes with the GPU and is therefore much more accurate than enabling the limiter. That said, if you have a 144 Hz display, you could still enable the limiter by setting it to something like 139 FPS.

I've tested this on Windows now and it's unfortunately very, very stuttery since even NtDelayExecution is on a 0.5ms granularity with random delays, unlike on wine where it is accurate down to a couple of microseconds assuming you have CPU resources to spare.

aufkrawall:
> We already have a mechanism in place for that, and that is setting `maxFrameLatency` to 1. It synchronizes with the GPU and is therefore much more accurate than enabling the limiter. That said, if you have a 144 Hz display, you could still enable the limiter by setting it to something like 139 FPS.
 
Yeah, maxFrameLatency is the equivalent to "Ultra low latency" (though the "ultra" setting of the latter is probably 0.5 and not 1, which unfortunately exceeds my technical understanding). Though the Nvidia feature has some additional black magic with VRR active by automatically turning on the driver fps limiter as well (combined with the real tight CPU -> GPU sync of the prerender <1 frame). However, I think maxFrameLatency doesn't affect vsync backbuffer queue, it just mitigates it by lowering latency in general?

I probably was a bit too worried about the vsync detection of DXVK's fps limiter anyway, at least additional vsync doesn't seem to make the frame time graph look any worse with an fps limit just 1 frame below refresh rate. Anecdote: I can test this with VRR, as my brother, who owns the VRR capable display, is starting to hate Radeon Chill (AMD Windows driver paper cuts everywhere...)  and would really appreciate a precise limiter in DXVK on Windows. :)

K0bin:
I think the ultra setting in the low latency option also synchronizes with the GPU every frame.

doitsujin:
`maxFrameLatency` synchronizes with the GPU at the same point in time when our frame rate limiter would kick in and is the closest we can get to actual sync-to-vblank. It should result in much more stable frame times than our limiter.

aufkrawall:
> On Windows there's also Riva Tuner which does a very good job and gets you stable frame times. Don't use that with DXVK though.

Yes, it also can have issues with native D3D11. E.g. in Heroes of the Storm native D3D11, RTSS limiter causes crazy flicker with VRR on both AMD & Nvidia. I also suspect it to sometimes have a bad effect on input processing, e.g. I got infrequent mouse input jumps in Battlefield 4 native D3D11 (even without VRR).
The Nvidia driver limiter so far is the only one that has been entirely flawless in all of my tests, across all APIs and with ULL ultra, also with extremely low latency for D3D9/10/11. Well, I'd say on Linux Vulkan, libstrangle works similarly well with regards to reliability and frame time variance.


> Edit, wrt frame times, do note that the frame times displayed by the DXVK HUD are merely the CPU-side present intervals on.... one of our internal threads, they don't really correspond to when the rendered picture will actually be displayed.

Isn't a low CPU frame time variance the premise to get to a clean GPU render variance + output/present? The HUD of DXVK shows some frame time spikes that are completely missed by the graph of RTSS, but are definitely shown on screen. In those moments, the 60Hz VA panels flickers an awful lot, which makes it clear which graph is closer to reality.  VRR with a 60Hz VA panel btw. makes even the slightest deviation from a perfect presentation well noticeable, usually even tiny spikes in DXVK's graph can be noticed as stutter/flicker. The same of course is true for having any additional input lag at 60Hz. Despite of VRR, it's always laggy as hell with vsync enabled and no fps limiter (also with `dxgi.maxFrameLatency = 1`).

Anyhow, I don't want to annoy you further with synthetic scenarios or anecdotes. I think the only thing that matter is that there are no perceivable issues. When I tried with the WIP commit v2, scrolling over the map in HotS via constant slow keyboard input came with not too severe, but still noticeable micro stutter with VRR on Windows, which is not there with Chill. I can try the once again updated PR again with VRR tomorrow.

aufkrawall:
This is on Linux with mangohud and latest PR version, it shows the same pattern as dxvk's graph (flat with libstrangle):
Screenshot_20210610_212527
It's also flat with dxvk in the game's menu, which might suggest CPU/GPU power saving feature's to be the culprit, but I double checked the CPU governor to be really set to performance and power_dpm_force_performance_level set to high for amdgpu.

doitsujin:
I made a new branch here: <https://github.com/doitsujin/dxvk/tree/fps-limiter-test>

This one works more like libstrangle/mangohud in that it stalls after Vulkan presentation and not on the DXGI/D3D9 side, which should hopefully fix wonky frame times.

aufkrawall:
Seems to be entirely flat now on both Linux and Windows, fantastic! :)
Really looking forward to testing it with VRR tomorrow.

PR #2094, opened 2021-06-09
[dxvk] only set strict fp mode if supported and requested

ryanizer:
Improves performance on typical shaders by ~5-7%

### DIFF
--- a/src/dxbc/dxbc_compiler.cpp
+++ b/src/dxbc/dxbc_compiler.cpp
@@ -7456,29 +7456,36 @@ namespace dxvk {
     const uint32_t width32 = 32;
     const uint32_t width64 = 64;
 
-    m_module.enableExtension("SPV_KHR_float_controls");
+    bool enableFPControls = false;
 
     if (flags.test(DxbcFloatControlFlag::DenormFlushToZero32)) {
       m_module.enableCapability(spv::CapabilityDenormFlushToZero);
       m_module.setExecutionMode(m_entryPointId, spv::ExecutionModeDenormFlushToZero, 1, &width32);
+      enableFPControls = true;
     }
 
-    if (flags.test(DxbcFloatControlFlag::PreserveNan32)) {
+    if (flags.test(DxbcFloatControlFlag::PreserveNan32) && m_precise) {
       m_module.enableCapability(spv::CapabilitySignedZeroInfNanPreserve);
       m_module.setExecutionMode(m_entryPointId, spv::ExecutionModeSignedZeroInfNanPreserve, 1, &width32);
+      enableFPControls = true;
     }
 
     if (m_module.hasCapability(spv::CapabilityFloat64)) {
       if (flags.test(DxbcFloatControlFlag::DenormPreserve64)) {
         m_module.enableCapability(spv::CapabilityDenormPreserve);
         m_module.setExecutionMode(m_entryPointId, spv::ExecutionModeDenormPreserve, 1, &width64);
+        enableFPControls = true;
       }
 
-      if (flags.test(DxbcFloatControlFlag::PreserveNan64)) {
+      if (flags.test(DxbcFloatControlFlag::PreserveNan64) && m_precise) {
         m_module.enableCapability(spv::CapabilitySignedZeroInfNanPreserve);
         m_module.setExecutionMode(m_entryPointId, spv::ExecutionModeSignedZeroInfNanPreserve, 1, &width64);
+        enableFPControls = true;
       }
     }
+
+    if (enableFPControls)
+      m_module.enableExtension("SPV_KHR_float_controls");
   }
 
 


liam-middlebrook:
The `Reviewed-by` line for my name has a typo. `nvida.com` where it should be `nvidia.com`

liam-middlebrook:
Rather than saying "typical shaders" it's probably a good idea for us to call out that we found this change nets a ~8.2% increase on RTX 2060 and ~3.1% increase on RTX 3070 for a DX11 apitrace of World of Warcraft.

Also the description should note that this fixes a regression introduced by https://github.com/doitsujin/dxvk/commit/f5fa7a909961692431306f0c07f583a54cfece49

Joshua-Ashton:
I guess my question is, why do we get a hit from enabling the extension alone without using any of the modes?

Is the m_precise stuff documented anywhere?

Joshua-Ashton:
Also has this been tested in the Unity titles where we have NaN framebuffer output problems?

liam-middlebrook:
Also I think the commit description should be prefixed with `[dxbc]` rather than `[dxvk]` since this is a change in the `dxbc` directory.

ryanizer:
Enabling the extension *shouldn't* cause code changes without any modes, but enabling extensions that aren't needed seems like a bad policy.

m_precise was pre-existing in the codebase, I just noticed the usage and that it seemed to be what I needed to key off. Suggestions?

Joshua-Ashton:
Well, I don't see why the fp rules set wouldn't apply just because we're not m_precise.

liam-middlebrook:
I think this should be dependent on whether `DxbcGlobalFlag::RefactoringAllowed` is set, rather than `m_precise`

Joshua-Ashton:
But that is always set for 5_0 -- it's only not set for 4_0 when the position is marked as precise afaik.

Maybe @doitsujin can correct that.

liam-middlebrook:
Then maybe we should just check the global flags directly again? Note the following spec language:
```
After refactoring, symbolic arithmetic must remain equivalent to the original, albeit without having to respect limits such as where NaN or INF may be produced as a byproduct of the order of operations.
```

From: https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#22.3.2%20Global%20Flags%20Declaration%20Statement

Joshua-Ashton:
```
When REFACTORING_ALLOWED is used on a shader, individual instructions can opt out
                by using the precise(22.19.2) modifier.
```

I guess this part is not handled for the float mode then.

Joshua-Ashton:
I guess we don't have a decoration in SPIR-V for this per-instruction...

doitsujin:
We're using float controls to guarantee the [floating point rules](https://docs.microsoft.com/en-us/windows/win32/direct3d11/floating-point-rules) which apply regardless of the global flags.

In particular, we ran into an issue with Mesa where drivers would not preserve NaNs on texture fetches, which is legal behaviour on Vulkan but something that just works in D3D11 regardless of shader settings.

doitsujin:
FWIW, the D3D11 spec language suggests that NaN may still be generated whereas Vulkan without `SignedZeroInfNanPreserve` assumes that no instruction will emit NaN. This is not the same thing.

doitsujin:
One more thing - like Josh said earlier, `RefactoringAllowed` is almost always set, there is no way to generate Shader Movel 5.0 code without it set. The only exception I found so far are SM4.1 vertex shaders.

doitsujin:
So apparently current Mesa doesn't require this anymore (at least in *some* cases?)

I'd really appreviate if someone could actually tell us what to about this do because this issue is blocking an important release right now. This PR clearly doesn't do the right thing, so I'm just tempted to do `if (Nvidia) disable_float_controls=true` at this point.

liam-middlebrook:
> I'd really appreviate if someone could actually tell us what to do about this because this issue is blocking an important release right now. This PR clearly doesn't do the right thing, so I'm just tempted to revert the chnage completely or just conditionally not enable it on Nvidia at this point. Problem is that all of that will require exhaustive testing which pushes our release back by several weeks at least.

Please don't feel that you need to hold the release on our behalf. I'd say the best course of action for now is to leave the behavior as-is, and any changes can be left for a future release.

liberodark:
This increase only affects Nvidia or Mesa too ?
I myself have not yet tested.
If this does not imply regression it may be interesting.
But if you have any doubts then you can push back to a later version.

doitsujin:
@liberodark you can test for yourself by setting `d3d11.floatControls = False` in dxvk.conf.

From what I can tell it doesn't matter much; Mesa got some optimizations for it around the same time when we started using the extension.

dustContributor:
When setting `d3d11.floatControls = False` like suggested, I get a 3% to 4% improvement on Deus Ex Mankind Divided benchmark with a GeForce 1660 Ti, 465.19.01 driver and DXVK 1.9.0. I had a similar improvement when trying the artifact from this PR.

Very GPU bound (100% usage during the benchmark). Specifically, from 67-68 fps avg to 70 fps avg.

ryanizer:
@doitsujin + @Joshua-Ashton -

Sorry for the extreme delay!

[Here](https://github.com/doitsujin/dxvk/pull/2094#issuecomment-858083039), it is mentioned that NaNs returned from texture fetches are guaranteed in DX11 and potentially ignored in VK. I can’t find anything about this in the DX, VK, or SPV_KHR_float_controls extension specs. Can you please point to these?

Also mentioned:
“We're using float controls to guarantee the floating point rules **which apply regardless of the global flags**.” 

AFAICT, this is in contradiction with the global flags and precision [documentation](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx-graphics-hlsl-variable-syntax):

> [When precise keyword is applied] IEEE unsafe operations are restricted. Where the compiler might have used fast math operations that don't account for NaN (not a number) and INF (infinite) values, precise forces IEEE requirements concerning NaN and INF values to be respected. **Without precise, these optimizations and mathematical operations are not IEEE safe**.

and from the [D3D11 specs](https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#22.3.2%20Global%20Flags%20Declaration%20Statement), which Liam mentioned previously:

> After refactoring, symbolic arithmetic must remain equivalent to the original, albeit without having to respect limits such as where NaN or INF may be produced as a byproduct of the order of operations.

My interpretation of this is that the floating point rules do not apply regardless of the global flags. These rules can be broken when REFACTORING_ALLOWED is present.

IMO, the per-instruction precise flag is a non-issue. It is easy enough to parse the DXBC to look for per-instruction precise flags and enable float controls globally if one is found since there is no SPIR-V per instruction flag.

If there were a disconnect between the DX and VK specs about NaN/Inf/SignedZero handling, then I can make the performance tweak change vendor specific, but I do not think there is a conflict.

doitsujin:
I'd argue that what the D3D spec says is somewhat academic when the shader I posted above proves that NaNs are not ignored in practice on native drivers regardless of any flags, and a very large number of shipping games rely on this behaviour.

As for the Vulkan side of things, the spec says:
> By default, the implementation may perform optimizations on half, single, or double-precision floating-point instructions that 
> ignore sign of a zero, **or assume that arguments and results are not NaNs or infinities**. If the entry point is declared with the 
> SignedZeroInfNanPreserve execution mode, then NaNs, infinities, and the sign of zero must not be ignored.

liam-middlebrook:
@doitsujin you stated above:
> We ran into an issue with Mesa where drivers would not preserve NaNs on texture fetches

As you note later it's legal behavior for Vulkan to do that, but it just works in D3D11 regardless of shader settings.

I tried to find this issue on the DXVK or Mesa issue trackers, as well as chat archives from Discord, and wasn't able to find anything that quite matches what's been described above. If you could point me towards the related issues here (or what timestamps to look for in our archives) I'd be interested in reading more on the discussion that led to this change.


From what was stated on this PR, it sounds like this change in behavior was added to WAR a bad interaction with Mesa where NaNs were not preserved on texture fetch operations. [Appendix A of the Vulkan Specification states the following about SignedZeroInfNanPreserve](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#spirvenv-precision-operation):

> By default, the implementation may perform optimizations on half, single, or
> double-precision floating-point instructions that ignore sign of a zero, or
> assume that arguments and results are not NaNs or infinities. If the entry
> point is declared with the SignedZeroInfNanPreserve execution mode, then NaNs,
> infinities, and the sign of zero must not be ignored.
> 
>     The following core SPIR-V instructions must respect the
>     SignedZeroInfNanPreserve execution mode: OpPhi, OpSelect, OpReturnValue,
>     OpVectorExtractDynamic, OpVectorInsertDynamic, OpVectorShuffle,
>     OpCompositeConstruct, OpCompositeExtract, OpCompositeInsert, OpCopyObject,
>     OpTranspose, OpFConvert, OpFNegate, OpFAdd, OpFSub, OpFMul, OpStore. This
>     execution mode must also be respected by OpLoad except for loads from the
>     Input storage class in the fragment shader stage with the floating-point
>     result type. Other SPIR-V instructions may also respect the
>     SignedZeroInfNanPreserve execution mode.

From the above, the operative statement here is:
> Other SPIR-V instructions may also respect the SignedZeroInfNanPreserve

Meaning that there is no expectation for image/texture operations to be affected by SignedZeroInfNanPreserve, and whether or not they are is left up to the implementation to determine.


Would it make sense to verify if this issue is still hit on latest Mesa drivers, and conditionality enable SignedZeroInfNanPreserve depending on whether or not a known affected Mesa driver is in-use?

doitsujin:
I still don't see which part of the spec guarantees that NaNs are preserved without `SignedZeroInfNanPreserve`? The problem was never that the texture load itself was doing the wrong thing but that the instructions using the loaded value *assumed* that the load was NaN.

liam-middlebrook:
> I still don't see which part of the spec guarantees that NaNs are preserved without SignedZeroInfNanPreserve?
Conversely, there is no part of the specification which guarantees they will not be preserved.

From the spec text which I quoted above, the normative **may** is used indicating:
> By default, the implementation **may** perform optimizations on half, single, or double-precision floating-point instructions that ignore sign of a zero, or assume that arguments and results are not NaNs or infinities.

It's behavior that is up to the implementation, so I understand why you would like DXVK to always set it; it's the safe thing to do on your part. However, on implementations where these more ambitious optimizations are not preformed this can result in degraded performance.

How would you feel about changing the logic here so that the more strict floating-point rules are only applied for Mesa? Or if you know that after a certain mesa version these ambitious optimizations are no longer being performed, the check could have version bounds added to it.

Venemo:
@liam-middlebrook It was some time ago, so i don't remember the exact details, but:

> > By default, the implementation **may** perform optimizations on half, single, or double-precision floating-point instructions that ignore sign of a zero, or assume that arguments and results are not NaNs or infinities.
> 
> It's behavior that is up to the implementation, so I understand why you would like DXVK to always set it

As far as I remember (please correct me if I'm wrong) this was added to DXVK because otherwise we can't guarantee that the driver won't break games by optimizing with the assumptions you quote above. There were some games (I don't remember which exactly) which broke because of this on RADV/ACO.

The worst part of it is that they can interact with other optimizations in unexpected ways.

I remember one case when we bisected a rendering glitch to a loop unrolling modification, and nobody understood what was happening until we realized that nothing was wrong with loop unrolling, just that the unrolled loop was now subject to one of these "more ambitious" optimizations.

> However, on implementations where these more ambitious optimizations are not preformed this can result in degraded performance.

If you don't do optimizations like that, then why does this flag affect your performance at all?  
I think if you don't have these "more ambitious optimizations" then you can just emit the exact same code with and without the different float controls, no?

> How would you feel about changing the logic here so that the more strict floating-point rules are only applied for Mesa? Or if you know that after a certain mesa version these ambitious optimizations are no longer being performed, the check could have version bounds added to it.

Maybe @cwabbott0 remembers the discussion that lead to this better? I think it's still a good idea to use `SignedZeroInfNanPreserve` with mesa. At least just quickly checking the code, these flags are used by the NIR algebraic optimizations, and ACO instruction selection and optimizer.

cwabbott0:
> It's behavior that is up to the implementation, so I understand why you would like DXVK to always set it; it's the safe thing to do on your part. However, on implementations where these more ambitious optimizations are not preformed this can result in degraded performance.

It's not "the safe thing to do," it's required and has always been required by D3D since at least D3D10, as already explained up-thread. Your Windows D3D driver also does not perform those "more ambitious optimizations," because they are explicitly illegal and have always been, and if DXVK on top of your Vulkan stack produces worse results than your D3D driver then that's your bug and we can't fix it. This D3D11 quote:

> After refactoring, symbolic arithmetic must remain equivalent to the original,
albeit without having to respect limits such as where NaN or INF may be
produced as a byproduct of the order of operations.

does *not* mean that all NaN-eliminating optimizations are allowed, it only applies to reassociating floating-point adds and multiplies. There are tons of other optimizations that can only be done if we assume the input isn't NaN which aren't allowed in D3D.

The precise flag in D3D seems to be exactly the same as `NoContraction` in SPIR-V. Since we already have `NoContraction`, I think the intent of the Vulkan spec is that `SignedZeroInfNanPreserve` doesn't disallow refactoring. In other words, `SignedZeroInfNanPreserve` should only block those other optimizations, corresponding to "base D3D rules" and `NoContraction` should be the same as the precise flag in D3D11. This is what we do in mesa. It's also the only way to correctly and optimally handle D3D-emulated shaders without yet another Vulkan extension. I think someone should file a bug against the Vulkan spec to incorporate a note similar to the above D3D11 quote about `NoContraction` in the "Precision and Operation of SPIR-V Instructions" section and nvidia should fix their driver.

cwabbott0:
For those in Khronos, I've filed an internal bug to track this: https://gitlab.khronos.org/vulkan/vulkan/-/issues/2795. I guess you could disable it on nvidia for a short-term workaround, but I think it's on nvidia to change their driver or propose an extension, because just disabling `SignedZeroInfNanPreserve` is not the correct solution.

cwabbott0:
There will be a future spec update to clarify that reassociation and fusing/un-fusing is allowed even with `SignedZeroInfNanPreserve`, so I believe this can now be closed. Nvidia needs to update their driver to be more permissive around `SignedZeroInfNanPreserve` if they haven't already, in order to get the performance gains claimed by this MR.

PR #2096, opened 2021-06-10
Frame rate limiter (v2)

doitsujin:
Turns out that stalling the queue thread works significantly better than stalling the app thread.

Supersedes #2091.

### DIFF
--- a/src/util/util_fps_limiter.cpp
+++ b/src/util/util_fps_limiter.cpp
@@ -1,0 +1,159 @@
+#include <thread>
+
+#include "thread.h"
+#include "util_fps_limiter.h"
+#include "util_string.h"
+
+#include "./log/log.h"
+
+namespace dxvk {
+  
+  FpsLimiter::FpsLimiter() {
+
+  }
+
+
+  FpsLimiter::~FpsLimiter() {
+
+  }
+
+
+  void FpsLimiter::setTargetFrameRate(double frameRate) {
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    m_targetInterval = frameRate > 0.0
+      ? NtTimerDuration(int64_t(double(NtTimerDuration::period::den) / frameRate))
+      : NtTimerDuration::zero();
+
+    if (isEnabled() && !m_initialized)
+      initialize();
+  }
+
+
+  void FpsLimiter::setDisplayRefreshRate(double refreshRate) {
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    m_refreshInterval = refreshRate > 0.0
+      ? NtTimerDuration(int64_t(double(NtTimerDuration::period::den) / refreshRate))
+      : NtTimerDuration::zero();
+  }
+
+
+  void FpsLimiter::delay(bool vsyncEnabled) {
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    if (!isEnabled())
+      return;
+
+    // If the swap chain is known to have vsync enabled and the
+    // refresh rate is similar to the target frame rate, disable
+    // the limiter so it does not screw up frame times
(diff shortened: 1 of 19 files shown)


aufkrawall:
There seems to be a difference in vsync detection between D3D9 and 11 (VRR off yet): In D3D9 games like Mirror's Edge, I have to go down to 82fps to make the fps limiter still enable with 85Hz vsync. When I use a value of 83fps, it is disabled and fps are capped by vsync at 85fps instead. In D3D11 Witcher 3, I can enable vsync and the limiter still caps fps at 84 when configured to.

Imho it'd be nice if the vsync detection doesn't turn of the limiter as long as the defined value for the limiter is at least 1fps lower than the display's refresh rate to allow a bit of flexibility for VRR usage.
Blurbusters tested input-to-pixel input lag of vsync vs. vsync + fps limiter and it showed how atrocious the input lag is that is caused by vsync capping fps, even with VRR enabled (with the driver's CPU prerender option set to 1, I think both values would be higher/lower by the same amount, not changing the relative distance):
https://blurbusters.com/gsync/gsync101-input-lag-tests-and-settings/4
https://blurbusters.com/gsync/gsync101-input-lag-tests-and-settings/7

aufkrawall:
Works like a charm with VRR. The display is overclocked to 64.8Hz and Heroes of the Storm and Days Gone run really well with vsync + an fps limit of 64fps.
Of course stupid Radeon software in shape of Radeon Overlay had to interfere by causing weird stutter, which could be solved by turning it off. So, once again DXVK can be a savior for AMD Windows users when the driver features let you down. :)

doitsujin:
> There seems to be a difference in vsync detection between D3D9 and 11

There is not; since this is implemented in the backend, the vsync detection just checks the present mode.

Note that using the environment variable (`DXVK_FRAME_RATE`) bypasses this and always enables the limiter, while using the config file does not (because we do enable this by default for some games).

src/util/util_fps_limiter.cpp
```diff
@@ -0,0 +1,159 @@
...
+    std::lock_guard<std::mutex> lock(m_mutex);
+
+    m_refreshInterval = refreshRate > 0.0
+      ? NtTimerDuration(int64_t(double(NtTimerDuration::period::den) / refreshRate))
+      : NtTimerDuration::zero();
+  }
+
+
+  void FpsLimiter::delay(bool vsyncEnabled) {
+    std::lock_guard<std::mutex> lock(m_mutex);
```
Joshua-Ashton, line 55: Why does this need a mutex?

setTargetFrameRate should only happen at startup, right?

Joshua-Ashton: I guess also on swapchain mode change but we aren't presenting then.

doitsujin: We can have a present queued up or executing on the queue thread while the app thread performs a mode change and calls `setDisplayRefreshRate`.

This does mean that not every frame during mode chages is necessarily perfect, but I'd rather do this than make presentation itself more complicated than it already is.

Joshua-Ashton:
I guess i am not super happy about having to lock a third mutex for presenting even in the case where this is disabled.

It probably doesn't matter (especially if it ends up being an SRW mutex) but... it does bug me a bit.

Anyway, other than that, lgtm.

aufkrawall:
In Watch Dogs Legion, the limiter still doesn't work correctly on Windows. A limit of 50 runs with 50fps on Linux/Wine, whereas on Windows it's something in the 30ies (and RTSS graph looks fat). It's really weird, as it definitely works correctly in other games.

doitsujin:
Not sure if I can be arsed to debug that, there's nothing wrong on our end.

PR #2102, opened 2021-06-15
[dxso] Remove old RADV/LLVM workaround for early discard

hakzsam:


Joshua-Ashton:
We can probably remove this for NV as well, and the whole old code path and just use demote to helper invocation everywhere.

hakzsam:
Phillip only removed it for RADV, so I only applied the same patch for DXSO.

Joshua-Ashton:
Oh right, I see. Sorry I am half-functional today.

Can probably do what I said anyway. I'll look into that later...

PR #2109, opened 2021-06-18
[dxvk] Adjust a comment about useEarlyDiscard on RADV

hakzsam:
Trivial.


doitsujin:
I'll remove the option entirely since it's not relevant on drivers that support `demote`.

PR #2111, opened 2021-06-19
Don't use staging buffer for first uploads

K0bin:
This change will make DXVK only use a staging buffer for D3D9 resources that have been locked twice before. It should improve memory usage in games that only copy data to their textures once. (which should be the vast majority of games)

It also contains a tiny optimization for buffer locking because once we hit the switch to use staging buffers, we can omit the CS thread sync point. (Supersedes #2039)

### DIFF
--- a/src/d3d9/d3d9_common_buffer.h
+++ b/src/d3d9/d3d9_common_buffer.h
@@ -173,6 +173,15 @@ namespace dxvk {
      */
     bool NeedsUpload() { return m_desc.Pool != D3DPOOL_DEFAULT && !m_dirtyRange.IsDegenerate(); }
 
+    bool DoesStagingBufferUploads() const { return m_uploadUsingStaging; }
+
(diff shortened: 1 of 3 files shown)


dnblank123:
Character rendering issue on Dragon Nest when using `d3d9.evictManagedOnUnlock = True`
https://imgur.com/a/kKJIHf3

K0bin:
@dnblank123 That only happens with this PR, right?

dnblank123:
> @dnblank123 That only happens with this PR, right?

Yes, I already tried #2039 and new release (1.9) not issue so far.

K0bin:
Any chance you could make an apitrace?

dnblank123:
@K0bin https://drive.google.com/file/d/1vHR24meYzsuoeOQEEYhOLI4OqDLIk9EH/view?usp=sharing

K0bin:
Thanks a lot.

Issue #2110, opened 2021-06-18
[win10] [amd] [d3d9] [dxvk 1.9] Mirror's Edge regression


Ryusennin:
Hello. I'm using Windows 10.

With the latest dxvk 1.9 release, Mirror's Edge display is corrupted. The game runs fine with dxvk 1.8.1.

I tried to run an apitrace, but Origin launcher crashed the game with the following message:
"Invalid license. Reason code = Missing DLL [d3d9.dll]. Error: 0x7E."

### Software information
Mirror's Edge (max settings, except PhysX turned off)

### System information
- GPU: Radeon Vega
- Driver: 21.5.2
- Wine version: N/A
- Windows version: 10-21H1
- DXVK version: 1.9

### Log files
[MirrorsEdge_d3d9_dxvk1.8_ok.log](https://github.com/doitsujin/dxvk/files/6679522/MirrorsEdge_d3d9_dxvk1.8_ok.log)
[MirrorsEdge_d3d9_dxvk1.9_bad.log](https://github.com/doitsujin/dxvk/files/6679523/MirrorsEdge_d3d9_dxvk1.9_bad.log)

dxvk 1.8.1
![unknown_2021 06 16-13 55](https://user-images.githubusercontent.com/9519544/122620626-baf01c00-d093-11eb-8c49-187ef132e8c5.png)

dxvk 1.9
![unknown_2021 06 16-13 50](https://user-images.githubusercontent.com/9519544/122620649-c8a5a180-d093-11eb-8570-35cb764d6ca0.png)

K0bin:
![Mirror's Edge Screenshot 2021 06 19 - 03 11 13 75](https://user-images.githubusercontent.com/1131720/122638892-21f2ec80-d0f7-11eb-80f6-8cf4be97c6e2.png)

Most likely a driver bug.

K0bin:
We found and fixed a minor bug. Please retest with the latest master.

Ryusennin:
Indeed the issue remains with the latest build.

I remembered I also have a Steam version, maybe I'll have better luck getting an apitrace.

Ryusennin:
The last version that worked with the game is this build from April 30th (commit 4a4b966):
https://github.com/doitsujin/dxvk/actions/runs/800364047

The issue started with the following builds (commit f9beaa3):
https://github.com/doitsujin/dxvk/actions/runs/801820890

K0bin:
Does it work if you revert that commit?

[...]

PR #2114, opened 2021-06-19
[d3d9] Make sure clear extent does not exceed rt size

K0bin:
Might help with #2110

Mirrors Edge calls a clear with the extents 1600x1200 on a RT that is only 1600x900.


Joshua-Ashton:
Need to account for offset too

PR #2126, opened 2021-06-27
Fix for EvictManagedOnUnlock & SWTOR performance

### LINKED ISSUE #2125: Regression: Black textures in Borderlands 2/Pre-Sequel with dxvk 1.9 and d3d9.evictManagedOnUnlock
There was a regression between dxvk 1.8.1 and 1.9 which results in black textures (or at least the appearance of black textures) when playing Borderlands 2 or TPS with `evictManagedOnUnlock` turned on, and you really need to turn it on as you'll hit #1318 otherwise.

I did a git bisect and identified change b9512a40630c1e9030aa7958ce2456664bc3b059 as the cause. I don't pretend to really understand any of this, but the description is `Only copy dirty parts of managed/sysmem resources` and the config option relates to managed resources, so perhaps the issue is something along the lines of evicted resources no longer being tracked for dirtiness so it never bothers copying them. Or something like that.

I'll also note that this change looks like it's part of a series and the actual tracking of dirtiness of managed resources shows up in earlier changes - this is the just the change where there is visible impact.

### Software information
- Borderlands 2 or Borderlands TPS. Any in-game graphical settings.
- d3d9.evictManagedOnUnlock turned on

### System information
- GPU: nvidia RTX 3090 (but also repros with my Intel 630 iGPU)
- Driver: 465.31
- Wine version: Proton 6.3-5
- DXVK version: Any build with b9512a40630c1e9030aa7958ce2456664bc3b059 present

### Log files
[Borderlands2_d3d9.log](https://github.com/doitsujin/dxvk/files/6720543/Borderlands2_d3d9.log)

### API Trace
https://www.dropbox.com/s/vtz699ok45b8202/borderlands2.trace.7z?dl=0

### LINKED ISSUE #2130: Star Wars - The Old Republic - Regression
This is run via Steam. Star Wars - The Old Republic retail client.
The issue is severe stuttering especially when turning around and/or running.

Upon updating Proton to Proton-6.3-5, dxvk was also upgraded from DXVK-1.8.1 -> DXVK-1.9. There was no issues before DXVK-1.9, and replacing the dxvk dll's in the Proton-6.3-5 folders with those of DXVK-1.8.1 brings performance back to what it was.

I did some regression testing for this, and found that after  https://github.com/doitsujin/dxvk/commit/4261ff6ec1aa7efdc6d7bf6a5b9434c2eb6cc184 this issue started happening. The issue does improve a bit with DXVK-1.9+GIT, but not enough for it to be playable. 
Mentioned this here: https://github.com/doitsujin/dxvk/pull/1987
https://github.com/ValveSoftware/Proton/issues/4326#issuecomment-869030604

### Software information
Star Wars - The Old Republic - retail client
Steam - Proton-6.3-5 (Also tested with Proton-GE-6.10-1)

### System information
- GPU: nVidia RTX2070
- Driver: 470.42.01 (Also happens with 465.31)
- Wine version: Proton-6.3-5
- DXVK version: DXVK-1.9

### Apitrace file(s)
https://filebin.net/aa7561eozmq5zzb0
This was made in Windows 10 without DXVK. I was not able to do a apitrace with latest or 10, but used apitrace-7. It did playback fine.
Hopefully it is usable.

### Log files
[swtor_d3d9.zip](https://github.com/doitsujin/dxvk/files/6729202/swtor_d3d9.zip)
[swtor_dxgi.zip](https://github.com/doitsujin/dxvk/files/6729203/swtor_dxgi.zip)

K0bin:
Fixes #2125

Borderlands 2 locks textures like this:
```
Lock(Mip 0)
Lock(Mip 1)
Unlock(Mip 0)
Unlock(Mip 1)
```
which leads to us clearing the dirty box too soon and not copying the smaller mip maps over.
______
Fixes #2130

SWTOR locks a 1024x1024 D3DPOOL_DEFAULT, D3DFORMAT_A8, USAGE_DYNAMIC texture very often. (Probably cached font glyphs?). It properly sets the dirty rect whenever it makes a change.

On current master we never clear the dirty so it stays at the initial 1024x1024 rect. That means we keep repacking (from 4 aligned pitch managed buffer into tightly packed staging buffer) the entire image a bunch of times per frame. That's really slow.
If we just clear the dirty box like we do for managed textures, all we have to repack and copy are <30x30 parts of the texture.

### DIFF
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -308,6 +308,8 @@ namespace dxvk {
 
     bool GetLocked(UINT Subresource) const { return m_locked.get(Subresource); }
 
+    bool IsAnySubresourceLocked() const { return m_locked.any(); }
+
     void SetWrittenByGPU(UINT Subresource, bool value) { m_wasWrittenByGPU.set(Subresource, value); }
 
     bool WasWrittenByGPU(UINT Subresource) const { return m_wasWrittenByGPU.get(Subresource); }
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4166,10 +4166,12 @@ namespace dxvk {
           cPackedFormat = packedFormat
         ] (DxvkContext* ctx) {
           if (cSubresources.aspectMask != (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT)) {
-            ctx->copyImageToBuffer(cImageBuffer, 0, 0, 0,
+            ctx->copyImageToBuffer(cImageBuffer, 0, 4, 0,
               cImage, cSubresources, VkOffset3D { 0, 0, 0 },
               cLevelExtent);
           } else {
+            // Copying DS to a packed buffer is only supported for D24S8 and D32S8
+            // right now so the 4 byte row alignment is guaranteed by the format size
             ctx->copyDepthStencilImageToPackedBuffer(
               cImageBuffer, 0,
               VkOffset2D { 0, 0 },
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4223,7 +4223,7 @@ namespace dxvk {
       }
     }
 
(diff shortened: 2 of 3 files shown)


philipl:
Did a quick test with the build artifacts and it looks correct. Thanks!

PR #2127, opened 2021-06-27
[util] Set dxgi.syncInterval to 0 for Earth Defense Force 5

junglerobba:
The game stutters heavily with vsync enabled, even on windows, and
offers no option to disable it.

Tested on Wayland (sway) and xorg (Plasma) with both 60 and 144 Hz, and got the same behavior in all cases.
Tbf, I don't know if this is really "just" normal stutter or whatever the game is doing, since MangoHud does not even show the framerate correctly.
Either way, setting dxgi.syncInterval to 0 makes the game run fine again.

edf5
edf2

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -436,6 +436,11 @@ namespace dxvk {
     { R"(\\FarCry\.exe$)", {{
       { "d3d9.customVendorId",              "10de" },
     }} },
+    /* Earth Defense Force 5 */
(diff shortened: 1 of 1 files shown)


aufkrawall:
Did you try `dxgi.numBackBuffers = 3` with `dxgi.syncInterval = 1` and RADV?

doitsujin:
This isn't really a solution, if there's a performance issue with the game then we should try to find the root cause.

PR #2129, opened 2021-06-28
Introduce custom sync primitives

doitsujin:
This mostly aims to reduce potential inefficiencies with the winpthread implementation.
- Uses SRW locks for `mutex`.
- Uses Critical Sections for `recursive_mutex`.
- Uses Win32 condition variables for `condition_variable`.

All of these use `futex` on wine.

Note that this does *not* remove the winpthreads dependency, so MinGW builds with threading support are still required.


Joshua-Ashton (approved):
LGTM

PR #2134, opened 2021-06-30
[d3d9] Ensure that the bound UP vertex buffer region is large enough

doitsujin:
If the buffer size is less than (vertexCount * stride), the last vertex may be considered out-of-bounds, even if all attribute data is included in the bound region.


Joshua-Ashton:
May make more sense to just enable `robustBufferAccess2` for d3d9.

doitsujin:
Doesn't fix the problem on amdvlk though.

PR #2138, opened 2021-07-02
Initial winelib support (for hangover)

cwabbott0:
For bringing up DXVK on hangover, we want to compile it in the host for performance reasons, and that entails compiling it as a winelib since there is no usable aarch64 mingw toolchain. This MR adds back support for winelib. However using it requires using a winegcc built with some patches I just sent to wine-devel https://www.winehq.org/pipermail/wine-devel/2021-July/190100.html. If you also don't like applying patches from mailing lists, they're available at https://github.com/cwabbott0/wine/tree/winegcc-meson.

This just gets things building on x86 as a first step. Later I will fix compilation with aarch64, and we also need to somehow deal with https://github.com/AndreRH/hangover/issues/28#issuecomment-749122396.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -66,8 +66,10 @@ dxvk_extradep = [ ]
 if dxvk_is_msvc
   wrc = find_program('rc')
 else
-  add_global_link_arguments('-static', '-static-libgcc', language: 'c')
-  add_global_link_arguments('-static', '-static-libgcc', '-static-libstdc++', language: 'cpp')
+  if not dxvk_is_winelib
+    add_global_link_arguments('-static', '-static-libgcc', language: 'c')
+    add_global_link_arguments('-static', '-static-libgcc', '-static-libstdc++', language: 'cpp')
+  endif
 
   wrc = find_program('windres')
 endif
--- a/meson.build
+++ b/meson.build
@@ -120,6 +120,12 @@ endif
 
 def_spec_ext = '.def'
 
+if dxvk_is_winelib
+  library_ext = 'dll.so'
+else
+  library_ext = []
+endif
+
 glsl_compiler = find_program('glslangValidator')
 glsl_args = [ '-V', '--vn', '@BASENAME@', '@INPUT@', '-o', '@OUTPUT@' ]
 if run_command(glsl_compiler, [ '--quiet', '--version' ]).returncode() == 0
--- a/meson.build
+++ b/meson.build
@@ -142,7 +142,7 @@ if dxvk_is_msvc
 else
   wrc_generator = generator(wrc,
   output    : [ '@BASENAME@' + res_ext ],
-  arguments : [ '-i', '@INPUT@', '-o', '@OUTPUT@' ])
+  arguments : meson.get_cross_property('windres_args', []) + [ '-i', '@INPUT@', '-o', '@OUTPUT@' ])
 endif
 
 dxvk_version = vcs_tag(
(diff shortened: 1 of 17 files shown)


doitsujin:
> since there is no usable aarch64 mingw toolchain

does this mean this is going to be running on non-x86 hardware? I don't think DXVK would even compile on that right now, and even when sorting that out I'm almost certain that we'll have to fix a few hundred memory visibility bugs and similar things first, because there has been no way to test this so far.

Anyway, not particularly thrilled about having to add this maintenance burden back, but I guess we don't have much of a choice.

Joshua-Ashton:
Wouldn't it make more sense to just get MinGW working?

You're going to **need** that going forward anyway so it makes more sense than bringing that maintenance burden here imo.

cwabbott0:
> Wouldn't it make more sense to just get MinGW working?
> 
> You're going to **need** that going forward anyway so it makes more sense than bringing that maintenance burden here imo.

I would absolutely love to have a working MinGW toolchain, after doing this I now know exactly how much the current setup sucks, but I just have no idea how one would go about achieving that and I don't think that's something I could reasonably take on myself. Also there are apparently thorny ABI issues (Microsoft reserved a register for the TEB that linux uses freely iiuc) that need to be solved somehow.

cwabbott0:
> does this mean this is going to be running on non-x86 hardware? I don't think DXVK would even compile on that right now, and even when sorting that out I'm almost certain that we'll have to fix a few hundred memory visibility bugs and similar things first, because there has been no way to test this so far.

Yes, the plan is to get it to replace wined3d with dxvk in hangover. The stack would use hangover for DLL thunking, with DXVK + turnip for D3D emulation, and eventually FEX instead of qemu for x86 emulation. On the turnip side we've implemented pretty much everything DXVK could want. I have a hacky setup with an older version of DXVK where it actually works for basic things, at least I got to the point of being able to replay a trace of a real game with a Windows version of apitrace running on top of this stack (except still with qemu). This is my attempt to start to clean it up and make it a coherent patch series that doesn't perturb the normal case. It's definitely still very experimental ATM, but at least it's a start.

doitsujin:
Pulled in some of the commits that don't add any winelib-specific stuff since they are useful, but I guess the question is whether want winelib support upstream or whether the whole thing could be maintained in an external fork.

If a fork is an option for you I'd also pull in 1053d5c975ea85cf18b481c977c21963eec20aca without the `__WINE__` bit to make life a bit easier.

YihaoPeng:
I wish someone can help me to port dxvk to aarch64 native dll. Because there is indeed a native aarch64 Windows game - World of Warcraft: Shadowlands. (https://us.forums.blizzard.com/en/wow/t/windows-10-arm64-support-now-available/961628)

And I have a PC with an 4-core arm64 CPU and a regular PCI-e x16 slot, so I can install some Vulkan compatible NVIDIA or AMD graphics cards. The AMD Radeon HD 8570 graphics card it is now installed seems to be Vulkan compatible.

image

So if there is an aarch64 native DXVK, I can test whether the game can be launched on my computer.

It does not work in wined3d now (with wine 6.13 aarch64 build):

image

I know that we may have a lot of work elsewhere in or outside of DXVK, but adding aarch64 support to DXVK builds will be a good start.

driver1998:
> since there is no usable aarch64 mingw toolchain
Maybe take a look at https://github.com/mstorsjo/llvm-mingw? Haven't tried DXVK with it but a lot of sizable projects (ffmpeg, qt, vlc, etc) builds just fine targeting aarch64.

Joshua-Ashton:
People are using ARM64EC for this now. https://github.com/doitsujin/dxvk/pull/3900

mirh:
Are there prons or cons in using ARM64EC compared to winelib?
Aside then of the obvious (and certainly important) win of being able to ship this directly in native windows.

PR #2156, opened 2021-07-13
D3D9: Initial data pitch

K0bin:


src/dxvk/dxvk_util.cpp
```diff
@@ -71,6 +71,41 @@ namespace dxvk::util {
       }
     }
   }
+
+
+  void packImageData(
```
Joshua-Ashton, line 76: You should make the above packImageData just call this with the info it gets.

Joshua-Ashton (changes requested):

src/dxvk/dxvk_util.cpp
doitsujin: If anything, I'd want everything to use the path below (the one with 8 parameters), if that works. Was too lazy to deal with it yet though.

Joshua-Ashton: I think having some helper that just calculates and passes it in with less args is fine. Calculating these values everywhere would kinda suck.

K0bin: I agree with Josh here.

doitsujin: I'm not really sure what's going on with the d3d9 side but originally these functions were meant to just tightly pack image data so we could use it with `vkCmdCopyBufferToImage`. Looking at what this does in more detail, it's a completely different use case.

K0bin: We're kinda doing the opposite here. "unpacking" tightly packed data. Would you prefer a separate function for that?

doitsujin: My problem with this whole thing is that we still have *two* functions doing more or less the thing with different capabilities. I'd rather just have one.

K0bin: So basically extend the multi-plane variant and use that instead?

src/dxvk/dxvk_util.cpp
```diff
@@ -142,6 +142,90 @@ namespace dxvk::util {
...
+          extent.height /= plane->blockSize.height;
+          elementSize = plane->elementSize;
+        }
+
+        auto blockCount = computeBlockCount(extent, formatInfo->blockSize);
+
+        VkDeviceSize bytesTotal = blockCount.depth * std::min(dstRowPitch, srcRowPitch);
+
+        const bool directCopy = ((dstRowPitch   == srcRowPitch  ) || (blockCount.height == 1))
+                             && ((dstSlicePitch == srcSlicePitch) || (blockCount.depth  == 1));
```
doitsujin, line 178: this is wrong, need to check if both src and dst pitches are equal to the byte count.

src/d3d9/d3d9_initializer.cpp
```diff
@@ -131,18 +132,53 @@ namespace dxvk {
...
+            pitch,
+            pitch * blockCount.height,
+            alignedPitch,
+            alignedPitch * blockCount.height,
+            type,
+            mipExtent,
+            resourceType == D3DRTYPE_CUBETEXTURE ? 6 : 1,
+            formatInfo,
+            VK_IMAGE_ASPECT_COLOR_BIT
+          );
```
doitsujin, line 176: what's so important about `);` that it needs its own line?

K0bin: Nothing, I just prefer it like that because it's horizontally aligned with the function call. I'll change that.

artemyto: Kotlin style

K0bin: I like Kotlin.
¯\\\_(ツ)\_/¯

src/dxvk/dxvk_util.cpp
```diff
@@ -142,6 +142,99 @@ namespace dxvk::util {
...
+                dstData + j * dstRowPitch,
+                srcData + j * srcRowPitch,
+                bytesPerRow);
+
+              if (dstRowPitch > bytesPerRow) {
+                std::memset(dstData + j * dstRowPitch + bytesPerRow, 0, dstRowPitch - bytesPerRow);
+              }
+            }
+            if (dstSlicePitch > bytesPerSlice) {
+              std::memset(dstData + bytesPerSlice, 0, dstSlicePitch - bytesPerSlice);
```
doitsujin, line 213: why are we memsetting anything here?

K0bin: Alright, gonna get rid of that.

src/dxvk/dxvk_util.cpp
```diff
@@ -142,6 +142,99 @@ namespace dxvk::util {
...
+                srcData += srcRowPitch;
+                dstData += dstRowPitch;
+                break;
+              case VK_IMAGE_TYPE_2D:
+                srcData += blockCount.height * srcRowPitch;
+                dstData += blockCount.height * dstRowPitch;
+                break;
+              case VK_IMAGE_TYPE_3D:
+                srcData += srcSlicePitch;
+                dstData += dstRowPitch;
```
doitsujin, line 227: dstSlicePitch

Joshua-Ashton:
Not technically correct given what we know about this host importing the pointer but this is a good stop gap that fixes some games :)

src/d3d9/d3d9_initializer.cpp
```diff
@@ -131,18 +131,41 @@ namespace dxvk {
...
+
+        if (pInitialData != nullptr) {
+          VkExtent3D mipExtent = pTexture->GetExtentMip(m);
+          const DxvkFormatInfo* formatInfo = imageFormatInfo(pTexture->GetFormatMapping().FormatColor);
+          VkExtent3D blockCount = util::computeBlockCount(mipExtent, formatInfo->blockSize);
+          uint32_t pitch = blockCount.width * formatInfo->elementSize;
+          uint32_t alignedPitch = align(pitch, 4);
+
+          D3DRESOURCETYPE resourceType = pTexture->GetType();
+          VkImageType type = resourceType == D3DRTYPE_VOLUME || resourceType == D3DRTYPE_VOLUMETEXTURE
```
Joshua-Ashton, line 148: pTexture->GetImage()->info().type

K0bin:
Yup, I'd go with this for 1.9 and do a proper solution in 1.9.1.

Alternatively we can also drop it, given that csgo and Desolation are basically the only games that actually use this as far as I know.

src/d3d9/d3d9_initializer.cpp
```diff
@@ -131,18 +131,41 @@ namespace dxvk {
...
+          util::packImageData(
+            mapSlice.mapPtr,
+            pInitialData,
+            pitch,
+            pitch * blockCount.height,
+            alignedPitch,
+            alignedPitch * blockCount.height,
+            type,
+            mipExtent,
+            resourceType == D3DRTYPE_CUBETEXTURE ? 6 : 1,
```
Joshua-Ashton, line 161: `pTexture->Desc()->ArraySize`

Joshua-Ashton (changes requested):

src/d3d9/d3d9_initializer.cpp
K0bin: That doesn't work because this is only called for POOL_SYSTEMMEM textures which don't have an image.

PR #2180, opened 2021-07-24
Misc. set of D3D9 optimizations and cleanup (part 1)

Joshua-Ashton:
A set of small optimizations and code cleanup for d3d9.


K0bin (approved):
Looks fine. You should squash a bunch of those commits though.

src/d3d9/d3d9_subresource.h
```diff
@@ -56,23 +58,23 @@ namespace dxvk {
       m_texture->PreLoadSubresource(GetSubresource());
     }
 
-    D3D9CommonTexture* GetCommonTexture() {
```
doitsujin, line 59: why `inline`? All of these functions are implicitly `inline` already.

PR #2181, opened 2021-07-24
[build] Enable strict dwarf2 debug symbols for Wine back traces

DadSchoorse:
Sometimes simple backtraces are useful and wine doesn't understand newer dwarf.


doitsujin:
any reason not to do this by default for debug builds?

DadSchoorse:
Presumably newer dwarf provides better info and e.g. gdb could work with it, but I have to admit I'm not an expert on debugging symbols.

doitsujin:
I think we could just pass these if we know that `debug` is true (i.e. non-release build) and we're building for a win32 target (i.e. not dxvk-native). This really shouldn't be an option we need to enable.

doitsujin (approved):

PR #2213, opened 2021-08-06
[d3d9] Mark texture dirty on SetLod

DadSchoorse:
Previously the image view was not updated. Fixes wine tests.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5817,8 +5817,13 @@ namespace dxvk {
     m_dirtyTextures = 0;
   }
 
-  void D3D9DeviceEx::MarkSamplersDirty() {
-    m_dirtySamplerStates = 0x001fffff; // 21 bits.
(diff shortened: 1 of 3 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -5817,8 +5817,13 @@ namespace dxvk {
     m_dirtyTextures = 0;
   }
 
-  void D3D9DeviceEx::MarkSamplersDirty() {
-    m_dirtySamplerStates = 0x001fffff; // 21 bits.
+  void D3D9DeviceEx::MarkTextureBindingDirty(IDirect3DBaseTexture9* texture) {
+    D3D9DeviceLock lock = LockDevice();
+
+    for (uint32_t i = 0; i < std::size(m_state.textures); i++) {
```
Joshua-Ashton, line 5823: Just use m_state.textures.size()

Joshua-Ashton (changes requested):

src/d3d9/d3d9_device.cpp
DadSchoorse: Done.

PR #2214, opened 2021-08-07
[d3d9] Only look at the last bit for D3DSAMP_SRGBTEXTURE

DadSchoorse:
Fixes a wine test that was added for https://bugs.winehq.org/show_bug.cgi?id=37705.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5761,7 +5761,7 @@ namespace dxvk {
       DxsoBindingType::DepthImage, uint32_t(shaderSampler.second));
 
     const bool srgb =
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Does the wine test also test values like 2 and 3?

DadSchoorse:
Yes, they test 1, 2, 3, 0x7e41882a, 100, 0 :frog:.

DadSchoorse:
fwiw
```
        /* AMD uses the LSB of the D3DSAMP_SRGBTEXTURE value.
         * NVIDIA ignores any values other than 0 and 1, leaving the previous
         * D3DSAMP_SRGBTEXTURE state.
         * Intel, WARP treat the value as boolean. */
```

PR #2218, opened 2021-08-07
[d3d9] Disable primitive restart

DadSchoorse:
There's a wine test that proves primitive restart does not exist for triangle strips and 16bit indices, and I guess that it does not exists in general because the docs recommend using degenerative triangles to restart a triangle strip. The question is if this correctness change is worth it, because it will probably invalidate a lot of the state cache.


Joshua-Ashton:
Given we've already invalidated FOZ for the spec constant fix for alpha test, may as well do this too.

PR #2224, opened 2021-08-09
[dxvk] Avoid needless refcounts

Joshua-Ashton:


doitsujin:
actually, different idea: `const DxvkImageView& view = *fb->...` etc.

Joshua-Ashton:
Attachment could be null?

doitsujin:
bad example, just grab the raw pointer for those I guess.

but doesn't matter much in the end anyway, so might as well just merge

PR #2225, opened 2021-08-09
[d3d9, dxso] Remove AlphaTestEnable spec constant

Joshua-Ashton:
Go based on the func we already have. Avoids wasting a spec const.

### DIFF
--- a/src/d3d9/d3d9_fixed_function.cpp
+++ b/src/d3d9/d3d9_fixed_function.cpp
@@ -2211,12 +2211,7 @@ namespace dxvk {
     uint32_t floatPtr = m_module.defPointerType(m_floatType, spv::StorageClassPushConstant);
 
     // Declare spec constants for render states
-    uint32_t alphaTestId = m_module.specConstBool(false);
     uint32_t alphaFuncId = m_module.specConst32(m_module.defIntType(32, 0), 0);
-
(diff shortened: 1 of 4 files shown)


src/d3d9/d3d9_fixed_function.cpp
```diff
@@ -2241,8 +2236,9 @@ namespace dxvk {
     uint32_t atestSkipLabel = m_module.allocateId();
 
     // if (alpha_test) { ... }
+    uint32_t isNotAlways = m_module.opINotEqual(boolType, alphaFuncId, m_module.constu32(VK_COMPARE_OP_ALWAYS));
     m_module.opSelectionMerge(atestSkipLabel, spv::SelectionControlMaskNone);
-    m_module.opBranchConditional(alphaTestId, atestBeginLabel, atestSkipLabel);
+    m_module.opBranchConditional(isNotAlways, atestBeginLabel, atestSkipLabel);
```
DadSchoorse, line 2241: Why is that branch needed at all?

Joshua-Ashton: It technically isn't, but we had it set up like this in the past to make some drivers happier.

We can get rid of it at some point once we guarantee folding everywhere.

K0bin (approved):

PR #2226, opened 2021-08-09
[d3d9] Optimize and clean SetSamplerState

Joshua-Ashton:


src/d3d9/d3d9_device.cpp
```diff
@@ -3644,48 +3644,47 @@ namespace dxvk {
 
     auto& state = m_state.samplerStates;
 
-    bool changed = state[StateSampler][Type] != Value;
+    if (state[StateSampler][Type] == Value)
```
K0bin, line 3647: Why don't you just make the early return unlikely instead of wrapping everything in the if block?

K0bin (approved):

PR #2228, opened 2021-08-09
Use bsf instead of tzcnt where mask != 0

Joshua-Ashton:
For when we know we aren't going to put in a mask of 0, we can use this and get better codegen.

There actually isn't anywhere we actually care about mask == 0 as we already check for that case to end the loop ourselves.

### DIFF
--- a/src/d3d11/d3d11_util.h
+++ b/src/d3d11/d3d11_util.h
@@ -13,7 +13,7 @@ namespace dxvk {
     uint32_t count = 0;
     
     while (Mask != 0) {
-      uint32_t id = bit::tzcnt(Mask);
+      uint32_t id = bit::bsf(Mask);
       pData[count++] = pData[id];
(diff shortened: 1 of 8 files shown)


doitsujin:
This kind of looks like it's changing something for the sake of changing something, I don't really like functions with undefined behaviour in some cases in the first place and saving two instructions in these places won't really affect performance.

Joshua-Ashton:
Then can we change tzcnt to use the builtin with a check instead of inline asm? That gets optimized away on GCC.

doitsujin:
Can't we just implement a proper iterator rather than doing some weird hacks? Like https://gcc.godbolt.org/z/v7M6oY4M6

DadSchoorse:
here's a branch that uses @doitsujin's iterator https://github.com/doitsujin/dxvk/compare/master...DadSchoorse:bititer

PR #2230, opened 2021-08-09
Add DXVK_NVAPIHACK env variable

SveSop:
Use DXVK_NVAPIHACK=0 to disable nvapi workaround for nVidia adapters.

Is this a viable change for DXVK for use with DXVK_NVAPI that is included in Proton Experimental/Proton-GE?
This could avoid the need for a dxvk.conf file to disable nvapiHack if you want to use nvapi with a nVidia adapter. (Atleast for games that actually tries to do it right and not just use nvapi.dll regardless of GPU)

Eg: `PROTON_ENABLE_NVAPI=1 DXVK_NVAPIHACK=0 %command%`

### DIFF
--- a/src/dxgi/dxgi_options.cpp
+++ b/src/dxgi/dxgi_options.cpp
@@ -41,6 +41,8 @@ namespace dxvk {
     this->maxSharedMemory = VkDeviceSize(config.getOption<int32_t>("dxgi.maxSharedMemory", 0)) << 20;
 
     this->nvapiHack   = config.getOption<bool>("dxgi.nvapiHack", true);
+    if (env::getEnvVar("DXVK_NVAPIHACK") == "0")
+      this->nvapiHack = false;
   }
   
 }


doitsujin:
we already provide a way to disable this with dxvk.conf.

jp7677:
With Proton in mind, disabling the nvapi hack by environment variable would actually be really cool.
As a follow up to this PR, a PR in Proton could set this environment variable automatically when dxvk-nvapi is enabled in Proton for certain games. This would avoid any cumbersome dealings with writing dxvk configurations within the Proton script.

PS: I would probably name this DXVK_DISABLE_NVAPI_HACK which disables the nvapi hack when being set to 1, since being enabled is the default. Though that’s just my (not very strong) personal preference.

SveSop:
> we already provide a way to disable this with dxvk.conf. I don't really see why you would want to change the option so frequently that it warrants an environment variable.

What is easiest for the "regular user"? To type `DXVK_NVAPIHACK=0` (or `DXVK_DISABLE_NVAPI_HACK=1`) on the steam game commandline... or to figure out what folder the binary of the game is in...

But yeah, it might make better sense to use this (or something similar) as a proton "only" setting thing maybe together with `PROTON_ENABLE_NVAPI`. I'm happy for you (or me) to close this PR and see if i can figure something out towards suggesting this to Proton.

doitsujin:
This isn't something a "regular user" should ever have to do outside of trying to fix a game that doesn't work. If a game needs this disabled for whatever reason then we'll add an app profile.

jp7677:
> This isn't something a "regular user" should ever have to do outside of trying to fix a game that doesn't work. If a game needs this disabled for whatever reason then we'll add an app profile. People have a tendency to just enable as many environment variables as they possibly can without understanding what they do, just to end up complaining on reddit that their game no longer works.
> 

Yeah, I get that point, unfortunately GitHub issue quality has rather going down than up recently :(. I don’t want to contribute to that trend even more.

> As for dxvk-nvapi shenanigans, I don't think that having that controlled via an env var is the correct solution either, I'd rather just have DXVK detect it and disable the hack if it is present and if the game is on some sort of whitelist.

Detecting DXVK-NVAPI is certainly doable, but unfortunately a bit more work since I guess you want to distinguish between other NVAPI implementations (wine-staging or NVIDIAs own on Windows). 
Not sure though how feasible a white list will work. I remember the time when DXVK hadn’t had the nvapi hack yet, the list of games that needed spoofing AMD grew rather fast. I could imagine that a new whitelist of titles that work with DXVK-NVAPI might end up fast with a lot of entries (e.g. lots of UE4 titles). 
That said just blindly trusting that every title works with DXVK-NVAPI once is detected, is indeed a bit optimistic as well.
May be having an env var set together with a successful DXVK-NVAPI detection could convince DXVK to disable the nvapi hack (thus both conditions needs to be met)?

SveSop:
> Not sure though how feasible a white list will work. I remember the time when DXVK hadn’t had the nvapi hack yet, the list of games that needed spoofing AMD grew rather fast. I could imagine that a new whitelist of titles that work with DXVK-NVAPI might end up fast with a lot of entries (e.g. lots of UE4 titles).
> That said just blindly trusting that every title works with DXVK-NVAPI once is detected, is indeed a bit optimistic as well.
> May be having an env var set together with a successful DXVK-NVAPI detection could convince DXVK to disable the nvapi hack (thus both conditions needs to be met)? Something like this would I guess work nicely for Proton when people enable nvapi there.

Maintaining a "whitelist" to be able to do this in a proper manner - and awaiting next release to merge into proton - is a tedious affaire i think.
And there would be no huge reason to use/enable this unless a game would benefit from it - Like having Interactive smoke/fog and debris in Batman Arkham Knight that some people might want to enjoy... It is not only to make a game work, since in the example of Arkham Knight, it works fine without using PhysX functions.

Using a pure "if present" detection for Proton would work, since the default is NOT to use nvapi unless `PROTON_ENABLE_NVAPI` is set, but for regular wine-staging users you would most likely have the old implementation of nvapi that does not work to well with DXVK.

Maybe a check for a particular symbol/functioncall if NVAPI.DLL is present to determine what version, and "trust" that whatever regular wineprefix containing DXVK_NVAPI actually have it for a reason vs. the non-optional install of wine-staging's nvapi. Not sure how easy that would be, but something in the lines of what is done for the VR module in DXVK? (And a logline of "nvapi detected, but too old - Faking AMD adapter" type in dxgi.log in the case of wine-staging nvapi?) 
Staging builds nvapi.dll as PE-lib now, so should be doable without too much problems?

doitsujin:
> I could imagine that a new whitelist of titles that work with DXVK-NVAPI might end up fast with a lot of entries (e.g. lots of UE4 titles).

I'd only add games that actually benefit in some way (e.g. support DLSS). I don't trust that dxvk-nvapi will work for all games that interact with nvapi.

jp7677:
> > I could imagine that a new whitelist of titles that work with DXVK-NVAPI might end up fast with a lot of entries (e.g. lots of UE4 titles).
> 
> I'd only add games that actually benefit in some way (e.g. support DLSS). I don't trust that dxvk-nvapi will work for all games that interact with nvapi.

Yeah, I agree with not trusting game interaction with the current state of dxvk-nvapi. Hence my proposal above with two conditions met before disabling nvapi hack. 

There are more benefits than just DLSS, but they are indeed small. 

Anyway, following your idea with a whitelist instead of env var sounds fine too. Let’s see then how things turn out…

SveSop:
Closing then

PR #2235, opened 2021-08-11
[dxso] Implement TexDepth

DadSchoorse:
Fixes a wine test.


DadSchoorse:
Actually I'm not sure what the correct behavior for 0/0 should be.

DadSchoorse:
@K0bin confirmed on native that the docs are wrong x <= 0.0, x/0.0 results in a depth of 0, not 1.

PR #2243, opened 2021-08-16
for each bit cleanup

DadSchoorse:
Replaces manual for each bit loops with a `bit::BitMask` iterator solution. Supersedes #2228 
This should result in marginally better code because gcc now optimizes the tzcnt.


doitsujin (approved):

doitsujin:
one include nit, other than that, LGTM.

DadSchoorse:
I just removed the include, I can't remember why I added it. The file used uint32_t before anyway.

PR #2251, opened 2021-08-20
[d3d9] Disable explicit front buffer for Fantasy Grounds

### LINKED ISSUE #1554: Fantasy Grounds window painting regresssion in dxvk 1.5.2
When using dxvk, sections of the Fantasy Grounds main menu window and in-game tabletop window fail to draw correctly. They are often filled with black or with whatever was on that section of the screen previously. Moving the mouse around the window sometimes cause the incorrectly drawn areas to flicker.

These glitches are not present with dxvk 1.5.1, nor when dxvk is not used. They are present with dxvk 1.5.2 and 1.6.

As instructed, I captured an API trace without dxvk. The glitches did not appear in this case, of course, but I launched the program and moved the mouse in the same way that would cause the glitches if dxvk had been in use.

### Software information

Fantasy Grounds Classic 3.3.10

https://www.fantasygrounds.com/filelibrary/FGWebInstall.exe

### System information
- GPU: AMD RX 480
- Driver: amdgpu (linux kernel 5.3.0) + Valve's ACO Mesa fork from their PPA, which reports itself as "Mesa 20.0.0-devel (LLVM 8.0.0)"
- Wine version: tkg-protonified-4.6, from Lutris
- DXVK version: 1.6 and 1.5.2

### Apitrace file(s)

[FantasyGrounds.zip](https://github.com/doitsujin/dxvk/files/4440162/FantasyGrounds.zip)

### Log files

This log file was from a session that reproduced the glitches, not the apitrace session.

[FantasyGrounds_d3d9.log](https://github.com/doitsujin/dxvk/files/4440220/FantasyGrounds_d3d9.log)

### Screenshots

Startup screen without dxvk: (correct)

Screenshot_2020-04-03_12-23-47

Startup screen with dxvk: (incorrect; black areas)

Screenshot_2020-04-03_12-23-18

Screenshot_2020-04-03_12-23-34

Tabletop without dxvk: (correct)

Screenshot_2020-04-03_12-28-07

Tabletop with dxvk: (incorrect; black areas & leftovers)

Screenshot_2020-04-03_12-26-25
Screenshot_2020-04-03_12-27-56

DadSchoorse:
The game expect the same, unchanged back buffer every frame.

Closes #1554


Joshua-Ashton:
We shouldn't need to do this, we should make the number of backbuffers exposed should be the same, only the actual number of swapchain images should change.

DadSchoorse:
I don't understand what you are saying. :frog:

Joshua-Ashton:
https://docs.microsoft.com/en-us/windows/win32/direct3d9/d3dswapeffect

I think we should enable this by default for `_COPY`... I'm also now curious as to if we should be doing other stuff to the buffer count for frame pacing reasons on FLIP/DISCARD... Will test some stuff out later.

Going to close this and mess around.

PR #2256, opened 2021-08-22
Enable Apitrace Mode for Crysis 3

pingubot:
Apitrace mode fixes the partly bad performance of Crysis 3. In the "Welcome to the jungle" level, my fps improves from 45fps->105fps in that location.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -145,9 +145,12 @@ namespace dxvk {
     { R"(\\Raft\.exe$)", {{
       { "dxvk.enableOpenVR",                "False" },
     }} },
-    /* Crysis 3 - slow if it notices AMD card     */
+    /* Crysis 3 - slower if it notices AMD card     *
+     * Apitrace mode helps massively in cpu bound   *
+     * game parts                                   */
     { R"(\\Crysis3\.exe$)", {{
       { "dxgi.customVendorId",              "10de" },
+      { "d3d11.apitraceMode",               "True" },
     }} },
     /* Atelier series - games try to render video *
      * with a D3D9 swap chain over the DXGI swap  *


aufkrawall:
It probably should be enabled for Ryse and Wolcen too, as they are/might be hit by the same weird Cryengine 3 CPU bound performance regression since Vega and later GPUs.

doitsujin:
@aufkrawall do you happen to know the exe names for both games? We should probably just bunch all the Cryengine games together in one app profile, the Nvidia thing affects most of them as well.

aufkrawall:
According to Nvidia driver profile database (should be correct), it is `ryse.exe` and `wolcen.exe`.

Though I wonder about Sniper Ghost Warrior Contracts 2. If it still uses Cryengine 3, it likely would be affected too. Allegedly the issue does not exist with CryEngine V games like Crysis Remastered, though I didn't check this myself.

K0bin:
> with CryEngine V games like Crysis Remastered

Crysis Remastered doesn't use CEV, it's basically CryEngine 2 ported to D3D11 with backported SVOGI and a couple of other enhancements.

burningtuna:
https://steamdb.info/app/443110/ or https://aw.my.com

Can you add Armored Warfare game also? 

Though It's online game. Official CM(Content Manager) allowed to use that. This game also uses CryEngine3. and like other CryEngine games. It shows poor performance from AMD graphics card.

after AMD driver 21.7.2, In game shadow inside is not drawn problem fixed so It can use seems like original version without a visible problem. (I tested DXVK from 1.8.1 version and the problem was depends on the AMD driver, not the DXVK)

And I manually added d3d11.apitraceMode = true, customVendorId = 10de via dxvk.conf like this branch, and that solution also works well with AW game. 

> @aufkrawall do you happen to know the exe names for both games? We should probably just bunch all the old Cryengine games together in one app profile, the Nvidia thing affects most of them as well.

the binary name is `armoredwarfare.exe` (All characters are lowercase. Steam just downloads original game launcher so there is no database about the real binary.)

K0bin:
@burningtuna Could you make a separate issue for this so we don't forget about it?

burningtuna:
> @burningtuna Could you make a separate issue for this so we don't forget about it?

Uh, Because There is only bug report for Issue category So I made a reply. 

Is it okay with that contents for issue, or should I make a new branch like this to report it?

I'm afraid to make code (even from new branch) because I don't have build enviroment right now So I cannot verify it build without error.

K0bin:
Just open an issue. Bug report is fine.

As a comment in a merged PR, it'll probably end up forgotten.

Issue #2255, opened 2021-08-22
Payday: The Heist visual glitches in reflections.


YellowOnion:
### Software information
Payday: The Heist, default settings.

The reflections artifacts look like I'm inside an icosahedron.

### System information
- GPU: Radeon RX580 4GB
- Driver:
- Wine version: Proton-GE-15.2
- DXVK version: 1.9~

### Apitrace file(s)
- https://drive.google.com/file/d/1kofo5sb6cRm9rqMgCiNpIYbWSp4QhjXX/view?usp=sharing

### Log files
- d3d9.log: https://gist.github.com/YellowOnion/3eeeea9dc05e07acdd8a5c6f0d9a7edb
- d3d11.log: 0kb
- dxgi.log: 0kb

https://user-images.githubusercontent.com/364160/130344037-4c8a8b4f-d8ca-4941-a270-de828c5b4d4c.mp4

doitsujin:
> info:  DXVK: v1.9.1-async

Please test with an official release, or current `master` (builds can be found [here](https://github.com/doitsujin/dxvk/actions)).

LingMan:
Can confirm using the official Proton 6.3-6 as well as latest master (https://github.com/doitsujin/dxvk/commit/d48b33bf951d2e7b0a42ae962ce6778473018099).
Card is an R9 390 with mesa 21.1.6.

CME42:
there are the following spir-v validation errors:
```
error: line 425: Expected Image Operand Grad dx to have 3 components, but given 2
  %261 = OpImageSampleExplicitLod %v4float %260 %249 Grad %255 %256

FS_13800625aaa63e7194eebd52f0c94a4fe46a6ebf.spv
error: line 407: Expected Image Operand Grad dx to have 3 components, but given 2
  %245 = OpImageSampleExplicitLod %v4float %244 %233 Grad %239 %240

FS_47a2bd375bfdd7fb2ac5ef277039aa496aace011.spv
```
[payday.zip](https://github.com/doitsujin/dxvk/files/7028310/payday.zip)

YellowOnion:
Thanks for the quick turn around 👍🏻

YellowOnion:
I upgraded to Proton-GE 16.1 and reflections are fixed, but there's a big shadow following me that wasn't there before...I guess I'll have to file another bug report tomorrow lol.

PR #2259, opened 2021-08-23
[dxso] Fix gradient instructions for cube maps

doitsujin:
See #2255, not sure if it'll actually fix the bug but does fix the SPIR-V validation errors.

### DIFF
--- a/src/dxso/dxso_compiler.cpp
+++ b/src/dxso/dxso_compiler.cpp
@@ -2732,7 +2732,7 @@ void DxsoCompiler::emitControlFlowGenericLoop(
       }
 
       if (opcode == DxsoOpcode::TexLdd) {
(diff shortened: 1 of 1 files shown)


src/dxso/dxso_compiler.cpp
```diff
@@ -2732,7 +2732,7 @@ void DxsoCompiler::emitControlFlowGenericLoop(
       }
 
       if (opcode == DxsoOpcode::TexLdd) {
-        DxsoRegMask gradMask(true, true, false, false);
+        DxsoRegMask gradMask(true, true, samplerType == DxsoSamplerType::SamplerTypeTextureCube, false);
```
Joshua-Ashton, line 2735: I would just use sampler.dimensions == 3 for that

Joshua-Ashton (changes requested):

PR #2260, opened 2021-08-24
Implement DXVK pieces required for DX11 DLSS support.

adamnv:
Notably, fairly generic functions to create/launch/destroy Cuda kernels,
and methods to fetch GPU virtual addresses and handles for DX11 resources.

Depends upon https://github.com/doitsujin/dxvk/pull/2287

Note: Also requires some corresponding dxvk-nvapi [changes](https://github.com/jp7677/dxvk-nvapi/pull/33) for [DLSS](https://developer.nvidia.com/dlss) to
be initialized successfully.

### DIFF
--- a/src/d3d11/d3d11_context_ext.cpp
+++ b/src/d3d11/d3d11_context_ext.cpp
@@ -3,9 +3,10 @@
 #include <cstring>
 
 #include "d3d11_device.h"
-#include "../util/log/log.h"
-
 #include "d3d11_context.h"
+#include "d3d11_cuda.h"
+
+#include "../util/log/log.h"
 
 namespace dxvk {
   
@@ -145,72 +146,51 @@ namespace dxvk {
   }
 
 
-  class CubinShaderLaunchInfo {
-  public:
-    Com<CubinShaderWrapper> cuShader;
-    std::vector<uint8_t> Params;
-    size_t ParamSize;
-    VkCuLaunchInfoNVX nvxLaunchInfo = { VK_STRUCTURE_TYPE_CU_LAUNCH_INFO_NVX };
-    const void* CuLaunchConfig[5];
-    std::vector<std::pair<Rc<DxvkBuffer>, DxvkAccessFlags>> buffers;
-    std::vector<std::pair<Rc<DxvkImage>, DxvkAccessFlags>> images;
-  };
-
-
-  bool STDMETHODCALLTYPE D3D11DeviceContextExt::LaunchCubinShaderNVX(IUnknown* hShader, uint32_t GridX, uint32_t GridY, uint32_t GridZ, const void* pParams, uint32_t ParamSize, const void** pReadResources, uint32_t NumReadResources, const void** pWriteResources, uint32_t NumWriteResources) {
+  bool STDMETHODCALLTYPE D3D11DeviceContextExt::LaunchCubinShaderNVX(IUnknown* hShader, uint32_t GridX, uint32_t GridY, uint32_t GridZ,
+      const void* pParams, uint32_t ParamSize, const void** pReadResources, uint32_t NumReadResources, const void** pWriteResources, uint32_t NumWriteResources) {
     D3D10DeviceLock lock = m_ctx->LockContext();
 
-    CubinShaderWrapper* cubinShader = (CubinShaderWrapper*)hShader;
-    auto rcLaunchInfo = new CubinShaderLaunchInfo();
+    CubinShaderWrapper* cubinShader = static_cast<CubinShaderWrapper*>(hShader);
+    CubinShaderLaunchInfo launchInfo;
 
     const uint32_t maxResources = NumReadResources + NumWriteResources;
-    rcLaunchInfo->buffers.reserve(maxResources);
-    rcLaunchInfo->images.reserve(maxResources);
+    launchInfo.buffers.reserve(maxResources);
+    launchInfo.images.reserve(maxResources);
 
-    auto InsertResource =
-    [&images = rcLaunchInfo->images,
-     &buffers = rcLaunchInfo->buffers](ID3D11Resource* res, DxvkAccessFlags access) {
-      auto buffer = GetCommonBuffer(res);
-      auto texture = GetCommonTexture(res);
-      if (buffer) buffers.emplace_back(std::make_pair(buffer->GetBuffer(), access));
-      else if (texture) images.emplace_back(std::make_pair(texture->GetImage(), access));
-    };
+    for (uint32_t i = 0; i < NumReadResources; i++)
+      launchInfo.insertResource(const_cast<ID3D11Resource*>(static_cast<const ID3D11Resource*>(pReadResources[i])), DxvkAccess::Read);
 
-    for (uint32_t i = 0; i < NumReadResources; i++) {
-      InsertResource(const_cast<ID3D11Resource*>(static_cast<const ID3D11Resource*>(pReadResources[i])), DxvkAccess::Read);
-    }
-    for (uint32_t i = 0; i < NumWriteResources; i++) {
-      InsertResource(const_cast<ID3D11Resource*>(static_cast<const ID3D11Resource*>(pWriteResources[i])), DxvkAccess::Write);
-    }
+    for (uint32_t i = 0; i < NumWriteResources; i++)
+      launchInfo.insertResource(const_cast<ID3D11Resource*>(static_cast<const ID3D11Resource*>(pWriteResources[i])), DxvkAccess::Write);
 
-    rcLaunchInfo->ParamSize = ParamSize;
-    rcLaunchInfo->Params.resize(rcLaunchInfo->ParamSize);
-    std::memcpy(rcLaunchInfo->Params.data(), pParams, ParamSize);
+    launchInfo.paramSize = ParamSize;
+    launchInfo.params.resize(launchInfo.paramSize);
+    std::memcpy(launchInfo.params.data(), pParams, ParamSize);
 
-    rcLaunchInfo->CuLaunchConfig[0] = reinterpret_cast<void*>(0x01); // CU_LAUNCH_PARAM_BUFFER_POINTER
-    rcLaunchInfo->CuLaunchConfig[1] = rcLaunchInfo->Params.data();
-    rcLaunchInfo->CuLaunchConfig[2] = reinterpret_cast<void*>(0x02); // CU_LAUNCH_PARAM_BUFFER_SIZE
-    rcLaunchInfo->CuLaunchConfig[3] = &rcLaunchInfo->ParamSize; // yes, this actually requires a pointer to a size_t containing the parameter size
-    rcLaunchInfo->CuLaunchConfig[4] = reinterpret_cast<void*>(0x00); // CU_LAUNCH_PARAM_END
+    launchInfo.cuLaunchConfig[0] = reinterpret_cast<void*>(0x01); // CU_LAUNCH_PARAM_BUFFER_POINTER
+    launchInfo.cuLaunchConfig[1] = launchInfo.params.data();
+    launchInfo.cuLaunchConfig[2] = reinterpret_cast<void*>(0x02); // CU_LAUNCH_PARAM_BUFFER_SIZE
+    launchInfo.cuLaunchConfig[3] = &launchInfo.paramSize; // yes, this actually requires a pointer to a size_t containing the parameter size
+    launchInfo.cuLaunchConfig[4] = reinterpret_cast<void*>(0x00); // CU_LAUNCH_PARAM_END
 
-    rcLaunchInfo->nvxLaunchInfo.function       = cubinShader->Function;
-    rcLaunchInfo->nvxLaunchInfo.gridDimX       = GridX;
-    rcLaunchInfo->nvxLaunchInfo.gridDimY       = GridY;
-    rcLaunchInfo->nvxLaunchInfo.gridDimZ       = GridZ;
-    rcLaunchInfo->nvxLaunchInfo.blockDimX      = cubinShader->BlockDimX;
-    rcLaunchInfo->nvxLaunchInfo.blockDimY      = cubinShader->BlockDimY;
-    rcLaunchInfo->nvxLaunchInfo.blockDimZ      = cubinShader->BlockDimZ;
-    rcLaunchInfo->nvxLaunchInfo.sharedMemBytes = 0;
-    rcLaunchInfo->nvxLaunchInfo.paramCount     = 0;
-    rcLaunchInfo->nvxLaunchInfo.pParams        = nullptr;
-    rcLaunchInfo->nvxLaunchInfo.extraCount     = 1;
-    rcLaunchInfo->nvxLaunchInfo.pExtras        = &rcLaunchInfo->CuLaunchConfig[0];
+    launchInfo.nvxLaunchInfo.function       = cubinShader->cuFunction();
+    launchInfo.nvxLaunchInfo.gridDimX       = GridX;
+    launchInfo.nvxLaunchInfo.gridDimY       = GridY;
+    launchInfo.nvxLaunchInfo.gridDimZ       = GridZ;
+    launchInfo.nvxLaunchInfo.blockDimX      = cubinShader->blockDim().width;
+    launchInfo.nvxLaunchInfo.blockDimY      = cubinShader->blockDim().height;
+    launchInfo.nvxLaunchInfo.blockDimZ      = cubinShader->blockDim().depth;
+    launchInfo.nvxLaunchInfo.sharedMemBytes = 0;
+    launchInfo.nvxLaunchInfo.paramCount     = 0;
+    launchInfo.nvxLaunchInfo.pParams        = nullptr;
+    launchInfo.nvxLaunchInfo.extraCount     = 1;
+    launchInfo.nvxLaunchInfo.pExtras        = &launchInfo.cuLaunchConfig[0];
 
-    rcLaunchInfo->cuShader = cubinShader;
+    launchInfo.shader = cubinShader;
 
-    m_ctx->EmitCs([cRcLaunchInfo = rcLaunchInfo] (DxvkContext* ctx) {
-        ctx->launchCuKernelNVX(cRcLaunchInfo->nvxLaunchInfo, cRcLaunchInfo->buffers, cRcLaunchInfo->images);
-        delete cRcLaunchInfo;
+    /* Need to capture by value in case this gets called from a deferred context */
+    m_ctx->EmitCs([cLaunchInfo = std::move(launchInfo)] (DxvkContext* ctx) {
+      ctx->launchCuKernelNVX(cLaunchInfo.nvxLaunchInfo, cLaunchInfo.buffers, cLaunchInfo.images);
     });
 
     return true;
--- a/src/d3d11/d3d11_context_ext.cpp
+++ b/src/d3d11/d3d11_context_ext.cpp
@@ -1,3 +1,10 @@
+#include <vector>
+#include <utility>
+#include <cstring>
+
+#include "d3d11_device.h"
+#include "../util/log/log.h"
+
 #include "d3d11_context.h"
 
 namespace dxvk {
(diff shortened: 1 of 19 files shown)


src/d3d11/d3d11_context_ext.cpp
```diff
@@ -134,4 +139,459 @@ namespace dxvk {
...
+      rcDxvkDevice->vkd()->vkDestroyCuFunctionNVX(vkDevice, Function, nullptr);
+      rcDxvkDevice->vkd()->vkDestroyCuModuleNVX(vkDevice, Module, nullptr);
+    };
+    VkCuModuleNVX Module;
+    VkCuFunctionNVX Function;
+    uint32_t BlockDimX, BlockDimY, BlockDimZ;
+    Rc<DxvkDevice> rcDxvkDevice;
+  };
+
+  class CubinShaderLaunchInfo : public RcObject {
```
doitsujin, line 236: Does this need to be ref-counted? I know this is a somewhat large struct, but it's probably better to just capture by value using `std::move` in order to avoid vector copies.

doitsujin:
Also, can you please split the commit a bit? At least the Vulkan header update should be separate.

src/dxvk/dxvk_context.cpp
```diff
@@ -2609,6 +2610,129 @@ namespace dxvk {
...
+
+    bool need_exec_barrier_flush = false;
+    for (auto& r : allBuffers) {
+      if (m_execBarriers.isBufferDirty(r->getSliceHandle(), DxvkAccess::Write)) {
+        need_exec_barrier_flush = true;
+        break;
+      }
+    }
+    if (!need_exec_barrier_flush) {
+      for (auto& r : allImages) {
```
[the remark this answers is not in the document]
adamnv: Thanks - done (will revisit when I've factored-out getAvailableSubresources() )

src/dxvk/dxvk_context.cpp
```diff
@@ -2609,6 +2610,129 @@ namespace dxvk {
     m_cmd->trackResource<DxvkAccess::None>(event);
   }
   
+
+  void DxvkContext::launchCuKernel(VkCuLaunchInfoNVX& launchInfo, const std::vector<Rc<DxvkBuffer>>& readBuffers, const std::vector<Rc<DxvkBuffer>>& writeBuffers, const std::vector<Rc<DxvkImage>>& readImages, const std::vector<Rc<DxvkImage>>& writeImages)
+  {
```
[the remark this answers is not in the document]
adamnv: Thanks, done.

src/d3d11/d3d11_context_ext.cpp
```diff
@@ -134,4 +139,459 @@ namespace dxvk {
...
+    if (!*pCudaTextureHandle) {
+      Logger::warn("GetCudaTextureObject() handle==0 - failed");
+      return false;
+    }
+
+    return true;
+  }
+
+
+  class CubinShaderProxy : public RcObject {
```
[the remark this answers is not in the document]
adamdmoss: Good idea.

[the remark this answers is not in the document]
adamnv: Good idea - will investigate.

src/d3d11/d3d11_context_ext.cpp
adamnv: You're right, refcounting was a leftover from before reorganization.  I've removed that now.  I believe I'm capturing by pointer.

src/dxvk/dxvk_context.cpp
```diff
@@ -2609,6 +2610,129 @@ namespace dxvk {
...
+    }
+    for (auto& r : writeBuffers) {
+      allBuffers.insert(r.ptr());
+    }
+    for (auto& r : readImages) {
+      allImages.insert(r.ptr());
+    }
+    for (auto& r : writeImages) {
+      allImages.insert(r.ptr());
+    }
```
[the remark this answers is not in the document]
adamnv: I'm happy to do that but I'm not sure how that'd affect the resulting barriers - do you mean that if an image is in both the readable and writable lists it's better to just treat it as DxvkAccess::Write only?  Or am I misunderstanding?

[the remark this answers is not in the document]
doitsujin: There's two concerns here:
a) we're doing an additional heap allocation just to combine two lists into one, when none of the code really relies on the two lists being separate in the first place. I'd strongly prefer the `(image,access)` tuples since that removes the need for pretty much all of this.
b) wrt barriers, we're currently treating *all* resources as read-write, which means that if a resource from the read list has previously been used in a read-only way as well (some sort of compute shader, copy, you name it), we may insert an unnecessary barrier anyway.

Same goes for buffers, fwiw.

src/d3d11/d3d11_interfaces.h
```diff
@@ -88,6 +90,60 @@ ID3D11VkExtContext : public IUnknown {
   virtual void STDMETHODCALLTYPE SetBarrierControl(
           UINT                    ControlFlags) = 0;
   
+  virtual bool STDMETHODCALLTYPE GetCudaTextureObject(
```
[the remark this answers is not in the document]
adamnv: Thanks, addressed in 55edad2e270883891705c0c08f5334f2c85935c3 and 885dcaa698d67a1a72686140d38e9c4df178a7aa I believe.

src/d3d11/d3d11_context_ext.h
```diff
@@ -57,10 +57,73 @@ namespace dxvk {
...
+     bool STDMETHODCALLTYPE CreateSamplerStateAndGetDriverHandle(
+            const D3D11_SAMPLER_DESC* pSamplerDesc,
+            ID3D11SamplerState**    ppSamplerState,
+            uint32_t*               pDriverHandle);
+
   private:
     
     D3D11DeviceContext* m_ctx;
     
+    static void add_sampler_and_handle(ID3D11SamplerState *sampler, uint32_t handle);
```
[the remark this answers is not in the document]
adamnv: done I believe

src/d3d11/d3d11_context_ext.h
```diff
@@ -57,10 +57,73 @@ namespace dxvk {
...
+    static ID3D11SamplerState *handle_to_sampler(uint32_t handle);
+    static void add_srv_and_handle(ID3D11ShaderResourceView *srv_imageview, uint32_t handle);
+    static ID3D11ShaderResourceView *handle_to_srv(uint32_t handle);
+    
+    static std::unordered_map<uint32_t, ID3D11SamplerState *> s_SamplerHandleToPtr;
+    static std::unordered_map<uint32_t, ID3D11ShaderResourceView *> s_SrvHandleToPtr;
```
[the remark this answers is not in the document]
adamnv: ~done I believe

adamnv:
> Also, can you please split the commit a bit? At least the Vulkan header update should be separate, especially for review purposes I'd like to get that out of the way.

Done - https://github.com/doitsujin/dxvk/pull/2287

adamnv:
I've left things unsquashed for a while to make incremental review easier but lemme know if you'd prefer a re-squash at some point.  Cheers.

src/dxvk/dxvk_cmdlist.h
```diff
@@ -340,6 +340,9 @@ namespace dxvk {
         pSizes, pStrides);
     }
     
+    void cmdLaunchCuKernel(VkCuLaunchInfoNVX launchInfo) {
```
adamnv, line 343: reminder to self - NVX suffix

adamnv:
-> Converted to draft PR since ToT isn't proven until I've gotten the corresponding dxvk-nvapi changes updated.

adamnv:
(dxvk-nvapi in https://github.com/jp7677/dxvk-nvapi/pull/33 works again with this branch now.)

src/d3d11/d3d11_context_ext.cpp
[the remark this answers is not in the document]
adamnv: done

src/dxvk/dxvk_extensions.h
```diff
@@ -283,6 +283,9 @@ namespace dxvk {
     DxvkExt khrSamplerMirrorClampToEdge       = { VK_KHR_SAMPLER_MIRROR_CLAMP_TO_EDGE_EXTENSION_NAME,       DxvkExtMode::Optional };
     DxvkExt khrShaderFloatControls            = { VK_KHR_SHADER_FLOAT_CONTROLS_EXTENSION_NAME,              DxvkExtMode::Optional };
     DxvkExt khrSwapchain                      = { VK_KHR_SWAPCHAIN_EXTENSION_NAME,                          DxvkExtMode::Required };
+    DxvkExt nvxBinaryImport                   = { VK_NVX_BINARY_IMPORT_EXTENSION_NAME,                      DxvkExtMode::Optional };
+    DxvkExt nvxImageViewHandle                = { VK_NVX_IMAGE_VIEW_HANDLE_EXTENSION_NAME,                  DxvkExtMode::Optional };
+    DxvkExt khrBufferDeviceAddress            = { VK_KHR_BUFFER_DEVICE_ADDRESS_EXTENSION_NAME,              DxvkExtMode::Optional };
```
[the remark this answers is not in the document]
adamnv: addressed now, I believe

src/dxvk/dxvk_context.cpp
```diff
@@ -2609,6 +2610,129 @@ namespace dxvk {
...
+
+    auto AutoImageSubresourceRange = [](DxvkImage* image) {
```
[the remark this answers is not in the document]
adamnv: done - and used it everywhere I was sure about.

src/dxvk/dxvk_context.cpp
```diff
@@ -2609,6 +2610,129 @@ namespace dxvk {
...
+      VkImageSubresourceRange result;
+      result.aspectMask     = image->formatInfo()->aspectMask;
+      result.baseMipLevel   = 0;
+      result.levelCount     = image->info().mipLevels;
+      result.baseArrayLayer = 0;
+      result.layerCount     = image->info().numLayers;
+      return result;
+    };
+
+    bool need_exec_barrier_flush = false;
```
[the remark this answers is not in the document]
adamnv: fixed

src/dxvk/dxvk_context.cpp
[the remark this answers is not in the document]
adamnv: To explain where the lists are coming from - they're coming from the launch entrypoint in NvAPI, which get two arrays (read, write) of opaque object handles... which in reality for DXVK are two lists of ID3D11Resource*s but not differentiated by image-vs-buffer.  Someone somewhere along the chain should ref them before submission to the CS thread, detect entries that are in both lists (to avoid double-transition) and differentiate image vs buffer.

I could do more of this before we get to launchCuKernelNVX(), likely remove one of the steps, though it's attractive (to me anyway) to take the bulk off of the application thread.

What d'ya think?  I'll look at this more deeply tomorrow, see what seems relatively elegant.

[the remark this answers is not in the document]
doitsujin: > I could do more of this before we get to launchCuKernelNVX(), likely remove one of the steps, though it's attractive (to me anyway) to keep the bulk off of the application thread.

That's what I mean, we're doing a fair bit of work there anyway so might as well process the two lists in a way that is more easy for the backend to work with. If duplicates can occur, even a naive O(n²) algorithm would be fine as long as the number of resources of each type that we're dealing with is no more than ~a dozen or so, this is how our barrier code works as well.

The primary bottleneck should be our worker thread anyway rather than the app thread, as long as the app doesn't spend more time in its own code than it does submitting D3D commands.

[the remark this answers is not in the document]
adamnv: Thinking about building the ID3D11Resource*s straight into these vectors (or sets, but probably vectors) in D3D11DeviceContextExt::LaunchCubinShaderNVX() and then using them as-is in DxvkContext::launchCuKernelNVX()
```
    std::vector<std::tuple<Rc<DxvkImage>, bool /*read*/, bool /*write*/>> images;
    std::vector<std::tuple<Rc<DxvkBuffer>, bool /*read*/, bool /*write*/>> buffers;
```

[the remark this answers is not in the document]
adamnv: Oops, didn't mean to resolve.
I've been pursuing this approach though and it seems to be working out neatly.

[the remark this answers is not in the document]
adamnv: I've pushed the rejig of this.  There are only two lists and they're only built once now, and then deduped in-place later on the CS thread.  The barriers are now finer-grained.

I haven't verified yet that this is doing what I expect (I will - next), just that nothing seems broken!

src/d3d11/d3d11_context_ext.cpp
```diff
@@ -133,5 +139,76 @@ namespace dxvk {
...
+
+  class CubinShaderLaunchInfo {
+  public:
+    Com<CubinShaderWrapper> cuShader;
+    std::vector<uint8_t> Params;
+    size_t ParamSize;
+    VkCuLaunchInfoNVX nvxLaunchInfo = { VK_STRUCTURE_TYPE_CU_LAUNCH_INFO_NVX };
+    const void* CuLaunchConfig[5];
+    std::vector<std::tuple<Rc<DxvkBuffer>, bool /*read*/, bool /*write*/>> buffers;
+    std::vector<std::tuple<Rc<DxvkImage>, bool /*read*/, bool /*write*/>> images;
```
doitsujin, line 152: Could use `DxvkAccessFlags` instead of two bools for clarity (supported bits are `DxvkAccess::Read and `DxvkAccess::Write`, respectively). It's a bit more clunky than it should be though (no operators, have to use .set/.test methods).

(2nd attempt at actually commenting because the first one for some reason didn't show up in the thing...)

doitsujin (commented):
Getting there, one final nit though.

src/d3d11/d3d11_context_ext.cpp
```diff
@@ -133,5 +139,76 @@ namespace dxvk {
...
+
+    CubinShaderWrapper* cubinshader = (CubinShaderWrapper*)hShader;
+    auto rcLaunchInfo = new CubinShaderLaunchInfo();
+
+    const uint32_t maxResources = numReadResources + numWriteResources;
+    rcLaunchInfo->buffers.reserve(maxResources);
+    rcLaunchInfo->images.reserve(maxResources);
+
+    auto InsertResource =
+    [&images = rcLaunchInfo->images, &buffers = rcLaunchInfo->buffers](ID3D11Resource* res, bool bRead, bool bWrite) {
```
Joshua-Ashton, line 167: no Hungarian here

Joshua-Ashton, line 167: newline between [] args that are this long

src/d3d11/d3d11_context_ext.cpp
```diff
@@ -133,5 +139,76 @@ namespace dxvk {
...
+      auto buffer = GetCommonBuffer(res);
+      auto texture = GetCommonTexture(res);
+      if (buffer) buffers.emplace_back(std::make_tuple(buffer->GetBuffer(), bRead, bWrite));
+      else if (texture) images.emplace_back(std::make_tuple(texture->GetImage(), bRead, bWrite));
+    };
+
+    for (uint32_t i=0; i<numReadResources; ++i) {
+      InsertResource(const_cast<ID3D11Resource*>(static_cast<const ID3D11Resource*>(pReadResources[i])), true, false);
+    }
+    for (uint32_t i=0; i<numWriteResources; ++i) {
```
Joshua-Ashton, line 177: i = 0

Joshua-Ashton: i < numWriteResources

Joshua-Ashton, line 177: i++

src/d3d11/d3d11_context_ext.cpp
```diff
@@ -133,5 +139,76 @@ namespace dxvk {
...
+      InsertResource(const_cast<ID3D11Resource*>(static_cast<const ID3D11Resource*>(pWriteResources[i])), false, true);
+    }
+
+    rcLaunchInfo->ParamSize = paramSize;
+    rcLaunchInfo->Params.clear();
+    rcLaunchInfo->Params.resize(rcLaunchInfo->ParamSize);
+    memcpy(&rcLaunchInfo->Params[0], pParams, paramSize);
```
Joshua-Ashton, line 184: .data() not &foo[0]

Joshua-Ashton: std::memcpy

src/d3d11/d3d11_context_ext.cpp
```diff
@@ -133,5 +139,76 @@ namespace dxvk {
...
+
+    rcLaunchInfo->CuLaunchConfig[0] = ((void*)0x01); // CU_LAUNCH_PARAM_BUFFER_POINTER
```
Joshua-Ashton, line 186: reinterpret_cast

src/d3d11/d3d11_context_ext.cpp
```diff
@@ -133,5 +139,76 @@ namespace dxvk {
...
+    rcLaunchInfo->nvxLaunchInfo.gridDimY = gridY;
+    rcLaunchInfo->nvxLaunchInfo.gridDimZ = gridZ;
+    rcLaunchInfo->nvxLaunchInfo.blockDimX = cubinshader->BlockDimX;
+    rcLaunchInfo->nvxLaunchInfo.blockDimY = cubinshader->BlockDimY;
+    rcLaunchInfo->nvxLaunchInfo.blockDimZ = cubinshader->BlockDimZ;
+    rcLaunchInfo->nvxLaunchInfo.sharedMemBytes = 0;
+    rcLaunchInfo->nvxLaunchInfo.paramCount = 0;
+    rcLaunchInfo->nvxLaunchInfo.pParams = nullptr;
+    rcLaunchInfo->nvxLaunchInfo.extraCount = 1;
+    rcLaunchInfo->nvxLaunchInfo.pExtras = &rcLaunchInfo->CuLaunchConfig[0];
```
Joshua-Ashton, line 203: i would line up these = signs

src/d3d11/d3d11_context_ext.h
```diff
@@ -56,11 +56,23 @@ namespace dxvk {
     
     void STDMETHODCALLTYPE SetBarrierControl(
             UINT                    ControlFlags);
-    
+
+    bool STDMETHODCALLTYPE LaunchCubinShaderNVX(
```
Joshua-Ashton, line 60: for interfaces like this, do `GridX` as msft style if youre going to use pParams and stuff

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+  // wants to use the handle.
+  // No attempt to shrink or depopulate the maps; number of
+  // entries is expected to be very limited and using a
+  // dangling handle is an app bug.
+  void D3D11DeviceExt::AddSamplerAndHandleNVX(ID3D11SamplerState *sampler, uint32_t handle) {
+    auto mapslock = m_multithreadMapsLock.AcquireLock();
+    m_SamplerHandleToPtr[handle] = sampler;
+  }
+  ID3D11SamplerState *D3D11DeviceExt::HandleToSamplerNVX(uint32_t handle) {
+    auto mapslock = m_multithreadMapsLock.AcquireLock();
```
Joshua-Ashton, line 2475: Just call this auto lock =

Joshua-Ashton: or mapLock

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+
+
+  bool STDMETHODCALLTYPE D3D11DeviceExt::GetCudaTextureObjectNVX(uint32_t srvDriverHandle, uint32_t samplerDriverHandle, uint32_t* pCudaTextureHandle) {
+    ID3D11ShaderResourceView* srv = HandleToSrvNVX(srvDriverHandle);
+    if (!srv) {
+      Logger::warn(str::format("GetCudaTextureObjectNVX() failure - srv handle wasn't found: ", srvDriverHandle));
+      return false;
+    }
+
+    ID3D11SamplerState* samplerstate = HandleToSamplerNVX(samplerDriverHandle);
```
Joshua-Ashton, line 2501: samplerState

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+      Logger::warn("GetCudaTextureObjectNVX() handle==0 - failed");
+      return false;
+    }
+
+    return true;
+  }
+  
+
+  bool STDMETHODCALLTYPE D3D11DeviceExt::CreateCubinComputeShaderWithNameNVX(const void* pCubin, uint32_t size, uint32_t blockX, uint32_t blockY, uint32_t blockZ, const char* pShaderName, IUnknown** phShader) {
+    dxvk::Rc<dxvk::DxvkDevice> dxvkdevice = m_device->GetDXVKDevice();
```
Joshua-Ashton, line 2537: dont need dxvk namespace

Joshua-Ashton: dxvkDevice

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+    VkDevice vkDevice = dxvkdevice->handle();
+
+    VkCuModuleCreateInfoNVX moduleCreateInfo = { VK_STRUCTURE_TYPE_CU_MODULE_CREATE_INFO_NVX };
+    moduleCreateInfo.pData = pCubin;
+    moduleCreateInfo.dataSize = size;
+
+    VkCuModuleNVX Module;
+    VkCuFunctionNVX Function;
```
Joshua-Ashton, line 2545: no caps please, i know module is a keyword now but its conditional and should work here, but MSVC is ass and doesnt implement it properly. so call this cuModule and cuFunction

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+    functionCreateInfo.pName = pShaderName;
+
+    result = dxvkdevice->vkd()->vkCreateCuFunctionNVX(vkDevice, &functionCreateInfo, NULL, &Function);
+    if (result != VK_SUCCESS) {
+      dxvkdevice->vkd()->vkDestroyCuModuleNVX(vkDevice, Module, nullptr);
+      Logger::warn(str::format("CreateCubinComputeShaderWithNameNVX() - failure to create function - result=", result));
+      return false; // failure
+    }
+
+    CubinShaderWrapper* cubinshader = new CubinShaderWrapper(dxvkdevice);
```
Joshua-Ashton, line 2563: cubinShader

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+      break;
+    case D3D11_RESOURCE_DIMENSION_TEXTURE1D:
+    case D3D11_RESOURCE_DIMENSION_TEXTURE3D:
+    case D3D11_RESOURCE_DIMENSION_UNKNOWN:
+    default:
+      Logger::warn(str::format("GetResourceHandleGPUVirtualAddressAndSize(?) - failure - unsupported dimension: ", resourceDesc.Dim));
+      return false;
+    }
+
+    dxvk::Rc<dxvk::DxvkDevice> dxvkdevice = m_device->GetDXVKDevice();
```
Joshua-Ashton, line 2602: no need for namespace

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+    VkDevice vkDevice = dxvkdevice->handle();
+
+    if (resourceDesc.Dim == D3D11_RESOURCE_DIMENSION_TEXTURE2D) {
+      D3D11CommonTexture *texture = GetCommonTexture(pResource);
+      Rc<DxvkImage> dxvk_image = texture->GetImage();
```
Joshua-Ashton, line 2607: dxvkImage or just image

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+
+      *gpuVAStart = imageViewAddressProperties.deviceAddress;
+      *gpuVASize = imageViewAddressProperties.size;
+    }
+    else if (resourceDesc.Dim == D3D11_RESOURCE_DIMENSION_BUFFER) {
+      D3D11Buffer *buffer = GetCommonBuffer(pResource);
+      const DxvkBufferSliceHandle dxvkh = buffer->GetBuffer()->getSliceHandle();
+      VkBuffer vk_buffer = dxvkh.handle;
+
+      VkBufferDeviceAddressInfoKHR bdainfo = { VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO_KHR };
```
Joshua-Ashton, line 2652: bdaInfo

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+        return false;
+      }
+
+      *gpuVAStart = imageViewAddressProperties.deviceAddress;
+      *gpuVASize = imageViewAddressProperties.size;
+    }
+    else if (resourceDesc.Dim == D3D11_RESOURCE_DIMENSION_BUFFER) {
+      D3D11Buffer *buffer = GetCommonBuffer(pResource);
+      const DxvkBufferSliceHandle dxvkh = buffer->GetBuffer()->getSliceHandle();
+      VkBuffer vk_buffer = dxvkh.handle;
```
Joshua-Ashton, line 2650: vkBuffer

src/d3d11/d3d11_device.cpp
```diff
@@ -2445,12 +2445,334 @@ namespace dxvk {
...
+
+      VkBufferDeviceAddressInfoKHR bdainfo = { VK_STRUCTURE_TYPE_BUFFER_DEVICE_ADDRESS_INFO_KHR };
+      bdainfo.buffer = vk_buffer;
+      VkDeviceAddress bufaddr = dxvkdevice->vkd()->vkGetBufferDeviceAddressKHR(vkDevice, &bdainfo);
+      *gpuVAStart = (uint64_t)bufaddr + dxvkh.offset;
```
Joshua-Ashton, line 2655: uint64_t(bufAddr)

src/d3d11/d3d11_device.cpp
Joshua-Ashton: I'll let you fix the rest there's quite a few

src/d3d11/d3d11_device.h
```diff
@@ -513,11 +513,80 @@ namespace dxvk {
...
+            ID3D11Resource*         pResource,
+            const D3D11_UNORDERED_ACCESS_VIEW_DESC* pDesc,
+            ID3D11UnorderedAccessView** ppUAV,
+            uint32_t*               pDriverHandle);
+
+     bool STDMETHODCALLTYPE CreateShaderResourceViewAndGetDriverHandleNVX(
+            ID3D11Resource*         pResource,
+            const D3D11_SHADER_RESOURCE_VIEW_DESC* pDesc,
+            ID3D11ShaderResourceView** ppSRV,
+            uint32_t*               pDriverHandle);
```
Joshua-Ashton, line 545: weird indent

src/d3d11/d3d11_context_ext.cpp
adamdmoss: Hah!  I considered that but having seen that Read=0 and Write=1 I assumed that they were never meant to be OR'd.  I didn't look deeper into Flags<> which is... pretty wild!  I'll amend.

adamnv: Hah! I considered that but having seen that Read=0 and Write=1 I assumed that they were never meant to be OR'd. I didn't look deeper into Flags<> which is... pretty wild! I'll amend.

adamnv: done!  it made some other stuff simpler too.

adamnv:
rebased onto master and squashed.

adamnv:
I think this is ready to go.

doitsujin:
Found some more stuff that needed addressing, and made a branch [here](https://github.com/doitsujin/dxvk/tree/dlss).

92faf5fb82ca90afa0d4f73656040247ab118be1 : Since we're always inserting a barrier that blocks `VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` anyway, we can fold that into the exec barriers which removes a whole bunch of code. Separate acquisition barriers are only necessary if we need the image to be in a specific layout, but that does not appear to be the case.

ed8b48314c95cb7bce4163018f3c67ca8c7c1230 : Other than moving some code around and cleaning up formatting, this does capture the `CubinShaderLaunchInfo` structure by value rather than by pointer, and also moves the resource list deduplication to the app thread. We don't really have a choice here; if this is called on a deferred context then everything submitted via `EmitCs` **may** be called multiple times, which would previously lead to potential race conditions and use-after-free.

If you're okay with these changes, I'd squash them into the original commit and merge.

adamnv:
Thanks Philip.
Hmm, those changes make DLSS-using apps TDR... I'll have to look into why.
First guess would be std::move()'ing the CubinShaderLaunchInfo, since it contains a pointer to its own internals - if that's it, I'll see if I can whip up a fix.

adamnv:
>  I don't know if the NVAPI side of CreateCubinComputeShaderWithNameNVX currently does an extra AddRef or not but if it does, that may need to be removed.
It does - it seemed weird at the time to have nvapi responsible for the release but not the ref, but of course an implicit first ref is a pretty normal COM thing.

adamnv:
> Since we're always inserting a barrier that blocks VK_PIPELINE_STAGE_ALL_COMMANDS_BIT anyway

You're right, that simplifies a lot of stuff.  I eventually wanted to revisit VK_PIPELINE_STAGE_ALL_COMMANDS_BIT; I did it this way because it's what the vulkan-native DLSS code does, along with making every barrier a VK_ACCESS_SHADER_READ_BIT|VK_ACCESS_SHADER_WRITE_BIT regardless of actual access, which all seemed stricter than necessary - but life is short and I really like how this simplifies things. ;)

adamnv:
> First guess would be std::move()

adamnv:
I've included your changes in this PR now and added this fix:
https://github.com/doitsujin/dxvk/commit/92fb01e9d6305a53fbf2412a5eb62dc0f8db9e7b

With that, everything appears to work.

doitsujin:
Thanks, the pointer thing was indeed an oversight.

Merged manually.

PR #2263, opened 2021-08-24
Fix DxvkShaderConstData garbage pointer for default c'tor

tayarani:
The current default c'tor DxvkShaderConstData doesn't set the m_data
pointer. The copy c'tor takes over that pointer when assigning, which
results in seg faults if an object is ever assigned. This fixes it by
initializing the correct data.

### DIFF
--- a/src/dxvk/dxvk_shader.cpp
+++ b/src/dxvk/dxvk_shader.cpp
@@ -6,7 +6,8 @@
 
 namespace dxvk {
   
-  DxvkShaderConstData::DxvkShaderConstData() : m_size(0), m_data(nullptr) {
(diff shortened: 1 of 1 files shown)


src/dxvk/dxvk_shader.cpp
```diff
@@ -6,7 +6,7 @@
 
 namespace dxvk {
   
-  DxvkShaderConstData::DxvkShaderConstData() {
+  DxvkShaderConstData::DxvkShaderConstData() : m_size(0), m_data(nullptr) {
```
doitsujin, line 9: nit, we put initializers on a new line.

tayarani: Fixed.

PR #2273, opened 2021-09-02
#include <utility> for d3d10_multithread.cpp

shoober420:


shoober420:
Fortunately, this is the last missing #include GCC12 picks up.

PR #2287, opened 2021-09-10
Update Vulkan headers to v1.2.182

adamnv:
Verbatim header snapshot from the KhronosGroup/Vulkan-Headers git repo.
Required for the extensions used by https://github.com/doitsujin/dxvk/pull/2260


adamnv:
Since I'm here I'm happy to update these to the now-latest version - ~1.2.191 rather than this PR's 1.2.182.  But I've been building and testing against 1.2.182 and haven't really proven 1.2.191 so :shrug:

Issue #2107, opened 2021-06-16
[d3d9] - DARK SOULS™ II (236430) - Grass in "Things Betwixt" does not render (most of the time).


ZereoX:
### Description:
In **_DARK SOULS™ II_** (236430) grass in **Things Betwixt** (starting location) does not render, but some will ocasionally pop-in depending on distance and camera angle.

- Issue is also present on WineD3D
- Issue occurs on both Nvidia and AMD GPUs.
- Issue occurs on both Proton 6.3-4 and Windows on all major DXVK versions since 1.5, including master (6b8ab4fd385cc3cdeb272e4b4a6b69a1d4100d59). 

| Windows - DXVK 1.9 | Proton - WineD3D | Windows - Native
| ---- | ---- | ---- |
|![Windows - DXVK](https://user-images.githubusercontent.com/761853/122217477-738f4300-ce7b-11eb-94f4-924828769262.png)|![Proton - WineD3D](https://user-images.githubusercontent.com/761853/122217527-80139b80-ce7b-11eb-8af1-2ee9f4e53a69.png)|![Windows - Native](https://user-images.githubusercontent.com/761853/122217554-83a72280-ce7b-11eb-8260-dd40a7e45b1f.png)|

Please let me know if any additional information is required or if additional testing is needed.

#### Notes
- Issue occurs at all graphical preset.
- Issue is not present in DARK SOULS™ II: Scholar of the First Sin (uses DirectX 11 instead of 9).

### Software information
[DARK SOULS™ II](https://store.steampowered.com/app/236430/DARK_SOULS_II/) (236430)

### System information
OS: Ubuntu 20.04 / Windows 10 Pro 20H2
GPU: Radeon RX 5700 XT / Nvidia GeForce GTX 1060
Driver: 21.2\~git2106150600.639579\~oibaf\~f / 465.27-0ubuntu0.20.04.2
Wine version: Proton 6.3-4
DXVK version: 1.5.5, 1.6.1, 1.7.3, 1.8.1, 1.9

### Apitrace file
- Link: https://drive.google.com/file/d/13j_-lX7BoJ0AHcGm_vbr6LNvwwZgtcUC/view?usp=sharing

### Log files
d3d9.log: [DarkSoulsII_d3d9.log](https://github.com/doitsujin/dxvk/files/6662914/DarkSoulsII_d3d9.log)

```
warn:  ConvertFormat: Unknown format encountered: D3D9Format::AL16
[...]

PR #2294, opened 2021-09-14
Rework D3D9 float emulation

doitsujin:
Replaces almost every occurence of `a * b` with `(b == 0 ? 0 : a) * (a == 0 ? 0 : b)` if we know that both operands may be zero. I've omitted the TexBem instructions for now since I don't know what those do exactly and if there are any NaNs to filter out there, but everythig else should be handled.

The existing `d3d9.floatEmulation` option still works, and setting it to `false` effectively disables this.

Fixes stuff like #2107, but will obviously need a fair bit of testing.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -6330,11 +6330,7 @@ namespace dxvk {
     }
 
     UpdateStateConstants<ProgramType, ConstantType, T>(
-      &m_state,
-      StartRegister,
-      pConstantData,
-      Count,
-      m_d3d9Options.d3d9FloatEmulation);
+      &m_state, StartRegister, pConstantData, Count);
 
     return D3D_OK;
   }
--- a/src/d3d9/d3d9_state.h
+++ b/src/d3d9/d3d9_state.h
@@ -241,20 +241,12 @@ namespace dxvk {
           D3D9CapturableState* pState,
           UINT                 StartRegister,
     const T*                   pConstantData,
-          UINT                 Count,
-          bool                 FloatEmu) {
+          UINT                 Count) {
     auto UpdateHelper = [&] (auto& set) {
       if constexpr (ConstantType == D3D9ConstantType::Float) {
+        size_t size = Count * sizeof(Vector4);
 
-        if (!FloatEmu) {
-          size_t size = Count * sizeof(Vector4);
-
-          std::memcpy(set.fConsts[StartRegister].data, pConstantData, size);
-        }
-        else {
-          for (UINT i = 0; i < Count; i++)
-            set.fConsts[StartRegister + i] = replaceNaN(pConstantData + (i * 4));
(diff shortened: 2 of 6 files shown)


DadSchoorse:
This really needs some work in the drivers for low end hardware, I'm losing up to 20% performance on my renoir apu.

pendingchaos:
Not sure if it matters but `(b == 0 ? 0 : a) * (a == 0 ? 0 : b)` is `-0.0` for `a,b=-0.0,-1.0`, unlike AMD's legacy multiplication opcode, which is `0.0` IIRC

doitsujin:
We can change this to `(b == 0 ? b : a) * (a == 0 ? a : b)` if that more closely emulates v_mul_legacy behaviour.

Joshua-Ashton:
fwiw we do not enabled Signed Zero Preserve in D3D9

pendingchaos:
> Does the instruction always return positive zero or does it return the correct sign? That's not clear to me right now.

I tested this a while ago and IIRC it's always positive zero if either source is negative or positive zero

K0bin (approved):

DadSchoorse:
This is probably more a theoretical problem, but doesn't this need SignedZeroInfNanPreserve to ensure correctness? Without it it's allowed to optimize `(b == 0 ? 0 : a) * (a == 0 ? 0 : b)` to `a * b`, I think.

pendingchaos:
> This is probably more a theoretical problem, but doesn't this need SignedZeroInfNanPreserve to ensure correctness? Without it it's allowed to optimize `(b == 0 ? 0 : a) * (a == 0 ? 0 : b)` to `a * b`, I think.

I don't think that's a valid optimization for the compiler to do. If one of the operands is 0 and the other is nan/inf, the expression has no multiplication by nan/inf

DadSchoorse:
Okay, you're the compiler expert not me. :smile: My understanding of the default vulkan floating point rules is that the compiler is allowed to assume that no arguments are NaN or Inf tho.

pendingchaos:
`nmin(abs(a), abs(b)) == 0.0 ? 0.0 : a * b` might be faster on devices where min can take two abs modifiers (most devices?), though the size of the spirv might be larger. It should also match `v_mul_legacy_f32` exactly

pendingchaos:
I think this (work in progress) branch should mostly restore RADV performance: https://gitlab.freedesktop.org/pendingchaos/mesa/-/commits/radv_zerowins_misc

I haven't actually tested this with DXVK though

DadSchoorse:
> I think this (work in progress) branch should mostly restore RADV performance: https://gitlab.freedesktop.org/pendingchaos/mesa/-/commits/radv_zerowins_misc
> 
> I haven't actually tested this with DXVK though

These patches fully restore performance in my handful of test cases.

K0bin:
Merged with #2359

PR #2305, opened 2021-09-23
Update setup_dxvk.sh

Ahmed-E-86:
Add an optional WINE variable that can be used with custom wine builds.

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -56,9 +56,18 @@ export WINEDEBUG=-all
 # wine gecko and mono
 export WINEDLLOVERRIDES="mscoree,mshtml="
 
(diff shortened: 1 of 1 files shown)


doitsujin:
We've rejected this at least once in the past, `WINE` is not a standard wine environment variable, and I don't see why you couldn't just adjust `PATH` or set `WINELOADER`, which would both work without changing the script.

Ahmed-E-86:
In order for `PATH` to work when installing DXVK to Proton or Proton-GE wineprefix `wineboot="wineboot"` needs to be changed to `wineboot="wine wineboot"`, that will probably work for `WINELOADER` as well.

PR #2317, opened 2021-09-27
[d3d9] Don't clamp clears to the size of RT 0

K0bin:
Fixes shadows in Black Mesa.

Black Mesa clears with RT | Depth. The depth buffer is 4096x4096 and the RT is 2048x2048.
It used to work by accident because DXVK mistakenly did a fast full clear.

@DadSchoorse added a check for the DS size which broke it. https://github.com/doitsujin/dxvk/commit/a75cf2f39d3ee134406e06b6d6e2e7f54f81dc9b

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1445,12 +1445,25 @@ namespace dxvk {
     }
 
     auto ClearImageView = [this](
-      bool                     fullClear,
+      uint32_t                 alignment,
       VkOffset3D               offset,
       VkExtent3D               extent,
       const Rc<DxvkImageView>& imageView,
       VkImageAspectFlags       aspectMask,
       VkClearValue             clearValue) {
+      
+      VkExtent3D imageExtent = imageView->mipLevelExtent(0);
+      extent.width = std::min(imageExtent.width, extent.width);
+      extent.height = std::min(imageExtent.height, extent.height);
+
+      if (unlikely(uint32_t(offset.x) >= imageExtent.width || uint32_t(offset.y) >= imageExtent.height))
+        return;
+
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -1445,12 +1445,17 @@ namespace dxvk {
...
-      bool                     fullClear,
+      uint32_t                 alignment,
       VkOffset3D               offset,
       VkExtent3D               extent,
       const Rc<DxvkImageView>& imageView,
       VkImageAspectFlags       aspectMask,
       VkClearValue             clearValue) {
+      
+      VkExtent3D imageExtent = imageView->image()->info().extent;
+      const bool fullClear = align(extent.width, alignment) == imageExtent.width
```
DadSchoorse, line 1456: I would clamp the extent before this, then we can still do fast clears if the extent is too big.

DadSchoorse: Also you need to check if the offset is 0

K0bin: Good call.

src/d3d9/d3d9_device.cpp
```diff
@@ -1445,12 +1445,22 @@ namespace dxvk {
...
     auto ClearImageView = [this](
-      bool                     fullClear,
+      uint32_t                 alignment,
       VkOffset3D               offset,
       VkExtent3D               extent,
       const Rc<DxvkImageView>& imageView,
       VkImageAspectFlags       aspectMask,
       VkClearValue             clearValue) {
+      
+      VkExtent3D imageExtent = imageView->image()->info().extent;
```
DadSchoorse, line 1455: This is the wrong extent if the image view isn't mip 0

PR #2323, opened 2021-10-05
[dxvk] fix barrier tracking code

oltolm:
Fix for a *possible* problem. I'm not 100% sure.


doitsujin (approved):

doitsujin:
This was indeed a bug, albeit a bit of a weird one since it will cause barriers to be too conservative rather than missing (`entry->data.overlaps(slice)` is always true and `entry->data.getAccess()` is as pessimistic as it can get), so it went unnoticed.

PR #2359, opened 2021-11-12
Rework D3D9 float emulation - as an option

K0bin:
Basically #2294 except that it's an option only enabled for specific games. (the ones in #2195)

The option `d3d9.floatEmulation` gets a third option besides `True` and `False` called `Strict` which enables the more accurate implementation. Once Mesa has landed an optimization for this, we can just make it automatically pick `Strict`.

### DIFF
--- a/src/d3d9/d3d9_options.h
+++ b/src/d3d9/d3d9_options.h
@@ -7,6 +7,12 @@
 
 namespace dxvk {
 
+  enum class D3D9FloatEmulation {
+    Disabled,
+    Enabled,
+    Strict
+  };
+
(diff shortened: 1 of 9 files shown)


src/d3d9/d3d9_options.h
```diff
@@ -7,6 +7,12 @@
 
 namespace dxvk {
 
+  enum D3D9FloatEmulation {
```
DadSchoorse, line 10: This should be an enum class.

DadSchoorse (changes requested):
You should also change dxvk.conf.

PR #2379, opened 2021-11-25
[meta] Be clearer about mingw threading support

DadSchoorse:


pchome:
Maybe use [compiler.compiles()](https://mesonbuild.com/Reference-manual_returned_compiler.html#compilercompiles) (or more suitable compiler.* function if any) and warn user?

Similar code was there a while ago:
```meson
code = '''#ifndef __WINE__
#error 1
#endif'''

dxvk_winelib = dxvk_compiler.compiles(code, name: 'winelib check')
```

PR #2396, opened 2021-12-13
Support per-application install in setup_dxvk.sh

DianaNites:
Adds an `--app` command-line option that installs as normal, but sets the dll overrides for the named application instead of all applications.

Does have the perhaps notable side effect of `setup_dxvk uninstall` not removing every override, as per-application ones will remain, but I don't think thats a big issue, and those are easily removed in `winecfg` or by running `setup_dxvk uninstall --app game.exe`.

If thats an issue though I suppose `wine reg query "HKEY_CURRENT_USER\Software\Wine\AppDefaults" 2>/dev/null` would list all the per-application overrides that exist and the script could simply try them all, though that might have unintended consequences.

----

The motivating use-case for this is a wineprefix containing the epic games store, where a game has a launcher that does not work with dxvk, but the game doesnt work without it. EGS itself doesnt mind either way.

This PR allows easily resolving this situation, as one need only do `setup_dxvk install --app game.exe` and both the launcher and game will work perfectly, the launcher without dxvk and the game with it, all from within EGS, or whatever other program gives issues like this.

The new option checks to make sure that an argument has been passed for `<name>` and that it doesn't look like a command-line flag, and displays a helpful error, so that it wont try something silly like using `--symlink` as the application if a user mistypes.

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -17,7 +17,7 @@ uninstall)
   ;;
 *)
   echo "Unrecognized action: $action"
-  echo "Usage: $0 [install|uninstall] [--without-dxgi] [--with-d3d10] [--symlink]"
+  echo "Usage: $0 [install|uninstall] [--without-dxgi] [--with-d3d10] [--symlink] [--app <name>]"
   exit 1
 esac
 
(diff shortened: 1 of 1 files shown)


doitsujin:
which "launcher that doesn't work with DXVK" are we talking about?

I'd rather have that fixed than adding crude workarounds to the install script.

DianaNites:
> which "launcher that doesn't work with DXVK" are we talking about?

Paradox launcher for the windows version of Prison Architect, which is currently free on epic games. PA natively supports linux, but not on epic games, which is why the windows version is being used.

> I'd rather have that fixed

Of course i'd love that too, It'd be great if everything just worked with dxvk, but things don't, so why not make it a little easier to workaround them using the very useful and well supported wine feature of per-application dll overrides?

It could be done manually, but thats rather troublesome, and this is hardly a big or high maintenance change.

> explicitly disable DXVK for the broken launcher rather than manually having to enable it for every single game in the prefix.

Oh yeah, thats true.

If you're at all open to this change, I could change it so that `--app` disables dxvk for an application instead of enabling it?

Blisto91:
Closure reason https://github.com/doitsujin/dxvk/pull/3097

PR #2399, opened 2021-12-14
Add dxgi.customSubSysId and dxgi.customRevision configs

imaami:
Admittedly a hack, but a relatively noninvasive one. This makes it possible to inform apps about the GPU's exact board partner vendor and model. The use case I encountered was Superposition not including details about the card in its results when run in Proton. And I couldn't find a proper way to add these automatically in `dxvk::Dxgi::GetDesc3()` either, since it looks like libvulkan simply has no API to query for these.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -48,6 +48,18 @@
 # d3d9.customDeviceDesc = ""
 
 
+# Override the PCI subsystem ID and revision reported to the application.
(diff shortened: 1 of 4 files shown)


imaami:
To illustrate that it works:

image

doitsujin:
Considering that the only real use cases for this are extremely niche and not really intended to be covered by DXVK, I'm going to close this.

PR #2405, opened 2021-12-18
[util] Don't call GetCurrentThreadId() on every spin

imaami:
Signed-off-by: Juuso Alasuutari <juuso.alasuutari@gmail.com>

### DIFF
--- a/src/util/sync/sync_recursive.cpp
+++ b/src/util/sync/sync_recursive.cpp
@@ -4,7 +4,8 @@
 namespace dxvk::sync {
 
   void RecursiveSpinlock::lock() {
-    spin(2000, [this] { return try_lock(); });
+    uint32_t threadId = GetCurrentThreadId();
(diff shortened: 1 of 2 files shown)


imaami:
Updated to restore compliance with C++11's Lockable requirement should it become an issue in the future. Thanks to Torge for pointing this out on Discord.

doitsujin:
does this lead to any tangible improvements anywhere?

imaami:
> does this lead to any tangible improvements anywhere?

I'd be surprised if it did. On the other hand there's no reason I can think of why `GetCurrentThreadId()` should be called on every loop. @Guy1524 said on Discord that he checked and found out the call doesn't get inlined by MingGW, so there's at least a nano-optimization there.

doitsujin:
Yeah, don't really see the point to be honest other than making the implementation unnecessarily clunky. We're literally just spinning faster, in a way this even the opposite of what we actually want.

Issue #2036, opened 2021-04-22
Issue with GPU (wrong GPU chosen)


gabriele2000:
I honestly don't know where to ask this, because I don't know if it's a DXVK issue or a Wine one...
Long story short, the game "Curse of the Dead Gods" launches on my integrated GPU instead on the Nvidia GPU which is strange because the game runs on DX11 as far as I know...
I'll upload the log, that's the only game that has this "issue". Yes, issue because the framerate is pure garbage without DXVK
[wine.log](https://github.com/doitsujin/dxvk/files/6358525/wine.log)

doitsujin:
Are you using Proton? If so, `WINEDLLOVERRIDES=dxgi=n` should fix that problem (which switches to DXVK's DXGI implementation).

gabriele2000:
> Are you using Proton? If so, `WINEDLLOVERRIDES=dxgi=n` should fix that problem (which switches to DXVK's DXGI implementation).
> 
> DXVK always prioritizes dedicated GPUs over integrated ones, but if the game insists on using the integrated one for whatever reason, there's always the option to use `DXGI_FILTER_DEVICE_NAME` as well.

I'm using wine-staging-tkg 6.5, a custom build, and that's the only game that has this problem...
I'll wait until 6.7 comes out, maybe it'll fix the problem

Joshua-Ashton:
It will not.

gabriele2000:
> It will not.

where do I have to write "DXGI_FILTER_DEVICE_NAME" if there isn't any .log file in the game's folder? Asking since the log files are a living proof that DXVK is working and for that game, it's not

SveSop:
[https://github.com/doitsujin/dxvk#device-filter](https://github.com/doitsujin/dxvk#device-filter)
Either in your launcher script with eg. `export DXVK_FILTER_DEVICE_NAME="My GPU"` or just as part of the wine launcher line.

You can also redirect log files to a separate folder of your choosing (so you do not have to find the correct folder where the .exe file is or whatnot).
`export DXVK_LOG_PATH="/home/yourname/gamelogs"` (or whatever you would like).

N-Accumbens:
Related to this, is it possible to choose the GPU based on something other than the name? I have two of the same GPUs that have the same name, but they're not in SLI...

meladath:
[...]

PR #2408, opened 2021-12-22
Add deviceUUID filter with DXVK_FILTER_DEVICE_UUID

meladath:
In relation to https://github.com/doitsujin/dxvk/issues/2036. Maybe not the best solution, but just letting people know its a solution.

### DIFF
--- a/src/dxvk/dxvk_device_filter.cpp
+++ b/src/dxvk/dxvk_device_filter.cpp
@@ -1,13 +1,25 @@
 #include "dxvk_device_filter.h"
 
+std::string convertUUID(const uint8_t* uuid) {
+    std::ostringstream convert;
+    for (int a = 0; a < 16; a++) {
+        convert << static_cast<int>(uuid[a]);
+    }
+    std::string key_string = convert.str();
+    return key_string;
+}
+
 namespace dxvk {
   
   DxvkDeviceFilter::DxvkDeviceFilter(DxvkDeviceFilterFlags flags)
   : m_flags(flags) {
     m_matchDeviceName = env::getEnvVar("DXVK_FILTER_DEVICE_NAME");
+    m_matchDeviceUUID = env::getEnvVar("DXVK_FILTER_DEVICE_UUID");
     
     if (m_matchDeviceName.size() != 0)
       m_flags.set(DxvkDeviceFilterFlag::MatchDeviceName);
+    if (m_matchDeviceUUID.size() != 0)
+      m_flags.set(DxvkDeviceFilterFlag::MatchDeviceUUID);
   }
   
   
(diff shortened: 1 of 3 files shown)


Tiny-Paws:
Shouldn't ConvertUUID take a reference to uuid rather than a pointer ?

src/dxvk/dxvk_device_filter.cpp
```diff
@@ -1,13 +1,25 @@
 #include "dxvk_device_filter.h"
 
+std::string convertUUID(const uint8_t* uuid) {
```
Joshua-Ashton, line 3: Doesn't seem to match how UUIDs are commonly formatted at all?

meladath: Thanks for taking a look at this. Do you have a reference that I can look at so that I can fix the implementation? As far as I was aware, UUID's were simply a string of chars, so I assumed this would be an ok, albeit ugly solution.

kitlith: From [wikipedia](https://en.m.wikipedia.org/wiki/Universally_unique_identifier#Format):

> In its canonical textual representation, the 16 [octets](https://en.m.wikipedia.org/wiki/Octet_(computing)) of a UUID are represented as 32 [hexadecimal](https://en.m.wikipedia.org/wiki/Hexadecimal) (base-16) digits, displayed in five groups separated by hyphens, in the form 8-4-4-4-12 for a total of 36 characters (32 hexadecimal characters and 4 hyphens). For example:
> 
>     123e4567-e89b-12d3-a456-426614174000
>     xxxxxxxx-xxxx-Mxxx-Nxxx-xxxxxxxxxxxx

I'm not aware of a format for UUIDs that only uses the digits 0-9, but that doesn't mean it doesn't exist. (i suspect it would be along the lines of interpreting the uuid as a single 128 bit number to avoid ambiguity, though.)

SveSop: @meladath 
Somewhat of a related UUID issue from CUDA: [https://stackoverflow.com/questions/68823023/set-cuda-device-by-uuid](https://stackoverflow.com/questions/68823023/set-cuda-device-by-uuid)

Sample code snippit there to select adapter based on UUID.

So, since it is octets/base-16 the number: 3f12  is not the same as 3f 12 😄 I have seen uuid/luid being treated like a char array - `char uuid[16]`

meladath: We now use uint32_t, this will preserve any hex chars and allow UUID's to correctly be captured and compared.

SveSop: [https://github.com/KhronosGroup/Vulkan-Tools/blob/master/vulkaninfo/outputprinter.h#L35-L43](https://github.com/KhronosGroup/Vulkan-Tools/blob/master/vulkaninfo/outputprinter.h#L35-L43)
Is that not the function Vulkaninfo use to output this UUID?

I am not sure that uint32_t would "preserve" the hex numbers any better... char == uint8_t != uint32_t.

But tbh, i am by no means any authority on that matter.

meladath: Hey, thats correct, I was looking at some old source where it simply cast each array element into a uint32_t and output it to cout using a formatting string which added the - chars in.

You are correct about it not being any better to preserve data than a char, the problem was in my original implementation I was casting it to an int, resulting in erroneous values for UUID's that contains any character from a-f.

meladath:
Could I get a review on this please? I have edited the OP to be more representative of the PR's current state.

joshhansen:
Just pinging this as I have four identical GPUs which are not distinguishable using DXVK_FILTER_DEVICE_NAME....

PR #2422, opened 2021-12-30
Fix off by one error in string utility functions

vxiiduu:
Microsoft documentation for WideCharToMultiByte (https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte) very clearly states, if cbMultiByte parameter is 0, "the function returns the required buffer size for lpMultiByteStr". The same goes for the cchWideChar parameter to function MultiByteToWideChar.
image

However for some sort of reason, tomb function in dxvk then goes on to subtract 1 from this returned value, before passing it straight back to WideCharToMultibyte parameter cbMultiByte. When executed, this causes an error (OS: Windows 7 x64). I don't know if it causes the same error when run under Wine. Regardless, this usage is clearly nonconformant to the official documentation for this function.
image

Please ignore the kernel33.DLL and pretend it is kernel32.DLL.

### DIFF
--- a/src/util/util_string.cpp
+++ b/src/util/util_string.cpp
@@ -8,8 +8,6 @@ namespace dxvk::str {
     if (len <= 1)
       return "";
 
-    len -= 1;
-
     std::string result;
     result.resize(len);
     ::WideCharToMultiByte(CP_UTF8, 0, ws, -1,
@@ -30,8 +28,6 @@ namespace dxvk::str {
     
     if (len <= 1)
       return L"";
-
-    len -= 1;
 
     std::wstring result;
     result.resize(len);


doitsujin:
We subtract 1 because we don't need nor want a terminating null byte.

TheLastRar:
Looking at the documentation of WideCharToMultiByte, it also states for the parameter `cchWideChar`;

> If this parameter is set to a positive integer, the function processes exactly the specified number of characters. If the provided size does not include a terminating null character, the resulting character string is not null-terminated, and the returned length does not include this character.

If you don't want to deal with a null character then the solution would probably be to get the length of the input string yourself before calling WideCharToMultiByte with cchWideChar set to that value (and similarly for MultiByteToWideChar)

PR #2485, opened 2022-02-06
[d3d11] Add D3D11On12CreateDevice stub

Saancreed:
"Solves" issues like https://github.com/cursey/reframework-d2d/issues/5 by allowing callers to gracefully handle the failure instead of crashing outright.

### DIFF
--- a/src/d3d11/d3d11.def
+++ b/src/d3d11/d3d11.def
@@ -3,3 +3,4 @@ EXPORTS
     D3D11CoreCreateDevice @18
     D3D11CreateDevice @22
     D3D11CreateDeviceAndSwapChain @23
+    D3D11On12CreateDevice @24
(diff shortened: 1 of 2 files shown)


doitsujin:
Any idea why the msvc build is broken? Would that change with your pending PR?

Saancreed:
> Any idea why the msvc build is broken? Would that change with your pending PR?

Yup, it's the same issue where `meson --internal symbolextractor` just fails with no error message whatsoever, which my PR is supposed to work around. [Your own branch exhibits the same behavior](https://github.com/doitsujin/dxvk/runs/5084709051?check_suite_focus=true), and so does vkd3d-proton for the last few days. Still no idea why because I can't repro this on my Windows VM.

PR #2488, opened 2022-02-09
[build] Avoid meson warning.

DadSchoorse:
WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in future releases of meson.

Stupid change, stupid warning, stupid fix.


doitsujin:
We should set this to `true` if anything, just to make it a bit more robust when shader compiling fails for whatever reason.

eli-schwartz:
When the run_command succeeds, it sets `glsl_args += [ '--quiet' ]`.
Currently when the run_command fails, it does nothing.

My assumption is the purpose of this check is to see whether the `--quiet` argument is supported. It probably shouldn't just error out otherwise.

The purpose of the Meson change is because it isn't obvious that run_command may fail and you need to manually check the returncode. In the common case, when doing work based on the stdout of a run_command, it's unexpected to have it do the wrong thing in bizarre "it somehow failed" scenarios. It is admittedly pointless in cases such as here, where you immediately check the returncode, but Meson as implemented doesn't know what you're going to do later on.

DadSchoorse:
Yeah, exactly. We don't want meson to error out while we check if an glslang argument is supported, that would mean configure would always fail on old versions.

PR #2492, opened 2022-02-12
[d3d9] Add HUD items for WaitForResource & CS thread syncs

K0bin:
Given that 90% of D3D9 performance problems with DXVK come down to WaitForResource, this is something we should have done a lot earlier.


K0bin:
Closed in favor of doitsujins backend implementation.

PR #2497, opened 2022-02-14
RFC: Free unused managed buffers

K0bin:
I'd like to reduce the amount of address space DXVK wastes because of D3DPOOL_MANAGED mapping buffers.

I have two slightly different proposals to do that:

- As soon as we upload a managed resource we start keeping track of the last time it was used. Lock and AddDirtyRect mark it as used.
Once per frame we iterate over the textures and free all managed buffers that haven't been used for 16 (arbitrary) frames. I also remove textures that are marked as needing upload (locked but not yet used) to keep the list short.

- The slightly more aggressive variant (implemented by a7de9d85c33648fe109721a4509496933b2828a9 on top of the above) starts tracking managed resources as soon as they get locked. We also keep textures that haven't been uploaded yet in the list and just upload those after 256 (again arbitrary) frames. This has the advantage that we don't waste memory on textures that got locked but never get used.

The commits after that do the same thing with buffers.

It works very well in Witcher 2 in my very limited testing. 550 -> 54 MB. So less than 10% of what it used before.
image

I hope that it works with broken games that keep the map pointer around if we make the number of frames it takes before we free the managed buffer large enough. Especially given that I update it in AddDirtyRect & dont free buffers that need to get uploaded.
If it does not, I'd be inclined to add an app option to disable the cleanup.

Also, right now I've disabled all this on 64bit builds. Not sure about that, unnecessarily wasting memory isn't great in 64bit games either even if it shouldnt really be a problem.

### DIFF
--- a/src/d3d9/d3d9_common_buffer.cpp
+++ b/src/d3d9/d3d9_common_buffer.cpp
@@ -18,6 +18,10 @@ namespace dxvk {
       m_dirtyRange = D3D9Range(0, m_desc.Size);
   }
 
+  D3D9CommonBuffer::~D3D9CommonBuffer() {
+    m_parent->RemoveManagedBuffer(this);
+  }
+
 
   HRESULT D3D9CommonBuffer::Lock(
           UINT   OffsetToLock,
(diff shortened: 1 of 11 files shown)


K0bin:
This really doesn't work well with Borderlands 2 (and likely other UE3 games). It seems to create textures for texture streaming, reuse them after some time (long enough to trigger the cleanup) and then throw them away and repeat the process. Thus we get lots of hitches with this. :(

PR #2516, opened 2022-02-23
Proton Shared Texture Resources Implementation

Guy1524:
This Pull Request implements shared texture resources for d3d9 and d3d11, using some Proton-Specific extensions to store texture metadata alongside the shared memory handles that back them.  The goal here is just to get the code reviewed if needed, not to get it upstream.

The first commit is a bit of a megacommit, so if it proves a pain to review I'll split it up.

Link to Wine branch for Proton: https://github.com/Guy1524/wine/commits/shared-resources

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -23,6 +23,8 @@
 #include "d3d11_texture.h"
 #include "d3d11_video.h"
 
+#include "../util/util_shared_res.h"
+
 namespace dxvk {
   
   constexpr uint32_t D3D11DXGIDevice::DefaultFrameLatency;
@@ -207,7 +209,7 @@ namespace dxvk {
       return S_FALSE;
     
     try {
-      Com<D3D11Texture2D> texture = new D3D11Texture2D(this, &desc);
+      Com<D3D11Texture2D> texture = new D3D11Texture2D(this, &desc, nullptr);
       m_initializer->InitTexture(texture->GetCommonTexture(), pInitialData);
       *ppTexture2D = texture.ref();
       return S_OK;
@@ -1388,10 +1390,7 @@ namespace dxvk {
           HANDLE      hResource,
           REFIID      ReturnedInterface,
           void**      ppResource) {
-    InitReturnPtr(ppResource);
-    
-    Logger::err("D3D11Device::OpenSharedResource: Not implemented");
-    return E_NOTIMPL;
+    return OpenSharedResourceGeneric(true, hResource, ReturnedInterface, ppResource);
   }
   
   
@@ -1399,10 +1398,7 @@ namespace dxvk {
           HANDLE      hResource,
           REFIID      ReturnedInterface,
           void**      ppResource) {
-    InitReturnPtr(ppResource);
-    
-    Logger::err("D3D11Device::OpenSharedResource1: Not implemented");
-    return E_NOTIMPL;
+    return OpenSharedResourceGeneric(false, hResource, ReturnedInterface, ppResource);
   }
 
   
@@ -2257,6 +2253,62 @@ namespace dxvk {
     }
 
     return ~0u;
+  }
+
+
+  HRESULT D3D11Device::OpenSharedResourceGeneric(
+          BOOL        isKmtResource,
+          HANDLE      hResource,
+          REFIID      ReturnedInterface,
+          void**      ppResource) {
+    InitReturnPtr(ppResource);
+
+    if (ppResource == nullptr)
+      return S_FALSE;
+
+    HANDLE ntHandle = isKmtResource ? openKmtHandle(hResource) : hResource;
+
+    if (ntHandle == INVALID_HANDLE_VALUE) {
+      Logger::warn(str::format("D3D11Device::OpenSharedResourceGeneric: Handle not found: ", hResource));
+      return E_INVALIDARG;
+    }
+
+    DxvkSharedTextureMetadata metadata;
+    bool ret = getSharedMetadata(ntHandle, &metadata, sizeof(metadata), NULL);
+
+    if (isKmtResource)
+      ::CloseHandle(ntHandle);
+
+    if (!ret) {
+      Logger::warn("D3D11Device::OpenSharedResourceGeneric: Failed to get shared resource info for a texture");
+      return E_INVALIDARG;
+    }
+
+    D3D11_COMMON_TEXTURE_DESC d3d11Desc;
+
+    d3d11Desc.Width          = metadata.Width;
+    d3d11Desc.Height         = metadata.Height;
+    d3d11Desc.Depth          = 1,
+    d3d11Desc.MipLevels      = metadata.MipLevels;
+    d3d11Desc.ArraySize      = metadata.ArraySize;
+    d3d11Desc.Format         = metadata.Format;
+    d3d11Desc.SampleDesc     = metadata.SampleDesc;
+    d3d11Desc.Usage          = metadata.Usage;
+    d3d11Desc.BindFlags      = metadata.BindFlags;
+    d3d11Desc.CPUAccessFlags = metadata.CPUAccessFlags;
+    d3d11Desc.MiscFlags      = metadata.MiscFlags;
+    d3d11Desc.TextureLayout  = metadata.TextureLayout;
+
+    // Only 2D textures may be shared
+    try {
+      const Com<D3D11Texture2D> texture = new D3D11Texture2D(this, &d3d11Desc, hResource);
+      texture->QueryInterface(ReturnedInterface, ppResource);
+      return S_OK;
+    }
+    catch (const DxvkError& e) {
+      Logger::err(e.message());
+      return E_INVALIDARG;
+    }
   }
 
 
--- a/src/d3d11/d3d11_texture.cpp
+++ b/src/d3d11/d3d11_texture.cpp
@@ -2,6 +2,8 @@
 #include "d3d11_gdi.h"
 #include "d3d11_texture.h"
 
+#include "../util/util_shared_res.h"
+
 namespace dxvk {
   
   D3D11CommonTexture::D3D11CommonTexture(
@@ -10,7 +12,8 @@ namespace dxvk {
     const D3D11_COMMON_TEXTURE_DESC*  pDesc,
           D3D11_RESOURCE_DIMENSION    Dimension,
           DXGI_USAGE                  DxgiUsage,
-          VkImage                     vkImage)
+          VkImage                     vkImage,
+          HANDLE                      hSharedHandle)
   : m_interface(pInterface), m_device(pDevice), m_dimension(Dimension), m_desc(*pDesc), m_dxgiUsage(DxgiUsage) {
     DXGI_VK_FORMAT_MODE   formatMode   = GetFormatMode();
     DXGI_VK_FORMAT_INFO   formatInfo   = m_device->LookupFormat(m_desc.Format, formatMode);
@@ -37,6 +40,22 @@ namespace dxvk {
     imageInfo.layout          = VK_IMAGE_LAYOUT_GENERAL;
     imageInfo.initialLayout   = VK_IMAGE_LAYOUT_UNDEFINED;
     imageInfo.shared          = vkImage != VK_NULL_HANDLE;
+
+    // Normalise hSharedhandle to INVALID_HANDLE_VALUE to allow passing in nullptr
+    if (hSharedHandle == nullptr)
+      hSharedHandle = INVALID_HANDLE_VALUE;
+
+    if (m_desc.MiscFlags & (D3D11_RESOURCE_MISC_SHARED|D3D11_RESOURCE_MISC_SHARED_NTHANDLE)) {
+      if (m_desc.MiscFlags & D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX)
+        Logger::warn("D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX: not supported.");
+
+      imageInfo.shared = true;
+      imageInfo.sharing.mode = hSharedHandle == INVALID_HANDLE_VALUE ? DxvkSharedHandleMode::Export : DxvkSharedHandleMode::Import;
+      imageInfo.sharing.type = m_desc.MiscFlags & D3D11_RESOURCE_MISC_SHARED_NTHANDLE
+        ? VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT
+        : VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT_BIT;
+      imageInfo.sharing.handle = hSharedHandle;
+    }
 
     if (!pDevice->GetOptions()->disableMsaa)
       DecodeSampleCount(m_desc.SampleDesc.Count, &imageInfo.sampleCount);
@@ -175,7 +194,7 @@ namespace dxvk {
     // We must keep LINEAR images in GENERAL layout, but we
     // can choose a better layout for the image based on how
     // it is going to be used by the game.
-    if (imageInfo.tiling == VK_IMAGE_TILING_OPTIMAL && !isMultiPlane)
+    if (imageInfo.tiling == VK_IMAGE_TILING_OPTIMAL && !isMultiPlane && imageInfo.sharing.mode == DxvkSharedHandleMode::None)
       imageInfo.layout = OptimizeLayout(imageInfo.usage);
 
     // For some formats, we need to enable sampled and/or
@@ -210,6 +229,9 @@ namespace dxvk {
       m_image = m_device->GetDXVKDevice()->createImage(imageInfo, memoryProperties);
     else
       m_image = m_device->GetDXVKDevice()->createImageFromVkImage(imageInfo, vkImage);
+
+    if (imageInfo.sharing.mode == DxvkSharedHandleMode::Export)
+      ExportImageInfo();
   }
   
   
@@ -410,7 +432,7 @@ namespace dxvk {
 
     // Use the maximum possible mip level count if the supplied
     // mip level count is either unspecified (0) or invalid
-    const uint32_t maxMipLevelCount = pDesc->SampleDesc.Count <= 1
+    const uint32_t maxMipLevelCount = (pDesc->SampleDesc.Count <= 1)
       ? util::computeMipLevelCount({ pDesc->Width, pDesc->Height, pDesc->Depth })
       : 1u;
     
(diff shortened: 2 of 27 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -427,7 +427,7 @@ namespace dxvk {
       return D3DERR_INVALIDCALL;
 
     try {
-      const Com<D3D9Texture2D> texture = new D3D9Texture2D(this, &desc);
```
K0bin, line 430: pSharedHandle can be a regular void Pointer here. Basically just move up the if in line 434 and use that to decide whether or not you pass the handle to the constructor.

> In Windows Vista CreateTexture can create a texture from a system memory pointer allowing the application more flexibility over the use, allocation and deletion of the system memory. For example, an application could pass a GDI system memory bitmap pointer and get a Direct3D texture interface around it. Using a system memory pointer with CreateTexture has the following restrictions.

> The pitch of the texture must be equal to the width multiplied by the number of bytes per pixel.
    Only textures with a single mipmap level are supported. The Levels argument must be 1.
    The Pool argument must be D3DPOOL_SYSTEMMEM.
    The pSharedHandle argument must be a valid pointer to a buffer that can hold the system memory point; *pSharedHandle must be a valid pointer to system memory with a size in bytes of texture width * texture height * bytes per pixel of the texture format.

Algebro7:
I'm currently debugging a game that crashes and tested this branch out against it. Before adding these commits I was getting:

```
err:   D3D11DXGIResource::GetSharedHandle: Stub
```

After building with those commits I'm still crashing but I'm getting the following error instead:

```
err:   Failed to create shared resource: VK_KHR_EXTERNAL_MEMORY_WIN32 not supported
```

I'm not sure if this information is helpful or even related but let me know if there's anything else I can do to help. Thanks!

src/d3d9/d3d9_device.cpp
Guy1524: Should be resolved

Guy1524:
Algebro7: You need to wine patches linked in the PR description

Algebro7:
> Algebro7: You need to wine patches linked in the PR description

Ah OK, thanks, I misunderstood the description. So this feature will only work for Proton and not wine-staging?

K0bin:
The patches aren't in Proton either right now.

src/dxvk/dxvk_memory.cpp
```diff
@@ -563,4 +563,59 @@ namespace dxvk {
     }
   }
 
+
+  bool DxvkMemoryAllocator::canShareImage(const VkImageCreateInfo&  createInfo, const DxvkSharedHandleInfo& sharingInfo) const {
```
doitsujin, line 567: why ist his part of DxvkMemoryAllocator and not DxvkImage? This seems to makes zero sense design-wise.

src/dxvk/dxvk_memory.h
```diff
@@ -17,8 +17,26 @@ namespace dxvk {
...
+   * \brief Shared handle info
+   *
+   * The shared resource information for a given resource.
+   * IMPORT:     type = [handle type], handle = [shared handle]
+   * EXPORT:     type = [handle type], handle = nullptr
+   * NOT SHARED: type = [VK_EXTERNAL_MEMORY_HANDLE_TYPE_FLAG_BITS_MAX_ENUM], handle = anything
+   */
+  struct DxvkSharedHandleInfo {
+    VkExternalMemoryHandleTypeFlagBits type   = VK_EXTERNAL_MEMORY_HANDLE_TYPE_FLAG_BITS_MAX_ENUM;
+    HANDLE                             handle = INVALID_HANDLE_VALUE;
```
doitsujin, line 32: Hardcoding win32 types here is going to cause a lot of pain and suffering for dxvk-native work, this should really be a union type with `HANDLE` being `#ifdef`'d out based on whether we're compiling for windows. Perhaps add a `void* dummy;` field or something to make it compile on non-win32.

All shared handle interactions in the backend should be built around that so we can easily ifdef out the win32 parts and perhaps support something like host memory import, should it become necessary.

It's not necessary to make the whole thing compile on native straight away, but we shouldn't make it harder than necessary.

src/util/util_shared_res.h
```diff
@@ -0,0 +1,29 @@
...
+
+    struct DxvkSharedMetadata {
+      enum class DxvkSharedMetadataSourceApi{
+        D3D9,
+        D3D11,
+      } sourceApi;
+
+      union {
+        D3D9_COMMON_TEXTURE_DESC d3d9;
+        D3D11_COMMON_TEXTURE_DESC d3d11;
```
doitsujin, line 25: Can we just make this a new struct instead and translate to and from it in the D3D11 and D3D9 frontends, respectively, so that we won't end up with a combinatorial explosion of metadata translations? D3D12 could share that format.

This could be pretty much a copy of `D3D11_COMMON_TEXTURE_DESC`, except without the `Depth` field since these APIs do not allow sharing 3D textures. If we ever have to implement D3D11<->Vulkan/D3D12 sharing, we'll have to make significant changes to the entire thing anyway, so making this a union is not immediately beneficial.

src/d3d11/d3d11_device.cpp
```diff
@@ -2260,6 +2258,46 @@ namespace dxvk {
   }
 
 
+  HRESULT D3D11Device::OpenSharedResourceGeneric(
+      BOOL        nt,
```
doitsujin, line 2262: `nt` should a) have a more descriptive name and b) perhaps be a template parameter instead, so we have more options to use things like `if constexpr`.

src/d3d11/d3d11_device.cpp
```diff
@@ -1388,21 +1390,17 @@ namespace dxvk {
           HANDLE      hResource,
           REFIID      ReturnedInterface,
           void**      ppResource) {
-    InitReturnPtr(ppResource);
-    
-    Logger::err("D3D11Device::OpenSharedResource: Not implemented");
-    return E_NOTIMPL;
+    Logger::info("D3D11Device::OpenSharedResource");
```
doitsujin, line 1393: We don't normally do this.

src/d3d11/d3d11_device.cpp
```diff
@@ -1388,21 +1390,17 @@ namespace dxvk {
...
   
   HRESULT STDMETHODCALLTYPE D3D11Device::OpenSharedResource1(
           HANDLE      hResource,
           REFIID      ReturnedInterface,
           void**      ppResource) {
-    InitReturnPtr(ppResource);
-    
-    Logger::err("D3D11Device::OpenSharedResource1: Not implemented");
-    return E_NOTIMPL;
+    Logger::info("D3D11Device::OpenSharedResource1");
```
doitsujin, line 1402: Same here.

src/d3d11/d3d11_texture.cpp
```diff
@@ -408,9 +442,12 @@ namespace dxvk {
     if (pDesc->MiscFlags & (D3D11_RESOURCE_MISC_TILE_POOL | D3D11_RESOURCE_MISC_TILED))
       return E_INVALIDARG;
 
+    // Shared resources may only have 1 mip. Native clamps this.
```
doitsujin, line 445: Is this really true? The documentation for [Extended Resource Sharing](https://docs.microsoft.com/en-us/windows/win32/direct3d11/direct3d-11-1-features#extended-support-for-shared-texture2d-resources) disagrees here and says that more than one mip is allowed.

src/d3d11/d3d11_texture.h
```diff
@@ -58,7 +58,8 @@ namespace dxvk {
       const D3D11_COMMON_TEXTURE_DESC*  pDesc,
             D3D11_RESOURCE_DIMENSION    Dimension,
             DXGI_USAGE                  DxgiUsage,
-            VkImage                     vkImage);
+            VkImage                     vkImage,
+            HANDLE                      hSharedHandle = INVALID_HANDLE_VALUE);
```
doitsujin, line 62: I'd like to avoid default parameter values.

src/d3d11/d3d11_texture.h
```diff
@@ -579,7 +580,8 @@ namespace dxvk {
     
     D3D11Texture2D(
             D3D11Device*                pDevice,
-      const D3D11_COMMON_TEXTURE_DESC*  pDesc);
+      const D3D11_COMMON_TEXTURE_DESC*  pDesc,
+            HANDLE                      hSharedHandle = INVALID_HANDLE_VALUE);
```
doitsujin, line 584: Same here.

src/d3d11/d3d11_texture.cpp
```diff
@@ -38,6 +41,17 @@ namespace dxvk {
     imageInfo.initialLayout   = VK_IMAGE_LAYOUT_UNDEFINED;
     imageInfo.shared          = vkImage != VK_NULL_HANDLE;
 
+    if (m_desc.MiscFlags & (D3D11_RESOURCE_MISC_SHARED|D3D11_RESOURCE_MISC_SHARED_NTHANDLE)) {
```
doitsujin, line 44: We should always keep shared images in `GENERAL` layout as well, so the `OptimizeLayout` call below needs to be conditional on the image not being shared.

doitsujin (changes requested):

src/dxvk/dxvk_memory.cpp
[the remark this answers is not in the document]
Guy1524: I wasn't sure how to access the list of enabled extensions from a DxvkImage function, as it only has the function pointer table, not the DxvkDevice object which holds that list.

src/dxvk/dxvk_image.cpp
```diff
@@ -83,10 +94,35 @@ namespace dxvk {
...
+      dedMemoryAllocInfo.pNext = &importInfo;
+    }
+
     m_vkd->vkGetImageMemoryRequirements2(
       m_vkd->device(), &memReqInfo, &memReq);
 
+    if (m_shared) {
+      dedicatedRequirements.prefersDedicatedAllocation  = VK_TRUE;
+      dedicatedRequirements.requiresDedicatedAllocation = VK_TRUE;
+    }
```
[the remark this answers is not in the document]
Guy1524: I think the spec specifies it as a "may"

https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkMemoryDedicatedRequirementsKHR.html

> requiresDedicatedAllocation may be VK_TRUE if the pNext chain of [VkImageCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkImageCreateInfo.html) for the call to [vkCreateImage](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkCreateImage.html) used to create the image being queried included a [VkExternalMemoryImageCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkExternalMemoryImageCreateInfo.html) structure, and any of the handle types specified in [VkExternalMemoryImageCreateInfo](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/VkExternalMemoryImageCreateInfo.html)::handleTypes requires dedicated allocation, as reported by [vkGetPhysicalDeviceImageFormatProperties2](https://www.khronos.org/registry/vulkan/specs/1.3-extensions/man/html/vkGetPhysicalDeviceImageFormatProperties2.html) in VkExternalImageFormatProperties::externalMemoryProperties.externalMemoryFeatures. Otherwise, requiresDedicatedAllocation will be VK_FALSE.

src/dxvk/dxvk_image.cpp
```diff
@@ -42,6 +42,17 @@ namespace dxvk {
     info.queueFamilyIndexCount = 0;
     info.pQueueFamilyIndices   = nullptr;
     info.initialLayout         = createInfo.initialLayout;
+
+    m_shared = memAlloc.canShareImage(info, createInfo.sharing);
```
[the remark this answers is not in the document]
Guy1524: What do you mean by weird parameters, DXVK and Vulkan requirements for shared images may be different, and there could be an app that specifies the shared parameter for an image but doesn't actually use it.  Either way I don't have a strong opinion, leaving it up to you.

src/dxvk/dxvk_memory.cpp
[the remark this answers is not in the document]
doitsujin: You can change the image constructors (both for consistency) to take a DxvkDevice pointer instead.

src/dxvk/dxvk_image.cpp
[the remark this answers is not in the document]
doitsujin: Yeah, it's a good point especially when running on systems that don't support shared resources.

src/dxvk/dxvk_image.cpp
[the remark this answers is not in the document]
Joshua-Ashton: Yes, so if the driver needs it (which they all do), it will set it.

src/d3d11/d3d11_device.cpp
Guy1524: Unfortunately I'm not familiar enough with c++ to know how the template suggestion would work.

src/dxvk/dxvk_memory.h
Guy1524: Let me know if the changes I made are enough on this front.

src/d3d9/d3d9_common_texture.cpp
```diff
@@ -39,20 +46,71 @@ namespace dxvk {
...
+              .MipLevels      = m_desc.MipLevels,
+              .ArraySize      = m_desc.ArraySize,
+              .Format         = dxgiFormat,
+              .SampleDesc     = {.Count = 1, .Quality = 0},
+              .Usage          = D3D11_USAGE_DEFAULT,
+              .BindFlags      = D3D11_BIND_SHADER_RESOURCE | D3D11_BIND_RENDER_TARGET,
+              .CPUAccessFlags = 0,
+              .MiscFlags      = D3D11_RESOURCE_MISC_SHARED,
+              .TextureLayout  = D3D11_TEXTURE_LAYOUT_UNDEFINED,
+            };
```
DadSchoorse, line 102: dxvk used C++17 so you can't use designated init sadly.

src/d3d11/d3d11_texture.cpp
```diff
@@ -210,6 +229,36 @@ namespace dxvk {
...
+        .MipLevels      = m_desc.MipLevels,
+        .ArraySize      = m_desc.ArraySize,
+        .Format         = m_desc.Format,
+        .SampleDesc     = m_desc.SampleDesc,
+        .Usage          = m_desc.Usage,
+        .BindFlags      = m_desc.BindFlags,
+        .CPUAccessFlags = m_desc.CPUAccessFlags,
+        .MiscFlags      = m_desc.MiscFlags,
+        .TextureLayout  = m_desc.TextureLayout,
+      };
```
DadSchoorse, line 253: no designated initializers

src/d3d11/d3d11_device.cpp
```diff
@@ -2260,6 +2256,56 @@ namespace dxvk {
...
+      .MipLevels      = metadata.MipLevels,
+      .ArraySize      = metadata.ArraySize,
+      .Format         = metadata.Format,
+      .SampleDesc     = metadata.SampleDesc,
+      .Usage          = metadata.Usage,
+      .BindFlags      = metadata.BindFlags,
+      .CPUAccessFlags = metadata.CPUAccessFlags,
+      .MiscFlags      = metadata.MiscFlags,
+      .TextureLayout  = metadata.TextureLayout,
+    };
```
DadSchoorse, line 2294: no designated initializers

src/d3d11/d3d11_texture.cpp
```diff
@@ -175,7 +194,7 @@ namespace dxvk {
     // We must keep LINEAR images in GENERAL layout, but we
     // can choose a better layout for the image based on how
     // it is going to be used by the game.
-    if (imageInfo.tiling == VK_IMAGE_TILING_OPTIMAL && !isMultiPlane)
+    if (imageInfo.tiling == VK_IMAGE_TILING_OPTIMAL && !isMultiPlane && !imageInfo.shared)
```
doitsujin, line 197: `imageInfo.shared` is used for some internal stuff as well such as swap chain images, where we don't want this, can we have a separate conditional for this?

src/d3d11/d3d11_texture.cpp
```diff
@@ -210,6 +229,36 @@ namespace dxvk {
       m_image = m_device->GetDXVKDevice()->createImage(imageInfo, memoryProperties);
     else
       m_image = m_device->GetDXVKDevice()->createImageFromVkImage(imageInfo, vkImage);
+
+    if (imageInfo.sharing.mode == DxvkSharedHandleMode::Export) {
```
doitsujin, line 233: The code below should probably be a separate function

src/d3d9/d3d9_common_texture.cpp
```diff
@@ -39,20 +46,72 @@ namespace dxvk {
...
           m_desc.MipLevels = 1;
-          m_image = CreatePrimaryImage(ResourceType, false);
+          m_image = CreatePrimaryImage(ResourceType, false, pSharedHandle);
         }
         else
           throw e;
       }
 
+      if (pSharedHandle) {
+        if (*pSharedHandle == nullptr) {
```
doitsujin, line 64: The code below should probably be a separate function as well

src/d3d9/d3d9_device.cpp
```diff
@@ -461,6 +463,9 @@ namespace dxvk {
     if (unlikely(ppVolumeTexture == nullptr))
       return D3DERR_INVALIDCALL;
 
+    if (pSharedHandle)
+        Logger::err("CreateVolumeTexture: pSharedHandle specified");
```
doitsujin, line 467: Is this an API usage error or are we not implementing something that we should support?

src/d3d9/d3d9_device.cpp
```diff
@@ -506,6 +511,9 @@ namespace dxvk {
     if (unlikely(ppCubeTexture == nullptr))
       return D3DERR_INVALIDCALL;
 
+    if (pSharedHandle)
+        Logger::err("CreateCubeTexture: pSharedHandle specified");
```
doitsujin, line 515: Same question here

src/d3d9/d3d9_device.cpp
```diff
@@ -550,6 +558,9 @@ namespace dxvk {
     if (unlikely(ppVertexBuffer == nullptr))
       return D3DERR_INVALIDCALL;
 
+    if (pSharedHandle)
+        Logger::err("CreateVertexBuffer: pSharedHandle specified");
```
doitsujin, line 562: and here

src/d3d9/d3d9_device.cpp
```diff
@@ -586,6 +597,9 @@ namespace dxvk {
     if (unlikely(ppIndexBuffer == nullptr))
       return D3DERR_INVALIDCALL;
 
+    if (pSharedHandle)
+        Logger::err("CreateIndexBuffer: pSharedHandle specified");
```
doitsujin, line 601: and here

src/d3d9/d3d9_common_texture.cpp
```diff
@@ -26,6 +29,10 @@ namespace dxvk {
       for (uint32_t i = 0; i < subresources; i++) {
         SetNeedsUpload(i, true);
       }
+      if (pSharedHandle) {
+        Logger::warn("D3D9CommonTexture: Shared resources must use D3DPOOL_DEFAULT");
+        pSharedHandle = nullptr;
```
doitsujin, line 34: Shouldn't we fail texture creation instead?

src/d3d9/d3d9_common_texture.cpp
```diff
@@ -230,6 +289,14 @@ namespace dxvk {
     imageInfo.tiling          = VK_IMAGE_TILING_OPTIMAL;
     imageInfo.layout          = VK_IMAGE_LAYOUT_GENERAL;
     imageInfo.shared          = m_desc.IsBackBuffer;
+    if (pSharedHandle) {
+      imageInfo.sharing.type = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_KMT_BIT;
+      if (*pSharedHandle) {
+        imageInfo.sharing.mode = *pSharedHandle == INVALID_HANDLE_VALUE ? DxvkSharedHandleMode::Export : DxvkSharedHandleMode::Import;
+        imageInfo.sharing.handle = *pSharedHandle;
+      }
+      // TODO: validate metadata?
```
doitsujin, line 298: `imageInfo.shared = true;`?

Also need to keep shared images in `GENERAL` layout to be consistent with D3D11.

doitsujin (changes requested):

src/d3d9/d3d9_device.cpp
Guy1524: My impression from MSDN is that sharing these types of resources within D3D9 is possible, but not to D3D11.  My idea here was to just put these log messages here until we find a game that actually uses this feature, instead of silently ignoring the parameter.

src/d3d9/d3d9_common_texture.cpp
Guy1524: Yeah, looks like we should.

src/d3d9/d3d9_device.cpp
doitsujin: Can you clarify the messages in that case? Something like "Shared volume textures not supported".

doitsujin (approved):

doitsujin:
Looks good now, I'd like to test this a bit before merging.

doitsujin:
Merged manually.

PR #2517, opened 2022-02-24
[d3d9] Fix sysmem readback

K0bin:
Upon further inspection, it turns out that 29d88127090997dcac329d35c45bcacf9ae40d37 is wrong. It prevents DXVK from synchronizing the GPU if the texture that gets locked has been used with GetFrontBufferData or GetRendertargetData.


mupuf:
Thanks for fixing this! Got 4 runs of dxvk ci with all the dx9 traces creating black frames:D

mupuf:
When I am happy with the visualization for pre-merge run results, how about I start automatically a new run for every pull request and write the result as a comment?

I want to add support for github actions in our CI system, but for now, we run in Gitlab so it requires this stupid dance :s

Issue #2417, opened 2021-12-26
Saints Row 2 DXVK Crash RTX 3080 Windows 10 x64


mrchrismania:
i added the x32 d3d9.dll and dxgi into the saints row 2 folder. started the game and it crashes on load and sometimes i can get into the main menu but as soon as i start a new game it crashes. heres a log file

[SR2_pc_d3d9.log](https://github.com/doitsujin/dxvk/files/7776435/SR2_pc_d3d9.log)

Windows 10 Version 21H1
OS Build 19043.1415
NVIDIA Driver version 497.09

K0bin:
Logs indicate this is another case of #1318

mrchrismania:
> Logs indicate this is another case of #1318

is there anything i have to do on my end or is it an issue with the DXVK Wrapper?

K0bin:
Judging by the driver version from the logs, I think you're using Windows (that would've been useful info to add to the original issue btw...). In that case run the game through one of those 3GB address space patchers and hope for the best.

mrchrismania:
> Judging by the driver version from the logs, I think you're using Windows (that would've been useful info to add to the original issue btw...). In that case run the game through one of those 3GB address space patchers and hope for the best.


how do i run the game through the 3GB Space patcher? i tried an NTCore 4GB Patcher for Saints Row 2 but it came up with an error when i tried to launch it.

lextra2:
try this one?
https://www.techpowerup.com/forums/threads/large-address-aware.112556/

mrchrismania:
> try this one? https://www.techpowerup.com/forums/threads/large-address-aware.112556/

no luck just comes up with an error saying. Application load error 3:0000065432

K0bin:
Closing this because it's tracked in 1318.

qinlili23333:
> @qinlili23333 Hope it's okay i ping you here as i didn't want to start a talk in the PR. Do you always crash when starting a saints row 2 game?
> 
> When i was testing #1814 on linux i didn't notice any issues and i loaded into the game a lot of times. Tho i did not actually play too much around.
> 
> Edit: I guess proton enables Large address aware by default now that i think about it so maybe that's why i experience a crash. Haven't looked at the actual memory consumption.


[...]

PR #2524, opened 2022-03-04
Use memory mapped files for managed textures

K0bin:
In order to finally fix some of our address space problems, I took a page out of Gallium Nines book.

Managed textures now use memory mapped files to store the internal data. The data is suballocated from 4MB (totally arbitrary) memory mapped files. This allows us to map and unmap that memory as we see fit. This only impacts managed textures, systemmem textures still use the DXVKBuffer to store their data because they can be used with `GetRenderTargetData` to read back data from the GPU. I don't think those are a problem anyway. Buffers are also unchanged. It's possible to extend this to buffers but I'd like to get some feedback first.

Unmapping follows a simple heuristic: We track the frame in which a texture was last used with `LockRect`, `AddDirtyRect` or when it was actually uploaded. If that was more than 16 frames ago or we've crossed the threshold of 512MB for mapped textures, we unmap. I hope that's good enough for games that keep the pointer around after calling Unlock. If it's not we can either try to tweak the heuristic, come up with something smarter or just disable it using a config option. I'd rather not punish decent D3D9 games that just happen to use too many high res managed textures because some do stupid things that aren't supposed to be allowed.

All of this is disabled in 64 bit builds of course.

The branch is basically a prototype. It works pretty well with Borderlands 2 and Witcher 2, both of which use a ton of managed textures. I'd like to get some feedback on this. Should we also extend this to buffers? Do you have any other ideas for the unmap heuristic?

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -7434,9 +7434,12 @@ namespace dxvk {
   void D3D9DeviceEx::UnmapTextures() {
     // Will only be called inside the device lock
 #ifdef D3D9_USE_MEM_FILE_FOR_MANAGED
+    if (m_d3d9Options.unmapDelay == 0)
+      return;
+
     const bool force = m_memoryAllocator.MappedMemory() > 512 << 20;
     for (auto iter = m_mappedTextures.begin(); iter != m_mappedTextures.end();) {
-      const bool mappingBufferUnused = (m_frameCounter - (*iter)->GetMappingFrame() > 16 || force) && !(*iter)->IsAnySubresourceLocked();
+      const bool mappingBufferUnused = (m_frameCounter - (*iter)->GetMappingFrame() > uint32_t(m_d3d9Options.unmapDelay) || force) && !(*iter)->IsAnySubresourceLocked();
       if (!mappingBufferUnused) {
          iter++;
         continue;
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -39,18 +39,19 @@ namespace dxvk {
           HWND                   hFocusWindow,
           DWORD                  BehaviorFlags,
           Rc<DxvkDevice>         dxvkDevice)
-    : m_parent         ( pParent )
-    , m_deviceType     ( DeviceType )
-    , m_window         ( hFocusWindow )
-    , m_behaviorFlags  ( BehaviorFlags )
-    , m_adapter        ( pAdapter )
-    , m_dxvkDevice     ( dxvkDevice )
-    , m_shaderModules  ( new D3D9ShaderModuleSet )
-    , m_d3d9Options    ( dxvkDevice, pParent->GetInstance()->config() )
-    , m_multithread    ( BehaviorFlags & D3DCREATE_MULTITHREADED )
-    , m_isSWVP         ( (BehaviorFlags & D3DCREATE_SOFTWARE_VERTEXPROCESSING) ? true : false )
-    , m_csThread       ( dxvkDevice, dxvkDevice->createContext() )
-    , m_csChunk        ( AllocCsChunk() ) {
+    : m_parent          ( pParent )
+    , m_deviceType      ( DeviceType )
+    , m_window          ( hFocusWindow )
+    , m_behaviorFlags   ( BehaviorFlags )
+    , m_adapter         ( pAdapter )
+    , m_dxvkDevice      ( dxvkDevice )
+    , m_memoryAllocator ( )
+    , m_shaderModules   ( new D3D9ShaderModuleSet )
+    , m_d3d9Options     ( dxvkDevice, pParent->GetInstance()->config() )
+    , m_multithread     ( BehaviorFlags & D3DCREATE_MULTITHREADED )
+    , m_isSWVP          ( (BehaviorFlags & D3DCREATE_SOFTWARE_VERTEXPROCESSING) ? true : false )
+    , m_csThread        ( dxvkDevice, dxvkDevice->createContext() )
+    , m_csChunk         ( AllocCsChunk() ) {
     // If we can SWVP, then we use an extended constant set
     // as SWVP has many more slots available than HWVP.
     bool canSWVP = CanSWVP();
@@ -4142,10 +4143,6 @@ namespace dxvk {
 
     auto& desc = *(pResource->Desc());
 
-    bool alloced = pResource->CreateBufferSubresource(Subresource);
-
-    const Rc<DxvkBuffer> mappedBuffer = pResource->GetBuffer(Subresource);
-
     auto& formatMapping = pResource->GetFormatMapping();
 
     const DxvkFormatInfo* formatInfo = formatMapping.IsValid()
@@ -4200,12 +4197,15 @@ namespace dxvk {
     bool needsReadback = pResource->NeedsReachback(Subresource) || renderable;
     pResource->SetNeedsReadback(Subresource, false);
 
-    DxvkBufferSliceHandle physSlice;
+    void* mapPtr;
 
-    if (Flags & D3DLOCK_DISCARD) {
+    if ((Flags & D3DLOCK_DISCARD) && !pResource->DoesStagingBufferUploads(Subresource)) {
       // We do not have to preserve the contents of the
       // buffer if the entire image gets discarded.
-      physSlice = pResource->DiscardMapSlice(Subresource);
+      pResource->CreateBufferSubresource(Subresource);
+      DxvkBufferSliceHandle physSlice = pResource->DiscardMapSlice(Subresource);
+      mapPtr = physSlice.mapPtr;
+      const Rc<DxvkBuffer> mappedBuffer = pResource->GetBuffer(Subresource);
 
       EmitCs([
         cImageBuffer = std::move(mappedBuffer),
@@ -4214,7 +4214,8 @@ namespace dxvk {
         ctx->invalidateBuffer(cImageBuffer, cBufferSlice);
       });
     } else {
-      physSlice = pResource->GetMappedSlice(Subresource);
+      const bool alloced = pResource->AllocLockingData(Subresource);
+      mapPtr = pResource->GetLockingData(Subresource);
 
       // We do not need to wait for the resource in the event the
       // calling app promises not to overwrite data that is in use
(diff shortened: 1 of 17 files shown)


src/d3d9/d3d9_mem.h
```diff
@@ -0,0 +1,98 @@
...
+      void FreeChunk(D3D9MemoryChunk* Chunk);
+      void NotifyMapped(uint32_t Size);
+      void NotifyUnmapped(uint32_t Size);
+      void NotifyFreed(uint32_t Size);
+      uint32_t MappedMemory();
+      uint32_t UsedMemory();
+
+    private:
+      dxvk::mutex m_mutex;
+      std::vector<D3D9MemoryChunk*> m_chunks;
```
DadSchoorse, line 92: Imo this should be a `std::vector<std::unique_ptr<D3D9MemoryChunk>>`, then you won't have to manually delete the chunk after removing it from `m_chunks`.

src/d3d9/d3d9_mem.cpp
```diff
@@ -0,0 +1,269 @@
...
+      return D3D9Memory(this, offset, size);
+
+    return {};
+  }
+
+  void D3D9MemoryChunk::Free(D3D9Memory *Memory) {
+    if constexpr (!env::is32BitHostPlatform()) {
+      // In 64bit builds we use a 1:1 mapping between chunks and memory
+      free(m_ptr);
+      delete this;
```
DadSchoorse, line 166: I hate everything about this. You can't simply commit suicide, you don't even know that this was allocated with `new`.

DadSchoorse: Imo, call `m_allocator->FreeChunk(this);` and that could call delete. But even that's still a bit eww.

src/d3d9/d3d9_mem.h
```diff
@@ -0,0 +1,98 @@
...
+  class D3D9MemoryChunk;
+
+  struct D3D9MemoryRange {
+    uint32_t offset;
+    uint32_t length;
+  };
+
+  class D3D9MemoryChunk {
+    public:
+      D3D9MemoryChunk();
```
DadSchoorse, line 26: where is the default allocator used?

src/d3d9/d3d9_mem.h
```diff
@@ -0,0 +1,98 @@
...
+      D3D9MemoryChunk(D3D9MemoryAllocator* Allocator, uint32_t Size);
```
DadSchoorse, line 27: Imo, this should be private and D3D9MemoryAllocator a friend.

src/d3d9/d3d9_mem.h
```diff
@@ -0,0 +1,98 @@
...
+  class D3D9MemoryAllocator;
+  class D3D9Memory;
+  class D3D9MemoryChunk;
+
+  struct D3D9MemoryRange {
+    uint32_t offset;
+    uint32_t length;
+  };
+
+  class D3D9MemoryChunk {
```
DadSchoorse, line 24: This should probably delete copy and move constructor/assignment.

DadSchoorse (changes requested):
Sorry for the technicalities, I saw the `delete this;` and got very annoyed about lifetimes :frog:

K0bin:
Thanks for the review. I guess my lack of C++ experience is really showing here.

K0bin:
Addressed the feedback.

DadSchoorse:
Honestly I would prefer something like this https://github.com/DadSchoorse/dxvk/commit/7a800f4d443b7f360cf64cff784e48e4c9706749, which also has the added advantage that it doesn't break dxvk-native.

wils0nyan:
every time I try to merge this in and run Return of Reckoning (warhammer online) with the latest wine and mesa the launcher crashes straight away

wine: Unhandled page fault on read access to 0000000C at address 6244DDDC (thread 0110), starting debugger...

Unhandled exception: page fault on read access to 0x0000000c in 32-bit code (0x6244dddc).
Register dump:
 CS:0023 SS:002b DS:002b ES:002b FS:0063 GS:006b
 EIP:6244dddc ESP:079df7b0 EBP:00000000 EFLAGS:00010256(  R- --  I  Z-A-P- )
 EAX:00000003 EBX:079df8b8 ECX:00c77278 EDX:00000000
 ESI:00000003 EDI:00000000
Stack dump:
0x079df7b0:  00c6a338 00d0b940 00010000 00000000
0x079df7c0:  00000023 079df8b8 00000000 6a2d961e
0x079df7d0:  0ae20000 e7c18000 00010000 00d059b0
0x079df7e0:  00000000 079df8b8 079df8f8 62462fa2
0x079df7f0:  079df8b8 00000003 00000000 00000000
0x079df800:  626a8ac4 626a92ec 626a8ac4 626a8884
Backtrace:
=>0 0x6244dddc in d3d9 (+0xdddc) (0x00000000)
0x6244dddc d3d9+0xdddc: movl    0xc(%edx),%eax


includes even the frog patch

K0bin:
@wils0nyan Can you make an apitrace?

wils0nyan:
[ror-apitrace.tar.gz](https://github.com/doitsujin/dxvk/files/8193816/ror-apitrace.tar.gz)

ok here it is I'm not sure if this is useful

this is unrelated but even though the launcher loads without the unmap merge the box is still blacked out/flashes details every now and then

K0bin:
The launcher crash should be fixed.

src/d3d9/meson.build
```diff
@@ -38,7 +38,8 @@ d3d9_src = [
   'd3d9_names.cpp',
   'd3d9_swvp_emu.cpp',
   'd3d9_format_helpers.cpp',
-  'd3d9_hud.cpp'
+  'd3d9_hud.cpp',
+  'd3d9_mem.cpp'
```
DadSchoorse, line 42: Please add a trailing comma.

wils0nyan:
can confirm that the crash is gone but also have the following going:
git revert 4e6db51230d78354e2a86a23943c590f72d43c5e
git revert 14df2f7251002892436b3e2dafb8db19be1624d9
(because they supposedly fixed that pretty quickly in radv)

and merged:
shader_replacement
dogma-fix
unmap-managed

no crash all good so far

K0bin:
@wils0nyan What game are you talking about and why do you randomly revert and merge branches of mine?

The shader_replacement branch was just a debugging experiment and will do absolutely nothing for you.

wils0nyan:
> @wils0nyan What game are you talking about and why do you randomly revert and merge branches of mine?
> 
> The shader_replacement branch was just a debugging experiment and will do absolutely nothing for you. Those reverts will only break games unless you're using a couple of very specific commits of RADV.

I just try whatever I see usually. dxvk doesn't take very long to compile and I don't have to worry about rebooting or reloading xservers. I haven't noticed any problems doing those reverts with the latest mesa but i've only been playing warhammer online.

K0bin:
Actually, the fixes you reverted will even cause crashes in some cases. Please don't mindlessly revert commits.

wils0nyan:
> Actually, the fixes you reverted will even cause crashes in some cases. Please don't mindlessly revert commits.

ok but they are simple reverts really and all the commits do is disable a feature I can disable thru the conf file anyway.

K0bin:
> ok but they are simple reverts really and all the commits do is disable a feature I can disable thru the conf file anyway.

No, 4e6db51230d78354e2a86a23943c590f72d43c5e doesn't disable the feature. It fixes a crash. The feature will still get enabled with that commit reverted.

Also, why are you even disabling that? It's essentially free on newer RADV drivers and on anything else we only enable it when it's necessary.

K0bin:
Alternatively we could unmap based on LRU and only past a certain memory threshold?

K0bin:
I played through half of Mirrors Edge with this on the Steam Deck. It works great with texture streaming in UE3.

Blisto91:
Great work.
If you have any specific games that needs testing just say so

K0bin:
Basically anything that's 32bit and currently crashes.

TheIronWolfModding:
Hi,
If I understand correctly, this approach saves process' address space when MANAGED pool is used?

Is it possible to extend this approach to optionally do this for D3DPOOL_SYSTEMMEM?  There's a family of games that uses SYSTEMMEM texture for each DEFAULT one, perhaps because it does some CPU processing on them, sometimes.   If those textures could be pushed out of process' address space it would be a huge win there.   Thanks!

K0bin:
> Is it possible to extend this approach to optionally do this for D3DPOOL_SYSTEMMEM? There's a family of games that mirrors SYSTEMMEM texture for each DEFAULT one, perhaps because it does some CPU processing on them, sometimes. If those textures could be pushed out of process' address space it would be a huge win there.

It's definitely possible but a bit more annoying because you can write to SYSTEMMEM textures on the GPU. Can you open an issue for that and attach an apitrace? I can look into extending this for SYSTEMMEM after landing the initial version.

Which games are we talking about anyway?

TheIronWolfModding:
I am messing with a 32bit game called GTR2.  However, same or similar game engine is used by rFactor, Automobilista and dozen of other games, so I am _guessing_ they might benefit as well.

One important detail about those games is that they are semi-open.  They are not opensource, but they are by design made in such a way, that people can create their own cars/tracks, and customize nearly any aspect of the simulation.  There are plenty of tools and a lot of stuff is adjustable via text configuration files.

The stock GTR2 won't run out of memory.  But if you use some modern community content with highly detailed meshes and high resolution textures that's where the problem surfaces.  It is not possible to simulate real events running, say 67 cars, because game runs out of memory at 30 cars.

I am in a middle of debugging, but it seems that game creates sysmem texture, and it keeps them around (creating the address space pressure), and dynamically manages matching default textures.  I see some calls to LockRect/AddDirtyRect/UpdateTexture etc.  I do not yet understand why that is done that way, but I think at least some of the processing is on the CPU side.  Buffers contribute to memory pressure too, because even though they seem to go to default pool, they are DYNAMIC usage, so they seem to eat up process' address space as well, but not sure how big of a problem that is.

All that said, would issue with an apitrace still help?

K0bin:
Renamed some symbols from _MANAGED to something a bit more generic in case we ever decide to extend this to other pool types.

src/d3d9/d3d9_device.cpp
```diff
@@ -7374,4 +7377,59 @@ namespace dxvk {
     return m_csChunk->empty() ? m_csSeqNum : m_csSeqNum + 1;
   }
 
+
+  void* D3D9DeviceEx::MapTexture(D3D9CommonTexture* pTexture, UINT Subresource) {
+    // Will only be called inside the device lock
+    void *ptr = pTexture->GetLockingData(Subresource);
+
+#ifdef D3D9_USE_MEM_FILE_FOR_MANAGED
```
TheIronWolfModding, line 7385: this is missing the rename :)  I noticed while updating my fork...

qinlili23333:
Some games really work better now with this version, but cannot solve https://github.com/doitsujin/dxvk/issues/2417
Saints Row 2 now stuck at loading screen (while master branch build will directly crash)
I tested it on Windows 11 Preview (22621.1). Spec:i7-11800H, RTX3060 6G, 64G DDR4, 1440P internal display with another 2160P monitor plugged in

K0bin:
Superseded by #2663

PR #2572, opened 2022-03-30
Fix link with dxgi

duck-rh:
When building 1.10 with mingw 2.37 it fails with:
  /usr/bin/x86_64-w64-mingw32-ld: src/d3d11/d3d11.dll.p/d3d11_main.cpp.obj:./build-win64/../src/d3d11/d3d11_main.cpp:137: u    ndefined reference to `CreateDXGIFactory1'
Same with d3d10 and d3d10_1.

### DIFF
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -28,7 +28,7 @@ d3d10_deps = [ lib_d3dcompiler_43, lib_dxgi, dxbc_dep, dxvk_dep, d3d10_core_dep 
 
 d3d10_dll = shared_library('d3d10'+dll_ext, d3d10_main_src, d3d10_res,
   name_prefix         : '',
-  dependencies        : [ d3d10_deps ],
+  dependencies        : [ d3d10_deps, dxgi_dep ],
   include_directories : dxvk_include_path,
   install             : true,
   vs_module_defs      : 'd3d10'+def_spec_ext,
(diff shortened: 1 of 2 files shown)


duck-rh:
I built initially with 1.10.1 (I made a mistake in the BR) but then rebuilt 1.10.0 and got the same problem. It seems the only thing that changed between the two is meson from 0.61.x to 0.62.

Joshua-Ashton:
What is Mingw 2.37? That doesn't exist.

duck-rh:
@Joshua-Ashton sorry, that's the version of binutils-mingw-w64-x86-64, now g++-mingw-w64-x86-64-posix is 10.3.0

niobium93:
Neither of those version numbers is the version of mingw. The first is the version of the binutils that come with mingw, the second is the version of gcc that's being used with mingw.

duck-rh:
@niobium93 Sorry, not used to their version numbering. It says `Source: gcc-mingw-w64 (24.3)` so maybe that's the one. If not then I have no idea where this can be found but I guess that can still be useful.

duck-rh:
Well, still the right version. We have `mingw-w64` version `9.0.0` and this time that matches the latest version on https://www.mingw-w64.org/changelog/ so that must be it.

duck-rh:
@Joshua-Ashton @niobium93 any comment on the patch itself? I'm really not in my field here so I'm eager to get opinions.

stefson:
compile works perfectly fine with this setup: 

```
binutils-mingw-2.36.1
binutils-gcc-10.3.1_p20211126
mingw64-runtime-8.0.0
```

doitsujin (changes requested):
We already have `lib_dxgi` in the dependency lists for all of these; why doesn't that work? We shouldn't have to link specifically against our own DXGI.

Anyway, duplicate dependencies right now, more or less.

PR #2573, opened 2022-03-30
Only bind frame buffer attachments that are actually going to get used

K0bin:
When changing Resolution in Metal Gear Rising Revengeance, I ended up with a 1440p render target (the "implicit" one) and a 1080p depth stencil attachment which it creates itself. DXVK then picks the smallest size of all attachments (which is 1080p) as the render area and we end up with this: 
image

This is technically a game bug but it seems to work with WineD3D. To work around this, I unbind the depth buffer if unless either the depth test, stencil test or depth write is enabled.

I also do the same thing for color targets now because Dead Space has a render path that relies on that. When DF texture formats are supported, it decides to not use the NULL hack format as a color RT and instead binds a 1x1 color RT when rendering shadow maps. If we unbind that if the color write mask is 0, it works. Granted, we already fixed shadows in DS1 by forcing it to use the NULL format code path, but there might be other games that rely on this behavior.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1795,8 +1795,9 @@ namespace dxvk {
       return m_recorder->SetRenderState(State, Value);
 
     auto& states = m_state.renderStates;
+    DWORD old = states[State];
 
-    bool changed = states[State] != Value;
+    bool changed = old != Value;
 
     if (likely(changed)) {
       const bool oldClipPlaneEnabled = IsClipPlaneEnabled();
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -1880,18 +1880,30 @@ namespace dxvk {
           break;
 
         case D3DRS_COLORWRITEENABLE:
+          if (likely(m_state.renderTargets[0] != nullptr))
```
Joshua-Ashton, line 1883: This is per-channel, we should only dirty it if ``!!old != !!new``

I guess same goes for UpdateActiveRTs

src/d3d9/d3d9_device.cpp
```diff
@@ -5551,12 +5563,18 @@ namespace dxvk {
       else if (unlikely(sampleCount != rtImageInfo.sampleCount))
         continue;
 
+      if (m_state.renderStates[ColorWriteIndex(i)] == 0)
```
Joshua-Ashton, line 5566: !

Joshua-Ashton (changes requested):

src/d3d9/d3d9_device.cpp
K0bin: Why `!`? We want to skip the RT if the color write mask is 0.

K0bin: Ooooh, you don't want the explicit `== 0`.

src/d3d9/d3d9_device.cpp
```diff
@@ -1880,19 +1881,39 @@ namespace dxvk {
...
+          }
           m_flags.set(D3D9DeviceFlag::DirtyBlendState);
           break;
         case D3DRS_COLORWRITEENABLE1:
-          UpdateActiveRTs(1);
+          if (!old != !Value) {
+            if (likely(m_state.renderTargets[1] != nullptr))
+              m_flags.set(D3D9DeviceFlag::DirtyFramebuffer);
+
+            UpdateActiveRTs(1);
```
Joshua-Ashton, line 1897: This should be in the if likely block above.

src/d3d9/d3d9_device.cpp
```diff
@@ -1880,19 +1881,39 @@ namespace dxvk {
           break;
 
         case D3DRS_COLORWRITEENABLE:
-          UpdateActiveRTs(0);
+          if (!old != !Value) {
+            if (likely(m_state.renderTargets[0] != nullptr))
```
Joshua-Ashton, line 1885: We always have a 0th rendertarget so this unnecessary.

Joshua-Ashton (changes requested):

PR #2588, opened 2022-04-06
[d3d9] window messages hook improvements

### LINKED ISSUE #2569: [d3d9] Metal Gear Rising Revengeance: blank screen and freeze after selecting Full Screen
If you select "Full Screen" on the graphics options, the screen blanks and the system freezes.  This is happening with Proton 7.0 with DXVK: 1.9.4. Using WineD3d the problems doesn't happen (and with Proton 6.3-8 using  DXVK [1.9.2-13-g714ca482]). 

There is a Proton issue for this problem: https://github.com/ValveSoftware/Proton/issues/787

### Software information
Game: Metal Gear Rising: Revengeance
Settings: default

### System information
- GPU: GTX 970
- Driver: 510.60.02
- Wine version: Proton 7.0-1
- DXVK version: 1.9.4

### Apitrace file(s)
- [api trace](https://mega.nz/file/NttjHDyR#shx6ASjNO2Nfie6k_jbWWjeuKxRd2dLTN_aCSJYmlAw)

### Log files
- [d3d9.log](https://github.com/doitsujin/dxvk/files/8373930/METAL.GEAR.RISING.REVENGEANCE_d3d9.tar.gz)

Let me know if more information is needed.

gofman:
Supposed to fix #2569 (the part which locks the game during fullscreen state change). I could not reproduce exactly the lock but I observed endlessly bouncing window messages between dxvk resetting window size and the game setting it back which are likely the cause of hangs under other WMs / conditions, and this bouncing should be fixed by the PR.

Here is the squeeze of some testing I did (an addition to existing Wine d3d9 test): https://gist.github.com/gofman/b89c0f2154f6d87123f038941641df51

The testing I did here involved studying message sequences which are never the same between native and dxvk d3d9 on Windows, primarily because of the window style changes. Also, part of the sequence depends on asynchronous messages triggered by display mode change, thus some added waits without which the test may fail on native as well.

The absence of activate_processed check for deactivation event processing is not a typo, that's how I read those observed message sequences.

### DIFF
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -128,6 +128,7 @@ namespace dxvk {
       windowData.swapchain->GetDevice()->GetCreationParameters(&create_parms);
 
       if (!(create_parms.BehaviorFlags & D3DCREATE_NOWINDOWCHANGES)) {
+        D3D9WindowMessageFilter filter(window);
         if (wparam) {
           // Heroes of Might and Magic V needs this to resume drawing after a focus loss
           D3DPRESENT_PARAMETERS params;
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -10,6 +10,7 @@ namespace dxvk {
   struct D3D9WindowData {
     bool unicode;
     bool filter;
+    bool activateProcessed;
     WNDPROC proc;
     D3D9SwapChainEx* swapchain;
   };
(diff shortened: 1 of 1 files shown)


Oschowa:
I can confirm that this fixes the game freezing after changing fullscreen state as described in #2569

ranplayer:
I can confirm it as well. Thank you @gofman

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -10,6 +10,7 @@ namespace dxvk {
   struct D3D9WindowData {
     bool unicode;
     bool filter;
+    bool activate_processed;
```
K0bin, line 13: Use camelCase please.

PR #2599, opened 2022-04-09
dxvk.conf cleanup

Blisto91:
Corrects `enableDebugUtils` default to False and removes `allowLockFlagReadonly` option.

I also noticed these options don't have a description or entry in `dxvk.conf`
```
d3d11.zeroInitWorkgroupMemory
d3d11.enableRtOutputNanFixup
d3d11.forceTgsmBarriers
d3d11.floatControls
```
But i don't know if that is on purpose and i wouldn't be able to describe their functionality anyway.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -426,16 +426,6 @@
 
 # d3d9.deviceLocalConstantBuffers = False
 
-# Allow Read Only
-#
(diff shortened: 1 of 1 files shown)


doitsujin:
The undocumented options are basically just workarounds for game bugs and/or old drivers, might as well leave them that way.

doitsujin (approved):

PR #2608, opened 2022-04-13
Add support for shared ID3D11Fence object

Guy1524:
Implements shared fences on top of D3D11_FENCE shareable Vulkan timeline semaphores, with the rewind support done in winevulkan.

Corresponding VKD3D-Proton PR: https://github.com/HansKristian-Work/vkd3d-proton/pull/1175

### DIFF
--- a/src/d3d11/d3d11_fence.cpp
+++ b/src/d3d11/d3d11_fence.cpp
@@ -6,12 +6,21 @@ namespace dxvk {
   D3D11Fence::D3D11Fence(
           D3D11Device*        pDevice,
           UINT64              InitialValue,
-          D3D11_FENCE_FLAG    Flags)
+          D3D11_FENCE_FLAG    Flags,
+          HANDLE              hFence)
   : D3D11DeviceChild<ID3D11Fence>(pDevice) {
     DxvkFenceCreateInfo fenceInfo;
     fenceInfo.initialValue = InitialValue;
+    m_flags = Flags;
 
-    if (Flags)
+    if (Flags & D3D11_FENCE_FLAG_SHARED) {
+      fenceInfo.sharedType = VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D11_FENCE_BIT;
+      if (hFence == nullptr)
+        hFence = INVALID_HANDLE_VALUE;
+      fenceInfo.sharedHandle = hFence;
+    }
+
+    if (Flags & ~D3D11_FENCE_FLAG_SHARED)
       Logger::err(str::format("Fence flags 0x", std::hex, Flags, " not supported"));
 
     m_fence = pDevice->GetDXVKDevice()->createFence(fenceInfo);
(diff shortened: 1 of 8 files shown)


src/d3d11/d3d11_fence.h
```diff
@@ -0,0 +1,98 @@
+#pragma once
+
+#include "../dxvk/dxvk_fence.h"
+#include "../dxvk/dxvk_gpu_query.h"
+#include "../dxvk/dxvk_context.h"
+
+#include "d3d11_device_child.h"
+
+extern "C" {
+  #include "shared_fence.h"
```
pchome, line 10: If you really need this then do in in header file like:
```c
#ifdef __cplusplus
extern "C" {
#endif
...
#ifdef __cplusplus
}
#endif
```
But, IMHO, better to have shared_fence.cpp

Guy1524: Right, not sure why I did it that way.  The reason the implementation is a .c is that it is shared with vkd3d-proton.

doitsujin:
Fixed some bad decisions I made with the first 5 commits and pulled those in. Will look at the final three in detail once this is rebased against master.

src/dxvk/dxvk_fence.cpp
```diff
@@ -11,6 +11,12 @@ namespace dxvk {
     typeInfo.semaphoreType = VK_SEMAPHORE_TYPE_TIMELINE;
     typeInfo.initialValue = info.initialValue;
     
+    VkExportSemaphoreCreateInfo exportInfo = { VK_STRUCTURE_TYPE_EXPORT_SEMAPHORE_CREATE_INFO };
+    exportInfo.handleTypes = VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE_BIT;
```
doitsujin, line 15: Why do we hard-code the handle type here when we pass it in through the create info?

src/dxvk/dxvk_fence.cpp
```diff
@@ -19,6 +25,17 @@ namespace dxvk {
     if (vr != VK_SUCCESS)
       throw DxvkError("Failed to create timeline semaphore");
 
+    if (info.sharedHandle != INVALID_HANDLE_VALUE) {
+      VkImportSemaphoreWin32HandleInfoKHR importInfo = { VK_STRUCTURE_TYPE_IMPORT_SEMAPHORE_WIN32_HANDLE_INFO_KHR };
+      importInfo.semaphore = m_semaphore;
+      importInfo.handleType = VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE_BIT;
```
doitsujin, line 31: Same here

src/dxvk/dxvk_fence.cpp
```diff
@@ -92,4 +109,20 @@ namespace dxvk {
     }
   }
 
+  HANDLE DxvkFence::sharedHandle() const {
+    if (m_info.sharedType == VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_FLAG_BITS_MAX_ENUM)
+      return INVALID_HANDLE_VALUE;
+
+    VkSemaphoreGetWin32HandleInfoKHR win32HandleInfo = { VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR };
+    win32HandleInfo.semaphore = m_semaphore;
+    win32HandleInfo.handleType = VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE_BIT;
```
doitsujin, line 118: And here

src/d3d11/d3d11_fence.cpp
```diff
@@ -49,14 +58,34 @@ namespace dxvk {
...
+    *pHandle = sharedHandle;
+    return S_OK;
   }
 
 
   HRESULT STDMETHODCALLTYPE D3D11Fence::SetEventOnCompletion(
           UINT64              Value,
           HANDLE              hEvent) {
+    if (m_flags & D3D11_FENCE_FLAG_SHARED) {
+      // TODO directly call vkWaitSemaphores here, enqueueWait assumes monotonic semaphore
```
doitsujin, line 84: Why is this an issue? The fence worker should already be able to deal with rewinding semaphores due to the timeout, I don't see any good reason to disable this functionality.

src/d3d11/d3d11_fence.cpp
```diff
@@ -66,6 +95,11 @@ namespace dxvk {
 
 
   UINT64 STDMETHODCALLTYPE D3D11Fence::GetCompletedValue() {
+    if (m_flags & D3D11_FENCE_FLAG_SHARED) {
+      // TODO directly call vkGetSemaphoreCounterValue here, getValue returns the monotonic tracked value
```
doitsujin, line 99: Same here, even if a fence is rewound then the reported fence value will be updated within 10ms.

We can improve this behaviour in the backend if necessary, but disabling this function seems unnecessary unless there is an extremely good reason to.

doitsujin (changes requested):

src/d3d11/d3d11_fence.cpp
Guy1524: I thought it could cause a nasty condition where after the fence is rewound, a wait is started that immediately signals because the timeout hasn't been reached yet.  If this is not too big of a concern for you I'd not be opposed to remove the disable.

doitsujin: Rewinding fences is always inherently race-y, we can make the backend a bit more robust there but for now I'd just put a TODO comment there and call it a day. We should definitely enable this functionality.

src/dxvk/dxvk_fence.cpp
Guy1524: Yeah that was an oversight.

PR #2610, opened 2022-04-15
[util] disable allowDoNotWait for Port Royale 3

### LINKED ISSUE #2148: [d3d9] Port Royale 3 hangs when transitioning to city view
The game hangs (background music still plays and the "loading" ship wheel still turns, yet the city view never renders on screen) while transitioning from map view to city view. Occasionally I have been able to get a city view after multiple tries, but hangs are the standard behavior. WineD3D works just fine. I've captured a trace with 3 transitions, but annoyingly it worked just fine (without any hangs) on replay with DXVK. I was able to reproduce the hangs both on an Nvidia card and an Intel iGPU. They only show up when using DXVK. 

Here's a [screen capture](https://drive.google.com/file/d/1Qcr7ftu2Q0BA5VcmKNn-3XU6TXzV-UH0/view?usp=sharing).

Initially I suspected a memory allocation issue and tried enabling LAA and setting various d3d9 memory limits without any results.

Hopefully you can figure out what's going on here, since I'm out of ideas.

Also, as far as I can tell, this is not a regression since I noticed the same problem a while ago with DXVK 1.8.1, but did not have the time to look into it until now.

### Software information
Port Royale 3

### System information
- GPU: GTX 1080
- Driver: 465.31
- Wine version: 6.12 Staging
- DXVK version: 1.9-f903400

### Apitrace file(s)
[PortRoyale3.trace.tar.xz](https://drive.google.com/file/d/1xnUgYD7Yb4WUhxa_G-Xap1KFd55aoZ4G/view?usp=sharing)

### Log files
- d3d9.log: [PortRoyale3_d3d9.log](https://github.com/doitsujin/dxvk/files/6777316/PortRoyale3_d3d9.log)
- d3d11.log: N/A
- dxgi.log: N/A

Blisto91:
Fixes the game getting stuck in infinite loading screens sometimes when entering city view

Fixes #2148


doitsujin:
Do we know what the root cause is?

Blisto91:
Not sure. Know almost nothing about debugging something like this.

Now that i'm trying again to get some more logs i can't reproduce it all of a sudden. So hold off with the PR anyway.

Blisto91:
I am so confused.

How i usually test different different dxvk versions along with proton is that i download a release from this github.
If i have the game set to proton 7.0 i then go into `teamapps/common/Proton 7.0/dist/lib/wine/dxvk/` and rename the files to something like d3d9_b.dll and then extract the downloaded files into the folder. Same for 64bit.

For this game i am launching using [proton 7.0-2rc which comes with 1.10.1](https://github.com/ValveSoftware/Proton/issues/5749) and if i use the ones that come along with proton i just cannot reproduce this issue. If i use 1.10.1 or a master artifact downloaded from this site i can reproduce it.
The files downloaded from here are also not the same size and sha256 checksum as the ones included with 7.0-2rc even tho they are supposedly both 1.10.1. I'm guessing that's because the compilation is slightly different.

If i am missing something here please enlighten me.
Note the OP of the original issue is using wine so i am guessing they are using the github downloaded version.

WinterSnowfall:
FYI, I stumbled across this fix completely by accident and have been happy enough that it let me play the game for more than a couple of days of today game time without any lockups. As to the "why" it works, no idea... and why the lockups don't happen with the proton dxvk build (I can also confirm the same behavior), also no idea.

pchome:
FWIW, I also noted difference between official and self-compiled version when I tried to use Mingw builds a while ago.
I used freshly compiled DXVK on a game I payed recently and it crashed after a few minutes of gameplay. I tried to localize the problem by recompiling toolchain with different binutils or gcc versions but with no luck. Meanwhile with official DXVK I was able to play for hours without any issues (same with winelib version).

It's not certain which build is broken, it can be correct official DXVK correctly failing where it should, but Proton DXVK not correct. Or vice-versa.

I googled there is a difference how toolchain should be compiled for cross-compilation, not only posix threads but some libraries should be linux binaries (not *.lib or *.dll) because otherwise compiler will produce broken binaries which will work but have incorrect memory management (or so). So I though it was the source of the problem in my toolchain, but I did not found what and how I should fix.

pchome:
BTW, there is also [llvm-mingw](https://github.com/mstorsjo/llvm-mingw)
Back then I was too busy testing self-compiled mingw toolchain to try this project, but it seems compile current DXVK master branch fine (despite a lot of warnings during compilation). I only checked dxvk triangle test (built with -Denable_tests=true) and it runs ok.

Could someone check if same issues present in DXVK built with llvm-mingw?

wils0nyan:
> FWIW, I also noted difference between official and self-compiled version when I tried to use Mingw builds a while ago. I used freshly compiled DXVK on a game I payed recently and it crashed after a few minutes of gameplay. I tried to localize the problem by recompiling toolchain with different binutils or gcc versions but with no luck. Meanwhile with official DXVK I was able to play for hours without any issues (same with winelib version).

i'm no expert but maybe it's a problem with environment variables CFLAGS etc. With Arch Linux the file /etc/makepkg.conf might be worth checking out for instance even just commenting out all the C and LD flags to see if it makes a difference. Also looking at an example package build https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=dxvk-winelib     you can see there they use sed a bit to make sure things are a certain way with the flags.

pchome:
> Could someone check if same issues present in DXVK built with llvm-mingw?

Here is the build of current git master branch: [dxvk-llvm-mingw.zip](https://github.com/doitsujin/dxvk/files/8501315/dxvk-llvm-mingw.zip)
I used llvm-mingw-20220323-msvcrt-ubuntu-18.04-x86_64.tar.xz. It can be extracted somewhere and used as is with DXVK, just `export PATH="/path/to/llvm-mingw:${PATH}"`, but for this build I added full paths in original cross-files just in case.

I'm on Gentoo, and llvm-mingw looks like all-in-one I wanted all this time for DXVK cross-compilation (instead of building and configuring crossdev environment just for DXVK). I did more extensive testing with some heavy games and was able to play with almost whole RAM/VRAM/swap full without issues.

WinterSnowfall:
> I did more extensive testing with some heavy games and was able to play with almost whole RAM/VRAM/swap full without issues.

Not really the issue in this particular case, since the game mostly stays under 1GB of RAM of system memory, and I doubt its VRAM usage is causing the lockups. 

One of the 3D wizards here needs to look into why the workaround I lucked my way into appears to be working. It ultimately needs to be fixed in the standard builds.

pchome:
>> I did more extensive testing with some heavy games and was able to play with almost whole RAM/VRAM/swap full without issues.

> Not really the issue in this particular case, since the game mostly stays under 1GB of RAM of system memory, and I doubt its VRAM usage is causing the lockups.

I was my case, where DXVK compiled on my system was not able to operate properly under heavy load. So I specifically tested llvm-mingw build for this, and since it passed the test I can say this build kind of legit (at least on my system).
Still worth to check how it differ from official and Proton builds in your case, maybe this will help pinpoint the problem.

Blisto91:
@pchome i tried using your build as a drop in replacement instead of the ones that come with proton and i could still reproduce the freeze :)

pchome:
@Blisto91 thanks.

This is how Proton builds differ from defaults:
1. they use their own *FLAGS, auto-generated cross-files and `--buildtype=plain` (similar to what Gentoo does for meson projects) 
https://github.com/ValveSoftware/Proton/blob/58eba7108f2b44a47a7c2ae7530363619228524f/Makefile.in#L177-L182
https://github.com/ValveSoftware/Proton/blob/58eba7108f2b44a47a7c2ae7530363619228524f/make/rules-meson.mk
2. there may be difference how they compile toolchain vs how it compiler by distros (flags/options/etc.)
https://github.com/ValveSoftware/Proton/blob/58eba7108f2b44a47a7c2ae7530363619228524f/docker/gcc.Dockerfile.in

So, it can be as "simple" as it was before (years ago) with -O2 vs -O3 compiler optimizations, which IIRC was single compiler flag triggered by -O3 mode and caused instability or performance hits. Or it can be effect from other flags, like `-mstackrealign` for 32bit builds. I doubt they use custom patches for DXVK.

Blisto91:
Either way it's not good proton dxvk and official dxvk produce different results. Makes it harder to debug and find issues xd.

@doitsujin Do you want me to close this PR if there is something else going on?

Joshua-Ashton:
> Do we know what the root cause is?

Probably worth looking at the disasm of what the game does when it calls LockRect with `DONOTWAIT`. The reason we have this hack is that one HDR demo ends up literally just doing:
```
HRESULT hr = device->LockRect(..., DONOTWAIT)
while (FAILED(hr)) {
  // Oops, I forgot to make the result equal to this!
  device->LockRect(..., DONOTWAIT)
}
```

Maybe this is another instance of that.

pchome:
If Proton and Wine D3D runs fine, it can be result of `-mstackrealign` flag in Proton, but in Wine they redefine `__stdcall` and `__cdecl` by adding `__attribute__((__force_align_arg_pointer__))`. This attribute has same effect as `-mstackrealign` but applies only to functions called by external applications (they mark them `WINAPI`), so do not cause big performance hit (just guessing).

> -mstackrealign
> Realign the stack at entry.  On the x86, the -mstackrealign option generates an alternate prologue and epilogue that realigns the run-time stack if necessary.  This supports mixing legacy codes that keep 4-byte stack alignment with modern codes that keep 16-byte stack alignment for SSE compatibility.  See also the attribute "force_align_arg_pointer", applicable to individual functions.

Even if something else in Proton "fixes" the problem maybe still worth to use `__attribute__((__force_align_arg_pointer__))`, at least for 32bit builds.

Blisto91:
@pchome WINED3D run fine in both wine and proton as far as i remember.
For dxvk if you run it with the official realease/artifact/build intructions in readme then it sometimes freezes when entering city view unless the conf option is set. With dxvk included in proton it runs fine without the variable in proton and wine.

WinterSnowfall:
Ehm, why close it? I've tested the workaround extensively and know it works... we just don't know why :sweat_smile:. I have no idea how to debug this either in order to find out, but that doesn't make it unwarranted IMHO... the game does hang otherwise on regular builds of dxvk.

Blisto91:
Omg...
I made a fresh fork of the dxvk repo and forget it would remove the branch... ffs :sweat_smile:

PR #2612, opened 2022-04-15
[util] Enforce a maxAvailableMemory limit for Majesty 2

### LINKED ISSUE #1542: [d3d9] Majesty 2 woes with iGPUs and system memory heaps
While trying to start Majesty 2 on a laptop with only 4GB of RAM (unfortunately not upgradable) and an iGPU, the game hangs on startup with the infamous: `DxvkMemoryAllocator: Memory allocation failed` exception.

I thought initially it's LAA related, but it's not. The game starts up and plays just fine if I use:

`d3d9.maxAvailableMemory = 1024`

A couple of other points:
- This is a 2008 game, so expecting it to run on 4GB of RAM is not unreasonable
- There's about 3GB of free RAM when the hang occurs
- The game works fine on my other gaming system which has 32GB of RAM (hardly a surprise)
- WineD3D handles itself ok (performance aside), the game starts as expected

I have a feeling this has the potential of opening a can of worms, but can anything be done to improve the memory allocation logic in such scenarios?

### Software information
Majesty 2 (GOG version), others may be affected as well.

### System information
- GPU: Intel(R) UHD Graphics 605 (Geminilake)
- Driver: 19.2.8
- Wine version: 5.5 Staging
- DXVK version: 1.6

### Apitrace file(s)
N/A

### Log files
- d3d9.log: [Majesty2_d3d9.log](https://github.com/doitsujin/dxvk/files/4414070/Majesty2_d3d9.log)
- d3d11.log: N/A
- dxgi.log: N/A

WinterSnowfall:
Fixes #1542

While a larger discussion on UMA may be beneficial, the game's VRAM allocation logic is clearly broken in this case. I've bumped the limit to 2048 just to be a bit less restrictive, since the said game pack is known for its mod support, which might complicate things in terms of actual memory usage.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -541,6 +541,14 @@ namespace dxvk {
     { R"(\\SWTFU2\.exe$)", {{
       { "d3d9.forceSamplerTypeSpecConstants",  "True" },
     }} },
+    /* Majesty 2 (Collection)                   *
+     * Crashes on UMA without a memory limit,   *
+     * since the game(s) will allocate all      *
+     * available VRAM on startup.               */
+    { R"(\\Majesty2\.exe$)", {{
+      { "d3d9.memoryTrackTest",             "True" },
+      { "d3d9.maxAvailableMemory",          "2048" },
+    }} },
   }};
 
 


Joshua-Ashton:
This won't restrict allocations unless you enable memory track test.

WinterSnowfall:
Hmmm... yet with this override alone I was able to get past the `err:   DxvkMemoryAllocator: Memory allocation failed` error and the game starts and plays just fine. Any idea why this works then without memory track test?

Joshua-Ashton:
Probably because its querying GetAvailableTextureMemory to determine it, but it makes sense to enable track test either way

PR #2616, opened 2022-04-17
[util] Implement env helpers on non-Windows platforms

Joshua-Ashton:


src/util/util_env.cpp
```diff
@@ -46,27 +59,38 @@ namespace dxvk::env {
...
   std::string getExePath() {
+#if defined(_WIN32)
     std::vector<WCHAR> exePath;
     exePath.resize(MAX_PATH + 1);
 
     DWORD len = ::GetModuleFileNameW(NULL, exePath.data(), MAX_PATH);
     exePath.resize(len);
 
     return str::fromws(exePath.data());
+#elif defined(__linux__)
```
pchome, line 82: `#else`

Also, there is so many platform-specific blocks in this file, maybe better to use whole separate blocks for linux and _WIN32, or surround whole function in block, or use `if constexpr ...` to make code less ugly.

src/util/util_env.cpp
```diff
@@ -36,7 +49,7 @@ namespace dxvk::env {
 
   std::string getExeName() {
     std::string fullPath = getExePath();
-    auto n = fullPath.find_last_of('\\');
+    auto n = fullPath.find_last_of(env::PlatformDirSlash);
```
pchome, line 52: `std::filesystem::path::preferred_separator` ?

src/util/util_env.cpp
Joshua-Ashton: This is not an "else", this is linux specific code. The others would work on BSD, this would not.

src/util/util_env.cpp
Joshua-Ashton: This is a wchar_t on Windows, so it is not super useful here.

src/util/util_string.h
```diff
@@ -39,5 +40,11 @@ namespace dxvk::str {
     format1(stream, args...);
     return stream.str();
   }
+
+  inline void strlcpy(char* dst, const char* src, size_t count) {
```
doitsujin, line 44: any particular reason why we're rolling our own implementation here?

Joshua-Ashton: strlcpy doesn't exist unless you're on BSD

doitsujin (approved):

Issue #2193, opened 2021-07-29
[feature-request] setup_dxvk.sh: use $WINE from system environment


jl452:
replace
> wine="wine"
> wine64="wine64"
> wineboot="wineboot"

to something like
> WINE="${WINE:-wine}"
> WINE64="${WINE}64"

example from Proton-6.13-GE-1/protonfixes/winetricks

doitsujin:
Feel free to submit a PR.

Ahmed-E-86:
```
wine="$WINE"
wine64="${WINE}64"
wineboot="${WINE}boot"
```
This will work with most custom wine builds, but it will not work with Proton or Proton-GE builds because `wineboot` does not exist with these builds, but the script will work without finding `wineboot`. The only remaining issue is if you do not use `WINE=/path_to_wine_binary`, because you will end up with this error:
`64: Not a wine executable. Check your $wine.`

but this issue can be fixed with this script:

```
if [ -z "$WINE" ]; then
   wine="wine"
   wine64="wine64"
   wineboot="wineboot"
fi

if [ -n "$WINE" ]; then
   wine="$WINE"
   wine64="${WINE}64"
   wineboot="${WINE}boot"
fi
```
I tested it, and it works without any issues.

Ahmed-E-86:
> Feel free to submit a PR.

https://github.com/doitsujin/dxvk/pull/2304

Ahmed-E-86:
> Feel free to submit a PR.

https://github.com/doitsujin/dxvk/pull/2305

PR #2617, opened 2022-04-18
[build] Support custom Wine versions

forkbomb9:
This allows people to use custom wine versions by exporting the WINE env variable.
re. https://github.com/doitsujin/dxvk/issues/2193

If undefined, it uses the `$PATH`'s `wine` using `command -v` instead of `which`.
This is POSIX compliant (also Debian 11 deprecated which)

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -56,36 +56,36 @@ export WINEDEBUG=-all
 # wine gecko and mono
 export WINEDLLOVERRIDES="mscoree,mshtml="
 
-wine="wine"
-wine64="wine64"
-wineboot="wineboot"
+# support custom Wine versions
+[ -z "$WINE" ] && WINE="$(command -v wine)"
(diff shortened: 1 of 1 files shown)


pchome:
or just export PATH
also try this PR with `export WINE=wine` or `export WINE=some_random_user_input/idk` or some random formatC command

forkbomb9:
Exporting a new PATH is not always as nice as just pointing `WINE`, scripts et all.
`WINE=wine` it works as expected (it's like reverting to the previous behavior), and a random input it doesn't work (this was already checked at line 75 & 76.

doitsujin:
We already rejected the exact same thing multiple times already. If you for whatever reason have such a non-standard setup that the script doesn't work, don't use the script.

forkbomb9:
Sorry, I hadn't seen this kind of change rejected. I just saw that issue I linked.

PR #2618, opened 2022-04-19
[util] Implement string helpers on non-Windows platforms

Joshua-Ashton:
Title.

I think we could probably just use these on Windows too tbh.

### DIFF
--- a/src/util/util_string.cpp
+++ b/src/util/util_string.cpp
@@ -1,43 +1,55 @@
 #include "util_string.h"
+#include "util_utf.h"
 
 namespace dxvk::str {
-  std::string fromws(const WCHAR *ws) {
-    size_t len = ::WideCharToMultiByte(CP_UTF8,
-      0, ws, -1, nullptr, 0, nullptr, nullptr);
 
-    if (len <= 1)
-      return "";
+  template <typename FromChar, typename ToChar>
+  size_t reencodeString(const FromChar* from, ToChar* to, size_t toLen = 0) {
+    using UTFFromChar = typename utf::UTFTypeTraits<FromChar>::CharType;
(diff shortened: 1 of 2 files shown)


src/util/util_string.cpp
```diff
@@ -15,16 +19,26 @@ namespace dxvk::str {
     ::WideCharToMultiByte(CP_UTF8, 0, ws, -1,
       &result.at(0), len, nullptr, nullptr);
     return result;
+#else
+    std::wstring_convert<std::codecvt_utf8<wchar_t>> converter;
```
doitsujin, line 23: This is deprecated in C++17.

doitsujin (changes requested):
See comments.

If we can get rid of the Win32-specific functions altogether I'd prefer that over ifdef sauce, but we need things to work properly and test with somewhat common stuff, such as Japanese game executable names. We could always roll our own UCS-2 or UTF-16 to UTF-8 conversion as well, the encodings are simple enough.

There's also potential issues on platforms where wchar_t is 32-bit.

src/util/util_string.cpp
Joshua-Ashton: I hate the C++ standards committee. *sigh*

Joshua-Ashton: Wow they really deprecated this and said "no replacement" huh.

Absolutely incredible work.

Joshua-Ashton:
In DXVK Native, it just uses the native wchar size.

pchome:
Any plans to use [str::filename patch](https://github.com/doitsujin/dxvk/commit/1053d5c975ea85cf18b481c977c21963eec20aca) from #2138 ?

doitsujin:
Superseded by #2758.

PR #2619, opened 2022-04-19
[build] Initial reworkings for non-Windows platform support in Meson

Joshua-Ashton:


DadSchoorse:
github won't let me comment on the line, but shouldn't all the `link_args` also only effect windows builds?

meson.build
```diff
@@ -28,57 +27,84 @@ link_args = [
...
+      output    : [ '@BASENAME@' + res_ext ],
+      arguments : [ '-i', '@INPUT@', '-o', '@OUTPUT@' ],
+    )
+  endif
+
+  dxvk_include_path = include_directories('./include')
+else
+  lib_vulkan = cpp.find_library('vulkan')
+  lib_sdl2   = cpp.find_library('SDL2')
+  wrc        = find_program('true')
```
DadSchoorse, line 96: what's this for, wouldn't you also need `wrc_generator`?

Joshua-Ashton: Fixed

Joshua-Ashton:
> github won't let me comment on the line, but shouldn't all the `link_args` also only effect windows builds?

Yes, fixed.

meson.build
```diff
@@ -19,66 +18,97 @@ compiler_args = [
...
+    dxvk_library_path = meson.source_root() + '/lib'
+  else
+    dxvk_library_path = meson.source_root() + '/lib32'
+  endif
+
+  lib_vulkan  = cpp.find_library('vulkan-1', dirs : dxvk_library_path)
+  lib_d3d9    = cpp.find_library('d3d9')
+  lib_d3d11   = cpp.find_library('d3d11')
+  lib_dxgi    = cpp.find_library('dxgi')
+  lib_d3dcompiler_43 = cpp.find_library('d3dcompiler_43', dirs : dxvk_library_path)
```
eli-schwartz, line 70: `meson.source_root()` is deprecated because it cannot sensibly be used while still supporting usage as subprojects.

On the other hand, I think what you wanted was `meson.current_source_dir()` anyway, and that only coincidentally happened to be the same.

Starting with Meson 0.56, which is too high for your minimum requirement, you can use `meson.project_source_root()` instead, but that's primarily useful when you want to get the location from inside a subdir.

Joshua-Ashton: This is only moving some.code around but ye we can fix it

meson.build
```diff
@@ -19,66 +18,97 @@ compiler_args = [
...
+    )
+  endif
+
+  dxvk_include_path = include_directories('./include')
+else
+  lib_vulkan = cpp.find_library('vulkan')
+  lib_sdl2   = cpp.find_library('SDL2')
+
+  wrc           = find_program('touch')
+  wrc_generator = generator(wrc, output : [ '@BASENAME@_ignored.h' ], arguments : [ '@OUTPUT@' ] )
```
doitsujin, line 100: shouldn't we stop using wrc unconditionally instead?

pchome: There is a thing in meson called ["disabler"](https://mesonbuild.com/Disabler.html), but idk if it can help here, will it disable only generator or a whole target too.

eli-schwartz: Its purpose is to cascade, it will convert all `wrc_generator.process()` to assign disablers instead of processed files, and consequently all `shared_library(..., foo_res)` are converted to disablers as well.

The typical use case is where, for example, an optional component of a build has a hard dependency on some dependency, and if that dependency cannot be found, the optional component shouldn't be built at all.

doitsujin: I think the point here is that we should conditionally execute the generator and only the resulting files to the respective file list if we're actually building for Windows.

pchome: My vote for leaving this stub as is (at least for now). Otherwise you'll likely end up adding even more stubs in every target which uses wrc_generator.

pchome:
I already asked for that several times (in different forms) but maybe better move all platform conditions into meson.build file in separate directory (platform/, or compat/, or you name it). It can be standalone file to "include" in main meson.build file (`subdir('compat')`) for "configuration" only, or it can generate an header file for target platform.

Also make it more linear maybe. Like
```meson
# We need thingA for platformB
if platform == 'platformB'
  varX += thingA
endif

# We must set thingB
if platform == 'platformA'
  varA += thingB
else
  varA += thingB + 'something'
endif
```
I mean not in huge separate blocks for each platform, but per feature/problem/etc. Should make new stuff additions easier.

Joshua-Ashton:
That sounds like a horrible idea

pchome:
Not folders (this time :) ), a single folder in project root and a single meson.build file. This is mainly because the main meson.build file will grow (if there more platforms planned). Not a big deal actually, will be just huge main meson.build file then.

pchome:
With kwargs support we can do something like
```meson
# Resource compiler
if platform == 'windows'
  if dxvk_is_msvc
    wrc_program = 'rc'
    wrc_args = {'output': ['@BASENAME@.res'], 'arguments': ['/fo', '@OUTPUT@', '@INPUT@']}
  else
    wrc_program = 'windres'
    wrc_args = {'output': ['@BASENAME@.o'], 'arguments': ['-i', '@INPUT@', '-o', '@OUTPUT@']}
  endif
elif platform == 'linux' and dxvk_is_winelib
  wrc_program = 'wrc'
  wrc_args = {'output': ['@BASENAME@.res'], 'arguments': ['-i', '@INPUT@', '-o', '@OUTPUT@']}
else
  wrc_program = 'touch'
  wrc_args = {'output': ['@BASENAME@_ignored.h'], 'arguments': ['@OUTPUT@']}
endif
wrc = find_program(wrc_program)
wrc_generator = generator(wrc, wrc_args)
```
Note: I did not tested if this code actually work.
I added winelib as an example, there is no any `dxvk_is_winelib` variable.

PR #2620, opened 2022-04-19
[util] Implement LUID helpers for non-Windows platforms

Joshua-Ashton:
Provides at least something here for an adapter.

### DIFF
--- a/src/util/util_luid.cpp
+++ b/src/util/util_luid.cpp
@@ -5,8 +5,19 @@
 
 #include <mutex>
 #include <vector>
+#include <atomic>
 
 namespace dxvk {
+
+#ifndef _WIN32
+  static BOOL AllocateLocallyUniqueId(LUID* luid) {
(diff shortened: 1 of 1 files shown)


src/util/util_luid.cpp
```diff
@@ -18,9 +18,12 @@ namespace dxvk {
     while (s_luids.size() < newLuidCount) {
       LUID luid = { 0, 0 };
 
+#ifdef _WIN32
```
doitsujin, line 21: Cleaner approach might be to add an implementation of `AllocateLocallyUniqueId` which just increments a static atomic or something within the `dxvk` namespace, ifdef the entire function definition, and remove the `::` from the call.

Joshua-Ashton: Ye, my initial impl was that. I will go back to it

Joshua-Ashton: Done.

PR #2659, opened 2022-05-31
[dxgi] Only enum outputs which belong to an adapter.

gofman:
That's what I see dxgi doing on Windows, and the method of selecting outputs seems to work both in Wine and Windows.

Fixes Monster Hunter Rise in case of multiple adapters which gets a bit mad when dxgi outputs don't match what it gets from QueryDisplayConfig(). Here on a dual GPU laptop (AMD) it keeps requerying display configs and resizing swapchain each frame which effectively results in not working alttabbing out of the game, unability to change display settings (like resolution, monitor) which gets reset to the previous value immediately and hang on exit. With the patch all that seems to be fixed, switching monitor through menu also works correctly.

### DIFF
--- a/src/dxgi/dxgi_adapter.cpp
+++ b/src/dxgi/dxgi_adapter.cpp
@@ -148,12 +148,15 @@ namespace dxvk {
     
     if (ppOutput == nullptr)
       return E_INVALIDARG;
-    
-    HMONITOR monitor = wsi::enumMonitors(Output);
-    
+
+    const auto deviceId   = m_adapter->devicePropertiesExt().vk11;
+
+    HMONITOR monitor = wsi::enumMonitors(
+      deviceId.deviceLUIDValid ? reinterpret_cast<const LUID *>(deviceId.deviceLUID) : nullptr, Output);
+
     if (monitor == nullptr)
       return DXGI_ERROR_NOT_FOUND;
-    
+
     *ppOutput = ref(new DxgiOutput(m_factory, this, monitor));
     return S_OK;
   }
--- a/src/dxgi/dxgi_adapter.cpp
+++ b/src/dxgi/dxgi_adapter.cpp
@@ -150,9 +150,28 @@ namespace dxvk {
       return E_INVALIDARG;
 
     const auto deviceId   = m_adapter->devicePropertiesExt().vk11;
+    const LUID *adapterLUIDs[2];
+    uint32_t numLUIDs = 0;
 
-    HMONITOR monitor = wsi::enumMonitors(
-      deviceId.deviceLUIDValid ? reinterpret_cast<const LUID *>(deviceId.deviceLUID) : nullptr, Output);
+    if (m_adapter->isLinkedToDGPU())
+      return DXGI_ERROR_NOT_FOUND;
+
+    if (deviceId.deviceLUIDValid)
+      adapterLUIDs[numLUIDs++] = reinterpret_cast<const LUID *>(deviceId.deviceLUID);
+
+    auto linkedAdapter = m_adapter->linkedIGPUAdapter();
+    /* If either LUID is not valid enumerate all monitors. */
+    if (numLUIDs && linkedAdapter != nullptr)
+    {
+      const auto *deviceId   = &linkedAdapter->devicePropertiesExt().vk11;
+
+      if (deviceId->deviceLUIDValid)
+        adapterLUIDs[numLUIDs++] = reinterpret_cast<const LUID *>(deviceId->deviceLUID);
+      else
+        numLUIDs = 0;
+    }
+
+    HMONITOR monitor = wsi::enumMonitors(adapterLUIDs, numLUIDs, Output);
 
     if (monitor == nullptr)
       return DXGI_ERROR_NOT_FOUND;
--- a/src/dxvk/dxvk_instance.cpp
+++ b/src/dxvk/dxvk_instance.cpp
@@ -251,10 +251,17 @@ namespace dxvk {
 
     DxvkDeviceFilter filter(filterFlags);
     std::vector<Rc<DxvkAdapter>> result;
+    uint32_t numDGPU = 0, numIGPU = 0;
 
     for (uint32_t i = 0; i < numAdapters; i++) {
       if (filter.testAdapter(deviceProperties[i]))
+      {
         result.push_back(new DxvkAdapter(m_vki, adapters[i]));
+        if (deviceProperties[i].deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU)
+          ++numDGPU;
+        else if (deviceProperties[i].deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU)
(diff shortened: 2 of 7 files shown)


gofman:
I've redid the original patch on top of QueryDisplayConfig() which is a stricter way to link the devices and doesn't depend on device names.

Then, during testing the patch with various games I figured it breaks Elden Ring on a dual GPU configuration in a way it doesn't see display resolutions and ends up with a broken one which can't be changed. The game depends on the first dxgi adapter having some outputs.

After testing on Windows with dual GPU (discrete / integrated) and plugging the additional external monitor to either adapter I figured the following. Windows always seem to have the first dxgi adapter with nonzero outputs. QueryDisplayConfig() reports the true paths corresponding to real monitor connections (that is, if it is plugged to discrete GPU port it will be always shown as such). dxgi interchanges the outputs between integrated and discrete GPU however based on the discrete / integrated GPU preference. Whatever GPU is used it will have both monitors as outputs from that GPU and that GPU will go first.

So currently dxvk's dxgi moves the discrete GPU first essentially suggesting to force the use of it ("Switchable graphics/ maximize performance" in Windows 10 terms). The second patch adds to this hack by also making all the iGPU outputs reported from the dGPU which will go first. This shouldn't change things for multiple discrete GPU configurations, the output should still be reported correctly (once we were able to get that from xrandr of course in winex11.drv).

gofman:
Since these patches are apparently controversial I am posting a teaser what I could achieve with that (apart from MHR, could make RDR2 sensibly work with multiple monitors for the first time):

IMG-0967

src/dxgi/dxgi_adapter.cpp
```diff
@@ -145,14 +145,71 @@ namespace dxvk {
...
+      deviceName.header.adapterId = path.sourceInfo.adapterId;
+      deviceName.header.id = path.sourceInfo.id;
+      deviceName.header.type = DISPLAYCONFIG_DEVICE_INFO_GET_SOURCE_NAME;
+      deviceName.header.size = sizeof(deviceName);
+      if ((result = DisplayConfigGetDeviceInfo(&deviceName.header)))
+      {
+        Logger::err(str::format("DisplayConfigGetDeviceInfo failed, result ", result));
+        return DXGI_ERROR_NOT_FOUND;
+      }
+      info.gdiDeviceName = deviceName.viewGdiDeviceName;
```
doitsujin, line 210: this seems to be unused by MonitorEnumProc, is this on purpose?

gofman: Oh, that is the consequences of too formal / quick rebase... The original patches (ad38417134fccaa2a8a377225151ed529681f324, 4fc55c2f90b0b560d8f5e33cbab3bdb72c249686) of course filter by this gdiDeviceName  in enum function:
```
    monitorInfo.cbSize = sizeof(monitorInfo);
    GetMonitorInfoW(hmon, (MONITORINFO *)&monitorInfo);
    if (wcscmp(data->gdiDeviceName, monitorInfo.szDevice))
      return TRUE;
```

I should probably move this part to wsi:: anyway, passing dxgi adapter luid to wsi::enumMonitors()?

doitsujin: > passing dxgi adapter luid to wsi::enumMonitors()

Sounds good, this should solve the portability issues.

src/dxvk/dxvk_instance.cpp
```diff
@@ -232,6 +232,15 @@ namespace dxvk {
       Logger::warn("DXVK: No adapters found. Please check your "
                    "device filter settings and Vulkan setup.");
     }
+    else if (result[0]->deviceProperties().deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU) {
+      for (uint32_t i = 1; i < result.size(); i++) {
+        if (result[i]->deviceProperties().deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU)
+        {
+          result[i]->linkToDGPU(result[0]);
```
doitsujin, line 239: what's the rationale behind any of this? what if there's more than one dedicated GPU present on a system?

src/dxgi/dxgi_adapter.cpp
gofman: Yeah, makes sense. Except for if can't find any GDI device in the QueryDisplayConfig, we have to report zero output count for this thing to work. QueryDisplayConfig queries only active display paths and if some adapter has no outputs it won't be returned there. It is possible to query more on Windows (inactive paths) but it is currently not supported under Wine and we are actually interested in active outputs only here.

src/dxvk/dxvk_instance.cpp
gofman: That's how discrete GPU reporting works on Windows, as far as my testing went. QueryDisplayConfig always reports the actual connection. That is, if a monitor is plugged to integrated GPU it will be always reported as such through QueryDisplayConfig. dxgi outputs to adapters binding however depends on whether discrete or integrated GPU is selected for the app (auto, or forced by "Powersave" settings I was changing for testing this). If discrete GPU is selected for the game the outputs will be reported from discrete GPU adapter. DXVK always tries to simulate the case when discrete GPU is selected, so to match Windows and not to confuse some games the outputs should be linked to discrete one.

I am afraid the case when there are multiple drivers for the same GPU device might be already broken for those games which are sensitive to correct adapter / displays configuration reporting. To have anything about that right we must end up with an adapter being configured in winex11.drv for it. If this part went well and we have an adapter this part here should probably be fine in general (we will have outputs linked to some of those), although I have no idea what is the "correct" adapter in this case, not sure if we have any good way to guess. That hopefully won't break the things completely, it will bind to one of, and if that will result in the game using not the adapter / driver user wants to use, the only way seems to be to choose / select just one adapter on host, or change their order somehow.

Can you maybe describe how exactly I can simulate such a configuration to see how it will look like?

gofman: PS Maybe it is actually best to just skip this redirection (at least for now) if more than one dGPU is found, for whatever reason (multiple drivers or real GPUs), and fallback to reporting displays bound as reported by QueryDisplayConfig. As it is indeed unclear what to do in this case, and I don't even have such a configuration with real GPUs to test, and that is apparently not a dual-GPU laptop config (or something special with multiple drivers where things either happen to work or user will have to disable one of those).

gofman:
I've updated the PR, now properly rebasing it and making the following changes:
- Fall back to returning all the monitors in case of invalid LUIDs or an unexpected error;
- Move the logic to wsi;
- Handle mirrored monitor case. Mirrored monitor is reported as the same gdi device name on Windows, only once in dxgi outputs but multiple times in QueryDisplayConfig (with different source ids). Since recently (and after I submitted the first version of the patches) this is also the case with upstream Wine and Proton;
- Refactor second patch: make it effective only when we have 1 dGPU and 1 iGPU; avoid extra class variables (just query dxvk adapter for linked adapter).

doitsujin:
Merged manually with some minor formatting changes.

Issue #1744, opened 2020-08-24
[d3d9] World in Conflict → crash "There is not enough memory to run this program"


NerosTie:
The crash occurs at the end of the second mission, before the cutescene when "Berlin-Ouest 11:30" is displayed.

The error message:

![Screenshot_20200824_173344](https://user-images.githubusercontent.com/4572452/91064752-f6ff4f80-e62f-11ea-94e6-14eaaf6c3769.png)

~~Also, I can't activate the option "Use DX10 Rendering". When I restart the game, the option is unchecked. No such issue with WineD3D.~~

### Software information
Important: the crash occurs when the settings are set to "high" or "very high". It doesn't crash when it's set to "normal" or below.

### System information
- GPU: Nvidia Geforce 730M
- Driver: 450.66
- Wine version: Wine Staging 5.15 
- DXVK version: 1344

### Apitrace file(s)
https://mega.nz/file/LjJxSK7b#ABsT2QDv6wtswaCKC5IpZAoYgJTTV9Cp3Myw3CouGaI

### Log files
[world-in-conflict.log](https://github.com/doitsujin/dxvk/files/5118823/world-in-conflict.log)

Iglu47:
> * DXVK version: 1344

it's not using master, only 743f309253f0512db9f2931b59499d6bd54e30f1

I have and check only gog version [pe_header_info_wic.exe.txt](https://github.com/doitsujin/dxvk/files/5120325/pe_header_info_wic.exe.txt)
this 32-bit game don't have set LARGE_ADDRESS_AWARE flag by default

so need a special wine-build with laa.patch -> [for staging](https://github.com/Frogging-Family/wine-tkg-git/blob/master/wine-tkg-git/wine-tkg-patches/proton/LAA-staging.patch) which will allow using the `WINE_LARGE_ADDRESS_AWARE=1`
or (without building wine) need to change 1 bit in the file-exe header yourself [for example](https://github.com/doitsujin/dxvk/issues/1318#issuecomment-570898353) or something

<details><summary>I prefer to use a simple python script</summary>
<p>

just manually make backup of original exe-file, `cp wic.exe wic.exe.bak`
save python-script (for example as `laa-flag-set.py`)
make it executable and run once

```python
#!/usr/bin/env python3.8
# See http://en.wikibooks.org/wiki/X86_Disassembly/Windows_Executable_Files#PE_Header
[...]

PR #2663, opened 2022-06-02
Use memory mapped files for D3D9 texture mapping data

K0bin:
In order to finally fix some of our address space problems, I took a page out of Gallium Nine's book.

Most resources require a copy in system memory that we have to keep around in order to avoid stalling in Lock* calls. Those take up a lot of address space leading to crashes when we exhaust the 2 or respectively 3GB (with LAA) of address space. 

To avoid that, we need to unmap resources once the application is done with them on the CPU. In theory that is possible with Vulkan memory, there is however two problems with doing it with Vulkan memory directly: 
- The DXVK memory allocator and buffer system is not designed with this in mind.
- Nvidia seems to map host visible Vulkan memory when it is allocated. (@DadSchoorse tested that.)

To work around those limitations, we use Win32 memory mapped files. Those can be mapped and unmapped as we please.

## Allocation & Mapping

We suballocate from 64MB memory mapped files to avoid overhead of allocating those. 64MB is quite a lot so we try to lock at the level of a suballocation. There is however a problem with that: MapViewOfFile requires the offset to be aligned to the memory allocation granularity which is 65k. To avoid a lot of wasted memory and address space and lots of MapViewOfFile calls for tiny resources, there are two strategies. We try to allocate tightly packed, so we cannot guarantee the alignment. Every mem file is split up into "mapping pages" that are 1MB (+- alignment to 65k) each. When a resource gets mapped, we either map the entire page or reuse the existing pointer. If a resource is either larger than such a page or crosses from one page to another, it gets a separate mapping. We round down the offset to 65k and increase the size accordingly.

## D3D resources

We use memory mapped files for all texture types, except ones placed in D3DPOOL_DEFAULT. I originally used it for buffers too but dropped that because Crysis apparently reads or writes from/to a locking pointer outside of the correct scope. That's not super rare either according to Joshua.

## GPU Readback

D3DPOOL_SYSTEMMEM textures can be written by the GPU with functions like GetRenderTargetData. When that happens, we lazily create a DXVK buffer and use that instead of the memory mapped file allocation for all future locking calls. That works well because GetRendertargetData and GetFrontbufferData usually overwrite the entire image. We compare the texture sizes and when the destination texture is smaller, we copy over the data from the memory mapped file. After that, we free that and use the buffer for everything else.

## Unmapping

Unmapping is done using a least recently used list. There is a configurable virtual memory budget which is set to 100MB by default. Once we cross that, we start unmapping old resources all the way until we are only using 3/4th of the budget.
We can only unmap resources that aren't currently locked, otherwise we could be potentially invalidating a pointer that the application is still going to use.

## 64 bit builds & DXVK Native

All of this is only enabled for 32bit Win32 builds. It gets removed by the preprocessor otherwise.

## Cleanup

I removed the direct upload path and because we pretty much always use the staging buffer upload path now anyway. It's necessary for the memory mapped files to work and the direct path was only implemented because I was worried about raised address space usage. We now have a better solution for that. Along with that, there's some nice cleanup in the LockImage/LockBuffer functions.

I also removed evictManagedOnUnlock. This PR basically solves the same problem without the downsides. The option has never really been useful as pretty much all games that tended to crash also relied on the system memory copies to avoid terrible hitches.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -298,17 +298,6 @@
 # d3d9.shaderModel = 3
 
 
-# Evict Managed on Unlock
-# 
-# Decides whether we should evict managed resources from
-# system memory when they are unlocked entirely.
-#
-# Supported values:
-# - True, False: Always enable / disable
-
-# d3d9.evictManagedOnUnlock = False
-
-
 # DPI Awareness
 # 
 # Decides whether we should call SetProcessDPIAware on device
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -560,3 +560,12 @@
 # - True/False
 
 # dxvk.enableDebugUtils = False
+
+# Memory limit for locked D3D9 textures
+#
+# How much virtual memory will be used for textures (in MB).
+# 0 to disable the limit.
+# THIS DOES NOT IMPACT ACTUAL MEMORY CONSUMPTION OR TEXTURE QUALITY.
+# DO NOT CHANGE THIS UNLESS YOU HAVE A VERY GOOD REASON.
+
+# d3d9.textureMemory = 100
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1,6 +1,7 @@
 #include "d3d9_device.h"
 
 #include "d3d9_annotation.h"
+#include "d3d9_common_texture.h"
 #include "d3d9_interface.h"
 #include "d3d9_swapchain.h"
 #include "d3d9_caps.h"
@@ -4201,6 +4202,7 @@ namespace dxvk {
         pResource->GetBuffer(Subresource, !needsReadback);
       }
 
+      // Don't use MapTexture here to keep the mapped list small while the resource is still locked.
       mapPtr = pResource->GetData(Subresource);
 
       if (needsReadback) {
@@ -4298,6 +4300,8 @@ namespace dxvk {
 
     pResource->SetLocked(Subresource, true);
 
+    UnmapTextures();
+
     const bool noDirtyUpdate = Flags & D3DLOCK_NO_DIRTY_UPDATE;
     if ((desc.Pool == D3DPOOL_DEFAULT || !noDirtyUpdate) && !readOnly) {
       if (pBox && MipLevel != 0) {
@@ -4354,6 +4358,7 @@ namespace dxvk {
     if (unlikely(!pResource->GetLocked(Subresource)))
       return D3D_OK;
 
+    MapTexture(pResource, Subresource); // Add it to the list of mapped resources
     pResource->SetLocked(Subresource, false);
 
     // Flush image contents from staging if we aren't read only
@@ -4380,6 +4385,7 @@ namespace dxvk {
       pResource->SetNeedsReadback(Subresource, true);
     }
 
+    UnmapTextures();
     return D3D_OK;
   }
 
@@ -4463,7 +4469,7 @@ namespace dxvk {
           + srcOffsetBlockCount.y * pitch
           + srcOffsetBlockCount.x * formatInfo->elementSize;
 
-      const void* mapPtr = pSrcTexture->GetData(SrcSubresource);
+      const void* mapPtr = MapTexture(pSrcTexture, SrcSubresource);
       VkDeviceSize dirtySize = extentBlockCount.width * extentBlockCount.height * extentBlockCount.depth * formatInfo->elementSize;
       D3D9BufferSlice slice = AllocStagingBuffer(dirtySize);
       const void* srcData = reinterpret_cast<const uint8_t*>(mapPtr) + copySrcOffset;
@@ -4489,7 +4495,7 @@ namespace dxvk {
     }
     else {
       const DxvkFormatInfo* formatInfo = lookupFormatInfo(pDestTexture->GetFormatMapping().FormatColor);
-      const void* mapPtr = pSrcTexture->GetData(SrcSubresource);
+      const void* mapPtr = MapTexture(pSrcTexture, SrcSubresource);
 
       // Add more blocks for the other planes that we might have.
       // TODO: PLEASE CLEAN ME
@@ -4522,6 +4528,7 @@ namespace dxvk {
         image, dstLayers,
         slice.slice);
     }
+    UnmapTextures();
   }
 
   void D3D9DeviceEx::EmitGenerateMips(
@@ -4643,6 +4650,7 @@ namespace dxvk {
     pResource->SetMapFlags(Flags | oldFlags);
     pResource->IncrementLockCount();
 
+    UnmapTextures();
     return D3D_OK;
   }
 
@@ -4675,7 +4683,8 @@ namespace dxvk {
     pResource->DirtyRange().Clear();
     TrackBufferMappingBufferSequenceNumber(pResource);
 
-	  return D3D_OK;
+    UnmapTextures();
+    return D3D_OK;
   }
 
 
@@ -7219,6 +7228,63 @@ namespace dxvk {
     // immediately after a flush, we need to use the sequence number
     // of the previously submitted chunk to prevent deadlocks.
     return m_csChunk->empty() ? m_csSeqNum : m_csSeqNum + 1;
+  }
+
+
+  void* D3D9DeviceEx::MapTexture(D3D9CommonTexture* pTexture, UINT Subresource) {
+    // Will only be called inside the device lock
+    void *ptr = pTexture->GetData(Subresource);
+
+#ifdef D3D9_ALLOW_UNMAPPING
+    if (likely(pTexture->GetMapMode() == D3D9_COMMON_TEXTURE_MAP_MODE_UNMAPPABLE)) {
+      m_mappedTextures.insert(pTexture);
+    }
+#endif
+
+    return ptr;
+  }
+
+  void D3D9DeviceEx::TouchMappedTexture(D3D9CommonTexture* pTexture) {
+#ifdef D3D9_ALLOW_UNMAPPING
+    if (pTexture->GetMapMode() != D3D9_COMMON_TEXTURE_MAP_MODE_UNMAPPABLE)
+      return;
+
+    D3D9DeviceLock lock = LockDevice();
+    m_mappedTextures.touch(pTexture);
+#endif
+  }
+
+  void D3D9DeviceEx::RemoveMappedTexture(D3D9CommonTexture* pTexture) {
+#ifdef D3D9_ALLOW_UNMAPPING
+    if (pTexture->GetMapMode() != D3D9_COMMON_TEXTURE_MAP_MODE_UNMAPPABLE)
+      return;
+
+    D3D9DeviceLock lock = LockDevice();
+    m_mappedTextures.remove(pTexture);
+#endif
+  }
+
+  void D3D9DeviceEx::UnmapTextures() {
+    // Will only be called inside the device lock
+
+#ifdef D3D9_ALLOW_UNMAPPING
+    uint32_t mappedMemory = m_memoryAllocator.MappedMemory();
+    if (likely(mappedMemory < uint32_t(m_d3d9Options.textureMemory)))
+      return;
+
+    uint32_t threshold = (m_d3d9Options.textureMemory / 4) * 3;
+
+    auto iter = m_mappedTextures.leastRecentlyUsedIter();
+    while (m_memoryAllocator.MappedMemory() >= threshold && iter != m_mappedTextures.leastRecentlyUsedEndIter()) {
+      if (unlikely((*iter)->IsAnySubresourceLocked() != 0)) {
+        iter++;
+        continue;
+      }
+      (*iter)->UnmapData();
+
+      iter = m_mappedTextures.remove(iter);
+    }
+#endif
   }
 
 }
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4142,10 +4142,6 @@ namespace dxvk {
     VkExtent3D levelExtent = pResource->GetExtentMip(MipLevel);
     VkExtent3D blockCount  = util::computeBlockCount(levelExtent, formatInfo->blockSize);
 
-    const bool systemmem = desc.Pool == D3DPOOL_SYSTEMMEM;
-    const bool managed   = IsPoolManaged(desc.Pool);
-    const bool scratch   = desc.Pool == D3DPOOL_SCRATCH;
-
     bool fullResource = pBox == nullptr;
     if (unlikely(!fullResource)) {
       VkOffset3D lockOffset;
@@ -4187,7 +4183,7 @@ namespace dxvk {
 
     DxvkBufferSliceHandle physSlice;
 
-    if (Flags & D3DLOCK_DISCARD) {
+    if ((Flags & D3DLOCK_DISCARD) && needsReadback) {
       // We do not have to preserve the contents of the
       // buffer if the entire image gets discarded.
       physSlice = pResource->DiscardMapSlice(Subresource);
@@ -4201,18 +4197,8 @@ namespace dxvk {
     } else {
       physSlice = pResource->GetMappedSlice(Subresource);
 
-      // We do not need to wait for the resource in the event the
-      // calling app promises not to overwrite data that is in use
-      // or is reading. Remember! This will only trigger for MANAGED resources
-      // that cannot get affected by GPU, therefore readonly is A-OK for NOT waiting.
-      const bool usesStagingBuffer = pResource->DoesStagingBufferUploads(Subresource);
-      const bool skipWait = (scratch || managed || systemmem) && !needsReadback
-        && (usesStagingBuffer || readOnly);
-
-      if (alloced && !needsReadback) {
-        std::memset(physSlice.mapPtr, 0, physSlice.length);
-      }
-      else if (!skipWait) {
+      if (needsReadback) {
+        const Rc<DxvkBuffer> mappedBuffer = pResource->GetBuffer(Subresource);
         if (unlikely(needsReadback) && pResource->GetImage() != nullptr) {
           Rc<DxvkImage> resourceImage = pResource->GetImage();
 
@@ -4283,12 +4269,12 @@ namespace dxvk {
             }
           });
           TrackTextureMappingBufferSequenceNumber(pResource, Subresource);
-        } else if (!(Flags & D3DLOCK_DONOTWAIT) && !WaitForResource(mappedBuffer, pResource->GetMappingBufferSequenceNumber(Subresource), D3DLOCK_DONOTWAIT)) {
-          pResource->EnableStagingBufferUploads(Subresource);
         }
 
         if (!WaitForResource(mappedBuffer, pResource->GetMappingBufferSequenceNumber(Subresource), Flags))
           return D3DERR_WASSTILLDRAWING;
+      } else if (alloced) {
+        std::memset(physSlice.mapPtr, 0, physSlice.length);
       }
     }
 
@@ -4324,7 +4310,7 @@ namespace dxvk {
       }
     }
 
-    if (managed && !readOnly) {
+    if (IsPoolManaged(desc.Pool) && !readOnly) {
       pResource->SetNeedsUpload(Subresource, true);
 
       for (uint32_t i : bit::BitMask(m_activeTextures)) {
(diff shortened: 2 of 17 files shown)


qinlili23333:
In my own test, this PR make Saints Rows 2 even worse. In #2524 , Saints Rows 2 will stuck in new game loading screen and then crash. In this PR's [artifacts](https://github.com/doitsujin/dxvk/runs/6700882717?check_suite_focus=true) , the game directly crash when choose new game in menu. And the mose weird thing is that no error can be found in log file. 


<details><summary>sr2_pc_d3d9.log</summary>

```
info:  Game: sr2_pc.exe
info:  DXVK: v1.10.1
info:  Found config file: dxvk.conf
info:    d3d9.maxFrameLatency = 1
info:    dxgi.maxFrameLatency = 1
info:  Built-in extension providers:
info:    Win32 WSI
info:    OpenVR
info:    OpenXR
info:  OpenVR: could not open registry key, status 2
info:  OpenVR: Failed to locate module
info:  Enabled instance extensions:
info:    VK_KHR_get_surface_capabilities2
info:    VK_KHR_surface
info:    VK_KHR_win32_surface
info:  D3D9: VK_FORMAT_D16_UNORM_S8_UINT -> VK_FORMAT_D24_UNORM_S8_UINT
info:  NVIDIA GeForce RTX 3060 Laptop GPU:
info:    Driver: 512.96.0
info:    Vulkan: 1.3.194
info:    Memory Heap[0]: 
info:      Size: 6023 MiB
info:      Flags: 0x1
info:      Memory Type[1]: Property Flags = 0x1
info:      Memory Type[4]: Property Flags = 0x7
info:    Memory Heap[1]: 
info:      Size: 32652 MiB
info:      Flags: 0x0
info:      Memory Type[0]: Property Flags = 0x0
info:      Memory Type[2]: Property Flags = 0x6
info:      Memory Type[3]: Property Flags = 0xe
info:  D3D9: VK_FORMAT_D16_UNORM_S8_UINT -> VK_FORMAT_D24_UNORM_S8_UINT
info:  NVIDIA GeForce RTX 3060 Laptop GPU:
info:    Driver: 512.96.0
info:    Vulkan: 1.3.194
info:    Memory Heap[0]: 
info:      Size: 6023 MiB
info:      Flags: 0x1
info:      Memory Type[1]: Property Flags = 0x1
info:      Memory Type[4]: Property Flags = 0x7
info:    Memory Heap[1]: 
info:      Size: 32652 MiB
info:      Flags: 0x0
info:      Memory Type[0]: Property Flags = 0x0
info:      Memory Type[2]: Property Flags = 0x6
info:      Memory Type[3]: Property Flags = 0xe
info:  Process set as DPI aware
info:  Device properties:
info:    Device name:     : NVIDIA GeForce RTX 3060 Laptop GPU
info:    Driver version   : 512.96.0
info:  Enabled device extensions:
info:    VK_EXT_4444_formats
info:    VK_EXT_conservative_rasterization
info:    VK_EXT_custom_border_color
info:    VK_EXT_depth_clip_enable
info:    VK_EXT_extended_dynamic_state
info:    VK_EXT_full_screen_exclusive
info:    VK_EXT_host_query_reset
info:    VK_EXT_memory_priority
info:    VK_EXT_robustness2
info:    VK_EXT_shader_demote_to_helper_invocation
info:    VK_EXT_shader_viewport_index_layer
info:    VK_EXT_transform_feedback
info:    VK_EXT_vertex_attribute_divisor
info:    VK_KHR_create_renderpass2
info:    VK_KHR_depth_stencil_resolve
info:    VK_KHR_draw_indirect_count
info:    VK_KHR_driver_properties
info:    VK_KHR_external_memory_win32
info:    VK_KHR_image_format_list
info:    VK_KHR_sampler_mirror_clamp_to_edge
info:    VK_KHR_shader_float_controls
info:    VK_KHR_swapchain
info:  Device features:
info:    robustBufferAccess                     : 1
info:    fullDrawIndexUint32                    : 1
info:    imageCubeArray                         : 1
info:    independentBlend                       : 1
info:    geometryShader                         : 1
info:    tessellationShader                     : 0
info:    sampleRateShading                      : 1
info:    dualSrcBlend                           : 0
info:    logicOp                                : 0
info:    multiDrawIndirect                      : 0
info:    drawIndirectFirstInstance              : 0
info:    depthClamp                             : 1
info:    depthBiasClamp                         : 1
info:    fillModeNonSolid                       : 1
info:    depthBounds                            : 1
info:    multiViewport                          : 1
info:    samplerAnisotropy                      : 1
info:    textureCompressionBC                   : 1
info:    occlusionQueryPrecise                  : 1
info:    pipelineStatisticsQuery                : 1
info:    vertexPipelineStoresAndAtomics         : 1
info:    fragmentStoresAndAtomics               : 0
info:    shaderImageGatherExtended              : 0
info:    shaderStorageImageExtendedFormats      : 0
info:    shaderStorageImageReadWithoutFormat    : 0
info:    shaderStorageImageWriteWithoutFormat   : 1
info:    shaderClipDistance                     : 1
info:    shaderCullDistance                     : 1
info:    shaderFloat64                          : 0
info:    shaderInt64                            : 0
info:    variableMultisampleRate                : 0
info:  VK_EXT_4444_formats
info:    formatA4R4G4B4                         : 1
info:    formatA4B4G4R4                         : 1
info:  VK_EXT_custom_border_color
info:    customBorderColors                     : 1
info:    customBorderColorWithoutFormat         : 1
info:  VK_EXT_depth_clip_enable
info:    depthClipEnable                        : 1
info:  VK_EXT_extended_dynamic_state
info:    extendedDynamicState                   : 1
info:  VK_EXT_host_query_reset
info:    hostQueryReset                         : 1
info:  VK_EXT_memory_priority
info:    memoryPriority                         : 1
info:  VK_EXT_robustness2
info:    robustBufferAccess2                    : 1
info:    robustImageAccess2                     : 0
info:    nullDescriptor                         : 1
info:  VK_EXT_shader_demote_to_helper_invocation
info:    shaderDemoteToHelperInvocation         : 1
info:  VK_EXT_transform_feedback
info:    transformFeedback                      : 0
info:    geometryStreams                        : 0
info:  VK_EXT_vertex_attribute_divisor
info:    vertexAttributeInstanceRateDivisor     : 1
info:    vertexAttributeInstanceRateZeroDivisor : 1
info:  VK_KHR_buffer_device_address
info:    bufferDeviceAddress                    : 0
info:  Queue families:
info:    Graphics : 0
info:    Transfer : 1
info:  DXVK: Read 630 valid state cache entries
info:  DXVK: Using 5 compiler threads
info:  D3D9DeviceEx::ResetSwapChain:
info:    Requested Presentation Parameters
info:      - Width:              3840
info:      - Height:             2160
info:      - Format:             D3D9Format::A8R8G8B8
info:      - Auto Depth Stencil: false
info:                  ^ Format: D3D9Format::Unknown
info:      - Windowed:           false
info:  Presenter: Actual swap chain properties:
info:    Format:       VK_FORMAT_B8G8R8A8_UNORM
info:    Present mode: VK_PRESENT_MODE_IMMEDIATE_KHR
info:    Buffer size:  3834x2120
info:    Image count:  2
info:    Exclusive FS: 0
info:  Setting display mode: 3840x2160@0
info:  Presenter: Actual swap chain properties:
info:    Format:       VK_FORMAT_B8G8R8A8_UNORM
info:    Present mode: VK_PRESENT_MODE_IMMEDIATE_KHR
info:    Buffer size:  3840x2160
info:    Image count:  2
info:    Exclusive FS: 0
```

</details>

Blisto91:
@qinlili23333 No new logging entries have been added in this pr yet (draft so i'm guessing it's not complete) from a quick code peek, if this this code specifically that causes the crash that is.
But i'm not a dev so dunno what makes sense to log and where.

K0bin:
Saints Row 2 crashes before reaching the main menu regardless of what I do. The Linux port crashes, stable Proton crashes, Proton experimental crashes. I doubt that's caused by DXVK.

dxvk.conf
```diff
@@ -557,3 +546,17 @@
...
+# 0 to disable unmapping.
+
+# d3d9.textureUnmapDelay = 16
+
+# Buffer unmapping delay
+#
+# DXVK will unmap D3D9 buffer data after a certain number of frames.
+# 0 to disable unmapping.
+
+# d3d9.bufferUnmapDelay = 16
```
pchome, line 562: `this->bufferUnmapDelay              = config.getOption<int32_t>     ("d3d9.bufferUnmapDelay",              256);`

Also, will e.g. `d3d9.presentInterval = 2` affect those delays? I didn't read the code, but delays will be 8 and 128 real frames in this case.

K0bin: > Also, will e.g. d3d9.presentInterval = 2 affect those delays? I didn't read the code, but delays will be 8 and 128 real frames in this case.

It should but that also shouldn't be a problem. This is just the first best thing I came up with and the values are somewhat arbitrary.

qinlili23333:
Tested on Windows 11 Preview 22621 with NVIDIA RTX 3060:
++ Most 3D D3D9 games got ~20% RAM decrease compared to 1.10.1 Release. 
++ Alan Wake got ~10% performance improvement.
== 2D D3D9 games have nothing difference compared to 1.10.1 Release. 
-- Saints Row 2 still crash at new game.

K0bin:
@qinlili23333 

The PR can't improve performance or reduce RAM usage.

qinlili23333:
> @qinlili23333 
> 
> The PR can't improve performance or reduce RAM usage.

That's weird. My test shows that Alan Wake got performance improvement and less memory usage. In my test scene that 1.10.1 Release used about 730M and this PR 's version used only about 510M.

Blisto91:
@qinlili23333 This is the master build this PR is built on, it's better to compare it to that to see if anything has changed.
https://github.com/doitsujin/dxvk/actions/runs/2415391587

qinlili23333:
> @qinlili23333 This is the master build this PR is built on, it's better to compare it to that to see if anything has changed. https://github.com/doitsujin/dxvk/actions/runs/2415391587

I compared with this build. Master build used the same memory but there exist performance difference. This PR really got ~5-10 more performance than master build in Alan Wake.
When force VSYNC at 2160P 60HZ master build will use 80% GPU in average, this PR will use 73% GPU in average. 
I'll test in lower resolution then.

K0bin:
Changes in GPU load make even less sense, I didn't touch any shader code or any GPU work.

qinlili23333:
This is this PR's version
Screenshot (4)

This is master build
Screenshot (5)

2560x1440 with all quality sets highest

jochuan:
I tested on the witcher 1 and it is still crashing but now looks like it only crashes when i try to enter in the temple quarter location which is a big area

Blisto91:
@jochuan What does it say in the d3d9 log when it crashes

jochuan:
[witcher_d3d9.log](https://github.com/doitsujin/dxvk/files/8838083/witcher_d3d9.log)

K0bin:
- Changed the mapping to be per resource rather than per memory mapped file. (+- the 65k alignment requirement)
- Increased the size of memory mapped files.
- Added the option to unmap resources after every use immediately. (config option: delay = 0)
- It also does that automatically if we cross 200MB of mapped data.
- Because of that I raised the unmap delay to 64 and 1024 frames for textures and buffers respectively.

K0bin:
@jochuan Please test the updated version.

jochuan:
> @jochuan Please test the updated version.

Same crash 
[errlog.txt](https://github.com/doitsujin/dxvk/files/8985847/errlog.txt)
[witcher_d3d9.log](https://github.com/doitsujin/dxvk/files/8985851/witcher_d3d9.log)

And this happened when i changed the graphics options of the game before it crashed
image

K0bin:
Must be something else entirely. The game does not crash for me, so I can't really do anything.

You should try it after uninstalling Riva Tuner and MSI Afterburner. Their Vulkan HUD is broken.

Blisto91:
Fixes https://github.com/doitsujin/dxvk/issues/1744 with latest build plus both options set to 0 close.

Saints Row 2 need `d3d9.textureUnmapDelay = -1` so graphics doesn't mess up.

Blisto91:
@jochuan Can you test the newest version? @K0bin seems to have found the issue.
I was able to reproduce it myself with your save game but with the latest version i have not been able to after a few tries.

jochuan:
> @jochuan Can you test the newest version? @K0bin seems to have found the issue. I was able to reproduce it myself with your save game but with the latest version i have not been able to after a few tries.

I played a little and no crashes so far

aufkrawall:
No issues to report with this, BL2 and Painkiller H&D run smoothly.
It might make sense to apply `d3d9.textureUnmapDelay = 4`  to Painkiller (`PKHDGame-Win32-Shipping.exe`) as well, as like BL2, it is super prone to oom crashing.

K0bin:
> it is super prone to oom crashing.

Shouldn't really be necessary. When DXVK uses more than 200MB, we go into full panic mode and unmap everything anyway.

aufkrawall:
> Shouldn't really be necessary. When DXVK uses more than 200MB, we go into full panic mode and unmap everything anyway.

So 253d403497444461e752b8c2416b432a7be05b43 might not survive draft, or is it intended to make unmapping event e.g. less harsh on performance?

K0bin:
Yeah, I might drop that commit. I haven't decided yet.

src/d3d9/d3d9_mem.cpp
```diff
@@ -0,0 +1,352 @@
+#include "d3d9_mem.h"
+#include "../util/util_string.h"
+#include "../util/util_math.h"
+#include "../util/log/log.h"
+#include "../util/util_likely.h"
+#include <utility>
+#include <sysinfoapi.h>
```
pchome, line 7: ?

K0bin: Uhhhh, I blame VSCode. :)

pchome: if `clangd` LSP then something like this may help (see `clangd --help`)
```
{
    "clangd.arguments": [
        "--header-insertion=never",
        "--suggest-missing-includes",
        ...
    ]
}
```

pchome: I'm on winelib, no problems if you really need sysinfoapi.h on mingw

dxvk.conf
```diff
@@ -581,3 +570,12 @@
 # - True/False
 
 # dxvk.enableDebugUtils = False
+
+# Memory limit for locked D3D9 resources
```
Blisto91, line 574: Probably mention that it's only relevant for 32bit. So people won't get confused.

K0bin: I don't really want users to touch that anyway.

jochuan:
Its been a long time since i was able to play the game because i was busy with college. But anyways i was able to complete chapter 4 of the game without problems but on chapter 5 the random crashes reappeared and rarely some things do not render right when the crash is near to happen.
image
I was using the .dll with the old commits where Blisto91 said that the issue was fixed. I will update and see if it changes anything.

Blisto91:
Just played for an hour ish or so of chapter 5 with latest version of the pr and didn't notice anything. That was with steam proton.
Gonna try gog version with regular wine

Blisto91:
@jochuan Haven't reproduced yet.
Does it happend if you make a `dxvk.conf` in `The Witcher Enhanced Edition/System/` and insert `d3d9.resourceMemory = 0` ?

jochuan:
> @jochuan Haven't reproduced yet. Does it happend if you make a `dxvk.conf` in `The Witcher Enhanced Edition/System/` and insert `d3d9.resourceMemory = 0` ?

I tried that and the crashes still happen and i also noticed that with the updated PR everytime that a NPC conversation begins, some pixels of the screen appear in wrong colors for 2 seconds, similar to artifacts or texture corruption (I don't know how to call it). Even so i was able to finally complete chapter 5 without the game crashing and it only crashed in the mid of the epilogue then i booted the game again and i was able to beat it. I am playing the extra adventure The Price of Neutrality now and my game only crashed 3 times, so while it is inconvenient, you can play the game without major problems if you save often and have luck.

Blisto91:
I am sadly not able to test this on windows since my card is too old for the new driver requirements there.
The crashes with `d3d9.resourceMemory = 0` have you still observed graphics bug out just before still?

Could you make a apitrace of a npc conversation where it usually happens? :slightly_smiling_face:

jochuan:
> I am sadly not able to test this on windows since my card is too old for the new driver requirements there. The crashes that happen with `d3d9.resourceMemory = 0` have you still observed the graphics bug out just before?
> 
> Could you make a apitrace of a npc conversation where it usually happens? 🙂 Might also be worth testing that with master since this pr have been updated to include most of the latest changes there. https://github.com/doitsujin/dxvk/actions/runs/2685945402

Before using `d3d9.resourceMemory = 0` this graphics bug still happened but not in the old version of this PR that i was using before and i also tested master and this doesn't happen there.
**apitrace:**
https://mega.nz/file/vkYzyRyK#wxK7pYAzQHMePXoyywJ6YtOgs0rBxozabxL8Wpl_I_k

src/d3d9/d3d9_mem.cpp
```diff
@@ -0,0 +1,350 @@
...
+  D3D9MemoryAllocator::D3D9MemoryAllocator() {
+    SYSTEM_INFO sysInfo;
+    GetSystemInfo(&sysInfo);
+    m_allocationGranularity = sysInfo.dwAllocationGranularity;
+  }
+
+  D3D9Memory D3D9MemoryAllocator::Alloc(uint32_t Size) {
+    std::lock_guard<dxvk::mutex> lock(m_mutex);
+
+    uint32_t alignedSize = align(Size, 64);
```
doitsujin, line 24: 64 == `CACHE_LINE_SIZE` or is there any other reason for this?

src/d3d9/d3d9_mem.cpp
```diff
@@ -0,0 +1,350 @@
...
+    for (auto& chunk : m_chunks) {
+      D3D9Memory memory = chunk->Alloc(alignedSize);
+      if (memory) {
+        m_usedMemory += memory.GetSize();
+        return memory;
+      }
+    }
+
+    alignedSize = align(Size, 4096);
```
doitsujin, line 33: If this is 4096 to match the page size, can we use `dwPageSize` from the system info struct instead?

src/d3d9/d3d9_mem.cpp
```diff
@@ -0,0 +1,350 @@
...
+      alignedOffset = alignDown(memory->GetOffset(), m_allocator->MemoryGranularity());
+      alignmentDelta = memory->GetOffset() - alignedOffset;
+      alignedSize = memory->GetSize() + alignmentDelta;
+
+      m_allocator->NotifyMapped(alignedSize);
+      uint8_t* basePtr = static_cast<uint8_t*>(MapViewOfFile(m_mapping, FILE_MAP_ALL_ACCESS, 0, alignedOffset, alignedSize));
+      if (unlikely(basePtr == nullptr)) {
+        DWORD error = GetLastError();
+        LPTSTR buffer = nullptr;
+        FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM, nullptr, error, MAKELANGID(LANG_NEUTRAL, SUBLANG_NEUTRAL), (LPTSTR)&buffer, 0, nullptr);
```
doitsujin, line 111: Does this provide any actually meaningful info compared to just printing the error code and maybe having a switch/case for common ones like `ERROR_OUTOFMEMORY`? These APIs seem rather cursed.

src/d3d9/d3d9_mem.cpp
```diff
@@ -0,0 +1,350 @@
...
+        Logger::err(str::format("Mapping non-persisted file failed: ", error, ", Mapped memory: ", m_allocator->MappedMemory(), ", Msg: ", buffer));
+        if (buffer) {
+          LocalFree(buffer);
+        }
+        return nullptr;
+      }
+      return basePtr + alignmentDelta;
+    }
+
+    // For small allocations we map the entire mapping page to minimize the overhead from having the align the offset to 65k bytes.
```
doitsujin, line 121: Does the allocator make sure that small allocations can't straddle two pages?

src/util/util_lru.h
```diff
@@ -0,0 +1,69 @@
...
+#include <list>
+#include <unordered_map>
+
+namespace dxvk {
+
+  template<typename T>
+  class lru_list {
+
+  public:
+    typedef typename std::list<T>::const_iterator lru_iterator;
```
doitsujin, line 11: why not just call this `const_iterator`? Might as well make this work with C++ forach thingy.

src/util/util_lru.h
```diff
@@ -0,0 +1,69 @@
...
+        return;
+
+      m_list.erase(cacheIter->second);
+      m_list.push_back(value);
+      auto iter = m_list.cend();
+      --iter;
+      m_cache[value] = iter;
+    }
+
+    lru_iterator leastRecentlyUsedIter() {
```
doitsujin, line 51: `cbegin`? This should also be const if we're returning a const_iterator.

src/util/util_lru.h
```diff
@@ -0,0 +1,69 @@
...
+      return m_list.begin();
+    }
+
+    lru_iterator leastRecentlyUsedEndIter() {
```
doitsujin, line 55: `cend` + const?

src/d3d9/d3d9_device.cpp
```diff
@@ -7223,4 +7232,61 @@ namespace dxvk {
...
+#ifdef D3D9_ALLOW_UNMAPPING
+    if (pTexture->GetMapMode() != D3D9_COMMON_TEXTURE_MAP_MODE_UNMAPPABLE)
+      return;
+
+    D3D9DeviceLock lock = LockDevice();
+    m_mappedTextures.remove(pTexture);
+#endif
+  }
+
+  void D3D9DeviceEx::UnmapResources() {
```
doitsujin, line 7269: Might as well call this `UnmapTexture` if there are no plans to re-add buffers to this.

doitsujin (changes requested):
Mostly nits, some questions about the allocation logic.

src/d3d9/d3d9_mem.cpp
K0bin: Not really. Might as well get rid of that.

src/d3d9/d3d9_mem.cpp
K0bin: Yes, the if just before this line takes care of that.

src/d3d9/d3d9_device.cpp
```diff
@@ -4187,7 +4183,7 @@ namespace dxvk {
 
     DxvkBufferSliceHandle physSlice;
 
-    if (Flags & D3DLOCK_DISCARD) {
+    if ((Flags & D3DLOCK_DISCARD) && needsReadback) {
```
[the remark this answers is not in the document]
K0bin: Yup, `needsReadback` basically tracks whether the resource has been written by the GPU. If that's not the case, we don't need to discard because we don't have to synchronize anyway. So we might as well save some memory.

src/d3d9/d3d9_mem.cpp
K0bin: This is a leftover from when I started this. I expected this API to require page aligned sizes but I dont think it has that requirement. So I'll just remove this line.

src/d3d9/d3d9_mem.cpp
doitsujin: Right, that's what all the alignDown mess etc is for.

Is there any downside to having the same page included in multiple mappings, or having overlapped mappings? Both can still happen with the current logic.

K0bin: The docs aren't super detailled except for saying that it's coherent. There's no mention of any downsides there and it works fine in practice, so I assume there isnt.

src/d3d9/d3d9_mem.cpp
K0bin: Yes, basically just avoid cursed alignments.

src/d3d9/d3d9_mem.cpp
doitsujin: Probably best to clean this up and just emit the error code, yeah. Anything that wants us to call `LocalFree` and similar bullshit is stuff that I'd rather avoid.

K0bin: Done.

src/d3d9/d3d9_mem.cpp
doitsujin: Please use the constant in that case so we know what this is for.

K0bin: Already done.

doitsujin (approved):

Blisto91:
@jochuan Try again with the latest changes if you can. I wasn't able to reproduce your issue on linux with mesa drivers. 
Part of the pr was pulled out since it was causing trouble for some games and it's probably soon gonna get merged into master so more will test it.

K0bin:
Found & fixed a bug with Shogun 2

PR #2682, opened 2022-06-12
[util] Strict float emulation for Supreme Commander

### LINKED ISSUE #2638: Missing particle effects in Supreme Commander
Supcom is missing a lot of particle effects, some of them critical to playing the game competitively. The easiest way to check if they're working properly is the gate in animation when you start a skirmish match. Here's an example of not working vs. working properly: https://imgur.com/a/WO2PJkc

There is a dll override you can load on windows called dgVoodoo that fixes this D3D9 to DX12 rendering issue. However on linux while this works, it causes random crashes after 10+ minutes of gametime. So it would be nice if this were fixed natively in DXVK since linux users can't use the dgVoodoo fix. dgVoodoo's creator comments on this when patching it: "I can reproduce it, and the reason of the erroneous rendering is as interesting as this game: it's a very special incompatibility between DX shader models (original and converted)."

Blisto91:
Fixes some missing effects for Supreme Commander and it's expansion.

Closes #2638

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -561,6 +561,10 @@ namespace dxvk {
     { R"(\\eoa\.exe$)", {{
       { "d3d9.customVendorId",              "10de" },
     }} },
+    /* Supreme Commander.                       */
+    { R"(\\SupremeCommander\.exe$)", {{
+      { "d3d9.floatEmulation",        "Strict" },
+    }} },
   }};
 
 


K0bin (approved):

K0bin:
We can't base it off the Steam ID but we could add another exe name.

Flypaste:
Alright, thanks. The name we'd need is "ForgedAlliance.exe". I have no idea if it's case sensitive lol.

PR #2683, opened 2022-06-12
Round down sample count to power of 2

### LINKED ISSUE #2590: [d3d9] The Godfather:The Game crashes at launch with DXVK, works with wined3d
### Software information

The Godfather: The Game, any  graphical setting

Game works with wined3d and dgVoodoo2+DXVK.

Attempting to use DXVK without any other wrappers leads to a crash at boot.

This patch/mod is needed in order to avoid movies playing at launch, which normally causes a crash at launch on any system that is newer than XP.

https://cookieplmonster.github.io/mods/godfather/#silentpatch

Tests were done with this applied, in both `wined3d only`, `dxvk only` and `dxvk+dgvoodoo2` cases.

### System information
- GPU: NV GTX 1050
- Driver: 510.54
- Wine version: 7.5
- DXVK version: 1.10.1 and `master`

### Apitrace file(s)
- [godfather.trace.tar.gz](https://github.com/doitsujin/dxvk/files/8431096/godfather.trace.tar.gz)

Trace was captured with wined3d. Reproduces crash with DXVK.

### Log files
- d3d9.log: [godfather_d3d9.log](https://github.com/doitsujin/dxvk/files/8431168/godfather_d3d9.log)
- Wine output: 
[wine-output.txt](https://github.com/doitsujin/dxvk/files/8431243/wine-output.txt)

K0bin:
Fixes #2590

Someone please check whether the bit logic is correct.

### DIFF
--- a/src/d3d9/d3d9_util.cpp
+++ b/src/d3d9/d3d9_util.cpp
@@ -1,4 +1,5 @@
 #include "d3d9_util.h"
+#include "d3d9_device.h"
 
 namespace dxvk {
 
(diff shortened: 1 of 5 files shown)


src/d3d9/d3d9_util.cpp
```diff
@@ -41,10 +41,8 @@ namespace dxvk {
           DWORD                     MultisampleQuality,
           VkSampleCountFlagBits*    pCount) {
     uint32_t sampleCount = std::max<uint32_t>(MultiSample, 1u);
-
-    // Check if this is a power of two...
-    if (sampleCount & (sampleCount - 1))
-      return D3DERR_INVALIDCALL;
+    sampleCount = std::min<uint32_t>(MultiSample, 8u);
```
DadSchoorse, line 44: Why? Intel for example supports MSAAx16, and I'm pretty sure it can be exposed in d3d9. Also, this makes the max before not do anything, you likely want something like `std::min<uint32_t>(sampleCount, 8u);`.

src/util/util_bit.h
```diff
@@ -96,6 +96,25 @@ namespace dxvk::bit {
     #endif
   }
 
+  inline uint32_t bsr(uint32_t n) {
+    #if defined(_MSC_VER) && !defined(__clang__)
+    unsigned long index;
+    _BitScanReverse(&index, n);
+    return uint32_t(index);
+    #elif defined(__GNUC__) || defined(__clang__)
+    return 31 - __builtin_clz(n);
```
DadSchoorse, line 105: __builtin_ctz is a thing iirc.

Blisto91:
Would it make sense to output a little info in the log when it rounds down as a fyi?

turol:
Non-loop solution for rounding up to power of two (and other interesting bit-twiddling hacks): http://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2

K0bin:
I'm rounding down, not up.

Joshua-Ashton:
Looks mostly good but I want to validate it against normal D3D behaviour

PR #2685, opened 2022-06-14
[util] Limit Bionic Commando to 60fps

Blisto91:
Game physics doesn't work properly when the game runs ~above 144fps~  at high fps. So might as well cap it.
See [ProtonDB](https://www.protondb.com/app/21670), [PCGamingWiki](https://www.pcgamingwiki.com/wiki/Bionic_Commando_(2009)#Frame_rate_.28FPS.29) or [steam forums](https://steamcommunity.com/app/21670/discussions/0/3050611812308087545/)

Also seems to help make the stuttering the game can have less noticeable when the gpu would have been maxed. Or what ever is the cause.

Edit: Originally capped to 144, but changed to 60. See comments below

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -573,6 +573,11 @@ namespace dxvk {
     { R"(\\swtor\.exe$)", {{
       { "d3d9.forceSamplerTypeSpecConstants", "True" },
     }} },
+    /* Bionic Commando                          
+       Physics break at high fps               */
+    { R"(\\bionic_commando\.exe$)", {{
+      { "d3d9.maxFrameRate",                "60" },
+    }} },
   }};
 
 


mirh:
144 seems an awfully specious limit tbh. 
Almost like some dumb user with a 144hz screen thought that vsync was the only way to limit framerate, and then from that called it a day,

Blisto91:
@mirh  Yes that's is probably what happened originally, i should probably not have worded it like that as it doesn't work above 144 specificall, and that does indeed seem like a odd number. Originally when i tested it i at least didn't notice any obvious issues.

So i have now been experimenting with this alot more than i had previously and i have unlocked some new physics mechanics to test out and tried playing with them at different fps caps.
I think the main reason the 144 limit was thought to be enough is that starting around 170 ish the throwing mechanic starts to mess up really visibly. So if you have a car wreck in front of you and you grapple throw it forward it might instead fly fast in another direction. This doesn't seem to be a noticeable issue at 144 or just above.

There is another mechanic where in some places there is a thin semi broken wall where you can punch a object towards it to break it, like a forklift. I have managed to get the forklift stuck half in the semi broken piece without it making a hole. This could just be a general general bug but it probably isn't unreasonable that it could be the physics messing up a bit.
I have managed to mess it up like that even at 144 a few times tho it doesn't happen the majority of the time and i haven't been able to do it consistently.

A last clue I've noticed about the physics acting up a bit is some of the objects on the main character himself. He have some hair at the back of his head and two straps hanging on his pants that can swing around. At above 60fps if you pay attention you will notice that they are stuttering and have this ghosting effect. This does not happen when i lock it to 60.


So maybe a fps cap of 60 would be more logical.
The pc version also have some quirks where even tho there are keyboard shortcuts and you can see them all in the settings the actual game hints are all showing controller buttons. So if the devs had more of a focus on the consoles when making the game the 60fps cap appearing most normal might seem right.

One of the game files have this line
```
<physics settings="\data\settings\physics_settings.xml" fps="60" physx_effects="hardware" multithreaded="true" physx_fluids="false"/>
```
Tho considering that the physics mess up at high values I'm in doubt that the devs implemented it solely on a 60 fps step.

mirh:
[Supposedly](https://gamefaqs.gamespot.com/boards/943680-bionic-commando/49627528), the ps3 version ran at 30fps, while xbox was unlocked and a bit faster, while the pc version was [dumpster fire](https://www.eurogamer.net/digitalfoundry-bionic-commando-tech-comparison). 
And it comes with a [hidden](https://tcrf.net/Bionic_Commando_\(Windows,_2009\)) 30fps toggle.

Blisto91:
I see thanks for the link.
I see this line snippet at the 30fps lock line `The game runs at 60 FPS by default.`
So hopefully that should be fine. 60 seemed to run fine physics wise from what i tried and observed.

mirh:
I couldn't really understand that tbh. If the game ran at 60 FPS by default, then certainly you wouldn't need an external limiter?

Blisto91:
Dunno. :man_shrugging: 
Maybe it's just wrong information or something is different with the steam version

PR #2692, opened 2022-06-21
[util] Fix built-in config options loging

pchome:
Before:
```
info:  Found built-in config:
...
```
After:
```
info:  Found built-in config:
info:    dxgi.nvapiHack = False
...
```

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -768,6 +768,10 @@ namespace dxvk {
     if (appConfig != g_appDefaults.end()) {
       // Inform the user that we loaded a default config
       Logger::info(str::format("Found built-in config:"));
+
+      for (auto& pair : appConfig->second.m_options)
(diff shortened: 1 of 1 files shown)


jp7677:
Originally the logging statements looked like this:
```
info:  Found built-in config: [appName]
    [all effective configurations, thus built-in and set externally]
```
The `appName` is no longer displayed since regular expressions where introduced https://github.com/doitsujin/dxvk/commit/54ca5900e11e30b422b7d9930e2080bd16b6d3af , so it looks a bit unexpected  due to the `:`. That said, your changes should display build-in settings twice (also below the effective configuration) which is also confusing. 
A good direction in my opinion would be to show the appName again, but also the regex that matched on it.

doitsujin:
This wasn't really a bug but I don't mind this, it'll log whether users changed any of the defaults this way.

PR #2696, opened 2022-06-28
Convert flags from VkShaderStageFlags to VkPipelineStageFlags in commitGraphicsBarriers

jukim-nv:
The `DxvkContext::commitGraphicsBarriers` function calls internal functions like `checkGfxBufferBarrier` and `checkGfxImageBarrier` which eventually generate vulkan barriers for buffers and images. These two functions take a `VkPipelineStageFlags` as an argument to determine which stages the barrier applies to. Sometimes the pipeline stages are hard coded, but sometimes they are passed using the `stages` field of a `DxvkPipelineLayout`. However, `DxvkPipelineLayout::stages` is a `VkShaderStageFlags`, not a `VkPipelineStageFlags`, and the two are not cross-compatible. For example, `VK_PIPELINE_STAGE_TESSELLATION_CONTROL_SHADER_BIT` and `VK_SHADER_STAGE_FRAGMENT_BIT` both have the same raw value of `0x10`. I believe that this is why we are seeing barriers on tessellation control even on applications that don't use tessellation at all.

DXVK already has a utility function to convert from `VkShaderStageFlags` to `VkPipelineStageFlags`, so all my fix does is use that function to make the conversion where appropriate.


doitsujin:
To be clear, while this was a bug, we don't actually use the info there to emit barriers. D3D11 applications will **always** emit barriers with `VK_PIPELINE_STAGE_TESSELLATION_CONTROL_SHADER_BIT` as the destination stage mask, regardless of whether tessellation is used or not, since we cannot know that in advance.

PR #2700, opened 2022-06-28
[util] Add workarounds for Garden Warfare 2

Blisto91:
Game won't launch with amd vendorId but works with `10de`. Everyone on ProtonDB sets it to `1c06` instead which also works, but i couldn't find whose vendorId that was so just went with 10de. Didn't find it to cause any issues from some short testing.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -296,6 +296,11 @@ namespace dxvk {
     { R"(\\AWayOut(_friend)?\.exe$)", {{
       { "dxgi.maxFrameLatency",                "1" },
     }} },
+    /* Garden Warfare 2
+       Won't start on amd Id without atiadlxx     */
+    { R"(\\GW2.Main_Win64_Retail\.exe$)", {{
+      { "dxgi.customVendorId",           "10de"   },
+    }} },
 
     /**********************************************/
     /* D3D9 GAMES                                 */


doitsujin (approved):

Blisto91:
DadSchoorse correctly pointed out disabling nvapiHack was redundant so I've removed that part.

Blisto91:
Reworded the comment a bit.

pchome:
@Blisto91 
> Then proton could rid this built in config, without the user having to make a dxvk.conf,

Or Proton could make their own dxvk.conf on the fly, like it's done in e.g. [protonfixes](https://github.com/simons-public/protonfixes) (_provides additional per-game configuration on top of Proton, also AFAIK it's integrated in GE builds_). If they don't want (or can't have, idk) the whole thing in Proton, they have my permission to freely use my commits from that project ([subj](https://github.com/simons-public/protonfixes/commit/04bc61212b098d0609063c30044f87b80a5bde6d)).

Blisto91:
@pchome Ye that is also possible ofc. I did have it on my mind, but didn't know how practical it was.
But if that method is already in use other places and works fine that's cool 🙂

pchome:
BTW for single overrides (no separate config file needed)  protonfixes supports "aliases" to its builtin functions in command line. So on Proton side such override could be something like
`WINEDLLOVERRIDES="atiadlxx=b" -pf_dxvk_set=dxgi.customVendorId=-1 %command%`

PR #2704, opened 2022-07-02
[util] Limit Dead Space to 60fps and fix vsync

Trevonn:
https://www.pcgamingwiki.com/wiki/Dead_Space#Issues_fixed

The game has mouse acceleration and physics issues above 60 FPS.

Also the game locks to 30 FPS using the built-in vsync. 
Setting `d3d9.presentInterval` to 1 blocks this and the game continues to run at 60 FPS

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -347,9 +347,13 @@ namespace dxvk {
       { "d3d9.memoryTrackTest",             "True" },
     }} },
     /* Dead Space uses the a NULL render target instead
-       of a 1x1 one if DF24 is NOT supported      */
+       of a 1x1 one if DF24 is NOT supported     
+       Mouse and physics issues above 60 FPS
+       Built-in Vsync Locks the game to 30 FPS    */
     { R"(\\Dead Space\.exe$)", {{
       { "d3d9.supportDFFormats",                 "False" },
+      { "d3d9.maxFrameRate",                     "60" },
+      { "d3d9.presentInterval",                  "1" },
     }} },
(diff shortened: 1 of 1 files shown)


doitsujin (approved):

K0bin (approved):
I played through it somewhat recently and used DXVKs frame limiter. Much nicer if we just make it work correctly out of the box.

RareMv:
This should be adjusted.

Finished Dead Space 1 and 2 on PC earlier in 2022, using DXVK (1.9.x I believe?) on Windows 11 and nothing breaks at 120 Hz from personal experience. There are tons of threads on Steam confirming144 Hz is also rock solid.

Game has mouse acceleration issues regardless of framerate. With the mouse look fix, EVERYTHING in the game is perfect aside from fleshy patches of floor, which boost mouse sensitivity WAY up. But that happens on the stock game with no mouse look fix anyway. It's a really really really minor issue.

Also: the default in-game Vsync is not a 30 FPS lock. It's a half-refresh rate lock. So 30 lock on 60 Hz, 60 lock on 120 Hz, 72 lock on 144 Hz etc.

K0bin:
We already discussed this on the issue:

https://github.com/doitsujin/dxvk/commit/4a0a9d62861abd4654a39152d1803bdbc3122212#comments

tl;dr. The game is broken, even if it didn't happen in your playthrough. If you don't like it you can use dxvk.conf to disable the fps limiter but we will keep it by default.

RareMv:
Oof, I completely forgot we already had that conversation, haha. My bad.

PR #2707, opened 2022-07-06
[dxvk] Don't synchronize device if going for DLL shutdown

Joshua-Ashton:
All our other threads have been destroyed and we can no longer synchronize with them properly.

Co-authored-by: Paul Gofman <pgofman@codeweavers.com>

### DIFF
--- a/src/dxvk/dxvk_device.cpp
+++ b/src/dxvk/dxvk_device.cpp
@@ -26,6 +26,13 @@ namespace dxvk {
   
   
   DxvkDevice::~DxvkDevice() {
+    // If we are being destroyed during/after DLL process detachment
+    // from TerminateProcess, etc, our CS threads are already destroyed
+    // and we cannot synchronize against them.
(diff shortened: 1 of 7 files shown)


src/dxvk/dxvk_device.cpp
```diff
@@ -26,6 +26,13 @@ namespace dxvk {
   
   
   DxvkDevice::~DxvkDevice() {
+    // If we are being destroyed during/after DLL process detachment
+    // from TerminateProcess, etc, our CS threads are already destroyed
+    // and we cannot synchronize against them.
+    // The best we can do is just wait for the Vulkan device to be idle.
```
doitsujin, line 32: This comment seems misleading since we're not actually doing that? Would requrie a CS thread sync anyway.

doitsujin (commented):
I really hate everything about this, but if we have to...

alasky17:
@doitsujin Is it possible for this to get in the 1.10.x branch?

PR #2709, opened 2022-07-09
 [util] Enable cached dynamic resources for DayZ

Snoop05:
In my testing, constant buffers improve roughly 100% and shader resources add another 25% on top of that.

Note: FPS drops this fixes are only in specific spots and dont seem to happen on native D3D11 Nvidia (but do on AMD)

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -597,6 +597,10 @@ namespace dxvk {
     { R"(\\PortRoyale3\.exe$)", {{
       { "d3d9.allowDoNotWait",           "False" },
     }} },
+    /* DayZ */
+    { R"(\\DayZ_x64\.exe$)", {{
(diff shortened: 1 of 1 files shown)


src/util/config/config.cpp
```diff
@@ -597,6 +597,10 @@ namespace dxvk {
     { R"(\\PortRoyale3\.exe$)", {{
       { "d3d9.allowDoNotWait",           "False" },
     }} },
+    /* DayZ */
+    { R"(\\DayZ_x64\.exe$)", {{
+      { "d3d11.cachedDynamicResources",     "cr" },
+    }} },
```
doitsujin, line 603: We generally have D3D11 and D3D9 options separated so this should probably go into the D3D11 section way above this.

Snoop05: I was looking if there is some ordering or anything and completely missed the DX9 banner, my bad

PR #2714, opened 2022-07-11
Fix for MWR H1-Mod

VladWinner:
Small change: applies the fix used for Call of Duty: Modern Warfare Remastered [d3d11]
also to a client for MWR called H1-Mod (`h1-mod.exe`): https://github.com/h1-mod/h1-mod/

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -634,6 +634,13 @@ namespace dxvk {
     { R"(\\(hammer(plusplus)?|mallet|wc)\.exe$)", {{
       { "d3d9.apitraceMode",                "True" },
     }} },
+    /* Dragon Age Origins                       *
+     * Keeps unmapping the same 3 1MB buffers   *
+     * thousands of times when you alt-tab out  *
+     * Causing it to crash OOM                  */
+    { R"(\\DAOrigins\.exe$)" , {{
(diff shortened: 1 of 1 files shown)


src/util/config/config.cpp
```diff
@@ -107,18 +107,6 @@ namespace dxvk {
...
-      { "d3d11.constantBufferRangeCheck",   "True" },
-    }} },
-    /* Grim Dawn                                  */
-    { R"(\\Grim Dawn\.exe$)", {{
-      { "d3d11.constantBufferRangeCheck",   "True" },
-    }} },
-    /* NieR:Automata                              */
-    { R"(\\NieRAutomata\.exe$)", {{
-      { "d3d11.constantBufferRangeCheck",   "True" },
-    }} },
```
K0bin, line 121: We need those config options.

K0bin (changes requested):
You accidentally removed some of the other config options.

src/util/config/config.cpp
VladWinner: > We need those config options.

Ok.

> You accidentally removed some of the other config options.

It's not my fault, it's the result of **force-push** action. I haven't deleted a single line, the file has been updated many times since July

VladWinner:
Reopened https://github.com/doitsujin/dxvk/pull/3025

PR #2716, opened 2022-07-12
call wineboot as a script

pabloveliz:
wineboot is a script, and in tkg when you call it directly from this setup script, I get

environment: : command not found

Therefore, to call a script from within a script let's use exec.

$ file /opt/wine-staging/bin/wineboot
/opt/wine-staging/bin/wineboot: POSIX shell script, ASCII text executable

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -80,7 +80,7 @@ fi
 
 # ensure wine placeholder dlls are recreated
 # if they are missing
-$wineboot -u
+exec $wineboot -u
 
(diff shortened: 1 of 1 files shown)


doitsujin:
Doesn't this completely break the entire script because `exec` never returns?

pabloveliz:
You're right. Please disregard this fix.
```
cat > a.sh << EOF
#!/bin/bash
echo 1
exec ./b.sh
echo 3
EOF
cat > b.sh << EOF
#!/bin/bash
echo 2
EOF
chmod 755 a.sh b.sh
```
```
./a.sh
1
2
```

PR #2726, opened 2022-07-14
call wineboot as a script,

pabloveliz:
wineboot is a script, and in tkg when you call it directly from this setup script, I get
```
environment: : command not found
```
Therefore, to call a script from within a script let's use bash because...
```
$ file /opt/wine-staging/bin/wineboot
/opt/wine-staging/bin/wineboot: POSIX shell script, ASCII text executable
```

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -80,7 +80,7 @@ fi
 
 # ensure wine placeholder dlls are recreated
 # if they are missing
-$wineboot -u
+bash $wineboot -u
 
 win64_sys_path=$($wine64 winepath -u 'C:\windows\system32' 2> /dev/null)
 win64_sys_path="${win64_sys_path/$'\r'/}"


Joshua-Ashton:
Works fine here with TKG.

`wineboot` has a shebang of `#!/bin/sh`, not sure why it doesn't work for you.

pchome:
Maybe something wrong in `~/.profile` / `~/.bash_profile` / `~/.bashrc` ?

airminer:
@pabloveliz @miocrime Could you post the output of the following commands?
```sh
$ which wineboot
$ command -v wineboot
$ ls -l $(readlink -f $(command -v wineboot))
```

WinterSnowfall:
> ```
> $ file /opt/wine-staging/bin/wineboot
> /opt/wine-staging/bin/wineboot: POSIX shell script, ASCII text executable
> ```

Out of curiosity, if you create a softlink for wineboot in /usr/bin, does it then work properly for you?

`ln -s /opt/wine-staging/bin/wineboot /usr/bin/wineboot `

PR #2736, opened 2022-07-20
Enable deferSurfaceCreation for Fallout: New Vegas

ThomasConnor:
Prevents occasional black screen on startup

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -579,6 +579,11 @@ namespace dxvk {
     { R"(\\PortRoyale3\.exe$)", {{
       { "d3d9.allowDoNotWait",           "False" },
     }} },
+    /* Fallout: New Vegas                       
+     * Prevents black screen on startup         */
+    { R"(\\FalloutNV\.exe$)", {{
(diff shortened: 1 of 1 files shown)


doitsujin:
Do we have a bug report for this anywhere? This game was among the first D3D9 games to run at all back in the day before this option even existed.

ThomasConnor:
Apologies, I didn't realise or expect that this was regarded a 100% working title as far as dxvk was concerned as the game is buggy af so I just submitted the fix that I found worked for me. As you expressed some suprise at that I did a bit more digging with that in mind, I had already reinstalled the game to a completely clean state just to be sure that there was nothing else influncing my results, drivers were up to date etc. It was still exactly the same. Tried it on another pc and wasn't getting the same result. What I didn't realise is that 1. rivatuner was running in the tray on startup and 2. I had at some point from more than a year ago made an application profile for fallout with custom d3d dupport enabled... I don't know why I did that as it's not needed at all. Leaving rivatuner running and disabling just that custom d3d option make this "fix" unnecessary/invalid. Thanks for all your work on this and sorry for wasting your time.

PR #2737, opened 2022-07-20
Update dxvk.conf add comment to numCompilerThreads

ThomasConnor:
Thread count estimated by https://github.com/doitsujin/dxvk/blob/ff39819086b616d175a8de179c4e292014975116/src/dxvk/dxvk_pipemanager.cpp#L14
Just adding some details in the conf file for clarity on how many threads it'll use by default.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -226,9 +226,6 @@
 
 
 # Sets number of pipeline compiler threads.
-# 
-# By default on systems with less than 4 cpu threads uses ~50% of available threads,
-# on systems with more than 4 threads uses ~67% of available threads, capped at 32 by default.
 # Auto values leave some cores to the application to avoid excessive stutter.
 #  
 # Supported values:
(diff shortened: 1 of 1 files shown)


ThomasConnor:
Could someone please clarify if, for example, on a system with 12 threads available would it create 7 or 8 compiler threads?

doitsujin:
I don't think it's useful to engrave this in the config file since it's an implementation detail that is subject to change.

> for example, on a system with 12 threads available would it create 7 or 8 compiler threads?

7, this is integer division so nubers are truncated.

ThomasConnor:
Yeah thats a fair point, if its still subject to change then I'd completely agree that it'd be best not to engrave it in the config file. I prefer to know what "auto" settings are actually doing as they occasionally (in many applications) aren't set to good values so I think it'd still be helpful to leave in the note about leaving some cores free to the application for anyone else who may go modifying them as that'll always remain true and more won't always be better.

doitsujin:
The behaviour changed now and I added some more descriptive text in 67614917c353b65e0d9b3428e34955077a2379a9.

PR #2742, opened 2022-07-22
[d3d9] Add Direct3D9ForceHybridEnumeration empty export function.

kdobrowo:
The WLK tests for d3d9 call Direct3D9ForceHybridEnumeration function. When it is missing the exception is thrown and the application is terminated.
With empty stub is added the test starts the operation.

### DIFF
--- a/src/d3d9/d3d9.def
+++ b/src/d3d9/d3d9.def
@@ -23,3 +23,5 @@ EXPORTS
 
   DXVK_RegisterAnnotation @ 28257 NONAME
   DXVK_UnRegisterAnnotation @ 28258 NONAME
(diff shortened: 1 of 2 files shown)


K0bin:
It's marked as a draft so I assume you were going to do that later but anyway: Please explain the rational behind your change, especially which game relies on this.

kdobrowo:
Who should I add as a reviewer?

Joshua-Ashton:
You don't need to add anyone, it would just be nice to know what this fixes

Joshua-Ashton:
You don't need to add a reviewer, looks good.

PR #2769, opened 2022-07-31
[dxvk] Disable graphics pipeline libraries on 32-bit builds

doitsujin:
Can't say I'm overly happy with this but unless we find a way to make all D3D9 games work reliably with the feature enabled we don't really have much of a choice.

### DIFF
--- a/src/dxvk/dxvk_options.cpp
+++ b/src/dxvk/dxvk_options.cpp
@@ -1,12 +1,15 @@
 #include "dxvk_options.h"
 
+#include "../util/util_env.h"
+
 namespace dxvk {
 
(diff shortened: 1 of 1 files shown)


K0bin:
Wouldn't it be a better idea to just disable it for RE6? Everything else I tested has a very low number of shaders anyway (<2000).

Joshua-Ashton:
We should reach out to NV first and see if there are easy address space wins that make this more viable.

doitsujin:
@K0bin I'd argue that those games with a low shader count also don't really need graphics_pipeline_library to be reasonably enjoyable.

K0bin:
Fair enough.

K0bin (approved):

PR #2770, opened 2022-07-31
[d3d9, dxso] Refactor spec constants to use a bitfield layout 

Joshua-Ashton:
Part of a series of work for GPL refactoring.

### DIFF
--- a/src/d3d9/d3d9_spec_constants.h
+++ b/src/d3d9/d3d9_spec_constants.h
@@ -90,17 +90,22 @@ namespace dxvk {
   class D3D9ShaderSpecConstantManager {
   public:
     uint32_t get(SpirvModule &module, D3D9SpecConstantId id) {
+      return get(module, id, 0, 32);
+    }
+
+    uint32_t get(SpirvModule &module, D3D9SpecConstantId id, uint32_t bitOffset, uint32_t bitCount) {
       const auto &layout = D3D9SpecializationInfo::Layout[id];
 
       uint32_t val = getSpecConstDword(module, layout.dwordOffset);
-      if (layout.sizeInBits == 32)
+      bitCount = std::min(bitCount, layout.sizeInBits - bitOffset);
+
+      if (bitCount == 32)
         return val;
 
       return module.opBitFieldUExtract(
-        module.defIntType(32, 0),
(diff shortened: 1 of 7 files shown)


src/d3d9/d3d9_spec_constants.h
```diff
@@ -1,31 +1,119 @@
...
+      { 1, 23, 2 },  // PointMode
+      { 1, 25, 2 },  // VertexFogMode
+      { 1, 27, 2 },  // PixelFogMode
+      { 1, 29, 1 },  // FogEnabled
+
+      { 2, 0,  20 }, // SamplerNull
+      { 2, 20, 6 },  // ProjectionType
+
+      { 3, 0,  16 }, // VertexShaderBools
+      { 3, 16, 16 }, // VertexShaderBools
```
K0bin, line 61: This should probably be PixelShaderBools.

K0bin (commented):
Tiny nit, looks okay to me over all.

src/d3d9/d3d9_spec_constants.h
Joshua-Ashton: Fixed.

K0bin (approved):

src/d3d9/d3d9_spec_constants.h
```diff
@@ -1,31 +1,119 @@
...
+
+      return (data[layout.dwordOffset] & layout.mask()) >> layout.bitOffset;
+    }
+
+    std::array<uint32_t, MaxSpecDwords> data = {};
+  };
+
+  class D3D9ShaderSpecConstantManager {
+  public:
+    uint32_t get(SpirvModule &module, D3D9SpecConstantId id) {
```
DadSchoorse, line 92: add an offset and range arguments to avoid the double extract for some of the bit fields?

doitsujin: Strongly agree.

cc636ed953f50785d42ce52b4fbbb003e91e5f36 should implement this, and also fixes a bug with depth and null samplers being broken in vertex shaders.

doitsujin (changes requested):
See Schoorse's comment.

PR #2773, opened 2022-08-01
[build] Try to not break with -Ofast

pchome:
In case someone will use -Ofast or -ffast-math (gcc),
which will enable -ffinite-math-only and likely will break
x != x, std::isfinite, std::isnan etc.


doitsujin:
Why do we need this? `-Ofast` isn't the default in any build environment, it's not conformant, and it's also completely unnecessary because we literally don't have any hot FPU-heavy code paths.

PR #2777, opened 2022-08-01
make destructor virtual

neheb:
Fixes usage with std::make_unique.

```
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/c++/v1/__memory/unique_ptr.h:54:5: warning: delete called on non-final 'dxvk::D3D10ShaderReflectionType' that has virtual functions but non-virtual destructor [-Wdelete-non-abstract-non-virtual-dtor]
    delete __ptr;
    ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/c++/v1/__memory/unique_ptr.h:315:7: note: in instantiation of member function 'std::default_delete<dxvk::D3D10ShaderReflectionType>::operator()' requested here
      __ptr_.second()(__tmp);
      ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/c++/v1/__memory/unique_ptr.h:269:19: note: in instantiation of member function 'std::unique_ptr<dxvk::D3D10ShaderReflectionType>::reset' requested here
  ~unique_ptr() { reset(); }
                  ^
../src/d3d10/d3d10_reflection.cpp:63:9: note: in instantiation of member function 'std::unique_ptr<dxvk::D3D10ShaderReflectionType>::~unique_ptr' requested here
        std::make_unique<D3D10ShaderReflectionType>(pMemberType) }).first;
        ^
1 warning and 1 error generated.
[6/24] Compiling C++ object src/d3d10/d3d10_1.dll.p/d3d10_reflection.cpp.obj
```

### DIFF
--- a/src/d3d10/d3d10_reflection.h
+++ b/src/d3d10/d3d10_reflection.h
@@ -17,7 +17,7 @@ namespace dxvk {
     D3D10ShaderReflectionType(
             ID3D11ShaderReflectionType*     d3d11);
     
-    ~D3D10ShaderReflectionType();
+    virtual ~D3D10ShaderReflectionType();
 
     HRESULT STDMETHODCALLTYPE GetDesc(
             D3D10_SHADER_TYPE_DESC*         pDesc);


Joshua-Ashton:
No. You cannot add a virtual destructor to an existing interface without breaking ABI.

We don't control the interface, and as this is a pointer to a parent type with a destructor so this is a non-issue anyway.

Probably makes sense to classify this type as `final` though.

doitsujin:
The "proper" fix is to not use make_unique and friends on COM objects. The correct way to free those is by calling `object->Release()`.

BNieuwenhuizen:
Sounds like one proper fix is to use a custom deleter that calls Release when called. Something like 

```
struct ReleaseDeleter {
   template<typename T>
   void operator()(T* t) {
      t->Release();
   }
};

std::unique_ptr<Type, ReleaseDeleter> ptr;
```

Guy1524:
> Sounds like one proper fix is to use a custom deleter that calls Release when called. Something like
> 
> ```
> struct ReleaseDeleter {
>    template<typename T>
>    void operator()(T* t) {
>       t->Release();
>    }
> };
> 
> std::unique_ptr<Type, ReleaseDeleter> ptr;
> ```

This is what DXVK's Com object does https://github.com/doitsujin/dxvk/blob/master/src/util/com/com_pointer.h#L98

Joshua-Ashton:
> The "proper" fix is to not use make_unique and friends on COM objects. The correct way to free those is by calling object->Release(), or use smart COM pointers such as ComPtr in MSVC, or our own Com.

`ID3D10ShaderReflectionType` is not COM, it does not inherit from `IUnknown`, so the correct solution is actually `final` like I said.

PR #2783, opened 2022-08-03
[util] Use raw tzcnt for BitMask iterator

DadSchoorse:
Dereferencing an end iterator is UB, so we don't have to care about the 0 case.


Joshua-Ashton:
I might move this out to a helper function so its cleaner for DXVK native on non-x86_64 platforms

PR #2786, opened 2022-08-04
[meta] It's 2022 now

artemyto:


artemyto:
Maybe it is a good idea to add Robin Kertels to the list of copyrights?

ArturVRSampaio (approved):

ojab:
*2023

PR #2776, opened 2022-08-01
remove dllexport on redeclaration and use WINAPI

neheb:
Fixes MinGW warning.

```
[172/277] Compiling C++ object src/d3d11/d3d11.dll.p/d3d11_main.cpp.obj
../src/d3d11/d3d11_main.cpp:206:31: warning: redeclaration of 'D3D11CreateDevice' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]
  DLLEXPORT HRESULT __stdcall D3D11CreateDevice(
                              ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/d3d11.h:10976:16: note: previous declaration is here
HRESULT WINAPI D3D11CreateDevice(IDXGIAdapter*,D3D_DRIVER_TYPE,HMODULE,UINT,const D3D_FEATURE_LEVEL*,
               ^
../src/d3d11/d3d11_main.cpp:225:31: warning: redeclaration of 'D3D11CreateDeviceAndSwapChain' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]
  DLLEXPORT HRESULT __stdcall D3D11CreateDeviceAndSwapChain(
                              ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/d3d11.h:10981:20: note: previous declaration is here
HRESULT __stdcall  D3D11CreateDeviceAndSwapChain(IDXGIAdapter *adapter,D3D_DRIVER_TYPE driver_type,HMODULE swrast,UINT flags,const D3D_FEATURE_LEVEL *feature_levels,UINT levels,UINT sdk_version,const DXGI_SWAP_CHAIN_DESC *swapchain_desc,IDXGISwapChain **swapchain,ID3D11Device **device,D3D_FEATURE_LEVEL *obtained_feature_level,ID3D11DeviceContext **immediate_context);
                   ^
2 warnings generated.
```

PR #2775, opened 2022-08-01
remove DLLEXPORT on redeclaration

neheb:
Fixes warning under MinGW.

```
../src/dxgi/dxgi_main.cpp:25:31: warning: redeclaration of 'CreateDXGIFactory2' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]
  DLLEXPORT HRESULT __stdcall CreateDXGIFactory2(UINT Flags, REFIID riid, void **ppFactory) {
                              ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/dxgi1_3.h:1941:20: note: previous declaration is here
HRESULT __stdcall  CreateDXGIFactory2(UINT flags,REFIID iid,void **factory);
                   ^
../src/dxgi/dxgi_main.cpp:30:31: warning: redeclaration of 'CreateDXGIFactory1' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]
  DLLEXPORT HRESULT __stdcall CreateDXGIFactory1(REFIID riid, void **ppFactory) {
                              ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/dxgi.h:1973:20: note: previous declaration is here
HRESULT __stdcall  CreateDXGIFactory1(REFIID riid,void **factory);
                   ^
../src/dxgi/dxgi_main.cpp:34:31: warning: redeclaration of 'CreateDXGIFactory' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]
  DLLEXPORT HRESULT __stdcall CreateDXGIFactory(REFIID riid, void **ppFactory) {
                              ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/dxgi.h:1971:20: note: previous declaration is here
HRESULT __stdcall  CreateDXGIFactory(REFIID riid,void **factory);
                   ^
../src/dxgi/dxgi_main.cpp:48:31: warning: redeclaration of 'DXGIGetDebugInterface1' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]
  DLLEXPORT HRESULT __stdcall DXGIGetDebugInterface1(UINT Flags, REFIID riid, void **ppDebug) {
                              ^
C:/Users/Mangix/scoop/apps/msys2/2022-01-28/clang32/include/dxgi1_3.h:1943:20: note: previous declaration is here
[...]

PR #2788, opened 2022-08-05
[build] Fix dllexport warnings for llvm-mingw

pchome:
e.g. `../src/dxgi/dxgi_main.cpp:34:31: warning: redeclaration of 'CreateDXGIFactory' should not add 'dllexport' attribute [-Wdll-attribute-on-redeclaration]`

### DIFF
--- a/src/dxgi/dxgi_include.h
+++ b/src/dxgi/dxgi_include.h
@@ -3,7 +3,7 @@
 //for some reason we need to specify __declspec(dllexport) for MinGW
 #if defined(__WINE__)
   #define DLLEXPORT __attribute__((visibility("default")))
-#elif defined(_MSC_VER)
+#elif defined(_MSC_VER) || defined(__clang__)
   #define DLLEXPORT
 #else
   #define DLLEXPORT __declspec(dllexport)


pchome:
Same warnings as in #2775 and #2776

Joshua-Ashton:
Does it break stdcall aliasing on 32-bir builds?

pchome:
what found in dlls
##### Before fix
`CreateDXGIFactory.CreateDXGIFactory1.CreateDXGIFactory1@8.CreateDXGIFactory2.CreateDXGIFactory2@12.CreateDXGIFactory@8.DXGIDeclareAdapterRemovalSupport.DXGIDeclareAdapterRemovalSupport@0.DXGIGetDebugInterface1.DXGIGetDebugInterface1@12`
##### After fix
`CreateDXGIFactory.CreateDXGIFactory1.CreateDXGIFactory2.DXGIDeclareAdapterRemovalSupport.DXGIGetDebugInterface1`
##### 1.9.1 release
`DXGI.DLL.CreateDXGIFactory.CreateDXGIFactory1.CreateDXGIFactory1@8.CreateDXGIFactory2.CreateDXGIFactory2@12.CreateDXGIFactory@8.DXGIDeclareAdapterRemovalSupport.DXGIDeclareAdapterRemovalSupport@0.DXGIGetDebugInterface1.DXGIGetDebugInterface1@12`

so seems it does.

Joshua-Ashton:
Sooo same reason I closed the last PR then...

pchome:
Then I guess it should be just `-Wno-dll-attribute-on-redeclaration` in meson.build.

pchome:
I missed this but there are other warnings
```
[94/126] Linking target src/dxgi/dxgi.dll
ld.lld: warning: duplicate /export option: CreateDXGIFactory
ld.lld: warning: duplicate /export option: CreateDXGIFactory1
ld.lld: warning: duplicate /export option: CreateDXGIFactory2
ld.lld: warning: duplicate /export option: DXGIDeclareAdapterRemovalSupport
ld.lld: warning: duplicate /export option: DXGIGetDebugInterface1
```
I don't know if this correct but if I add
```
    CreateDXGIFactory1@8  = CreateDXGIFactory1
    CreateDXGIFactory2@12 = CreateDXGIFactory2
    CreateDXGIFactory@8   = CreateDXGIFactory
    DXGIDeclareAdapterRemovalSupport@0 = DXGIDeclareAdapterRemovalSupport
    DXGIGetDebugInterface1@12          = DXGIGetDebugInterface1
```
in dxgi.def file then after compilation with this patch dxgi.dll file seem contain them. But 64bit dxgi.dll have them too.
I don't know which tool can show them so checked in `mcview`.

If you want to check llvm-mingw it can be easily installed in few minutes
* just download from https://github.com/mstorsjo/llvm-mingw/releases and extract somewhere (I used [msvcrt-ubuntu-18.04-x86_64](https://github.com/mstorsjo/llvm-mingw/releases/download/20220802/llvm-mingw-20220802-msvcrt-ubuntu-18.04-x86_64.tar.xz))
* copy cross-files into `~/.local/share/meson/cross/` and use by name as usual from everywhere (change /tmp/llvm-mingw path)
* build using one of provided cross-file names (named as .txt because of github)

[llvm-win64.txt](https://github.com/doitsujin/dxvk/files/9275152/llvm-win64.txt)
[llvm-win32.txt](https://github.com/doitsujin/dxvk/files/9275153/llvm-win32.txt)

pchome:
Well, I dug a bit and from what I understood we do not really need decorated aliases in symbol export table. What we really need is undecorated symbols and sometimes fixed ordinals, so that `Name@nn` exported symbols just a garbage left after all this hacks with --add-stdcall-alias, *def, etc.
This can be fixed with just `-Wl,--kill-at`.

And we should care only about dxvk linked ok internally, and it is, in most cases, because e.g. `-ldxgi` will use imports from `i686-w64-mingw32/lib/libdxgi.a` which have `undecorate` type (unlike `noprefix` in generated `dxgi.a`)
```
File: dxgi.dll
Format: COFF-import-file
Type: code
Name type: undecorate
Symbol: __imp__D3DKMTOpenResource@4
Symbol: _D3DKMTOpenResource@4
```
_command used_: `$ llvm-readobj ../i686-w64-mingw32/lib/libdxgi.a | less`

Command e.g. `i686-w64-mingw32-objdump -x  d3d11.dll | less` on unstripped dll file shows that undecorated exported symbols linked to decorated symbols in symbol table by addr + offset (0x1000 in my case).

So, dropping declspec and replacing `--add-stdcall-alias` and maybe `--enable-stdcall-fixup` with `--kill-at` should do the trick IMHO.

pchome:
So with only this patch and nothing else changed llvm-mingw generates perfectly fine dlls, even all generated *.a import files contain `undecorate` type (so `--enable-stdcall-fixup` is redundant here and `--add-stdcall-alias` is unsupported).

There can be only problem if something want to use dxvk dlls with decorated symbol names in `GetProcAddress`.

PR #2790, opened 2022-08-05
WIP: Support for VK_EXT_attachment_feedback_loop_layout

Joshua-Ashton:
Still need to figure out what to do for self dependency + `VK_DEPENDENCY_FEEDBACK_LOOP` with the dynamic rendering changes.


src/dxvk/dxvk_context.h
```diff
@@ -87,10 +87,23 @@ namespace dxvk {
      * \param [in] targets Render targets to bind
      */
     void bindRenderTargets(
-            DxvkRenderTargets&&   targets) {
+            DxvkRenderTargets&&   targets,
+            bool                  hazardColor = false,
+            bool                  hazardDepth = false) {
```
doitsujin, line 92: I suppose the plan is to not transition any attachment that's in `FEEDBACK_LOOP_OPTIMAL` back to `ATTACHMENT_OPTIMAL` at any point?

If so, keeping the function argument is fine, but this should really be something like `VkImageAspectFlags feedbackLoop` or something.

Also, please no default argument values here.

src/dxvk/dxvk_context.h
```diff
@@ -87,10 +87,23 @@ namespace dxvk {
...
       // Set up default render pass ops
       m_state.om.renderTargets = std::move(targets);
 
+      if (unlikely(m_state.gp.state.om.hazardColor() != hazardColor ||
+                   m_state.gp.state.om.hazardDepth() != hazardDepth)) {
+        m_state.gp.state.om = DxvkOmInfo(
```
doitsujin, line 98: Just add a function `setFeedbackLoop` to `DxvkOmInfo`.

src/dxvk/dxvk_graphics.cpp
```diff
@@ -396,14 +403,19 @@ namespace dxvk {
       dyInfo.pDynamicStates     = &dynamicState;
     }
 
+    VkPipelineCreateFlags flags = VK_PIPELINE_CREATE_LIBRARY_BIT_KHR;
+    if (device->features().extAttachmentFeedbackLoopLayout.attachmentFeedbackLoopLayout)
+      flags |= VK_PIPELINE_CREATE_COLOR_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT
+            |  VK_PIPELINE_CREATE_DEPTH_STENCIL_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT;
```
doitsujin, line 409: We know the exact bits here because they are stored in `DxvkGraphicsPipelineFragmentOutputState`, so we should check those rather than unconditionally enabling the flags.

This would have been reasonable for fragment *shader* pipelines, but we don't need to pass the flags there.

src/dxvk/dxvk_graphics.h
```diff
@@ -119,6 +119,9 @@ namespace dxvk {
     std::array<VkPipelineColorBlendAttachmentState, MaxNumRenderTargets> cbAttachments  = { };
     std::array<VkFormat,                            MaxNumRenderTargets> rtColorFormats = { };
 
+    VkBool32                                        hazardColor = VK_FALSE;
+    VkBool32                                        hazardDepth = VK_FALSE;
```
doitsujin, line 123: This should really be `VkImageAspectFlags feedbackLoop = 0;`

src/dxvk/dxvk_graphics_state.h
```diff
@@ -434,9 +434,13 @@ namespace dxvk {
 
     DxvkOmInfo(
             VkBool32          enableLogicOp,
-            VkLogicOp         logicOp)
+            VkLogicOp         logicOp,
+            bool              hazardColor,
+            bool              hazardDepth)
```
doitsujin, line 439: `VkImageAspectFlags feedbackLoop`

src/dxvk/dxvk_graphics_state.h
```diff
@@ -447,11 +451,21 @@ namespace dxvk {
...
+      return !!m_hazardDepth;
+    }
+
   private:
 
     uint16_t m_enableLogicOp          : 1;
     uint16_t m_logicOp                : 4;
-    uint16_t m_reserved               : 11;
+    uint16_t m_hazardColor            : 1;
+    uint16_t m_hazardDepth            : 1;
```
doitsujin, line 467: `uint16_t feedbackLoop : 2` and then add a function to get/set it as `VkImageAspectFlags`

doitsujin (changes requested):
The direction is good but I'm going to be rather picky about style things.

If you're going to add a parameter to `bindRenderTargets`, maybe do all of that in a separate commit and fix up all the spots where we call the function at the same time.

src/dxvk/dxvk_image.h
```diff
@@ -251,8 +251,13 @@ namespace dxvk {
      * \returns A compatible image layout
      */
     VkImageLayout pickLayout(VkImageLayout layout) const {
-      return m_info.layout == VK_IMAGE_LAYOUT_GENERAL
-        ? VK_IMAGE_LAYOUT_GENERAL : layout;
+      if (unlikely(m_info.layout == VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT))
+        return VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT;
+
+      if (unlikely(m_info.layout == VK_IMAGE_LAYOUT_GENERAL))
```
doitsujin, line 257: This isn't actually *that* unlikely. I'm fine keeping it for the feedback loop layout code above though.

src/dxvk/dxvk_image.h
```diff
@@ -251,8 +251,13 @@ namespace dxvk {
      * \returns A compatible image layout
      */
     VkImageLayout pickLayout(VkImageLayout layout) const {
-      return m_info.layout == VK_IMAGE_LAYOUT_GENERAL
-        ? VK_IMAGE_LAYOUT_GENERAL : layout;
+      if (unlikely(m_info.layout == VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT))
+        return VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT;
```
doitsujin, line 255: If `layout` is any of `TRANSFER_SRC_OPTIMAL`, `TRANSFER_DST_OPTIMAL` or `GENERAL`, we need to return `layout` for this.

src/dxvk/dxvk_graphics.cpp
```diff
@@ -396,14 +400,23 @@ namespace dxvk {
       dyInfo.pDynamicStates     = &dynamicState;
     }
 
+    VkPipelineCreateFlags flags = VK_PIPELINE_CREATE_LIBRARY_BIT_KHR;
+    if (m_device->features().extAttachmentFeedbackLoopLayout.attachmentFeedbackLoopLayout) {
```
doitsujin, line 404: Do we need this check? The front-end should never set the bits if the feature is unsupported.

The logic here is that all pipeline state should be normalized against whatever the device supports *before* we end up looking up pipelines, see other things like conservative rasterization etc.

src/dxvk/dxvk_graphics.cpp
```diff
@@ -1159,6 +1172,14 @@ namespace dxvk {
         stageInfo.addStage(VK_SHADER_STAGE_FRAGMENT_BIT, getShaderCode(m_shaders.fs, key.shState.fsInfo), &key.scState.scInfo);
     }
 
+    if (m_device->features().extAttachmentFeedbackLoopLayout.attachmentFeedbackLoopLayout) {
```
doitsujin, line 1175: Same here.

src/dxvk/dxvk_graphics_state.h
```diff
@@ -433,10 +433,12 @@ namespace dxvk {
...
-            VkBool32          enableLogicOp,
-            VkLogicOp         logicOp)
-    : m_enableLogicOp (uint16_t(enableLogicOp)),
-      m_logicOp       (uint16_t(logicOp)),
+            VkBool32           enableLogicOp,
+            VkLogicOp          logicOp,
+            VkImageAspectFlags feedbackLoop)
+    : m_enableLogicOp      (uint16_t(enableLogicOp)),
+      m_logicOp            (uint16_t(logicOp)),
+      m_feedbackLoop(uint16_t(feedbackLoop)),
```
doitsujin, line 441: formatting :frog:

Joshua-Ashton: fucking vscode

doitsujin (approved):
Assuming this works: LGTM.

Joshua-Ashton:
Works fine for me locally. Hopefully ends up working out on Intel and such whom need the flags.

PR #2792, opened 2022-08-06
[util] Disable unmapping for Saints Row 2

Blisto91:
Game doesn't like the unmapping changes. Might as well keep it semi broken (the native version that is) instead of fully broken.


K0bin (approved):
Stupid game. >:(

PR #2796, opened 2022-08-08
D3D9 flat shading rework

doitsujin:
This greatly reduces the number of shaders in some games, and simplifies a bunch of front-end code. The only downside is that pipelines with flat shading have to be late-compiled because we need to patch the fragment shader, but they'll still work with the state cache.

I only tested the flat shading part by hacking some code into `test_d3d9_triangle.cpp`, since I don't know any real-world game which uses this feature. The fixed-function path is untested, but the idea is the same.

With these changes, RE6 goes from 10.3k shaders in the menu down to 7.7k.


src/dxvk/dxvk_swapchain_blitter.cpp
```diff
@@ -124,6 +124,7 @@ namespace dxvk {
     rsState.depthBiasEnable    = VK_FALSE;
     rsState.conservativeMode   = VK_CONSERVATIVE_RASTERIZATION_MODE_DISABLED_EXT;
     rsState.sampleCount        = VK_SAMPLE_COUNT_1_BIT;
+    rsState.flatShading        = VK_FALSE;
```
Joshua-Ashton, line 127: I would just make this `DxvkRasterizerState blah = {}` like you did above

src/dxvk/hud/dxvk_hud.cpp
```diff
@@ -22,6 +22,7 @@ namespace dxvk::hud {
     m_rsState.depthBiasEnable   = VK_FALSE;
     m_rsState.conservativeMode  = VK_CONSERVATIVE_RASTERIZATION_MODE_DISABLED_EXT;
     m_rsState.sampleCount       = VK_SAMPLE_COUNT_1_BIT;
+    m_rsState.flatShading       = VK_FALSE;
```
Joshua-Ashton, line 25: Same here, avoids us forgetting it.

Joshua-Ashton (approved):
Looks fine

PR #2800, opened 2022-08-08
Limit D3D9 staging memory

K0bin:
Games like Witcher 2 and modded Mass Effect 2 peak at 190MB and 170MB of memory used for staging buffers during loading screens. In the case of UE3 games, that happens at the same time as a very high peak for mapped resource memory, so this could be potentially problematic. Those games don't crash for me but I'd still prefer to get this under control.

The idea of this PR is to add a limit to how much staging memory an application is allowed to use and stall once we've exceeded that. I've settled on 16MB which was never an issue during gameplay in my testing.

Ideally we'd just unmap the memory and just not care but that doesn't work on Nvidia.

To track how much memory we're using for staging buffers, I add a new blank class that inherits from DxvkResource.
That allows us to check whether it's still in use (ie the command list has not been reset by the fence thread).
It's preferable to keep those objects around compared to keeping references to the actual staging buffers and checking those.
A prettier version of this would somehow inform DxvkCsThread of the last sequence number completed by the GPU. That's tricky to implement however, especially with DeferredContexts.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4596,7 +4596,8 @@ namespace dxvk {
 
     DxvkBufferSliceHandle physSlice;
 
-    if ((Flags & D3DLOCK_DISCARD) && (directMapping || needsReadback)) {
+    bool smallEnoughForDiscard = !env::is32BitHostPlatform() || desc.Size < 2 << 20;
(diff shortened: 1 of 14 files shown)


doitsujin:
I don't think we need a custom staging buffer implementation for this. The `DxvkMarker` thing is kind of clever though.

I'll try something, will submit PR if it ends up working.

K0bin:
Superseded by #2808

PR #2802, opened 2022-08-08
Pipeline lifetime tracking on 32-bit

doitsujin:
In order to reduce address space issues with graphics pipeline libraries in 32-bit apps, we need to aggressively destroy libraries as well as linked pipelines.

Pipeline libraries that are compiled on the background threads after a `Create*Shader` call will be destroyed immediately, rather than written back. When recreating these, we'll rely on `VK_EXT_shader_module_identifier` and the driver's disk cache to give us a new one very quickly. In my testing, this is still orders of magitude faster than compiling a pipeline (100-200 µs per pipeline library) and does not lead to any noticeable stutter in games.

Base pipelines will be kept alive as long as they are potentially in use by the GPU **and** as long as there are compile jobs for optimized pipelines on the worker threads. This way we avoid extreme scenarios like compiling and destroying the same set of base pipelines every frame if the app is CPU-bound. We never destroy optimized pipelines.

This is fairly ugly and it is rather unfortunate that this is needed, but in RE6 we save ~500MB of address space in the menu on Nvidia, and even more during gameplay.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -286,6 +286,20 @@
 # dxvk.enableGraphicsPipelineLibrary = Auto
 
 
+# Controls pipeline lifetime tracking
+#
+# If enabled, pipeline libraries will be freed aggressively in order
+# save memory and address space. Has no effect if graphics pipeline
+# libraries are not supported or disabled.
+#
+# Supported values:
+# - Auto: Enable tracking for 32-bit applications only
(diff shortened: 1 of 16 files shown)


Joshua-Ashton:
I would make a single function somewhere so we can toggle this on/off for debugging instead of multiple env::blah. And maybe put behind a setting.

PR #2804, opened 2022-08-08
Fix shared textures that export both NT and KMT handles.

Guy1524:
This was being triggered in Crysis Remastered, but actually all games using NT handles were broken, it just so happens that the video paths seem to prefer NT handles.


src/d3d11/d3d11_texture.cpp
```diff
@@ -627,7 +627,7 @@ namespace dxvk {
       Logger::warn("D3D11: Failed to write shared resource info for a texture");
     }
 
-    if (hSharedHandle != INVALID_HANDLE_VALUE)
+    if (m_desc.MiscFlags & D3D11_RESOURCE_MISC_SHARED && hSharedHandle != INVALID_HANDLE_VALUE)
```
doitsujin, line 630: can we have some parentheses here? I don't want to have to wrap my head around operator precedence when looking at this code.

doitsujin (changes requested):
Smol nit, will merge when fixed.

Issue #2803, opened 2022-08-08
Saints Row 2 shadow/light regression


Blisto91:
After the unmap merge the character have a bug where half his face is darker than the other half and the textures or skin shader looks different.

Haven't tested specific unmap commits yet but it's broken with at least the latest commit after the merge (ea76bfd) and works with the master commit before the merge (116feca)

<details>
  <summary>Screenshots</summary>
  
![Screenshot_20220808_205100](https://user-images.githubusercontent.com/47954800/183496211-9d575cfb-5f64-47f7-8f98-3dd93542e3fe.png)
![Screenshot_20220808_205407](https://user-images.githubusercontent.com/47954800/183496230-e9f19b8d-97ae-41a3-b2c9-dab109d355c8.png)
</details>

### Software information
Saints Row2

<details>
  <summary>Settings</summary>
  
![Screenshot_20220808_211816](https://user-images.githubusercontent.com/47954800/183497029-d5a6e7dd-db98-4b4b-9ac3-235df2285851.png)
![Screenshot_20220808_211844](https://user-images.githubusercontent.com/47954800/183497045-8844af96-bceb-4975-bb0d-c995fb40f8a4.png)
</details>

### System information
- GPU: Nvidia 960
- Driver: Vulkan beta/dev 515.49.10
- Wine version: proton experimental
- DXVK version: ea76bfd (bugged) 116feca (working)

### Apitrace file(s)
https://mega.nz/file/Y05hgLRK#augKh68Aq6kJNS3EzLoIFMHPecymB1R1TXd8wggg9JY

PR #2805, opened 2022-08-08
Fix synchronization when copying from system mem texture to default texture 

K0bin:
Fixes: #2803

SR2 does the following:

- Render to texture
- GetRenderTargetData to copy that to a POOL_SYSTEMMEM texture
- UpdateSurface to copy from the POOL_SYSTEMMEM to a POOL_DEFAULT texture

I've never seen this pattern and it's honestly pretty stupid given that you could just call StretchRect and skip the system memory copy, so we'll just stall.


doitsujin (approved):

K0bin:
I removed a debug line.

PR #2806, opened 2022-08-09
[util] Bringup of utils on non-Windows platforms

Joshua-Ashton:
Gets everything building on non-Windows (aside from fps limiter + wsi to come later)

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -889,7 +889,7 @@ namespace dxvk {
       filePath = "dxvk.conf";
     
     // Open the file if it exists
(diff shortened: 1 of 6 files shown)


doitsujin (approved):

pchome:
What about something like this (to use in places with "double conversion")
```cpp
  inline const char*  tocpath(std::wstring mbs) { return tows(mbs.c_str()).c_str(); }
  inline const char*  tocpath(std::string mbs) { return mbs.c_str(); }
```
?

Joshua-Ashton:
There's a reasin that's not done and it's because that's returning a dead ptr

PR #2807, opened 2022-08-09
[wsi] Implement new WSI abstraction

Joshua-Ashton:
Implements a WSI abstraction and uses it for DXGI. This code has been used in dxvk native for a while, cleaned it up a little bit.

Primarily looking for feedback that this is going in the right direction.

Didn't bother bringing over D3D9 yet, need to rebase and rework some things there.


src/wsi/wsi_window.h
```diff
@@ -0,0 +1,125 @@
+#pragma once
+
+// The windows.h will define HWND/HMONITOR
+// to be types that it wants for the wsi we are compiling against.
+// ie. HWND     = their own window data ptr (eg. SDL_Window)
+//     HMONTIOR = their own monitor data ptr/a display index
```
doitsujin, line 6: This is **really** ugly.

I think it's better to have something like `wsi/win32/wsi_win32.h` where we include the real `windows.h`, and for other WSI implementations we could have `wsi/sdl2/wsi_sdl2.h` do something like
```c++
// this is kinda fine since both are pointer types
using HWND = SDL_Window*;

// this is more annoying since SDL doesn't have a display type and 0 is a valid display index
struct WsiSdl2Monitor;
using HMONITOR = WsiSdl2Monitor*;

inline HMONITOR toMonitorHandle(int index) {
  uintptr_t handleValue = index < 0 ? uintptr_t(0) : uintptr_t(index) + 1;
  return reinterpret_cast<HMONITOR>(index);
}

inline int fromMonitorHandle(HMONITOR handle) {
  auto handleValue = reinterpret_cast<uintptr_t>(handle);
  return (handleValue && handleValue != INVALID_HANDLE_VALUE)
    ? int32_t(handleValue - 1)
    : -1;
}
```

And then have a header `wsi/wsi.h` or something that includes the platform-specific header.

If we need more Win32 types (we probably do), those should be defined in another common header that we include on non-win32, and that is preferably **not** called `windows.h`.

src/wsi/wsi_window.h
```diff
@@ -0,0 +1,125 @@
...
+
+#include "wsi_mode.h"
+
+namespace dxvk::wsi {
+
+  /**
+    * \brief Impl-dependent state
+    */
+  struct DxvkWindowState {
+#ifdef DXVK_WSI_WIN32
```
doitsujin, line 18: This whole struct should be in platform-specific headers.

doitsujin (commented):
I think this is *mostly* going the right way, but the separation between Win32 and common stuff seems a bit wonky right now.

I'd also like to reduce the number of files, `wsi_mode` and `wsi_monitor` can really be one unit, and so can `wsi_presenter` and `wsi_window`.

All the places that actually use WSI functions should really just include a common `wsi.h` or `wsi_include.h` or whatever that just includes `wsi/win32/wsi_win32_*.h`, or `wsi/sd2/wsi_sdl2_*.h`, depending on configuration.

src/wsi/wsi_mode.h
```diff
@@ -0,0 +1,70 @@
...
+
+  /**
+   * \brief Display mode
+   */
+  struct WsiMode {
+    uint32_t     width;
+    uint32_t     height;
+    WsiRational  refreshRate;
+    uint32_t     bitsPerPixel;
+    bool         interlaced;
```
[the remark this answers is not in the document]
Joshua-Ashton: 1080i TVs are still common here... :(

src/wsi/wsi_window.h
Joshua-Ashton: Agree.

src/wsi/wsi_window.h
doitsujin: Can we at least get rid of the comment then to make it less obvious how utterly cursed and broken the whole thing is?

Joshua-Ashton: Yes, it is also not true anymore actually since this got refactored. HWND/HMONITOR Is always a HANDLE so size is consistent with Windows.

The WSI headers define helper functions to cast to/from.

doitsujin: That's okay then.

Joshua-Ashton:
> All the places that actually use WSI functions should really just include a common wsi.h or wsi_include.h or whatever that just includes wsi/win32/wsi_win32_*.h, or wsi/sd2/wsi_sdl2_*.h, depending on configuration.

This is handled by the fake `windows.h` for native already fwiw. The native headers are not included in this PR, just migrating Win32 over to it for now.

Joshua-Ashton:
I'll still add one for `DxvkWindowState` though.

Joshua-Ashton:
Implemented it for D3D9, and addressed the feedback. Opening up for proper review now.

src/dxgi/dxgi_monitor.h
```diff
@@ -55,4 +57,32 @@ namespace dxvk {
...
+   * \brief Converts two display modes
+   */
+  inline void ConvertDisplayMode(
+    const wsi::WsiMode&          WsiMode,
+          DXGI_MODE_DESC1*       pDxgiMode) {
+    pDxgiMode->Width            = WsiMode.width;
+    pDxgiMode->Height           = WsiMode.height;
+    pDxgiMode->RefreshRate      = DXGI_RATIONAL{ WsiMode.refreshRate.numerator, WsiMode.refreshRate.denominator };
+    pDxgiMode->Format           = DXGI_FORMAT_R8G8B8A8_UNORM_SRGB; // FIXME
+    pDxgiMode->ScanlineOrdering = DXGI_MODE_SCANLINE_ORDER_PROGRESSIVE;
```
doitsujin, line 70: Could set `DXGI_MODE_SCANLINE_ORDER_UPPER_FIELD_FIRST` if the mode at hand is interlaced

src/dxgi/dxgi_monitor.h
```diff
@@ -55,4 +57,32 @@ namespace dxvk {
...
+   * \brief Converts two display modes
+   */
+  inline void ConvertDisplayMode(
+    const DXGI_MODE_DESC1&        DxgiMode,
+          wsi::WsiMode*           pWsiMode) {
+    pWsiMode->width        = DxgiMode.Width;
+    pWsiMode->height       = DxgiMode.Height;
+    pWsiMode->refreshRate  = wsi::WsiRational{ DxgiMode.RefreshRate.Numerator, DxgiMode.RefreshRate.Denominator };
+    pWsiMode->bitsPerPixel = GetMonitorFormatBpp(DxgiMode.Format);
+    pWsiMode->interlaced   = false;
```
doitsujin, line 85:
```
pWsiMode->interlaced = DxgiMode.ScanlineOrdering == DXGI_MODE_SCANLINE_ORDER_UPPER_FIELD_FIRST
                    || DxgiMode.ScanlineOrdering == DXGI_MODE_SCANLINE_ORDER_LOWER_FIELD_FIRST;
```
would be more correct here

src/wsi/win32/wsi_window_win32.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+
+  VkResult createSurface(
+          HWND                hWindow,
+    const Rc<vk::InstanceFn>& vki,
+          VkSurfaceKHR*       pSurface) {
+    HINSTANCE hInstance = reinterpret_cast<HINSTANCE>(
+      GetWindowLongPtr(hWindow, GWLP_HINSTANCE));
+
+    VkWin32SurfaceCreateInfoKHR info;
```
doitsujin, line 275: info = { VK_STRUCTURE_TYPE_WIN32_SURFACE_CREATE_INFO_KHR };

src/dxgi/dxgi_monitor.h
```diff
@@ -55,4 +57,32 @@ namespace dxvk {
   uint32_t GetMonitorFormatBpp(
           DXGI_FORMAT             Format);
 
+  /**
+   * \brief Converts two display modes
+   */
+  inline void ConvertDisplayMode(
```
doitsujin, line 63: In general I'd prefer if this just returned a `DXGI_MODE_DESC1` directly

src/dxgi/dxgi_monitor.h
```diff
@@ -55,4 +57,32 @@ namespace dxvk {
...
+    pDxgiMode->Format           = DXGI_FORMAT_R8G8B8A8_UNORM_SRGB; // FIXME
+    pDxgiMode->ScanlineOrdering = DXGI_MODE_SCANLINE_ORDER_PROGRESSIVE;
+    pDxgiMode->Scaling          = DXGI_MODE_SCALING_UNSPECIFIED;
+    pDxgiMode->Stereo           = FALSE;
+  }
+
+  /**
+   * \brief Converts two display modes
+   */
+  inline void ConvertDisplayMode(
```
doitsujin, line 78: Same here, just return a `wsi::WsiMode`

src/dxgi/dxgi_output.cpp
```diff
@@ -300,28 +294,27 @@ namespace dxvk {
...
     uint32_t srcModeId = 0;
     uint32_t dstModeId = 0;
     
     std::vector<DXGI_MODE_DESC1> modeList;
     
-    while (GetMonitorDisplayMode(m_monitor, srcModeId++, &devMode)) {
+    while (wsi::getDisplayMode(m_monitor, srcModeId++, &devMode)) {
       // Skip interlaced modes altogether
-      if (devMode.dmDisplayFlags & DM_INTERLACED)
+      if (devMode.interlaced)
```
doitsujin, line 306: `&& !(Flags & DXGI_ENUM_MODES_INTERLACED)` for correctness

src/dxgi/dxgi_output.cpp
```diff
@@ -300,28 +294,27 @@ namespace dxvk {
...
-      if (devMode.dmBitsPerPel != GetMonitorFormatBpp(EnumFormat))
+      if (devMode.bitsPerPixel != GetMonitorFormatBpp(EnumFormat))
         continue;
       
       if (pDesc != nullptr) {
         DXGI_MODE_DESC1 mode;
-        mode.Width            = devMode.dmPelsWidth;
-        mode.Height           = devMode.dmPelsHeight;
-        mode.RefreshRate      = { devMode.dmDisplayFrequency * 1000, 1000 };
+        mode.Width            = devMode.width;
```
doitsujin, line 315: Any reason not to use `ConvertDisplayMode` here?
Can fix up the format afterwards.

src/d3d9/d3d9_adapter.cpp
```diff
@@ -768,30 +770,29 @@ namespace dxvk {
...
         continue;
 
       D3DDISPLAYMODEEX mode;
       mode.Size             = sizeof(D3DDISPLAYMODEEX);
-      mode.Width            = devMode.dmPelsWidth;
-      mode.Height           = devMode.dmPelsHeight;
-      mode.RefreshRate      = devMode.dmDisplayFrequency;
+      mode.Width            = devMode.width;
+      mode.Height           = devMode.height;
+      mode.RefreshRate      = devMode.refreshRate.numerator / devMode.refreshRate.denominator;
```
doitsujin, line 795: again why not use `ConvertDisplayMode` helper?

Might as well respect interlacing while we're at it (`D3DSCANLINEORDERING_INTERLACED`)

src/wsi/wsi_window.h
```diff
@@ -0,0 +1,134 @@
...
+    * \param [in] hWindow The window (may be unused on some platforms)
+    * \param [in] pMode The mode
+    * \param [in] EnteringFullscreen Are we entering fullscreen?
+    * \returns \c true on success, \c false on failure
+    */
+  bool setWindowMode(
+          HMONITOR         hMonitor,
+          HWND             hWindow,
+    const WsiMode*         pMode,
+          bool             EnteringFullscreen);
```
doitsujin, line 51: What is `EnteringFullscreen` needed for?

doitsujin (changes requested):

src/wsi/wsi_window.h
Joshua-Ashton: (It was a poor abstraction for something, that I can do better)

doitsujin: I don't like it either, which is why I was asking.

Joshua-Ashton:
Addressed feedback

doitsujin (approved):

PR #2808, opened 2022-08-09
Limit D3D9 staging memory v2

doitsujin:
Similar idea as PR #2800: We stall until the amount of staging memory in flight drops below a certain threshold.

The limit enforced here is not a hard limit, we can have up to two more staging buffers in flight: One that's being tracked by the command list if we had to allocate a new buffer in the meantime (this will pretty much always be the case), and potentially one from subsequent `AllocStagingBuffer` calls. This might be unfortunate, but calling `Flush` does not appear to be safe at the time we compute the amount of memory needed. With some work we should be able to sort this out and reduce the number of potential extra staging buffers to just one.

I don't know if 8M is a good limit or if we should try to allow more (maybe 12 once we can take the required memory into account, so peaks will be no more than 16M?).

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4018,11 +4018,90 @@ namespace dxvk {
 
 
   D3D9BufferSlice D3D9DeviceEx::AllocStagingBuffer(VkDeviceSize size) {
+    m_stagingBufferAllocated += size;
+
     D3D9BufferSlice result;
     result.slice = m_stagingBuffer.alloc(256, size);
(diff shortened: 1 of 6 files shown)


K0bin (changes requested):
Looks good.

Last commit message is missing a "y" ("memor"). 8MB is too low, that causes rare stalls during Borderlands 2 gameplay. 16 works fine in the games I've tested.

K0bin:
I've had occasional crashes in Borderlands 2 with this PR. I'll try to investigate.

K0bin (approved):

Issue #1899, opened 2021-01-15
Dragonshard looks really weird


Metalhead33:
The game looks really weird with DXVK being used:

[https://imgbox.com/gallery/edit/BbFvVnCdQc/2vtkeSYN6FwCQdfT](https://imgbox.com/gallery/edit/BbFvVnCdQc/2vtkeSYN6FwCQdfT)

When I use DgVoodoo2 in combination with DXVK, this problem does not exist, but instead all unit textures turn black. But with DgVoodoo2 removed, things become ugly, as demonstrated on the pictures.

### Software information

[Dungeons & Dragons: Dragonshard](https://en.wikipedia.org/wiki/Dungeons_%26_Dragons:_Dragonshard)

### System information

* Linux localhost.localdomain 5.9.11-gentoo #1 SMP x86_64 Intel(R) Core(TM) i3-6100 CPU @ 3.70GHz GenuineIntel GNU/Linux
* GeForce GTX 960/PCIe/SSE2 NVIDIA 455.38
* Wine version: lutris-6.0-rc1-x86_64
* dxvk-1.7.3

### Apitrace file

[https://drive.google.com/file/d/14symu8TUaFrpGUT0pKcZ41-UKPm4T2C2/view?usp=sharing](https://drive.google.com/file/d/14symu8TUaFrpGUT0pKcZ41-UKPm4T2C2/view?usp=sharing)

### Log files
- [d3d9.log](https://drive.google.com/file/d/1n2dN0ZJGmUnaUHKj1MCs5ye5iaTfTc32/view?usp=sharing)

#### DXVK 1.7.3 + DgVoodoo v2.72:

![image](https://cdn.discordapp.com/attachments/561359616135397396/799719598341357589/2021-01-15-201944-screenshot.png)
![image](https://cdn.discordapp.com/attachments/561359616135397396/799719603236110346/2021-01-15-201945-screenshot.png)

#### Just DXVK 1.7.3:

![image](https://cdn.discordapp.com/attachments/561359616135397396/799719956567687218/2021-01-15-202113-screenshot.png)
![image](https://cdn.discordapp.com/attachments/561359616135397396/799719961420759100/2021-01-15-202110-screenshot.png)

In conclusion:

**DXVK + DgVoodoo:** Bridge looks normal, but units are completely black :frog: 
**Just DXVK:** Units have normal textures, but units and bridges alike have weird screwed-up transparency :frog: 
**OpenGL renderer:** Let's found out!

[...]

PR #2817, opened 2022-08-12
Rework D3D9 alpha test

doitsujin:
In order to get rid of the `alphaTestWiggleRoom` option and have better behaviour out of the box without floating point math inaccuracies, we should pass in the alpha reference as an integer and quantize the `oC0` alpha value on the GPU.

This adjusts the alpha test precision based on the currently bound render target, which matches Nvidia's behaviour.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -414,16 +414,6 @@
 
 # d3d9.longMad = False
 
-# Alpha Test Wiggle Room
-#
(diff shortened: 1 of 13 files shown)


Blisto91:
Fixes https://github.com/doitsujin/dxvk/issues/1899
Found by @WinterSnowfall

PR #2823, opened 2022-08-14
[util] Support for fps limiter on non-Windows platforms

Joshua-Ashton:
Defaults to a sleep granularity of 0.5ms, which is slightly on the cautious side.

Calls through to std::this_thread::sleep_for directly, which calls through to nanosleep.

### DIFF
--- a/src/util/util_fps_limiter.cpp
+++ b/src/util/util_fps_limiter.cpp
@@ -33,8 +33,8 @@ namespace dxvk {
 
     if (!m_envOverride) {
       m_targetInterval = frameRate > 0.0
-        ? NtTimerDuration(int64_t(double(NtTimerDuration::period::den) / frameRate))
-        : NtTimerDuration::zero();
+        ? TimerDuration(int64_t(double(TimerDuration::period::den) / frameRate))
+        : TimerDuration::zero();
 
       if (isEnabled() && !m_initialized)
         initialize();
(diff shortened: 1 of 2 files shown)


doitsujin:
Can we factor out the platform-specific bits to separate functions? I hate when there's ifdef sauce and common code in the same place, it's completely unreadable.

src/util/util_fps_limiter.h
```diff
@@ -53,34 +53,34 @@ namespace dxvk {
...
-      return m_targetInterval != NtTimerDuration::zero();
+      return m_targetInterval != NativeTimerDuration::zero();
     }
 
   private:
 
     using TimePoint = dxvk::high_resolution_clock::time_point;
 
-    using NtTimerDuration = std::chrono::duration<int64_t, std::ratio<1, 10000000>>;
+    using NativeTimerDuration = std::chrono::duration<int64_t, std::ratio<1, 10000000>>;
```
doitsujin, line 65: why not just `TimerDuration`? "Native" doesn't immediately mean anything.

src/util/util_fps_limiter.cpp
```diff
@@ -161,8 +168,12 @@ namespace dxvk {
       }
     } else {
       // Assume 1ms sleep granularity by default
-      m_sleepGranularity = NativeTimerDuration(10000);
+      m_sleepGranularity = NativeTimerDuration(1ms);
     }
+#else
+    // Assume 0.5ms sleep granularity by default
+    m_sleepGranularity = NativeTimerDuration(500us);
+#endif
```
doitsujin, line 176: just make a function `getSleepGranularity` or something

src/util/util_fps_limiter.h
Joshua-Ashton: Sure

Joshua-Ashton:
Updated from feedback

PR #2824, opened 2022-08-14
Introduce array_of helper to make dealing with std::arrays easier

Joshua-Ashton:
Uses this for the devExtensionList, as this will be getting ifdefs in future, and we need the size to be dynamic.

### DIFF
--- a/src/dxvk/dxvk_adapter.cpp
+++ b/src/dxvk/dxvk_adapter.cpp
@@ -5,6 +5,8 @@
 #include "dxvk_device.h"
 #include "dxvk_instance.h"
 
+#include "../util/util_array.h"
+
 namespace dxvk {
   
   DxvkAdapter::DxvkAdapter(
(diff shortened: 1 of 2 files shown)


doitsujin:
if we need the size to be dynamic, use an `std::vector`? `array` isn't made for that in any way whatsoever.

Joshua-Ashton:
That's a useless heap allocation that is then tossed and is initialized with completely static data.
Like sure, it's probably meaningless in the grand scheme of things, but I don't think this is insane.

There are other places I would like to use this helper, mainly because it's really annoying to keep the count in sync when stuff changes and it has bit us before. (eg. remove an element, forget to lower count).

No reason to use a vector when we can just have the array size computed at compile time with this.

doitsujin:
The helper is useful but i **really** don't want to introduce it for something that could be more easily managed with a vector.

Smething like
```
  &extFrog,
#ifdef win32
  &extSomethingWin32,
#endif
```

just isn't something I want in my code.

Joshua-Ashton:
What do you propose then? Vector + push_back?

doitsujin:
For this particular case, yes.

soerengrunewald:
What about using `std::to_array` instead, or one of the proposed implementations (details on [cppreference](https://en.cppreference.com/w/cpp/container/array/to_array))

Joshua-Ashton:
I made it so we dont need to ifdef any extensions now

PR #2826, opened 2022-08-16
Add DXVK profiles to express Vulkan requirements

christophe-lunarg:
I created a Vulkan profiles JSON file to express DXVK minimum requirements for each D3D level. This JSON files are defined by Vulkan Profiles schemas avaiable [here](https://schema.khronos.org/vulkan/) and can be used as part of the Vulkan SDK Profiles Toolset. ([PDF](https://www.lunarg.com/wp-content/uploads/2022/03/The-Vulkan-Profiles-Toolset-Solution-FEB2022.pdf), [Presentation](https://www.lunarg.com/wp-content/uploads/2022/08/Vulkan-SDK-Tools-to-Use-and-Create-Vulkan-Profiles-AUG-SIGGRAPH-2022.pdf))

I based myself on https://github.com/doitsujin/dxvk/wiki/Driver-support but it would certainly be necessary to validate that my interpretation is correct.

This Vulkan profiles JSON file could help users of the project to track new requirements. For the case of IHVs, it will allows them to prioritize implementing new features and for ISVs it will help them to figure out the ecosystem of devices where the project versions could run.

Further use of this file could be to express future requirements by adding "roadmap" profiles or "optimal" profiles to express the requirement for the best level of performance with DXVK.

Finally, having these information baked into this file allows some level of automation to the users of the projects and make sure that each revision of the project comes the corresponding explicit set of Vulkan requirements.

I am creating this PR first for the purpose of discussion and evaluate whether you could be interested by such features.

### DIFF
--- a/VP_DXVK_requirements.json
+++ b/VP_DXVK_requirements.json
@@ -1,0 +1,585 @@
+{
+    "$schema": "https://schema.khronos.org/vulkan/profiles-0.8.0-224.json#",
+    "capabilities": {
+        "vulkan10requirements": {
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "robustBufferAccess": true
+                }
+            }
+        },
+        "vulkan11requirements": {
+            "features": {
+                "VkPhysicalDeviceVulkan11Features": {
+                    "multiview": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceVulkan11Properties": {
+                    "maxMultiviewViewCount": 6,
+                    "maxMultiviewInstanceIndex": 134217727
+                }
+            }
+        },
+        "vulkan12requirements": {
+            "features": {
+                "VkPhysicalDeviceVulkan12Features": {
+                    "uniformBufferStandardLayout": true,
+                    "subgroupBroadcastDynamicId": true,
+                    "imagelessFramebuffer": true,
+                    "separateDepthStencilLayouts": true,
+                    "hostQueryReset": true,
+                    "timelineSemaphore": true,
+                    "shaderSubgroupExtendedTypes": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceVulkan12Properties": {
+                    "maxTimelineSemaphoreValueDifference": 2147483647
+                }
+            }
+        },
+        "vulkan13requirements": {
+            "features": {
+                "VkPhysicalDeviceVulkan12Features": {
+                    "vulkanMemoryModel": true,
+                    "vulkanMemoryModelDeviceScope": true,
+                    "bufferDeviceAddress": true
+                },
+                "VkPhysicalDeviceVulkan13Features": {
+                    "robustImageAccess": true,
+                    "shaderTerminateInvocation": true,
+                    "shaderZeroInitializeWorkgroupMemory": true,
+                    "synchronization2": true,
+                    "shaderIntegerDotProduct": true,
+                    "maintenance4": true,
+                    "pipelineCreationCacheControl": true,
+                    "subgroupSizeControl": true,
+                    "computeFullSubgroups": true,
+                    "shaderDemoteToHelperInvocation": true,
+                    "inlineUniformBlock": true,
+                    "dynamicRendering": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceVulkan13Properties": {
+                    "maxBufferSize": 1073741824,
+                    "maxInlineUniformBlockSize": 256,
+                    "maxPerStageDescriptorInlineUniformBlocks": 4,
+                    "maxPerStageDescriptorUpdateAfterBindInlineUniformBlocks": 4,
+                    "maxDescriptorSetInlineUniformBlocks": 4,
+                    "maxDescriptorSetUpdateAfterBindInlineUniformBlocks": 4,
+                    "maxInlineUniformTotalSize": 4
+                }
+            }
+        },
+        "d3d9_baseline": {
+            "extensions": {
+                "VK_EXT_robustness2": 1
+            },
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "geometryShader": true,
+                    "shaderStorageImageWriteWithoutFormat": true,
+                    "imageCubeArray": true,
+                    "depthClamp": true,
+                    "depthBiasClamp": true,
+                    "fillModeNonSolid": true,
+                    "sampleRateShading": true,
+                    "shaderClipDistance": true,
+                    "shaderCullDistance": true,
+                    "textureCompressionBC": true,
+                    "occlusionQueryPrecise": true,
+                    "independentBlend": true,
+                    "fullDrawIndexUint32": true,
+
+                    "shaderImageGatherExtended": true
+                },
+                "VkPhysicalDeviceVulkan12Features": {
+                    "samplerMirrorClampToEdge": true
+                },
+                "VkPhysicalDeviceRobustness2FeaturesEXT": {
+                    "nullDescriptor": true,
+                    "robustBufferAccess2": true
+                }
+            }
+        },
+        "d3d9_optional": {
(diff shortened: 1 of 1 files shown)


VP_DXVK_baseline.json
```diff
@@ -0,0 +1,374 @@
...
+                }
+            }
+        },
+        "level9": {
+            "extensions": {
+                "VK_EXT_robustness2": 1
+            },
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                	"geometryShader": true,
```
Joshua-Ashton, line 158: indentation

Joshua-Ashton:
Looks like `samplerMirrorClampToEdge` is missing and maybe more.

I would check against the `GetDeviceFeatures` functions instead.

Joshua-Ashton:
I am also not super sure how one would express optional features with this.

eg. in your profile `shaderDemoteToHelperInvocation` is marked as true, but it is not actually required, we only use that if it is supported.

VP_DXVK_baseline.json
```diff
@@ -0,0 +1,374 @@
...
+                "vulkan10requirements",
+                "vulkan11requirements",
+                "level9"
+            ]
+        },
+        "VP_DXVK_level10_0_baseline": {
+            "version": 1,
+            "api-version": "1.3.204",
+            "label": "DXVK feature level 10.1 profile",
+            "description": "DXVK requirements for feature level 10.1",
```
DadSchoorse, line 255: 10.0 not 10.1

VP_DXVK_baseline.json
```diff
@@ -0,0 +1,374 @@
...
+                "VK_KHR_spirv_1_4": 1,
+                "VK_KHR_timeline_semaphore": 1,
+                "VK_KHR_uniform_buffer_standard_layout": 1,
+                "VK_KHR_vulkan_memory_model": 1,
+                "VK_EXT_descriptor_indexing": 1,
+                "VK_EXT_host_query_reset": 1,
+                "VK_EXT_sampler_filter_minmax": 1,
+                "VK_EXT_scalar_block_layout": 1,
+                "VK_EXT_separate_stencil_usage": 1,
+                "VK_EXT_shader_viewport_index_layer": 1
```
Joshua-Ashton, line 74: I don't understand these huge blocks of promoted extensions.

We only support Vulkan 1.3, and don't use any of them.

If this is a necessary quirk for profiles, then that should be fixed in the schema/spec/implementations/whatever.

doitsujin:
If this is useful for people then sure, why not. Seems easy enough for us to maintain.

See Josh's concerns about the extension lists though, since we really don't need any of the promoted extensions unless we explicitly enable the corresponding device features.

> I also created a file for DXVK 1.10.3 which could be useful to express the different is term of requirements with master. Maybe this could be part of a 1.10.4 release?

We don't plan to make another 1.x release.

VP_DXVK_baseline.json
christophe-lunarg: Ok removing the promoted extensions, there is no need for that.

VP_DXVK_baseline.json
christophe-lunarg: Fixed with upcoming commit

christophe-lunarg:
I am looking at the `GetDeviceFeatures` function to figure out the actual requirements

christophe-lunarg:
Note that if you are interested into that, we have a Python script that convert the JSON file into .cpp/.h code. This could be used for you initialization code and to ensure that the JSON files always match the source.

Joshua-Ashton:
> Note that if you are interested into that, we have a Python script that convert the JSON file into .cpp/.h code. This could be used for you initialization code and to ensure that the JSON files always match the source.

How would we encode optional features with this?

christophe-lunarg:
I am editing the level9 profile to give you an idea and pushing it when it's done... (in few minutes)

christophe-lunarg:
I updated `VP_DXVK_level9_baseline` and added `VP_DXVK_level9_optional` but I must admit that it's far from ideal.

doitsujin:
D3D10 is identical to D3D11 FL10_0 or FL10_1 respectively.

christophe-lunarg:
I separated D3D9 and D3D11 (and 10 then) into two different profiles files has there are not necessarilly requiring the same features. Let me know what you think.

Joshua-Ashton:
I would personally prefer if this was one json rather than two, I don't see why we need two.

christophe-lunarg:
Ok, I'll make only one.

doitsujin:
I think the whole D3D11 feature level stuff could also be stripped down. Feature levels below 11_0 have next to no practical relevance nowadays, and I'm thinking we could have something like this as profiles:
- D3D9 baseline
- D3D9 optimal
- *maybe* D3D10 FL10_1 baseline for the handful of D3D10 apps out there (it's really not many)
- D3D11 FL11_0 baseline
- D3D11 FL11_1 baseline
- D3D11 FL11_1 optimal

christophe-lunarg:
Thanks for all the feedback! 

I merged back the profiles in a single file. The D3D9 and D3D11 section remains separated in a way because the requirements expressed in the code actually diverge.

I also followed the profiles suggested by @doitsujin but note that I originally followed the project source which effectively expressed fine grain granularity of feature level down to feature level 9. But if it's what makes sense to you, I am fine with it!

Any more feedback ?

VP_DXVK_requirements.json
```diff
@@ -0,0 +1,585 @@
...
+    },
+    "profiles": {
+        "VP_DXVK_d3d9_baseline": {
+            "version": 1,
+            "api-version": "1.3.204",
+            "label": "DXVK D3D9 Baseline profile",
+            "description": "DXVK for D3D9 minimum requirements",
+            "contributors": {
+                "Philip Rebohle": {
+                    "company": "DXVK"
```
doitsujin, line 333: This should really say Valve for me.

doitsujin:
Just a nit.

One question: I'm currently implementing missing bits to report Feature Level 12_0, and we will require certain device **properties** (rather than just device features) to have specific values in order to report this. Can the profiles express this?

As an example, we'd want `residencyStandard2DBlockShape` to be `VK_TRUE` but `residencyAlignedMipSize` to be `VK_FALSE`.

VP_DXVK_requirements.json
christophe-lunarg: I'll update tomorrow, thanks.

christophe-lunarg:
> Just a nit.
> 
> One question: I'm currently implementing missing bits to support Feature Level 12_0, and we will require certain device **properties** (rather than just device features) to have specific values in order to expose that. Can the profiles express this?
> 
> As an example, we'd want `residencyStandard2DBlockShape` to be `VK_TRUE` but `residencyAlignedMipSize` to be `VK_FALSE`.

Yes all properties are supported but also formats if that's useful to you. I'll create tomorrow a capability JSON block for level12 without the corresponding profile just to show you.

Joshua-Ashton:
shaderDemoteToHelperInvocation is listed as optional but this is required for Vulkan 1.3 which we use.

christophe-lunarg:
Regarding shaderDemoteToHelperInvocation 
I would suggest to replace: 
```
enabled.vk13.shaderDemoteToHelperInvocation                   = supported.vk13.shaderDemoteToHelperInvocation;
```
By:
```
enabled.vk13.shaderDemoteToHelperInvocation = VK_TRUE;
```
To match the Vulkan 1.3 requirement.

I updated the profiles file accordingly.

christophe-lunarg:
@doitsujin such for d3d11_level12_0 in the JSON file to find the VkPhysicalDeviceSparseProperties example.

The block is never referenced in a profile so it can remain here as WIP doc without affecting the definision of the DXVK profiles / requirements.

christophe-lunarg:
@Joshua-Ashton, @doitsujin  Anything missing to approve this PR?

PR #2833, opened 2022-08-18
[d3d9] Only do one allocation for all texture subresources

K0bin:
Warlock - Master of the Arcane writes out of bounds of mip 0.
This crashes with the new memory mapped file memory allocator.
If we allocate one block of memory (or one DxvkBuffer) for all subresources, it just writes into mip 1 instead of crashing.

### DIFF
--- a/src/d3d9/d3d9_initializer.cpp
+++ b/src/d3d9/d3d9_initializer.cpp
@@ -113,41 +113,35 @@ namespace dxvk {
     // If the buffer is mapped, we can write data directly
     // to the mapped memory region instead of doing it on
     // the GPU. Same goes for zero-initialization.
-    const D3D9_COMMON_TEXTURE_DESC* desc = pTexture->Desc();
-    for (uint32_t a = 0; a < desc->ArraySize; a++) {
-      for (uint32_t m = 0; m < desc->MipLevels; m++) {
-        uint32_t subresource = pTexture->CalcSubresource(a, m);
-        void* mapPtr = pTexture->GetData(subresource);
-        uint32_t length = pTexture->GetMipSize(subresource);
+    void* mapPtr = pTexture->GetData(0);
+    if (pInitialData) {
+      // Initial data is only supported for textures with 1 subresource
+      VkExtent3D mipExtent = pTexture->GetExtentMip(0);
+      const DxvkFormatInfo* formatInfo = lookupFormatInfo(pTexture->GetFormatMapping().FormatColor);
+      VkExtent3D blockCount = util::computeBlockCount(mipExtent, formatInfo->blockSize);
+      uint32_t pitch = blockCount.width * formatInfo->elementSize;
+      uint32_t alignedPitch = align(pitch, 4);
 
-        if (pInitialData != nullptr) {
-          VkExtent3D mipExtent = pTexture->GetExtentMip(m);
-          const DxvkFormatInfo* formatInfo = lookupFormatInfo(pTexture->GetFormatMapping().FormatColor);
-          VkExtent3D blockCount = util::computeBlockCount(mipExtent, formatInfo->blockSize);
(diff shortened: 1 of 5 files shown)


Blisto91:
Also fixes crashes after unmap merge in Cryostasis & theHunter: Primal

src/d3d9/d3d9_initializer.cpp
```diff
@@ -113,41 +113,35 @@ namespace dxvk {
...
+    } else {
+      // All subresources are allocated in one chunk of memory.
+      // So we can just get the pointer for subresource 0 and memset all of them at once.
+      std::memset(
+        mapPtr, 0,
+        pTexture->GetTotalSize());
     }
-    if (pTexture->GetMapMode() == D3D9_COMMON_TEXTURE_MAP_MODE_UNMAPPABLE)
-      pTexture->UnmapData();
+    pTexture->UnmapData();
```
Joshua-Ashton, line 144: Looks like the logic here changed and now it always calls UnmapData instead of just when == D3D9_COMMON_TEXTURE_MAP_MODE_UNMAPPABLE

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -177,12 +177,9 @@ namespace dxvk {
     if (unlikely(dstTexInfo->Desc()->Pool != D3DPOOL_SYSTEMMEM && dstTexInfo->Desc()->Pool != D3DPOOL_SCRATCH))
       return D3DERR_INVALIDCALL;
 
-    VkExtent3D dstTexExtent = dstTexInfo->GetExtentMip(dst->GetMipLevel());
-    VkExtent3D srcTexExtent = srcTexInfo->GetExtentMip(0);
-
-    dstTexInfo->CreateBufferSubresource(dst->GetSubresource(), dstTexExtent.width > srcTexExtent.width || dstTexExtent.height > srcTexExtent.height);
-    Rc<DxvkBuffer> dstBuffer = dstTexInfo->GetBuffer(dst->GetSubresource());
-    Rc<DxvkImage>  srcImage  = srcTexInfo->GetImage();
+    dstTexInfo->CreateBuffer(true);
```
Joshua-Ashton, line 180: This always initializes now, ugh.

Joshua-Ashton: Its not the biggest deal, but deserving of a comment for us to come back to.

doitsujin: what's the problem with that exactly?

K0bin: The problem is that we're memsetting memory which will be overwritten immediately.

I changed it to only initialize if the texture has more than one mip level.
That should cover the majority of cases.

src/d3d9/d3d9_initializer.cpp
K0bin: Yeah, that's fine. Unmap will check if it's mapped and just exit out.

src/d3d9/d3d9_swapchain.cpp
Joshua-Ashton: You are missing the `dstTexExtent.width > srcTexExtent.width || dstTexExtent.height > srcTexExtent.height` check that used to be there.

K0bin: Done.

src/d3d9/d3d9_initializer.cpp
Joshua-Ashton: Sorry I don't see how that is equivalent.

K0bin: If the map mode isn't UNMAPPABLE, the memory is zero initialized. In that case all methods of m_memory exit out without doing anything.

Issue #2834, opened 2022-08-18
Regression - Alan Wake (108710) has graphical issues when aiming flashlight


alasky17:
Alan Wake has graphical issues when you aim the flashlight at an enemy.  The easiest way to recreate this problem is to run through the tutorial.  There is a point in the tutorial right after you are taught how to use the flashlight where there is an enemy immediately after a checkpoint.  You can launch the game, shine the flashlight at the enemy to check for graphical issues, then exit the game infinitely.  Screenshot of graphical problem: ![Experimental](https://user-images.githubusercontent.com/40211296/185479593-2f161307-d39a-4284-a3a9-bb71b4ea482e.png)

The last "good" proton version is 6.3-8 which shipped with v1.9.2-13-g714ca482 (head at 78ef4cfd92cb7f448292aaca83091914ab271257) and the first "bad" proton version is 7.0-1 which shipped with 1.9.4 Release (head at 1216466be7917ea9336cf3172a4d119a4c9f97f2).  

I ran a git bisect which blamed either:
possible first bad commit: [4e6db51230d78354e2a86a23943c590f72d43c5e] [d3d9] Fix float emulation detection.
possible first bad commit: [14df2f7251002892436b3e2dafb8db19be1624d9] [d3d9] Enable strict float emulation by default for new radv.

I tried experimental-bleeding-edge-7.0-22900-20220818-p3d9324-we97c6a-d85cc87-vd00d03 and the problem is still present.

### System information
- GPU: AMD RX 570
- Driver: 22.1.3
- Wine version: Proton 7.0-4
- DXVK version: head at 85cc87

### Log files
- d3d9.log: [AlanWake_d3d9.log](https://github.com/doitsujin/dxvk/files/9377014/AlanWake_d3d9.log)

Blisto91:
Happens with both `d3d9.floatEmulation = Strict` and `d3d9.floatEmulation = False`. Tho the latter have a extra issue with a floating light circle in the sky.
Doesn't happen with it set to `True`

Trace:
https://mega.nz/file/t0ZzjBJY#mRuZ21RQ7BE1uT9za6sW2pCTXcNz6jXSUFJO3nEjjpw

Blisto91:
Thanks for bisecting btw 👍

alasky17:
[...]

PR #2835, opened 2022-08-19
[dxso] normalize 0 to 0 with strict float emulation

DadSchoorse:
Fixes: #2834

### DIFF
--- a/src/dxso/dxso_compiler.cpp
+++ b/src/dxso/dxso_compiler.cpp
@@ -2070,18 +2070,18 @@ namespace dxvk {
         DxsoRegMask srcMask(true, true, true, false);
         auto vec3 = emitRegisterLoad(src[0], srcMask);
 
-        DxsoRegisterValue dot = emitDot(vec3, vec3);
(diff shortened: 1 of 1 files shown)


src/dxso/dxso_compiler.cpp
```diff
@@ -2070,18 +2070,22 @@ namespace dxvk {
...
         }
 
-        // r * rsq(r . r);
-        result.id = m_module.opVectorTimesScalar(
-          typeId,
-          emitRegisterLoad(src[0], mask).id,
-          dot.id);
+        std::array<uint32_t, 3> dotReplicated = {dot, dot, dot};
+
+        DxsoRegisterValue vecLength;
```
K0bin, line 2083: It's not really the length. Call it something like `invVecLength` instead.

K0bin:
You beat me to it.

src/dxso/dxso_compiler.cpp
DadSchoorse: names should be better now.

PR #2838, opened 2022-08-19
[d3d9] Fix crash when auto generating mip maps for unmappable textures

### LINKED ISSUE #2837: [d3d9] Venetica - Gold Edition - crash on startup
Game doesn't work with dxvk-master, fine with v.1.10.3.
Using dxvk-master, the ONLY "fix" I found was to set workaround option "d3d9.textureMemory = 0".
Other values: 128, 256, 512, 1024, 2048, doesn't work either.
I guess it's another "unmap memory" bug.
P.S. Game executable name: "Venetica.exe"

### Software information
Venetica - Gold Edition

### System information
- GPU: nVidia GTX 960
- Driver: 515.49.10
- Wine version: proton 7.0-4
- DXVK version: 1.10.3, 2.0 (master)

### Apitrace file(s)
- https://drive.google.com/file/d/1pjbEh0vYcl0OrlQy3aNCb75hN4YfRSru/view?usp=sharing

### Log files
- https://drive.google.com/file/d/1Tra4RdioSCceME-gbnHoETFktbXub4uD/view?usp=sharing

K0bin:
Fixes #2837

Locking textures that had just generated mip maps could crash in the following scenario:

* retrieve the pointer from the memory mapped file locking data
* then realize that it needs to read back data written by the GPU
* lazily create a DXVKBuffer to do so and free the memory mapped file allocation
* the pointer is no longer valid => BOOM

This PR fixes that and makes it more explicit when we are actually creating a readback buffer for textures.

### DIFF
--- a/src/d3d9/d3d9_common_texture.cpp
+++ b/src/d3d9/d3d9_common_texture.cpp
@@ -179,7 +179,11 @@ namespace dxvk {
     return memory.Ptr();
   }
 
-  void D3D9CommonTexture::CreateBufferSubresource(UINT Subresource) {
+  void D3D9CommonTexture::CreateBufferSubresource(UINT Subresource, bool Initialize) {
+    if (likely(m_buffers[Subresource] != nullptr)) {
+      return;
+    }
+
     DxvkBufferCreateInfo info;
     info.size   = GetMipSize(Subresource);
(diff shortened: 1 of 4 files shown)


kassindornelles (approved):
Seem's good

PR #2839, opened 2022-08-20
[util] Remove RE:REV2 and RE5 workarounds

kassindornelles:
There's no real impact, it doesn't fix stuttering (for non-GPL drivers) and since GPL got merged this is not needed anymore.


kassindornelles:
Still unsure if RE Revelations 1 needs it but i don't own the game, so i can't try it myself

PR #2842, opened 2022-08-20
[native] Add substitute Windows headers

Joshua-Ashton:
Adds substutite Windows + DirectX headers for use when building DXVK natively for Linux

### DIFF
--- a/include/native/windows/oaidl.h
+++ b/include/native/windows/oaidl.h
@@ -1,0 +1,3 @@
+#pragma once
+
+// Don't care.
--- a/include/native/windows/poppack.h
+++ b/include/native/windows/poppack.h
@@ -1,0 +1,8 @@
+/**
+ * This file has no copyright assigned and is placed in the Public Domain.
+ * This file is part of the mingw-w64 runtime package.
+ * No warranty is given; refer to the file DISCLAIMER.PD within this package.
+ */
+#if !(defined(lint) || defined(RC_INVOKED))
+#pragma pack(pop)
+#endif
--- a/include/native/windows/windows_base.h
+++ b/include/native/windows/windows_base.h
@@ -1,0 +1,314 @@
+#pragma once
+
+#ifdef __cplusplus
+#include <cstdint>
+#include <cstring>
+#else
+#include <stdint.h>
+#include <string.h>
+#include <wchar.h>
+#endif // __cplusplus
+
+// GCC complains about the COM interfaces
+// not having virtual destructors
+
+// and class conversion for C...DESC helper types
+#if defined(__GNUC__) && defined(__cplusplus)
+#pragma GCC diagnostic ignored "-Wnon-virtual-dtor"
+#pragma GCC diagnostic ignored "-Wclass-conversion"
+#endif // __GNUC__ && __cplusplus
+
+typedef int32_t INT;
+typedef uint32_t UINT;
+
+typedef int32_t LONG;
+typedef uint32_t ULONG;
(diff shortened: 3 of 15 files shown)


Joshua-Ashton:
If there is anything here to review at all really, it's the `winbase.h` I made which just has the substitute Windows defines.

There is also `native_sdl2.h` but that's not very interesting really.

Joshua-Ashton:
For any header license concerns please see `native/directx/readme.txt`.

The relevant part (copied exactly from their COPYING file):

```
DirectX and DDK headers are under GNU LGPLv2.1+ (see the file
COPYING.LGPLv2.1) and copyrighted by various people. Using these
headers doesn't make LGPLv2.1 apply to your code, because these
headers files contain only data structure definitions, short
macros, and short inline functions. Here is the relevant part
from LGPLv2.1 section 5 paragraph 4:

    If such an object file uses only numerical parameters, data
    structure layouts and accessors, and small macros and small
    inline functions (ten lines or less in length), then the use
    of the object file is unrestricted, regardless of whether it
    is legally a derivative work.
```

include/native/wsi/native_wsi.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
doitsujin, line 1: What is this header for and why don't we just include the sdl2 stuff from wsi_platform.h as necessary?

Joshua-Ashton: This header is for games/apps built against DXVK native to include.

include/native/windows/oaidl.h
```diff
@@ -0,0 +1,3 @@
+#pragma once
+
+// Don't care.
```
doitsujin, line 3: Any reason why we need these empty headers?

include/native/windows/windows_base.h
```diff
@@ -0,0 +1,318 @@
...
+// GCC complains about the COM interfaces
+// not having virtual destructors
+
+// and class conversion for C...DESC helper types
+#if defined(__GNUC__) && defined(__cplusplus)
+#pragma GCC diagnostic ignored "-Wnon-virtual-dtor"
+#pragma GCC diagnostic ignored "-Wclass-conversion"
+#endif // __GNUC__ && __cplusplus
+
+struct SDL_Window;
```
doitsujin, line 21: do we really need to forward-declare that here of all places?

include/native/windows/poppack.h
```diff
@@ -0,0 +1,8 @@
+/**
+ * This file has no copyright assigned and is placed in the Public Domain.
+ * This file is part of the mingw-w64 runtime package.
+ * No warranty is given; refer to the file DISCLAIMER.PD within this package.
+ */
+#if !(defined(lint) || defined(RC_INVOKED))
+#pragma pack(pop)
+#endif
```
doitsujin, line 8: does this do anything remotely useful for native builds?

Joshua-Ashton: Yes, the struct packing matters.

include/native/windows/windows_base.h
Joshua-Ashton: No, this is a left over, will fix.

include/native/windows/oaidl.h
Joshua-Ashton: The DX headers include them for common IDL stuff but they don't define anything useful or needed.

doitsujin (approved):

PR #2847, opened 2022-08-21
[util] Implement thread set_priority on non-Windows platforms

Joshua-Ashton:


src/util/thread.h
```diff
@@ -332,9 +332,43 @@ namespace dxvk {
   };
 
 #else
+  enum class ThreadPriority : int32_t {
+    Lowest      = -2,
+    BelowNormal = -1,
+    Normal      =  0,
+    AboveNormal =  1,
+    Highest     =  2
+  };
```
doitsujin, line 341: I think this whole thing should be cleaned up to only support the priorities we actually use, to something like the following:

```
// in common code, NOT platform-specific
enum class ThreadPriority : uint32_t {
  Normal,
  Lowest,
};
```

And then just have the respecive thread implementations resolve those.

doitsujin (changes requested):

src/util/thread.h
Joshua-Ashton: Can do

Joshua-Ashton: Done.

PR #2853, opened 2022-08-21
[native] Collection of misc. commits not worthy of their own PRs

Joshua-Ashton:
These commits are really trivial, and are not worthy of their own PRs.

### DIFF
--- a/src/d3d11/d3d11_options.cpp
+++ b/src/d3d11/d3d11_options.cpp
@@ -4,6 +4,14 @@
 
 namespace dxvk {
   
+  static bool IsAPITracingDXGI() {
+#ifdef _WIN32
+    return !!::GetModuleHandle("dxgitrace.dll");
+#else
+    return false;
+#endif
+  }
+
   D3D11Options::D3D11Options(const Config& config, const Rc<DxvkDevice>& device) {
     this->dcSingleUseMode       = config.getOption<bool>("d3d11.dcSingleUseMode", true);
     this->enableRtOutputNanFixup   = config.getOption<bool>("d3d11.enableRtOutputNanFixup", false);
@@ -39,7 +47,7 @@ namespace dxvk {
 
     auto cachedDynamicResources = config.getOption<std::string>("d3d11.cachedDynamicResources", std::string());
 
-    if (::GetModuleHandle("dxgitrace.dll")) {
+    if (IsAPITracingDXGI()) {
       // apitrace reads back all mapped resources on the CPU, so
       // allocating everything in cached memory is necessary to
       // achieve acceptable performance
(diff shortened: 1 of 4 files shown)


src/dxgi/dxgi_swapchain.cpp
```diff
@@ -171,7 +171,13 @@ namespace dxvk {
     pStats->PresentCount = m_presentCount;
     pStats->PresentRefreshCount = 0;
     pStats->SyncRefreshCount = 0;
+#ifdef _WIN32
     QueryPerformanceCounter(&pStats->SyncQPCTime);
+#else
+    auto now = dxvk::high_resolution_clock::now().time_since_epoch();
+    pStats->SyncQPCTime.QuadPart =
+      std::chrono::duration_cast<std::chrono::nanoseconds>(now).count();
```
doitsujin, line 179: Any reason not to unconditionally use this path? On win32 we use QPC anyway, and this *should* return the unmodified value.

doitsujin (changes requested):

src/dxgi/dxgi_swapchain.cpp
Joshua-Ashton: That would not be correct, because `::now()` is divided by the frequency, and Windows expects this value not to be:
```cpp
    static inline time_point now() noexcept {
      // Keep the frequency static, this doesn't change at all.
      static const int64_t freq = getFrequency();
      const int64_t counter     = getCounter();

      const int64_t whole = (counter / freq) * period::den;
      const int64_t part  = (counter % freq) * period::den / freq;

      return time_point(duration(whole + part));
    }
```

doitsujin: Meh, I forgot that QPC is some arbitrary bullshit.

I'd still rather have this abstracted away in util_time.h. Can do the same trick as with threads and subclass to add some new methods?

Joshua-Ashton: I'll remove this for now to not block as per discussion

Joshua-Ashton: Done.

FireBurn:
Hi, I'm seeing:

`../dxvk-9999/src/d3d11/meson.build:69:0: ERROR: Unknown variable "dxgi_dep".`

FireBurn:
That's with enable_dxgi=false

doitsujin:
dxgi is required to build d3d10/11.

FireBurn:
If it's now required, then the option not to build it should probably be removed, or altered to warn that it's now required

pchome:
I suppose it only required for native build, because there is no libdxgi.so or something in system.
Could be "fixed" like (maybe)
```
if platform != 'windows'
  lib_dxgi = dxgi_dep
endif
```
in src/d3d11/meson.build

PR #2859, opened 2022-08-22
[util] remove allowDoNotWait from example config

Blisto91:
Missed in https://github.com/doitsujin/dxvk/pull/2815


K0bin:
Thank you.

PR #2861, opened 2022-08-22
[tests] Get DXBC tests building natively.

Joshua-Ashton:
We can use what was built on here for shader compilation for D3D11/D3D9 tests.

### DIFF
--- a/include/native/windows/windows_base.h
+++ b/include/native/windows/windows_base.h
@@ -96,7 +96,18 @@ typedef HANDLE HWND;
 typedef HANDLE HKEY;
 typedef DWORD COLORREF;
 
+#if INTPTR_MAX == INT64_MAX
+typedef int64_t  INT_PTR;
+typedef uint64_t UINT_PTR;
+#else
+typedef int32_t  INT_PTR;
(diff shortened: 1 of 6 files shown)


doitsujin:
Is there any particular reason why we need this?

This just seems like a *huge* fucking mess for no real benefit, the dxbc and hlsl compiler stuff is mostly used for debugging purposes anyway and requires a properly functioning d3dcompiler implementation anyway.

Joshua-Ashton:
@doitsujin We still need to build the regular samples, and they all compile shaders and will use the same thing here. I just picked `dxbc` as a starting point because it was easy to get going

doitsujin:
I'd rather do a more radical thing and remove all the samples, except like one super basic triangle app that supports all three APIs or something. Would need to be written.

The tests we have right now aren't great anyway, and I'd rather move the more useful ones to a separate repo and drop the ones that no one uses anymore (things like d3d11-streamout and a whole bunch of d3d9 stuff).

doitsujin:
Solved by moving tests to https://github.com/doitsujin/dxvk-tests.

PR #2864, opened 2022-08-22
[dxso] Fix write mask for nrm

DadSchoorse:
ooops :frog:


K0bin:
This commit should also get picked for the 1.10.x branch.

K0bin (approved):

PR #2870, opened 2022-08-26
Bump native headers

Joshua-Ashton:
Useful for external projects that include these headers and want to use D3DCompiler.


pchome:
Maybe makes sense to optionally install those `windows` headers for native build?
For now `directx` ones could be obtained from Joshua-Ashton/mingw-directx-headers, but `windows` ones should be manually copied from DXVK repository.

PR #2871, opened 2022-08-26
[vulkan] Make LibraryLoader dynamically load vulkan-1

Joshua-Ashton:
This makes LibraryLoader actually load the library and moves ownership
of GetInstanceProcAddr into it, which means we pass through the
loaders into their parents to grab stuff.

### DIFF
--- a/src/vulkan/vulkan_loader.cpp
+++ b/src/vulkan/vulkan_loader.cpp
@@ -21,6 +21,10 @@ namespace dxvk::vk {
   PFN_vkVoidFunction LibraryLoader::sym(const char* name) const {
     return sym(nullptr, name);
   }
+
+  bool LibraryLoader::valid() const {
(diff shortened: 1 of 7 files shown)


src/vulkan/vulkan_loader.cpp
```diff
@@ -1,26 +1,45 @@
 #include "vulkan_loader.h"
 
+#include "../util/util_win32_compat.h"
+
 namespace dxvk::vk {
 
-  static const PFN_vkGetInstanceProcAddr GetInstanceProcAddr = vkGetInstanceProcAddr;
+  LibraryLoader::LibraryLoader()
+  : m_library(LoadLibraryA("vulkan-1"))
```
pchome, line 8: this breaks "native" builds, should be "libvulkan.so.1" on linux

PR #2878, opened 2022-08-27
[util] Add missing feature level in example config

kimmalmo:
Nit.


doitsujin:
12_1 too technically.

Blisto91:
@kimmalmo Friendly ping. I dunno if the feedback above was supposed to be addressed in your force push or not :slightly_smiling_face:

doitsujin:
See #2885.

Issue #2879, opened 2022-08-29
[d3d9] Heroes Of Annihilated Empires crashes on startup


WinterSnowfall:
The game crashes on startup with DXVK (either 1.10.3 or current master), works fine with WineD3D. Tried limiting d3d9 total available memory to 2048 MB + memorytracktest (otherwise the game will complain about "insufficient memory"), however the crash still persists.

Captured a trace of loading up the game's menu, which crashes apitrace when trying to replay with DXVK :sweat_smile:, but similarly plays just fine with WineD3D.

P.S.: It appears WineD3D also needs the "videomemorysize=2048" winetrick to run the game properly, but with that set up it's all good.

### Software information
Heroes Of Annihilated Empires (GOG version)

### System information
- GPU: Nvidia GTX 1080
- Driver: 515.65.01
- Wine version: 7.15 Staging
- DXVK version: 1.10.3 / master 7ebd359

### Apitrace file(s)
[engine.trace.tar.xz](https://drive.google.com/file/d/1oin7VSkjLAQze1Zbq15qynm_F17V2xf1/view?usp=sharing)

### Log files
- d3d9.log: [engine_d3d9.log](https://github.com/doitsujin/dxvk/files/9443414/engine_d3d9.log)
- d3d11.log: N/A
- dxgi.log: N/A

Blisto91:
Such a nice sight with a well filled out issue 😁

Joshua-Ashton:
Closed by https://github.com/doitsujin/dxvk/commit/c49b1ee3909fd48bde7598de9c60c3b53916d9dd

PR #2880, opened 2022-08-29
[util] Add a maxAvailableMemory limit for Heroes of Annihilated Empires

WinterSnowfall:
Sort of the second half of the fix for #2879. Josh fixed the crash, but the game won't render textures correctly without a video memory limit.

The regex should cover the following scenarios (I'd appreciate a second pair of :eyes: on it, but seems to work fine based on my testing):
- install dir can be either `Heroes of Annihilated Empires` or `Heroes Of Annihilated Empires` (default on GOG and Steam versions of the game, no idea what a retail copy uses)
- `engine.exe` can be present either directly in the above install dir (allegedly, Steam) or within a `Data` subdirectory (GOG). Thanks to @Blisto91 for his checks on the matter.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -595,6 +595,13 @@ namespace dxvk {
       { "d3d9.textureMemory",               "16" },
       { "d3d9.allowDirectBufferMapping",    "False" },
     }} },
+    /* Heroes of Annihilated Empires            *
+     * Has issues with texture rendering and    *
+     * video memory detection otherwise.        */
+    { R"(\\Heroes (o|O)f Annihilated Empires.*\\engine\.exe$)", {{
+      { "d3d9.memoryTrackTest",             "True" },
+      { "d3d9.maxAvailableMemory",          "2048" },
+    }} },
   }};
 
 


Blisto91:
I know this is already annoyingly long already, but it might be an idea to include the two demos on steam :sweat_smile: 
Their folders are called `Heroes of Annihilated Empires Demo` or `Heroes of Annihilated Empires Multiplayer Demo` depending on the demo

WinterSnowfall:
> Or maybe it can be made to work even when the demo part is omitted. so it just looks for a a folder that includes `Heroes of Annihilated Empires`. I know nothing about regex

Alright, this should take care of the additional `Demo` or `Multiplayer Demo` and has the added benefit of not caring that much about sub-folder structure, since it will match `engine.exe` anywhere within a folder that starts with `Heroes (o|O)f Annihilated Empires`.

Hope that's better... and I haven't missed anything that will otherwise explode spectacularly. Seems fine from testing :sweat_smile:.

WinterSnowfall:
For posterity, here's what the memory limit does:

image

... compared to default behavior:

image

PR #2881, opened 2022-08-29
Remove d3d10.dll/d3d10_1.dll

doitsujin:
Before Reddit flips out over this: No, this does **not** mean we're dropping d3d10 support. That will still be provided by d3d10core.dll.

These DLLs are incomplete and already not used by anything that's relevant (Proton, Lutris), so let's just drop them.


src/d3d10/meson.build
```diff
@@ -25,24 +17,3 @@ d3d10_core_dep = declare_dependency(
 )
 
 d3d10_deps = [ lib_d3dcompiler_43, lib_dxgi, dxbc_dep, dxvk_dep, d3d10_core_dep ]
```
pchome, line 19: this is part of block below (remove too)

pchome:
Also `lib_d3dcompiler_43` from the main `meson.build` and `lib{,32}/`.
`lib_d3dcompiler_47` used for tests.

Joshua-Ashton (approved):

PR #2889, opened 2022-09-01
[d3d9] Force D3DTADDRESS_CLAMP for cubes.

DadSchoorse:
matches wined3d

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3786,7 +3786,13 @@ namespace dxvk {
         m_depthTextures &= ~(1u << StateSampler);
         if (newDepth)
           m_depthTextures |= 1u << StateSampler;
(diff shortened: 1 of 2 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -3783,10 +3783,14 @@ namespace dxvk {
       const bool newDepth = newTexture->IsShadow();
 
       if (oldDepth != newDepth) {
-        m_depthTextures &= ~(1u << StateSampler);
-        if (newDepth)
-          m_depthTextures |= 1u << StateSampler;
+        m_depthTextures ^= 1u << StateSampler;
```
Joshua-Ashton, line 3786: Can we separate out xor change into another commit in case we mess something up here?

DadSchoorse: done

PR #2891, opened 2022-09-02
[dxvk] Ignore queue with VK_QUEUE_FAMILY_IGNORED index

Samsuper12:
Do not add queue to set (and create info too) for queue with
`VK_QUEUE_FAMILY_IGNORED` family index.

Some drivers doesn't support sparse queue. More information in Mesa MR:
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/18384

Signed-off-by: Mykhailo Skorokhodov <mykhailo.skorokhodov@globallogic.com>

### DIFF
--- a/src/dxvk/dxvk_adapter.cpp
+++ b/src/dxvk/dxvk_adapter.cpp
@@ -481,6 +481,10 @@ namespace dxvk {
     this->logQueueFamilies(queueFamilies);
     
     for (uint32_t family : queueFamiliySet) {
+
+      if (family == VK_QUEUE_FAMILY_IGNORED)
+        continue;
+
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Looks fine to me, just a minor nit is that it should be `[dxvk]` not `[dxvk]:`

Samsuper12:
> Looks fine to me, just a minor nit is that it should be `[dxvk]` not `[dxvk]:`

Done. :frog:

doitsujin:
Thanks for reporting, I fixed this in a slightly more obvious way since the sparse queue is the only one that we expect not to find on a device.

PR #2911, opened 2022-09-13
Support for ARM64

Joshua-Ashton:
Not much needed.

### DIFF
--- a/src/util/util_bit.h
+++ b/src/util/util_bit.h
@@ -1,5 +1,16 @@
 #pragma once
 
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
+  #define DXVK_ARCH_X86
+  #if defined(__x86_64__) || defined(_M_X64)
+    #define DXVK_ARCH_X86_64
+  #endif
+#elif defined(__aarch64__) || defined(_M_ARM64)
+  #define DXVK_ARCH_ARM64
+#else
+#error "Unknown CPU Architecture"
+#endif
+
 #ifndef _MSC_VER
 #if defined(__WINE__) && defined(__clang__)
 #pragma push_macro("_WIN32")
--- a/src/util/util_bit.h
+++ b/src/util/util_bit.h
@@ -11,17 +11,19 @@
 #error "Unknown CPU Architecture"
 #endif
 
-#ifndef _MSC_VER
-#if defined(__WINE__) && defined(__clang__)
-#pragma push_macro("_WIN32")
-#undef _WIN32
-#endif
-#include <x86intrin.h>
-#if defined(__WINE__) && defined(__clang__)
-#pragma pop_macro("_WIN32")
-#endif
-#else
-#include <intrin.h>
+#ifdef DXVK_ARCH_X86
+  #ifndef _MSC_VER
+    #if defined(__WINE__) && defined(__clang__)
+      #pragma push_macro("_WIN32")
+      #undef _WIN32
+    #endif
+    #include <x86intrin.h>
+    #if defined(__WINE__) && defined(__clang__)
+      #pragma pop_macro("_WIN32")
+    #endif
+  #else
+    #include <intrin.h>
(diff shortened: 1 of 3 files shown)


src/util/util_vector.h
```diff
@@ -151,12 +151,18 @@ namespace dxvk {
...
+    #ifdef DXVK_ARCH_X86
     Vector4 result;
     __m128 value = _mm_load_ps(a.data);
     __m128 mask  = _mm_cmpeq_ps(value, value);
            value = _mm_and_ps(value, mask);
     _mm_store_ps(result.data, value);
     return result;
+    #else
+    for (int i = 0; i < 4; i++)
+      a[i] = a[i] != a[i] ? 0.0f : a[i];
```
doitsujin, line 163: `std::isnan`?

src/util/util_bit.h
```diff
@@ -101,6 +116,8 @@ namespace dxvk::bit {
       : "r" (n)
       : "cc");
     return res;
+    #elif defined(__GNUC__) || defined(__clang__)
+    return n != 0 ? __builtin_ctz(n) : 64;
```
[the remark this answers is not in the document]
Joshua-Ashton: Good catch, I thought these were typeless like the atomic ones but they were not.

src/util/util_bit.h
```diff
@@ -114,6 +116,8 @@ namespace dxvk::bit {
       : "r" (n)
       : "cc");
     return res;
+    #elif defined(__GNUC__) || defined(__clang__)
+    return n != 0 ? __builtin_ctzl(n) : 64;
```
doitsujin, line 120: `ctzll`? `l` isn't exactly guaranteed to be 64 bits wide either.

src/util/util_bit.h
```diff
@@ -100,11 +100,11 @@ namespace dxvk::bit {
...
 
   inline uint32_t tzcnt(uint64_t n) {
-    #if defined(_M_X64) && defined(_MSC_VER) && !defined(__clang__)
+    #if defined(DXVK_ARCH_X86_64) && defined(_MSC_VER) && !defined(__clang__)
     return _tzcnt_u64(n);
-    #elif defined(__x86_64__) && defined(__BMI__)
+    #elif defined(DXVK_ARCH_X86_64) && defined(__BMI__)
     return __tzcnt_u64(n);
-    #elif defined(__x86_64__) && defined(__GNUC__) || defined(__clang__)
+    #elif defined(DXVK_ARCH_X86_64) && defined(__GNUC__) || defined(__clang__)
```
doitsujin, line 107: can this be fixup'd into the tzcnt commit?

doitsujin (changes requested):
nits

src/util/util_vector.h
Joshua-Ashton: Generates horrible code on MSVC on x86 and ARM, but it's identical for GCC/Clang because they're not shit so I am okay with using it here.

doitsujin (approved):

src/util/util_bit.h
```diff
@@ -1,16 +1,29 @@
...
-#undef _WIN32
-#endif
-#include <x86intrin.h>
-#if defined(__WINE__) && defined(__clang__)
-#pragma pop_macro("_WIN32")
-#endif
+#if defined(__x86_64__) || defined(_M_X64) || defined(__i386__) || defined(_M_IX86)
+  #define DXVK_ARCH_X86
+  #if defined(__x86_64__) || defined(_M_X64)
+    #define DXVK_ARCH_X86_64
```
dlshinobi, line 6: If I'm reading the code right, why are we having 2 defines of 64-bits: DXVK_ARCH_x86_64 and DXVK_ARCH_x86 ?

LingMan: DXVK_ARCH_x86 is defined for 64 bit and 32 bit.. DXVK_ARCH_x86_64 is defined for 64 bit only.

PR #2923, opened 2022-09-15
[d3d9] Fix for missing restriction check in UpdateSurface.

kdobrowo:
The spec of IDirect3DDevice9::UpdateSurface contains the following restriction:
- Neither surface can be created with multisampling.
  The only valid flag for both surfaces is D3DMULTISAMPLE_NONE.

This commit adds this check and returns D3DERR_INVALIDCALL
when source or destination surfaces are multisampled.


kdobrowo:
This code change fixes Colorfill WHQL test.

Joshua-Ashton:
Thanks! Looks good.

PR #2924, opened 2022-09-15
[d3d9] Fix for missing mapping of VertexElements declarations to FVF bits.

kdobrowo:
When Vertex declaration is created by CreateVertexDeclaration
and SetFVF is not called then GetFVF returns 0.
This code change implements mapping of D3D declarations to FVF mask
and sets it if FVF was not set previously.

### DIFF
--- a/src/d3d9/d3d9_vertex_declaration.cpp
+++ b/src/d3d9/d3d9_vertex_declaration.cpp
@@ -228,16 +228,19 @@ namespace dxvk {
     // D3DDECLTYPE_FLOAT3     D3DDECLUSAGE_POSITION       1               N / A
     // D3DDECLTYPE_FLOAT3     D3DDECLUSAGE_NORMAL         1               N / A
 
-    DWORD fvfRet = 0;
-
     if (element.Usage == D3DDECLUSAGE_POSITION && element.Type == D3DDECLTYPE_FLOAT3 && element.UsageIndex == 0)
       return D3DFVF_XYZ;
     else if (element.Usage == D3DDECLUSAGE_POSITIONT && element.Type == D3DDECLTYPE_FLOAT4 && element.UsageIndex == 0)
       return D3DFVF_XYZRHW;
-    else if (element.Usage == D3DDECLUSAGE_BLENDWEIGHT
-          && element.UsageIndex == 0
-          && (fvfRet = MapD3DDeclTypeFloatToFvfXYZBn(element.Type)) != 0)
-      return fvfRet;
+    else if (element.Usage == D3DDECLUSAGE_BLENDWEIGHT && element.UsageIndex == 0) {
+      DWORD fvfRet = MapD3DDeclTypeFloatToFvfXYZBn(element.Type);
+      if (likely(fvfRet != 0))
+        return fvfRet;
+      else {
+        Logger::warn("D3D9VertexDecl::MapD3DDeclToFvf: Unsupported set of D3DDECLUSAGE_BLENDWEIGHT / D3DDECLTYPE_*  / UsageIndex");
+        return 0;
+      }
+    }
     else if (element.Usage == D3DDECLUSAGE_BLENDINDICES && element.Type == D3DDECLTYPE_UBYTE4 && element.UsageIndex == 0)
       return D3DFVF_XYZB1;
     else if (element.Usage == D3DDECLUSAGE_NORMAL && element.Type == D3DDECLTYPE_FLOAT3 && element.UsageIndex == 0)
@@ -252,12 +255,19 @@ namespace dxvk {
       case 1:
         return D3DFVF_SPECULAR;
       default:;
+        Logger::warn("D3D9VertexDecl::MapD3DDeclToFvf: Unsupported set of D3DDECLUSAGE_COLOR / D3DDECLTYPE_D3DCOLOR / UsageIndex");
+        return 0;
       }
     }
-    else if (element.Usage == D3DDECLUSAGE_TEXCOORD
-          && element.UsageIndex < 8
-          && (fvfRet = MapD3DDeclUsageTexCoordToFvfTexCoordSize(element, fvf, texCountPostUpdate)) != 0)
-      return fvfRet;
+    else if (element.Usage == D3DDECLUSAGE_TEXCOORD && element.UsageIndex < 8) {
+      DWORD fvfRet = MapD3DDeclUsageTexCoordToFvfTexCoordSize(element, fvf, texCountPostUpdate);
+      if (likely(fvfRet != 0))
+        return fvfRet;
+      else {
+        Logger::warn("D3D9VertexDecl::MapD3DDeclToFvf: Unsupported set of D3DDECLUSAGE_TEXCOORD / D3DDECLTYPE_* / UsageIndex");
+        return 0;
+      }
+    }
 
     Logger::warn("D3D9VertexDecl::MapD3DDeclToFvf: Unsupported set of D3DDECLUSAGE_* / D3DDECLTYPE_* / UsageIndex");
     return 0;
@@ -269,15 +279,15 @@ namespace dxvk {
     switch (type)
     {
     default:;
-        return 0;
+      return 0;
     case D3DDECLTYPE_FLOAT1:
-        return D3DFVF_XYZB1;
+      return D3DFVF_XYZB1;
     case D3DDECLTYPE_FLOAT2:
-        return D3DFVF_XYZB2;
+      return D3DFVF_XYZB2;
     case D3DDECLTYPE_FLOAT3:
-        return D3DFVF_XYZB3;
+      return D3DFVF_XYZB3;
     case D3DDECLTYPE_FLOAT4:
-        return D3DFVF_XYZB4;
+      return D3DFVF_XYZB4;
     }
   }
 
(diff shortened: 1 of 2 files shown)


kdobrowo:
This code change fixes Fvf WHQL test for dx9.

src/d3d9/d3d9_vertex_declaration.cpp
```diff
@@ -205,6 +206,106 @@ namespace dxvk {
...
+          break;
+  
+        return localFvf;
+      } while (1);
+    }
+    Logger::warn("D3D9VertexDecl::MapD3DDeclToFvf: Unsupported set of D3DDECLUSAGE_* / D3DDECLTYPE_* / UsageIndex");
+    return 0;
+  }
+
+  void D3D9VertexDecl::MapD3D9VertexElementsToFvf() {
```
Joshua-Ashton, line 299: Just make this return a DWORD and initialise m_fvf with it.

src/d3d9/d3d9_vertex_declaration.cpp
```diff
@@ -205,6 +206,106 @@ namespace dxvk {
...
+    if (!m_fvf) {
+      DWORD texCountPostUpdate = 0;
+      for (const auto& element : m_elements)
+        m_fvf |= MapD3DDeclToFvf(element, m_fvf, texCountPostUpdate);
+
+      if (texCountPostUpdate)
```
Joshua-Ashton, line 305: Useless if

src/d3d9/d3d9_vertex_declaration.cpp
```diff
@@ -205,6 +206,106 @@ namespace dxvk {
...
+  // D3DDECLTYPE_FLOATm     D3DDECLUSAGE_TEXCOORD       n               D3DFVF_TEXCOORDSIZEm(n)
+  // D3DDECLTYPE_FLOAT3     D3DDECLUSAGE_POSITION       1               N / A
+  // D3DDECLTYPE_FLOAT3     D3DDECLUSAGE_NORMAL         1               N / A
+
+  if (element.Usage == D3DDECLUSAGE_POSITION && element.Type == D3DDECLTYPE_FLOAT3 && element.UsageIndex == 0)
+    return D3DFVF_XYZ;
+  else if (element.Usage == D3DDECLUSAGE_POSITIONT && element.Type == D3DDECLTYPE_FLOAT4 && element.UsageIndex == 0)
+    return D3DFVF_XYZRHW;
+  else if (element.Usage == D3DDECLUSAGE_BLENDWEIGHT && element.UsageIndex == 0) {
+    switch (element.Type)
```
Joshua-Ashton, line 236: Make a helper for this switch

src/d3d9/d3d9_vertex_declaration.cpp
```diff
@@ -205,6 +206,106 @@ namespace dxvk {
...
+          localFvf = D3DFVF_TEXCOORDSIZE2(element.UsageIndex);
+        else if (element.Type == D3DDECLTYPE_FLOAT3)
+          localFvf = D3DFVF_TEXCOORDSIZE3(element.UsageIndex);
+        else if (element.Type == D3DDECLTYPE_FLOAT4)
+          localFvf = D3DFVF_TEXCOORDSIZE4(element.UsageIndex);
+        else
+          break;
+  
+        return localFvf;
+      } while (1);
```
Joshua-Ashton, line 293: I'd put the contents of this loop into a helper.

src/d3d9/d3d9_vertex_declaration.cpp
```diff
@@ -205,6 +206,106 @@ namespace dxvk {
...
+    default:;
+    }
+  }
+  else if (element.Usage == D3DDECLUSAGE_TEXCOORD && element.UsageIndex < 8) {
+      DWORD localFvf = 0;
+      do {
+        // Check if bits of format for current UsageIndex are free in the fvf
+        // It is necessary to skip multiple initializations of the bitfield because
+        // returned value is bitwise or-ed to final fvf DWORD.
+        if (D3DFVF_TEXCOORDSIZE1(element.UsageIndex) & fvf != 0)
```
Joshua-Ashton, line 271: Make this a while loop with this as the condition? :P

src/d3d9/d3d9_vertex_declaration.cpp
```diff
@@ -205,6 +206,106 @@ namespace dxvk {
...
+          localFvf = D3DFVF_TEXCOORDSIZE3(element.UsageIndex);
+        else if (element.Type == D3DDECLTYPE_FLOAT4)
+          localFvf = D3DFVF_TEXCOORDSIZE4(element.UsageIndex);
+        else
+          break;
+  
+        return localFvf;
+      } while (1);
+    }
+    Logger::warn("D3D9VertexDecl::MapD3DDeclToFvf: Unsupported set of D3DDECLUSAGE_* / D3DDECLTYPE_* / UsageIndex");
```
Joshua-Ashton, line 295: Put this log into the switches, this will be easy when you put them into the helpers too.

Joshua-Ashton (changes requested):

src/d3d9/d3d9_vertex_declaration.cpp
kdobrowo: moved to the helper

src/d3d9/d3d9_vertex_declaration.cpp
kdobrowo: done

src/d3d9/d3d9_vertex_declaration.cpp
kdobrowo: My intention was to not multiply these logs when one at the end is necessary.
Do you ant to have this log in helpers too?

src/d3d9/d3d9_vertex_declaration.cpp
kdobrowo: done

src/d3d9/d3d9_vertex_declaration.cpp
kdobrowo: done

kdobrowo (commented):
Corrected code after the review.

Joshua-Ashton:
You had a loop in your original code... But now there just isn't one, what gives?

doitsujin:
@Joshua-Ashton it's still there?
```c++
  DWORD D3D9VertexDecl::MapD3D9VertexElementsToFvf() {
    DWORD fvf = 0;
    DWORD texCountPostUpdate = 0;

    for (const auto& element : m_elements)
      fvf |= MapD3DDeclToFvf(element, fvf, texCountPostUpdate);

    fvf |= texCountPostUpdate;

    return fvf;
  }
```

kdobrowo:
> You had a loop in your original code... But now there just isn't one, what gives?

That loop was not a loop actually:

      do {
...
        if (D3DFVF_TEXCOORDSIZE1(element.UsageIndex) & fvf != 0)
          **break;**
...  
        if (element.Type == D3DDECLTYPE_FLOAT1)
          localFvf = D3DFVF_TEXCOORDSIZE1(element.UsageIndex);
...
        else
          **break;**
  
        **return localFvf;**
      } while (1);

The do { } while(); was used to make break working.
It would be more clear when while(0) was used.
Anyway, the while checking was never executed because breaks jumped out of the loop at earlier stages and return went out of the function. (There was no continue;).

kdobrowo:
> @Joshua-Ashton it's still there?
> 
> ```c++
>   DWORD D3D9VertexDecl::MapD3D9VertexElementsToFvf() {
>     DWORD fvf = 0;
>     DWORD texCountPostUpdate = 0;
> 
>     for (const auto& element : m_elements)
>       fvf |= MapD3DDeclToFvf(element, fvf, texCountPostUpdate);
> 
>     fvf |= texCountPostUpdate;
> 
>     return fvf;
>   }
> ```

Do you mean call of MapD3DDeclToFvf?
If so, yes, it should be there.

Joshua-Ashton:
> The do { } while(); was used to make break working.


Oh, that was not clear... yes lets not do things like that.

kdobrowo:
Is there anything more I should do here?

src/d3d9/d3d9_vertex_declaration.cpp
```diff
@@ -205,6 +206,135 @@ namespace dxvk {
...
+    // returned value is bitwise or-ed to final fvf DWORD.
+    if ((D3DFVF_TEXCOORDSIZE1(element.UsageIndex) & fvf) != 0)
+      return 0;
+
+    // Update max texture's index in fvf
+    DWORD lastMaxTexCount = (fvf >> 8) & 0x7;
+    DWORD currentTexCount = element.UsageIndex + 1;
+
+    if (lastMaxTexCount < currentTexCount)
+      texCountPostUpdate = currentTexCount << 8;
```
Joshua-Ashton, line 311: This seems wrong because we don't OR in the `texCountPostUpdate` until after all this is done, yet we seem to try to pull it out here and check it?

kdobrowo: You mean that lastMaxTexCount will always be 0, right? 
Then we will always get the last counter instead of largest. 
Good point.
Thanks.

kdobrowo: New commit [e05092bc921937e4026d5196f7fd1bf9f23f9ee4](https://github.com/doitsujin/dxvk/pull/2924/commits/e05092bc921937e4026d5196f7fd1bf9f23f9ee4) fixes problem of texCountPostUpdate.
Additionally, function MapD3DDeclUsageTexCoordToFvfTexCoordSize may return 0 as valid value. 
Functionally it worked, but when 0 was treated as incorrect value then unwanted warning was generated.

kdobrowo:
Is there any chance for finish the review and merge of this PR?
Thanks.

K0bin:
@kdobrowo Is there a reason for why you don't turn POSITION + VEC4 into `D3DFVF_XYZW`?

Issue #2892, opened 2022-09-03
D3D9 Depth bias Mega-issue


K0bin:
### Depth bias in Vulkan:

`m * depthBiasSlopeFactor + r * depthBiasConstantFactor`

where r is:

> The minimum resolvable difference r is a parameter that depends on the depth attachment representation. It is the smallest difference in framebuffer coordinate z values that is guaranteed toremain distinct throughout polygon rasterization and in the depth attachment. All pairs of fragments generated by the rasterization of two polygons with otherwise identical vertices, but zf
values that differ by r, will have distinct depth values.

### Depth bias in D3D9:

`m * depthBiasSlopeFactor + depthBiasConstantFactor`

### r-Value

The constant part of the final depth bias value is scaled by the smallest resolvable difference in Vulkan and unscaled in D3D9. There is no way to make the driver switch to an unscaled depth bias mode or query the factor.

Right now DXVK uses a hardcoded r-value for each depth image format. Microsofts d3d9on12 curiously uses the exact same values. While that seems to work for 9on12, testing shows that it doesn't for DXVK, leading to various issues listed below.

Wine does a few off screen draws when initializing the graphics device to determine the depth bias factor for each depth format.

Nine just falls back to an r-value of 1 << 23 when the underlying driver doesn't support the unscaled depth bias.

On my RTX 3090, the r-value for 16 bit formats is 1 << 15 and 1 << 23 for everything else according to Wine.

### Floating point formats

> For floating-point depth attachment, there is no single minimum resolvable difference. In this case, the minimum resolvable difference for a given polygon is dependent on the maximum exponent, e, in the range of z values spanned by the primitive. If n is the number of bits in the floating-point mantissa, the minimum resolvable difference, r, for the given primitive is
defined as `r = 2^(e-n)`

This adds the additional problem that we emulate D24 formats with D32 on AMD hardware. So the scaling behavior is different.
This also prevents us from simply doing a few quick draws to determine the scaling factor and then apply that.

K0bin:
#1461

K0bin:
#2642

K0bin:
#1811

K0bin:
#1329

K0bin:
#1560

K0bin:
According to Wine, Mass Effect 2 relies on it as well.

I haven't personally noticed that yet, maybe the current solution is good enough for that particular game.

pchome:
[...]

PR #2931, opened 2022-09-16
[d3d9] Determine depth bias r-factors

K0bin:
Ref: #2892

I've only tested it on Nvidia and it seems to work.

### DIFF
--- a/src/d3d9/d3d9_depth_bias.cpp
+++ b/src/d3d9/d3d9_depth_bias.cpp
@@ -1,0 +1,254 @@
+#include "d3d9_depth_bias.h"
+#include "d3d9_find_depth_bias_factor.h"
+#include "../dxvk/dxvk_format.h"
+
+namespace dxvk {
+
+  D3D9DepthBias::D3D9DepthBias(const Rc<DxvkDevice>& Device)
+    : m_device(Device), m_context(Device->createContext(DxvkContextType::Supplementary)) {
+    DxvkBufferCreateInfo readbackBufferInfo;
+    readbackBufferInfo.size   = 4;
+    readbackBufferInfo.usage  = VK_BUFFER_USAGE_TRANSFER_DST_BIT;
+    readbackBufferInfo.stages = VK_PIPELINE_STAGE_TRANSFER_BIT | VK_PIPELINE_STAGE_HOST_BIT;
+    readbackBufferInfo.access = VK_ACCESS_TRANSFER_WRITE_BIT | VK_ACCESS_HOST_READ_BIT;
+    m_readbackBuffer = m_device->createBuffer(readbackBufferInfo, VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT);
+    uint32_t* ptr = reinterpret_cast<uint32_t*>(m_readbackBuffer->mapPtr(0));
+    *ptr = 0;
+
+    DxvkShaderCreateInfo shaderInfo;
+    shaderInfo.stage           = VK_SHADER_STAGE_VERTEX_BIT;
+    shaderInfo.bindingCount    = 0;
+    shaderInfo.bindings        = nullptr;
+    shaderInfo.pushConstOffset = 0;
+    shaderInfo.pushConstSize   = 0;
+    shaderInfo.inputMask       = 0;
+    m_vertexShader = new DxvkShader(shaderInfo, d3d9_find_depth_bias_factor);
+
+    DetermineFactors();
+
+    m_readbackBuffer = nullptr;
+    m_vertexShader   = nullptr;
+    m_context        = nullptr;
+    m_device         = nullptr;
+  }
+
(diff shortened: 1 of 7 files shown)


src/d3d9/d3d9_depth_bias.cpp
```diff
@@ -0,0 +1,200 @@
+#include "d3d9_depth_bias.h"
+#include "d3d9_find_depth_bias_factor.h"
+#include "../dxvk/dxvk_format.h"
+#include "vulkan/vulkan_core.h"
```
Joshua-Ashton, line 4: Vs code moment

src/d3d9/d3d9_depth_bias.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+      default: return 1.0f / float(1 << 23);
+    }
+
+    Logger::info(str::format("Using depth bias r-factor: ", val, " for format: ", Format));
+    return 1.0f / val;
+  }
+
+  float D3D9DepthBias::GetFactor(VkFormat Format) {
+    auto iter = m_depthBiasFactors.find(Format);
```
Joshua-Ashton, line 193: Just make this an array... We don't need to use a map here.

src/d3d9/d3d9_depth_bias.cpp
```diff
@@ -0,0 +1,200 @@
...
+    rts.depth.layout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
+    m_context->bindRenderTargets(std::move(rts), 0);
+
+    DxvkVertexAttribute attribute = { 0, 0, VK_FORMAT_R32G32B32A32_SFLOAT, 0};
+    DxvkVertexBinding binding = { 0, 0, VK_VERTEX_INPUT_RATE_VERTEX, 16 };
+    m_context->setInputLayout(1, &attribute, 1, &binding);
+
+    DxvkRasterizerState rsState;
+    rsState.polygonMode = VK_POLYGON_MODE_FILL;
+    rsState.cullMode = VK_CULL_MODE_NONE;
```
Joshua-Ashton, line 121: Would be nice to line up the = signs (use "Align" in vs code to do automatically)

src/d3d9/d3d9_depth_bias.cpp
K0bin: indeed

src/d3d9/shaders/d3d9_find_depth_bias_factor.vert
```diff
@@ -0,0 +1,13 @@
...
+
+void main() {
+  vec2 coord = vec2(
+    float(gl_VertexIndex & 2),
+    float(gl_VertexIndex & 1) * 2.0f);
+
+  gl_Position = vec4(
+    -1.0f + coord.x,
+    -1.0f + 3.0f * coord.y,
+    gl_VertexIndex == 2 ? 1.0 : 0.0,
```
Joshua-Ashton, line 11: Making a gradient?

Joshua-Ashton (changes requested):

src/d3d9/shaders/d3d9_find_depth_bias_factor.vert
K0bin: Yes. 

> For floating-point depth attachment, there is no single minimum resolvable
difference. In this case, the minimum resolvable difference for a given polygon is dependent on the
maximum exponent, e, in the **range of z values spanned by the primitive**. If n is the number of bits
in the floating-point mantissa, the minimum resolvable difference, r, for the given primitive is
defined as: 2^(e-n)

Emphasis mine. But because of that we can't really solve it with a simple factor. I went with the factor for a range from 0-1 as a best effort.

The top left vertex is supposed to be at depth 0 because that's what we read back for the depth bias.

K0bin:
Unfortunately it doesn't seem to fix some of the depth bias problems and returns rather weird values for AMD.

On Nvidia, I get 1/(1<<23) for D24 and D32 and 1/(1<<15) for D16. Wine calculates the same values.

CME got:

> info:  Using depth bias r-factor: 1.52588e-05 for format: VK_FORMAT_D16_UNORM
info:  Using depth bias r-factor: 1.52588e-05 for format: VK_FORMAT_D16_UNORM_S8_UINT
info:  Using depth bias r-factor: 1.21072e-07 for format: VK_FORMAT_D32_SFLOAT
info:  Using depth bias r-factor: 1.21072e-07 for format: VK_FORMAT_D32_SFLOAT_S8_UINT

which doesn't match what Wine calculates.

Joshua-Ashton:
@CME42 Can you try with nodcc

src/d3d9/d3d9_depth_bias.cpp
pchome: A little offtopic, but I noticed similar moment in [gamescope code](https://github.com/Plagman/gamescope/commit/1b34a570a51d0a65451d1f2a990487add8cde40a#diff-0ca179f8adb820354aba0b6a063350f9e32689950ead39e8bbeca021fa8ba0e0)

K0bin: Josh knows what he's talking about. :)

Issue #2901, opened 2022-09-09
GTA IV mirror bug 


MiesThies:
As seen in the screenshots, the mirrors look off, when DXVK is installed. 

Actually, when I wanted to create the trace file, I couldn't reproduce the problem. Before I only had d3d9.dll, dxgi.dll, and dxvk.conf in the game directory, then I put in d3d9.dll, d3d11.dll, dxgi.dll and dxgitrace.dll from the folder: apitrace-msvc/x86/lib/wrappers

This is weird. So if I just keep these files in the game directory, I guess my problem is "solved", but again, with only the aforementioned files from \dxvk-1.10.3.tar\dxvk-1.10.3\x32, the mirrors looked off. I hope you can work with what I provided.

### Software information
GTA IV, downgraded to patch 1.0.8.0, Max ingame settings

### System information
- GPU: RTX 2080
- Driver: 516.94
- Wine version: 
- DXVK version: Version 1.10.3

### Apitrace file(s)
(https://drive.google.com/file/d/1LfolKr0Izyb1WXXDF2bk7QLlMfk_9MmB/view?usp=sharing)

### Log files
- d3d9.log: 
[GTAIV_d3d9.log](https://github.com/doitsujin/dxvk/files/9537767/GTAIV_d3d9.log)

[PlayGTAIV_d3d9.log](https://github.com/doitsujin/dxvk/files/9537768/PlayGTAIV_d3d9.log)

- dxvk.conf: https://drive.google.com/file/d/1xFXOcM6HG5rTKr2xOY6T0eMDt4VO9IgS/view?usp=sharing

- screenshots: https://drive.google.com/drive/folders/1uoHgSh4uWM4wZUsmaLMjX6huaj_a7Bo0?usp=sharing

- d3d11.log:
- dxgi.log:

K0bin:
Like Blisto said, your apitrace is basically empty.

Please make another one where you actually run the game and look into the mirror. It doesn't matter that the issue isn't there when you make the apitrace, it will show up when we replay it.

renatopereyra:
A while ago, I tried to trace GTA V and my traces were broken in a similar way. IIRC, I determined that the launcher is 32-bit but the game is 64-bit. I didn't spend too much time on it but I wasn't able to get my traces to work in the end, though, even after trying to account for this.

K0bin:
The game is 32bit.

I cannot fix this without an apitrace.

Blisto91:
[...]

PR #2936, opened 2022-09-17
[util] Disable DF support for pretty mirrors in GTA IV

K0bin:
Fixes: #2901

The comment in the code explains the rational.


Blisto91:
As pchome said over in https://github.com/doitsujin/dxvk/issues/2901#issuecomment-1250132703 i guess we should include `EFLC.exe` which is the old standalone version of the two GTA IV expansions https://store.steampowered.com/app/12220/Grand_Theft_Auto_Episodes_from_Liberty_City/

PR #2949, opened 2022-09-22
Emulate depth bias

K0bin:
To get this out of the way: Yes, an extension would be the much nicer solution and when (if?) we ever get an extension for an unscaled depth bias this, I'll open a PR to remove these changes again.

 #2892

This PR emulates the constant depth bias in the fragment shader. D3D9 doesn't use depthBiasClamp, so we can just use the regular fixed function depth bias for the slope scaled part of it.

If the fragment shader doesn't already replace the fragment depth, the execution mode is set to `DepthUnchanged`. Once the game sets a non 0 depth bias constant, a SPIR-V pass will change that to DepthReplacing and the pipeline will be recompiled. That is an unfortunate late pipeline compile but IMO it's not the end of the world.

We could optionally merge this with a cleaned up version of #2931 and only emulate it in the shader if the depth buffer uses a float format. That might especially benefit Nvidia GPUs which natively support D24S8. Let me know if you think that's worth doing.

Another thing we could do is emit DepthGreater or DepthLess instead of DepthReplacing.

### DIFF
--- a/src/d3d11/d3d11_rasterizer.cpp
+++ b/src/d3d11/d3d11_rasterizer.cpp
@@ -38,6 +38,7 @@ namespace dxvk {
     m_state.conservativeMode  = DecodeConservativeRasterizationMode(desc.ConservativeRaster);
     m_state.sampleCount       = VkSampleCountFlags(desc.ForcedSampleCount);
     m_state.flatShading       = VK_FALSE;
+    m_state.depthReplacing    = VK_FALSE;
 
     m_depthBias.depthBiasConstant = float(desc.DepthBias);
     m_depthBias.depthBiasSlope    = desc.SlopeScaledDepthBias;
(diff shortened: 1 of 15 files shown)


Joshua-Ashton:
Absolutely not, this kills early z.

Having a few games broken >>>> not getting early z in anything anymore.

If we want to solve this let's do it properly and just push on the ext again.

K0bin:
> Having a few games broken >>>> not getting early z in a bunch of stuff anymor

I disagree. As long as the driver can work with `depthUnchanged`, it'll only impact early z in games that actually use depth bias and don't render correctly.

As I said, I consider this a temporary solution. We've been talking about a depth bias extension for at least 3 years now and I'd like to get this fixed one way or another.

Joshua-Ashton:
Then once I am done with pushing through this Wayland extension to fix the xwayland surface association race, let's actually pick up on the extension again.

PR #2952, opened 2022-09-22
Implement IDXGIKeyedMutex interface on top of VK_KHR_win32_keyedmutex

Guy1524:
Also fixes ::GetSharedHandle behavior for resources w/o MISC_SHARED flag.

Marked as draft until wine components reach Proton bleeding-edge.


K0bin:
Keyed Mutexes got implemented since then.

PR #2958, opened 2022-09-26
[d3d9] Use D32 not D32_S8 when upgrading formats without stencil aspect.

DadSchoorse:


WWSNK:
In One Piece Pirate Warriors 3 this caused black skies in the game.

oppw3_2022_09_27_00_28_18_776

DadSchoorse:
That's a unrelated regression caused by f2950953e0d03e62070c234db6edc7974211365f.

Blisto91:
But thanks for the report anyway :+1: 
Hadn't noticed

PR #2964, opened 2022-09-28
[d3d9] Fix for crashes on BACK buffer allocation failure due to OOM

adamjer:
Details:
To avoid nullptr access on back buffer, its allocation failure exception is now catched to return D3DERR_OUTOFVIDEOMEMORY error on IDirect3DDevice9::Reset. Also on  Back buffers destroy and Present paths, back buffer pointer validation was added to avoid accessing nullptr and crash.

### DIFF
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -492,9 +492,7 @@ namespace dxvk {
     if (changeFullscreen)
       SetGammaRamp(0, &m_ramp);
 
-    CreateBackBuffers(m_presentParams.BackBufferCount);
-
-    return D3D_OK;
+    return CreateBackBuffers(m_presentParams.BackBufferCount);
   }
 
 
@@ -644,6 +642,8 @@ namespace dxvk {
   void D3D9SwapChainEx::PresentImage(UINT SyncInterval) {
     m_parent->EndFrame();
     m_parent->Flush();
+
+    ValidateBackBuffersPtrs();
 
     // Retrieve the image and image view to present
     auto swapImage = m_backBuffers[0]->GetCommonTexture()->GetImage();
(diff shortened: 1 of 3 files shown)


src/d3d9/d3d9_swapchain.cpp
```diff
@@ -1160,4 +1170,13 @@ namespace dxvk {
...
 
+  void D3D9SwapChainEx::ValidateBackBuffersPtrs() {
+    for (uint32_t i = 0; i < m_backBuffers.size(); i++) {
+      if (!GetBackBuffer(i))
+      {
+        throw DxvkError("D3D9SwapChainEx: Back buffer pointer is invalid - nullptr");
+      }
+    }
+  }
+
```
Joshua-Ashton, line 1181: This shouldn't be possible to hit unless the app ignores the Reset error and goes on trying to present black frame buffers...

If you want to sanity check, just check `m_backBuffers[0]` being non-null at present time and log+error, all of them don't need to be checked.

Joshua-Ashton (changes requested):

Blisto91 (commented):
Could you remove the reference to your internal issue in the commit message?

src/d3d9/d3d9_swapchain.cpp
wdebkows-intel: > This shouldn't be possible to hit unless the app ignores the Reset error and goes on trying to present black frame buffers...

Yes, agree, that makes sense, but apparently MSFT Multisample Stability (WoW64) test behaves this way and goes with Present despite of OOM error on Reset.

> If you want to sanity check, just check `m_backBuffers[0]` being non-null at present time and log+error, all of them don't need to be checked.

Actually, in my scenario, only m_backBuffers[1] is null, m_backBuffers[0] is fine, so unless we're okay with crash on Present called after failed Reset, we should validate both backbuffers.

adamjer: Any progress with this?

Issue #11, opened 2018-01-19
Can not build on Ubuntu 17.10


pingubot:
I tried to build dxvk, i have installed all dependencies, during compiling it failes, here is the output:

chriundcla@chriundcla-K53SV:~/Downloads/dxvk/dxvk-master/build.w64$ ninja 
[0/1] Regenerating build files.
The Meson build system
Version: 0.42.1
Source dir: /home/chriundcla/Downloads/dxvk/dxvk-master
Build dir: /home/chriundcla/Downloads/dxvk/dxvk-master/build.w64
Build type: cross build
Project name: dxvk
Native C compiler: cc (gcc 7.2.0)
Cross C compiler: /usr/bin/x86_64-w64-mingw32-gcc (gcc 6.3.0)
Native C++ compiler: c++ (gcc 7.2.0)
Cross C++ compiler: /usr/bin/x86_64-w64-mingw32-g++ (gcc 6.3.0)
Host machine cpu family: x86_64
Host machine cpu: x86_64
Target machine cpu family: x86_64
Target machine cpu: x86_64
Build machine cpu family: x86_64
Build machine cpu: x86_64
Library vulkan-1 found: YES
Library d3d11 found: YES
Library dxgi found: YES
Library d3dcompiler_47 found: YES
Program glslangValidator found: YES (/home/chriundcla/Downloads/glslang/bin/glslangValidator)
Build targets in project: 12
[3/111] Compiling C++ object 'src/util/util@sta/log_log_debug.cpp.obj'.
FAILED: src/util/util@sta/log_log_debug.cpp.obj 
/usr/bin/x86_64-w64-mingw32-g++ -Og -gstabs -Isrc/util/util@sta -Isrc/util -I../src/util -I.././include -fdiagnostics-color=always -pipe -Wall -Winvalid-pch -Wnon-virtual-dtor -std=c++1z -O0 -g -MMD -MQ 'src/util/util@sta/log_log_debug.cpp.obj' -MF 'src/util/util@sta/log_log_debug.cpp.obj.d' -o 'src/util/util@sta/log_log_debug.cpp.obj' -c ../src/util/log/log_debug.cpp
In file included from ../src/util/log/log_debug.h:5:0,
                 from ../src/util/log/log_debug.cpp:1:
../src/util/log/log.h:43:10: error: ‘mutex’ in namespace ‘std’ does not name a type
     std::mutex    m_mutex;
          ^~~~~
[...]

PR #2966, opened 2022-09-28
[d3d9] Fix for missing restriction check in ProcessVertices (#11)

adamjer:


K0bin:
Why does the commit message mention #11?

PR #2971, opened 2022-10-01
[util] set Beyond Good and Evil allowDiscard to false

Blisto91:
In some scenes there can be graphical glitches unless allowDiscard is set to false
Tested on R9 380

<details>
  <summary>Example screenshot</summary>

This will correctly show her holding a blue orb when allowDiscard is set to false

Screenshot_20221001_183139
</details>

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -582,9 +582,11 @@ namespace dxvk {
       { "d3d9.enableDialogMode",          "True" },
     }} },
     /* Beyond Good And Evil                     *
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
This may be an AMD-only problem, I guess, just FYI... never seen in on the ol' Nvidia frog GPU.

Blisto91:
On further inspection this might be a driver issue as it doesn't happen on windows with dxvk.
Also i got feedback from the devs that `allowDiscard` is in general a bad workaround.

Issue #2278, opened 2021-09-05
Escape from Tarkov launcher (BsgLauncher.exe) broken since 54ed8f0bb0f0859e05b7b406b7e95fa71900ba40 (D9VK merge).


Vinjul1704:
The game launcher for Escape from Tarkov is broken since D9VK got added with commit 54ed8f0bb0f0859e05b7b406b7e95fa71900ba40. 

It either crashes immediately or is a black window while nothing is changing, causing heavy flickering even when moving the mouse and animations are playing. Video showing the issue: https://youtu.be/XpjSq5ZpNNc

The latest launcher installer can be download directly through this link (if it happens to be invalid in the future, please let me know and I can update it): http://cdn-11.eft-store.com/LauncherDistribs/12.9.0.1521_722d8cfcd97ab13601bae0f63f28dec2/BsgLauncher.12.9.0.1521.exe

An account is not required to test the launcher, since the issue is present in the login window too.

### System information
- GPU: Nvidia GeForce GTX 1060 6GB
- Linux Driver: `470.63.01`
- Windows Driver: `471.68`
- Wine version: `wine-tkg-staging-fsync-git` from chaotic-aur
- DXVK version: `54ed8f0bb0f0859e05b7b406b7e95fa71900ba40` up to master

### Apitrace file
Recorded under Windows without DXVK: https://www.swisstransfer.com/d/db4fe1cb-902e-4b32-9d69-02295c661b12

### Windows log file (dxgi log is empty)
[BsgLauncher_d3d9.log](https://github.com/doitsujin/dxvk/files/7111569/BsgLauncher_d3d9.log)

### Linux log files
[BsgLauncher_d3d9.log](https://github.com/doitsujin/dxvk/files/7111571/BsgLauncher_d3d9.log)
[BsgLauncher_dxgi.log](https://github.com/doitsujin/dxvk/files/7111572/BsgLauncher_dxgi.log)

Faalagorn:
I also had the same issue back on NVIDIA GTX 960 and I think under RX 580 too, though I'd have to retest it when I get home, could see Intel as well, I also have the game, let me know if any further testing is needed but I think it's unrelated. Thanks for bisecting and apitracing the issue properly!

Faalagorn:
[...]

PR #2974, opened 2022-10-03
[util] Force SM1 for the Escape from Tarkov launcher

Vinjul1704:
The launcher for Escape from Tarkov uses DX9 with partial present and therefore suffers from flickering, black screens and crashes without that config option. With the config option, it runs a bit sluggish, but at least works correctly and without the graphical issues and crashes.


Blisto91:
Closes https://github.com/doitsujin/dxvk/issues/2278

PR #2976, opened 2022-10-05
Fix for saving vertex shader constants with a StateBlock

adamjer:


Joshua-Ashton:
This code looks wrong. Why is it manually capturing all previous sets of 32 constants when it sets one?

Please tell me what you are trying to fix, and what you think is wrong with the old code.

Joshua-Ashton:
Ah, I think I see, we are setting some that are out of bounds of the initial range. We have a `.setAll` helper, maybe that works?

I know we have differing sizes for our constant layouts so may need to do something like this, but it should primarily be a helper in the class imo

Blisto91 (commented):
Same as the other PR's remove the reference to your internal issue in the commit message

src/util/util_bit.h
```diff
@@ -320,6 +320,21 @@ namespace dxvk::bit {
       return get(idx);
     }
 
+    constexpr void setN(uint32_t bits) {
+      uint32_t dwords = align(bits, 32) / 32;
```
Joshua-Ashton, line 324: You can get rid of this `align` and just use `dwords` for the loop in both paths.

As in the first path, it will always be aligned to 32, so nothing changes, and in the second, the un-aligned value of dwords will always be the value of aligned dwords - 1

Joshua-Ashton (changes requested):

src/util/util_bit.h
Joshua-Ashton: Maybe call it `fullDwords` instead or something

PR #2977, opened 2022-10-05
[build] Native build changes

Joshua-Ashton:


pchome:
Could be `gnu_symbol_visibility` [argument](https://mesonbuild.com/Reference-manual_functions.html#arguments10) by default (instead `-fvisibility=hidden`). 
> Specifies how symbols should be exported, see e.g [the GCC Wiki](https://gcc.gnu.org/wiki/Visibility) for more information. This value can either be an empty string or one of default, internal, hidden, protected or inlineshidden, which is the same as hidden but also includes things like C++ implicit constructors as specified in the GCC manual. Ignored on compilers that do not support GNU visibility arguments.

Joshua-Ashton:
> Could be `gnu_symbol_visibility` [argument](https://mesonbuild.com/Reference-manual_functions.html#arguments10) by default (instead `-fvisibility=hidden`).
> 
> > Specifies how symbols should be exported, see e.g [the GCC Wiki](https://gcc.gnu.org/wiki/Visibility) for more information. This value can either be an empty string or one of default, internal, hidden, protected or inlineshidden, which is the same as hidden but also includes things like C++ implicit constructors as specified in the GCC manual. Ignored on compilers that do not support GNU visibility arguments.

I moved to using symbol scripts instead (like our .def files) because we were still exporting lots of random stray symbols due to static library leakage.

DadSchoorse:
Maybe you can generate the .sym from the .def files at build time? That would avoid duplicating information and the risk of it getting out of sync.

pchome:
@DadSchoorse 
Can be both generated, but I guess it's more likely to forget that def/list should be updated after adding new function than to forget one of a files. Two file-types just simpler, but having generated files have some pros too.

@Joshua-Ashton 
> I moved to using symbol scripts instead (like our .def files) ...

If compiler do some optimizations for hidden symbols (inline, etc., while .sym processed by linker) then should be valid to use both methods. And meson unity builds, LTO, etc. can contribute. If will not be directly supported then good practice will be not braking this at least.

eli-schwartz:
> If compiler do some optimizations for hidden symbols (inline, etc., while .sym processed by linker) then should be valid to use both methods.

Indeed there are people who do both, for this exact reason.

Joshua-Ashton:
There is no potential for such optimizations with DXVK given we use COM objects everywhere.

doitsujin:
Do we not need a .sym for d3d10core entry points?

Joshua-Ashton:
We don't support d3d10 for native builds.

PR #2981, opened 2022-10-08
setup_dxvk.sh: install with reflinks if supported

Hi-Angel:
Some filesystems, such as XFS and BTRFS, may perform a lightweight copy with CoW. This is most useful for rarely changing content, and the usecase of installing DXVK is one of such usecases, as library files are usually not modified (might be removed, but it's not modification).

So make use of it whenever supported.

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -25,7 +25,7 @@ esac
 shift
 
 with_dxgi=true
-file_cmd="cp -v"
+file_cmd="cp -v --reflink=auto"
 
 while (($# > 0)); do
   case "$1" in


doitsujin:
Isn't this the default behaviour of `cp` already? Not that I'm opposed to pulling this in but this should only affect people who use BTRFS on old distros.

Hi-Angel:
> Isn't this the default behaviour of `cp` already? Not that I'm opposed to pulling this in but this should only affect people who use BTRFS on old distros.

Well, `cp` only recently started defaulting to `--relfink=always`. I didn't find offhand the exact version, but I found that [it should be after 8.32](https://unix.stackexchange.com/a/152639/59928), and the `8.32` release is the one used in latest Ubuntu 22.04. So I would say, this affects a lot of people as Ubuntu and its derivatives like Mint are pretty popular.

I should also note that reflinks are supported not only on BTRFS but on XFS as well.

Hi-Angel:
ping

PR #2989, opened 2022-10-13
[build] Update github actions

jp7677:
Fixes Node.js 12 deprecation warning.


jp7677:
`Joshua-Ashton/gcc-problem-matcher` needs to updated too to fully get rid of that deprecation warning

jp7677:
Depends on https://github.com/Joshua-Ashton/gcc-problem-matcher/pull/2

jp7677:
Should finally be good now.

PR #3002, opened 2022-10-20
[dxgi] Store device pointer in DxgiSwapChainDispatcher

gofman:
Helps REFramework.

### DIFF
--- a/src/dxgi/dxgi_swapchain_dispatcher.h
+++ b/src/dxgi/dxgi_swapchain_dispatcher.h
@@ -8,8 +8,9 @@ namespace dxvk {
 
   public:
 
-    DxgiSwapChainDispatcher(IDXGISwapChain4* dispatch)
-      : m_dispatch(dispatch) {
+    DxgiSwapChainDispatcher(IDXGISwapChain4* dispatch, IUnknown* device)
+      : m_device(device),
+        m_dispatch(dispatch) {
     }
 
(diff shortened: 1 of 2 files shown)


src/dxgi/dxgi_swapchain_dispatcher.h
```diff
@@ -296,6 +297,7 @@ namespace dxvk {
 
   private:
 
+    void* m_device_ptr;
```
Joshua-Ashton, line 300: IUnknown* m_device;

Probably worth adding a comment as to why its not necessary to ref it either.

Joshua-Ashton (changes requested):

src/dxgi/dxgi_swapchain_dispatcher.h
gofman: Eh, that's exactly why I made it void * and suffixed _ptr, to suggest we have nothing to do with this data and specifically do not want to refcount it. Can be changed this way ofc.

Joshua-Ashton: I will defer to @doitsujin's judgement then.

doitsujin: Agree with Josh, we'd want that comment anyway to make it clear why we need an unused data member in the first place.

PR #3009, opened 2022-10-24
Disable locking non-dynamic default textures

K0bin:
Sonic Generations is slow with DXVK because it syncs with the GPU every frame:

It has a single 4x4 D3DPOOL_DEFAULT texture with a 0 usage.
`400 @0 IDirect3DDevice9::CreateTexture(this = 0xef20020, Width = 4, Height = 4, Levels = 1, Usage = 0x0, Format = D3DFMT_A8R8G8B8, Pool = D3DPOOL_DEFAULT, ppTexture = [0x12973030], pSharedHandle = NULL) = D3D_OK`

That texture gets locked every frame without any lockflags such as discard. `D3DPOOL_DEFAULT` without `D3DUSAGE_DYNAMIC` means that we do not keep the texture in memory. Therefore it has to do a GPU copy every single time and then wait for that to finish.

The fun part is that:
- the game doesn't even use the texture for anything. It just writes data to it.
- that `LockRect` call doesn't succeed on the Nvidia D3D9 driver. Which the game is perfectly fine with because all of this is pointless.

So I guess the docs aren't lying after all:
> D3DPOOL_DEFAULT pool cannot be locked unless they are dynamic textures or they are private, FOURCC, driver formats.

This improves perf quite a lot in some scenes. The game still drops below 60 but that seems to be related to asset/level streaming and not caused by DXVK.

### DIFF
--- a/src/d3d9/d3d9_common_buffer.h
+++ b/src/d3d9/d3d9_common_buffer.h
@@ -229,7 +229,6 @@ namespace dxvk {
     const D3D9_BUFFER_DESC      m_desc;
     DWORD                       m_mapFlags;
     bool                        m_needsReadback = false;
-    bool                        m_uploadUsingStaging = false;
     D3D9_COMMON_BUFFER_MAP_MODE m_mapMode;
 
     Rc<DxvkBuffer>              m_buffer;
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -512,8 +512,6 @@ namespace dxvk {
 
     int64_t                       m_size = 0;
 
-    bool                          m_systemmemModified = false;
-
     bool                          m_hazardous = false;
 
     D3D9ColorView                 m_sampleView;
@@ -525,8 +523,6 @@ namespace dxvk {
     D3D9SubresourceBitset         m_needsReadback = { };
 
     D3D9SubresourceBitset         m_needsUpload = { };
-
-    D3D9SubresourceBitset         m_uploadUsingStaging = { };
 
     DWORD                         m_exposedMipLevels = 0;
 
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4241,6 +4241,15 @@ namespace dxvk {
 
     auto& desc = *(pResource->Desc());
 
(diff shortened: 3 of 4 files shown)


blueskythlikesclouds:
This is the randomized rainbow texture the game uses whenever a texture in a material is missing. The fact that it causes massive slowdown is funny.

K0bin:
@blueskythlikesclouds Are you sure? All calls to LockRect on that texture fail, so it should just be black.

blueskythlikesclouds:
> @blueskythlikesclouds Are you sure? All calls to LockRect on that texture fail, so it should just be black.

Now that I remember, the game fills the rainbow texture I'm talking about with D3DXFillTexture, so it's possible this is a different one.

Unless D3DXFillTexture uses LockRect internally?

K0bin:
There's a UpdateTexture call with that texture early on, that definitely works.
So they write some data to that texture once at the start and then *try* to update that every single frame for some reason which always fails and the game is like ¯\\\_(ツ)\_/¯.

blueskythlikesclouds:
Hmm, I managed to locate where this texture is being created & updated in the game code, and it seems the only thing utilized is D3DXFillTexture. I'm not quite familiar with DXVK's codebase, do these functions get custom implementations?

Not debating on the pull request BTW, just found this interesting.

K0bin:
> I'm not quite familiar with DXVK's codebase, do these functions get custom implementations?

No, `D3DXFillTexture` is part of D3D**X**9. Not to be confused with D3D9. D3D**X**9 is a helper library that just calls regular D3D9 calls internally. If the game ships the library, we just use the Microsoft implementation, if not, Wine implements that function too.

blueskythlikesclouds:
https://user-images.githubusercontent.com/19259897/197618625-50a8ef04-aafc-4d4b-9ffc-990c609e90ec.mp4

Egg Dragoon boss battle actually shows this texture if you want to check.

K0bin:
I'd probably have to play to that point first. I literally downloaded the game for the very first time two hours ago to check DXVK performance. :)

Joshua-Ashton:
@K0bin did you test with Wine or Native D3DX9?

Maybe the implementation is different and that's why it fails and why it was slow...

Issue #1854, opened 2020-12-13
Final Fantasy XIII - Resolution bug @1440p


DistantThunder:
The game has a bug on Windows where 1440p is badly handled.

* Windows: https://steamcommunity.com/sharedfiles/filedetails/?id=361822631

It appears as if the game internally renders to 720p at all the times and simply stretches the picture to whatever the resolution when above 1080p (supposedly).

Various workarounds were found, mostly through downscaling.

There also exists a particular fix for FF XIII (https://github.com/rebtd7/FF13Fix) where they worked around the issue by setting the internal resolution slightly below 1440p. 

When using WINE_D3D9 with the fix, this appears to work on Linux too:

![image](https://user-images.githubusercontent.com/3923810/102016791-f9162700-3d62-11eb-8376-20aa3586ea87.png)

However, going through DX/9VK the Windows behaviour is "kind of" reproduced, as the screen is split diagonally where one half renders correctly, and the other half renders like in Windows bugged state:

![image](https://user-images.githubusercontent.com/3923810/102016829-44303a00-3d63-11eb-90ae-ba0d3e2eaa3b.png)

This bug happens with and without the injected ffxiii-dll fix.

Api Trace included. 

Thank you for your work, hoping a solution could be found.

### Software information
Final Fantasy XIII - Steam version

### System information
- GPU: 5700 XT
- Driver: Mesa 21.0.0-devel (git-ac0d393eb1) (ACO)
- Wine version: Proton 5.0.10
- DXVK version: v1.7-8-g51531e4a7
- AMDGPU: 5.9.13 (Arch Linux)

### Apitrace file(s)
- http://dl.free.fr/o332gdN9c

### Log files
- d3d9.log: [ffxiiiimg_d3d9.log.zip](https://github.com/doitsujin/dxvk/files/5684926/ffxiiiimg_d3d9.log.zip)

Joshua-Ashton:
Thanks, will take a look later today!

DistantThunder:
Thanks, feel free to ask for details, I'll provide as best as I can.

Joshua-Ashton:
Fixed with 65635b23c1e8575362722c99e6ffdf9d9df64116.

DistantThunder:
Confirmed working on my end:

![image](https://user-images.githubusercontent.com/3923810/102261063-72537c80-3f11-11eb-8154-e6212e07851a.png)

Joshua-Ashton:
Yw!

[...]

PR #3011, opened 2022-10-26
[d3d9] remove Correctness Factor to PASS D3d9 WHQL test

eni-intc:


doitsujin:
This is going to break texture filtering in fullscreen passes on some hardware, which needs to be addressed.

Joshua-Ashton:
Yes, this will break rendering under AMD.

Joshua-Ashton:
https://github.com/doitsujin/dxvk/issues/1854

PR #3018, opened 2022-10-30
[d3d9] Don't advertise support for MS INTZ format

### LINKED ISSUE #3017: Empire Total War invisible game objects
While i was testing the `surface-factory` branch in random games i noticed a bug in the Windows version of Empire Total War.
If you turn on both Volumetric effects and Anti Aliasing then most of the terrain and objects become invisible.

This also happens with wined3d on Linux and with dxvk on Windows. Doesn't happen with native Windows without dxvk.
I think the trace i made through proton just have the issue baked into it so i also recorded one on Windows and when replaying that one on Linux with dxvk the issue doesn't show so i included both.

I haven't played with config options yet, so this might be something that can just be solved through that. Making the issue first so i won't forget about it.

<details>
  <summary>Screenshot</summary>

Screenshot_20221030_190747
</details>

### Software information
Empire Total War.
Volumetric effects and Anti Aliasing turned on

### System information
- GPU: R9 380
- Driver: mesa-git
- Wine version: Proton experimental & 7.0-3 

### Apitrace file(s)
https://mega.nz/file/A1wVSKoR#NmJqOV_pXo-_jOv-NhUZ0IuyT4MTSavA1ULyxlax7sA

### Log files
[Empire_d3d9.log](https://github.com/doitsujin/dxvk/files/9896589/Empire_d3d9.log)

K0bin:
Fixes #3017

D3D9 does not allow sampling the depth buffer directly. So the INTZ format driver hack exists as a hack to make that possible.
D3D9 also does not allow sampling multi sampled textures. `CreateTexture` doesn't even have a sample count parameter. So it makes no sense to advertise support for multisampled INTZ.

Empire Total War checks whether INTZ is supported with MSAA and if it is, it will always use that format for depth buffers. Problem is that it can only create a 1 sample texture with that format but tries to use that with the multi sampled render target regardless. This PR fixes that.

### DIFF
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -186,7 +186,8 @@ namespace dxvk {
     if (MultiSampleType != D3DMULTISAMPLE_NONE
      && (SurfaceFormat == D3D9Format::D32_LOCKABLE
       || SurfaceFormat == D3D9Format::D32F_LOCKABLE
-      || SurfaceFormat == D3D9Format::D16_LOCKABLE))
+      || SurfaceFormat == D3D9Format::D16_LOCKABLE
+      || SurfaceFormat == D3D9Format::INTZ))
       return D3DERR_NOTAVAILABLE;
 
     uint32_t sampleCount = std::max<uint32_t>(MultiSampleType, 1u);


Blisto91:
Confirmed fixed.
Give this man a medal!

Issue #3024, opened 2022-11-04
Add FPS limit to a few wayforward games.


Bitwolfies:
Shantae and the Pirate's Curse, and Mighty Switch Force! Collection are all broken when run above 60, but run at 144 by default. The exe's for the respective games are "ShantaeCurse.exe" and "MSFC.exe"

Blisto91:
I'm guessing this is also relevant for the other few Shantae games and the individual games within the `Mighty Switch Force! Collection` ? 👀

Bitwolfies:
> I'm guessing this is also relevant for the other few Shantae games and the individual games within the `Mighty Switch Force! Collection` ? 👀

The collection uses a single exe for all its games, so the one limiter will do. The other shantaes are surprisingly fine.

Have not tested if Mighty Switch Force Hose It Down needs a limit though.

PR #3033, opened 2022-11-06
[util] Limit fps to 60 for some WayForward games

Blisto91:
Closes https://github.com/doitsujin/dxvk/issues/3024

Will make a note to check if the couple individual games in the collection behave the same.


Blisto91:
@Bitwolfies It's correct that these games are DirectX 11 right?

Bitwolfies:
You'd be correct.

Issue #3038, opened 2022-11-08
[d3d9] Sonic & Sega All-Stars Racing Transformed Collection: performs poorly when Risezable BAR is enabled


ranplayer:
This old game performs poorly (inconsistent fps, and huge drops) when Resizable BAR is enabled on the BIOS (not a DXVK bug, since it's the same for Windows). I was just wondering if there is a possible workaround on the DXVK's side. This feature does improve performance in newer games, and it would be nice to keep it enable on the system. I've tried setting `RADV_PERFTEST=nosam`, but no success. Is there any DXVK's property that I could try ?

### Software information
Name: Sonic & Sega All-Stars Racing Transformed_Collection
SteamId: 212480

### System information
- GPU: RX 6700 XT 12 Gb
- Driver: Mesa 22.2.2
- Wine version: Proton 7.0-4 / 7.0-5 (Experimental)
- DXVK version: 1.10.3 and latest changes from master

### Log files
- [d3d9_resizable_bar.log](https://github.com/doitsujin/dxvk/files/9961661/ASN_App_PcDx9_Final_d3d9-resizable-bar.log)
- [d3d9_no_resizable_bar.log](https://github.com/doitsujin/dxvk/files/9961663/ASN_App_PcDx9_Final_d3d9-no-resizable-bar.log)

doitsujin:
For context, this is usually caused by games reading from memory that's only supposed to be written to, and sadly not uncommon at all. Like Josh said, `d3d9.apitraceMode = True` in dxvk.conf (see https://github.com/doitsujin/dxvk/wiki/Configuration) should help.

ranplayer:
Thanks for the explanation, folks. That property did trick ! It's running smoothly now with Resizable BAR enabled.

Blisto91:
@ranplayer Can you test if it has the same effect when you instead add `d3d9.allowDirectBufferMapping = False`
Use it with proton experimental.

ranplayer:
@Blisto91 , this property helps a little, but the apitrace mode provides a solid and consistent performance.

Blisto91:
Just noting here that https://github.com/doitsujin/dxvk/pull/3039 might be a temporary workaround until the devs have looked into the issue more

PR #3039, opened 2022-11-08
[util] Enable apitraceMode for Sonic & All-Stars Racing Transformed

Blisto91:
Temporary (maybe) workaround for https://github.com/doitsujin/dxvk/issues/3038 until devs have looked into it more.


K0bin:
Disabling direct buffer mapping should achieve the same thing and without impacting stuff like staging buffers.

PR #3041, opened 2022-11-08
[dxvk] Rework SetEventOnCompletion

K0bin:
DXVK equivalent to: https://github.com/HansKristian-Work/vkd3d-proton/pull/1284

* Create the waiter thread on demand
* Wait for the lowest requested value to avoid waking the thread up every 10ms when no events are queued.
* Wait on caller thread when hEvent = null

### DIFF
--- a/src/dxvk/dxvk_fence.cpp
+++ b/src/dxvk/dxvk_fence.cpp
@@ -56,28 +56,51 @@ namespace dxvk {
         Logger::warn(str::format("Importing semaphores of type ", info.sharedType, " not supported by device"));
       }
     }
-
-    m_thread = dxvk::thread([this] { run(); });
   }
 
 
   DxvkFence::~DxvkFence() {
-    m_stop.store(true);
-    m_thread.join();
-
+    if (m_thread.joinable()) {
+      {
+        std::unique_lock<dxvk::mutex> lock(m_mutex);
+        m_running = false;
+        m_condVar.notify_one();
+      }
+      m_thread.join();
+    }
     m_vkd->vkDestroySemaphore(m_vkd->device(), m_semaphore, nullptr);
   }
 
 
   void DxvkFence::enqueueWait(uint64_t value, DxvkFenceEvent&& event) {
-    std::unique_lock<dxvk::mutex> lock(m_mutex);
-
-    if (value > m_lastValue.load())
+    if (value > getValue()) {
+      std::unique_lock<dxvk::mutex> lock(m_mutex);
       m_queue.emplace(value, std::move(event));
-    else
+
+      if (!m_running) {
+        m_running = true;
+        m_thread = dxvk::thread([this] { run(); });
+      } else {
+        m_condVar.notify_one();
+      }
+      lock.unlock();
+    } else {
       event();
+    }
   }
-  
+
+  void DxvkFence::wait(uint64_t value) {
+    VkSemaphoreWaitInfo waitInfo = { VK_STRUCTURE_TYPE_SEMAPHORE_WAIT_INFO };
+    waitInfo.semaphoreCount = 1;
+    waitInfo.pSemaphores = &m_semaphore;
+    waitInfo.pValues = &value;
+    VkResult vr = m_vkd->vkWaitSemaphores(
+      m_vkd->device(), &waitInfo, ~0ull);
+
+    if (vr != VK_SUCCESS) {
+      Logger::err(str::format("Failed to wait for semaphore: ", vr));
+    }
+  }
 
   void DxvkFence::run() {
     uint64_t value = 0ull;
@@ -87,8 +110,10 @@ namespace dxvk {
     waitInfo.pSemaphores = &m_semaphore;
     waitInfo.pValues = &value;
 
-    while (!m_stop.load()) {
+    while (true) {
       std::unique_lock<dxvk::mutex> lock(m_mutex);
+
+      m_condVar.wait(lock, [&]() { return !m_queue.empty() || !m_running; });
 
       // Query actual semaphore value and start from there, so that
       // we can skip over large increments in the semaphore value
@@ -99,8 +124,6 @@ namespace dxvk {
         return;
       }
 
-      m_lastValue.store(value);
-
       // Signal all enqueued events whose value is not greater than
       // the current semaphore value
       while (!m_queue.empty() && m_queue.top().value <= value) {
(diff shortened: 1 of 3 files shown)


src/dxvk/dxvk_fence.cpp
```diff
@@ -56,80 +57,94 @@ namespace dxvk {
...
     }
-
-    m_thread = dxvk::thread([this] { run(); });
   }
 
 
   DxvkFence::~DxvkFence() {
-    m_stop.store(true);
-    m_thread.join();
+    if (m_running.exchange(false, std::memory_order_acq_rel)) {
```
doitsujin, line 64: need to lock here, or otherwise we'll risk setting this and notifying the cond var before the worker thread actually waits on the cond var.

src/dxvk/dxvk_fence.cpp
```diff
@@ -56,80 +57,94 @@ namespace dxvk {
...
+    if (value > getValue()) {
+      {
+        std::unique_lock<dxvk::mutex> lock(m_mutex);
+        m_events.emplace_back(value, std::move(event));
+      }
 
-    if (value > m_lastValue.load())
-      m_queue.emplace(value, std::move(event));
-    else
+      if (!m_running.exchange(true, std::memory_order_acq_rel)) {
```
doitsujin, line 80: we have to lock anyway so i don't really see any benefit to doing any of this outside the lock, especially since atomic exchange is more expensive than a simple `if (m_running)`.

src/dxvk/dxvk_fence.cpp
```diff
@@ -56,80 +57,94 @@ namespace dxvk {
...
+    VkSemaphoreWaitInfo waitInfo = { VK_STRUCTURE_TYPE_SEMAPHORE_WAIT_INFO };
+    waitInfo.semaphoreCount = 1;
+    waitInfo.pSemaphores = &m_semaphore;
+    waitInfo.pValues = nullptr;
 
-      if (vr != VK_SUCCESS) {
-        Logger::err(str::format("Failed to query semaphore value: ", vr));
-        return;
-      }
+    while (m_running.load(std::memory_order_acquire)) {
```
doitsujin, line 108: `acquire` doesn't make a huge amount of sense here, keeping the atomic is fine so we don't need to lock outside the loop but memory order literally doesn't matter for this one.

src/dxvk/dxvk_fence.cpp
```diff
@@ -56,80 +57,94 @@ namespace dxvk {
...
+      QueueItem smallestValueItem;
+      {
+        std::unique_lock<dxvk::mutex> lock(m_mutex);
 
-      m_lastValue.store(value);
+        m_condVar.wait(lock, [&]() { return !m_events.empty(); });
```
doitsujin, line 113: this also needs to check `m_running`.

doitsujin (changes requested):
Hmm, thinking about this, the issue here is that we'll be blocking any future waits for 10ms even if they could be signaled before the current smallest wait. Suppose something like

```c++
fence->SetEventOnCompletion(1000, event1);
// do enough work that the fence thread will start waiting on this
fence->SetEventOnCompletion(1, event2);
fence->Signal(1);
// this wait will stall until the vkWaitSemaphores call that waits for value 1000 times out
WaitForSingleObject(event2);
```

Avoiding situations like that was the main motivation behind the whole +1 thing, even if it is somewhat awkward. Unfortunately I think

src/dxvk/dxvk_fence.cpp
K0bin: Would that actually be a problem? The worker thread would check if m_running is set as part of the condition variable predicate and skip waiting.

HansKristian-Work: This series of events is possible:
- Thread reads m_running (true)
- This thread writes m_running.
- This thread calls notify_one().
- Thread calls wait().
- Deadlock

K0bin: If the write to m_running happens "in between" the m_running check on the worker thread and the wait call?

src/dxvk/dxvk_fence.cpp
```diff
@@ -56,59 +57,76 @@ namespace dxvk {
...
     uint64_t value = 0ull;
 
     VkSemaphoreWaitInfo waitInfo = { VK_STRUCTURE_TYPE_SEMAPHORE_WAIT_INFO };
     waitInfo.semaphoreCount = 1;
     waitInfo.pSemaphores = &m_semaphore;
-    waitInfo.pValues = &value;
+    waitInfo.pValues = &value;;
 
-    while (!m_stop.load()) {
+    while (m_running.load(std::memory_order_relaxed)) {
```
HansKristian-Work, line 109: Polling m_running here is somewhat meaningless, since it will be taken care of by m_condWait.

src/dxvk/dxvk_fence.cpp
```diff
@@ -56,59 +57,76 @@ namespace dxvk {
...
       std::unique_lock<dxvk::mutex> lock(m_mutex);
 
+      m_condVar.wait(lock, [&]() { return !m_queue.empty() && m_running.load(std::memory_order_acq_rel); });
```
HansKristian-Work, line 112: m_running is always modified in a lock, so does not need to be atomic.

HansKristian-Work (commented):
Looks correct to me now.

src/dxvk/dxvk_fence.cpp
K0bin: It was atomic because of the loop condition. But now that we got rid of that, it really doesnt need to be atomic.

src/dxvk/dxvk_fence.cpp
```diff
@@ -56,28 +56,55 @@ namespace dxvk {
...
-
-    m_thread = dxvk::thread([this] { run(); });
   }
 
 
   DxvkFence::~DxvkFence() {
-    m_stop.store(true);
-    m_thread.join();
+    std::unique_lock<dxvk::mutex> lock(m_mutex);
+    bool threadRunning = m_running;
```
doitsujin, line 64: this could be simplified quite a bit, just do something like
```c++
if (m_thread.joinable()) {
  { std::unique_lock /* ... */
    // set flag, notify cond
  }
  m_thread.join();
}
```

doitsujin (commented):
smol nit

doitsujin (approved):

PR #3051, opened 2022-11-11
[hud] Display build time on dirty working trees

AlpyneDreams:
Simple single commit to make debugging DXVK substantially easier based on what I'm currently doing in D8VK. If the working tree is dirty ('+' in the version on the HUD), the compile date and time of the current frontend are shown on the HUD. This helps us check if builds installed correctly and how long builds took, among other things.

Here is an example:
image

This also allows changing the D3D9 API name if desired, which is useful for D3D8 and lower.


Joshua-Ashton:
Exposing the API name thing is fine, but having date/time is bad because it breaks reproducible builds.

Joshua-Ashton:
I recommend exposing the API name thing the same way we do for DX10.

AlpyneDreams:
> Exposing the API name thing is fine, but having date/time is bad because it breaks reproducible builds.

> I recommend exposing the API name thing the same way we do for DX10.

I have modified it so that the date and timestamp is only included if the working tree is dirty. I removed the SetAPIName thing from D3D9 as that was a leftover from d8vk.

ViNi-Arco:
> I have modified it so that the date and timestamp is only included if the working tree is dirty.

Hey **Alpyne**, I think removing the date/timestamp  for dirty trees too, that didn't seem interesting, for I agree with what **Josh** said `it breaks reproducible builds`.

AlpyneDreams:
> > I have modified it so that the date and timestamp is only included if the working tree is dirty.
> 
> Hey **Alpyne**, I think removing the date/timestamp for dirty trees too, that didn't seem interesting, for I agree with what **Josh** said `it breaks reproducible builds`.

Non-timestamped builds can be reproduced by building with a non-dirty tree or by adding `&& 0 ` to line 5 of `version.h.in`. As far as I know there aren't usually cases where dirty tree builds have needed to be reproduced precisely, but correct me if I'm wrong.

If people want I can add a meson option for timestamping that's off by default too, or I can just close this PR. Up to the maintainers. I only made this PR at @K0bin's suggestion, otherwise I would probably have left it upstream in d8vk.

ViNi-Arco:
> If people want I can add a meson option for timestamping that's off by default

Good idea, I like it, an option for that, I think I would even use :relaxed:

AlpyneDreams:
Added as an option :)

Run `meson configure -Dbuild_date=true <build dir>` to enable it or set it to `always` if you don't care if the working tree is dirty (although at that point you can just go by commit ID).

ViNi-Arco:
Sup @AlpyneDreams, did you build the DXVK with this modification you made?

GCC-Mingw and CLANG-Mingw do not build DXVK with your Patch..

**GCC-Mingw:**
GCC dxvk hud patch

**CLANG-Mingw:**
patch HUD fail clang

And you will have to Bump the required meson version to >=0.55
Aviso Meson DXVK

PR #3072, opened 2022-11-16
[wsi] Add getMonitorEdid function

Joshua-Ashton:
What an unbelievable pain this is to do on Windows...

No-op on SDL2 right now, as there is nothing for that.

I plan to parse this out into IDXGIOutput for the colorimetry information when this lands, but I want to get this part reviewed first before doing that.


Joshua-Ashton:
cc: @gofman

PR #3085, opened 2022-11-20
Implement capturing lights in state blocks

K0bin:
Need For Speed Most Wanted (2005) needs this.


Joshua-Ashton (changes requested):
Missing the implementation in `CaptureType` to do an initial capture.

Joshua-Ashton (approved):

PR #3093, opened 2022-11-24
setup_dxvk.sh: switch to posix

wael444:
+ changed shebang to `/bin/sh`, as using env could mean a broken system, and we don't w
ant that. 
+ changed the need of `readlink` and dirname to realpath 
+ changed the need of `dirname` to shell substitution 
+ changed arthmetic non-posix 'calculations' to native shell
+ removed the need of removing `\r` as it didn't exist in the first place 
+ changed checking `if $?` to shell native `||` and `&&`
  + `comm || foo` run foo if comm fails (non 0 exit code)
  + `comm && foo` run foo if comm succeeds (0 exit code) 
+ properly tabbed/spaced lines 
+ check if script is not being ran from tarball 

this script overall has some issues, and ive simply shellchecked it and made it POSIX f
or systems such as mine and to remove the need for bash.

### DIFF
--- a/setup_dxvk.sh
+++ b/setup_dxvk.sh
@@ -1,40 +1,42 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # default directories
 dxvk_lib32=${dxvk_lib32:-"x32"}
 dxvk_lib64=${dxvk_lib64:-"x64"}
 
 # figure out where we are
-basedir="$(dirname "$(readlink -f "$0")")"
+basedir="$(realpath "${0%/*}")"
 
 # figure out which action to perform
 action="$1"
 
+
+{ [ -d "x32" ] || [ -d "x64" ] ; } || {
+	echo "Do not run this script from the git repository."
+	echo "Run it from the dxvk tarball provided in https://github.com/doitsujin/dxvk/releases" 
+    exit 1
+}
+
 case "$action" in
-install)
-  ;;
-uninstall)
-  ;;
-*)
-  echo "Unrecognized action: $action"
-  echo "Usage: $0 [install|uninstall] [--without-dxgi] [--symlink]"
-  exit 1
+  install) ;;
+  uninstall) ;;
+  *)
+    echo "Unrecognized action: $action"
+    echo "usage: $0 [install|uninstall] [--without-dxgi] [--symlink]"
+    exit 1
 esac
 
 # process arguments
 shift
 
+# default variables
 with_dxgi=true
 file_cmd="cp -v --reflink=auto"
 
-while (($# > 0)); do
+while [ $# -gt 0 ]; do
   case "$1" in
-  "--without-dxgi")
-    with_dxgi=false
-    ;;
-  "--symlink")
-    file_cmd="ln -s -v"
-    ;;
+    "--without-dxgi") with_dxgi=false ;;
+    "--symlink") file_cmd="ln -s -v" ;;
   esac
   shift
 done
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
I don't really see the advantage to moving to sh aside from it being more painful for us to maintain.

If a user has an esoteric enough config to not have bash (or bash-compatible substitute) then they are probably competent enough to install it into a prefix manually.

Joshua-Ashton:
I also have to say, I don't see why a user of DXVK wouldn't have bash, when it's like < 1MB given all of the other space redundancies going on with Wine/Proton prefixes :frog:

wael444:
> it being more painful for us to maintain.

Hmm? How is it gonna be any harder? Is a simple wine dll installation script gonna have to be complex?

> If a user has an esoteric enough config to not have bash 

Convenience... not everyone with a POSIX system wants to spend time making their own installation script.

> I don't see why a user of DXVK wouldn't have bash,

Not a valid excuse. By then DXVK can be portable and installable on wineprefixes without the need of a extra dependency for such a simple script.

What is there to lose? there is barely any 'loss' in this situation.

doitsujin:
> Not a valid excuse. 

Please give us a concrete example of a real-world setup that can run DXVK but does not have bash.

Sorry if this doesn't really sound all that diplomatic but I'm really tired of people constantly forcing their own ideas of what the setup script should look like on us. We get one of these like once every couple of months, we've *already* rejected one that switched to plain `sh` because it didn't seem useful, we've rejected a whole bunch more that would either straight-up break things or add options to work around people's questionable workflows, we've *accepted* one that cleaned up a bunch of stuff and switched to the current `/usr/bin/env` shebang thing, and now this one reverts all that again and just makes things annoying.

Might as well remove the script at this point because I swear to god, we can't ever make anybody happy and clearly it's better if everyone writes their own that suits their needs.

wael444:
1. re-write the script as simple and understandble so others dont complain and can simply add with pull requests
2. keep this script and have dxvk be portable.
3. have a member that can do the first one, and have him review pull rwquests for script changes.

K0bin:
The current one works fine for 99% of users and I don't really see a need to over-complicate it.

wael444:
the script by itself is already unneccessarily complicated. simply making the script *portable* over-complicates it?

just sounds like you want [1] to occur

doitsujin:
Again:

> Please give us a concrete example of a real-world setup that can run DXVK but does not have bash.

Portability isn't useful in a vacuum, there's zero value to having the script work on an unmaintained Solaris box from the 90s but I get the feeling that that's where we're going with this.

MacSlash:
"Yeah let's just completely ignore a bunch of people who don't want to use bash because they don't see the point in learning it or having it on their system for a single program" You guys do know people have simple distros like KISS right? Literally changes nothing except makes it more annoying for those people.

doitsujin:
> You guys do know people have simple distros like KISS right?

I don't know every single distro in the world, no. That's quite literally why I asked for a **real-world example**.

nvidiaLinuxUser:
Why add an extra dependency if you just don't need it? I don't have Bash installed on my system

doitsujin:
You're quite literally the first person I've seen to outright say that rather than just arguing about some abstract portability bullshit. I genuinely didn't expect there's people out there who don't have it installed given that bash isn't exactly a niche piece of software.

Which also raises the question what people do when there's something using Python, but well.

Anyway, not sure if we can/want to maintain the script like this. Even if I were to merge this, we'd just get more complaints about it anyway so I'd rather just get rid of it.

pixelcluster:
TBH, I don't get what the trouble with installing bash is. It's around 8MB, which is nearly nothing compared to wine or mingw (which I assume also have to be separately installed on distros focused on being lightweight). It would be near-zero effort to install bash alongside those dependencies, and if you hate it that much, uninstall it again afterwards.
It sounds like a PEBCAK problem to me at least.

Blisto91:
As a quick fyi https://github.com/doitsujin/dxvk/pull/3097

wael444:
Ok. since now the setup is manual and straight forward i guess it's fine, probably will break software utilizing dxvk..

Blisto91:
I and others from the community will try to give a heads up anywhere we know the script is being used.

PR #3097, opened 2022-11-25
Fix all issues with the setup script once and for all

doitsujin:
Clearly this is better than providing a script at all given that we're constantly getting complaints about it.

Most people don't use this anyway.


README.md
```diff
@@ -9,26 +9,17 @@ The most recent development builds can be found [here](https://github.com/doitsu
 Release builds can be found [here](https://github.com/doitsujin/dxvk/releases).
 
 ## How to use
-In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, run the following commands from within the DXVK directory:
-
+In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, copy or symlink the DLLs into the following directories as follows, then open `winecfg` and manually add DLL overrides for `d3d11`, `d3d10core`, `dxgi`, and `d3d9`:
 ```
-export WINEPREFIX=/path/to/.wine-prefix
-./setup_dxvk.sh install
+WINEPREFIX=/path/to/wineprefix
```
pendingchaos, line 14: I think this needs to be exported to be used by winecfg

shmerl:
Since you removed the script, may be it would be good to document on the project wiki how to actually "install" things manually then? I.e. where to place / symlink the library (such as 32-bit and 64-bit versions)? Not everyone is familiar with it, especially when 32-bit location is confusingly called `syswow64` and 64-bit location - `system32` in wow64 prerix, and just `system32` in 32-bit prefix.

Blisto91:
@shmerl feel free to open a PR. Worst that can happen is that it gets closed again. 🙂

pchome:
FYI: one more way to use DXVK:
```bash
export WINEPREFIX=/path/to/wineprefix
export WINEDLLOVERRIDES='*d3d9,*d3d11,*d3d10core,*dxgi=n'
export WINEPATH='Z:\\path\\to\\dxvk\\x64;Z:\\path\\to\\dxvk\\x32'
```
But, for this to work one should **(re)move original d3d9.dll, d3d11.dll, etc.** from wine prefix.

Likely an setup and/or launcher script needed for this trick, but it may come handy in cases like switching between multiple dxvk versions/builds for testing (by changing `WINEPATH`). Can't recommend this for regular users or for use on _important_ wine prefixes.

p/s: Same can be done for d3d**x**\*.dll, install all of them into separate wine prefix, move somewhere to /path/to/d3dx/x{32,64}, add to overrides and path (`;` separator for both) ...

shmerl:
Oh,  I didn't know about `WINEPATH`: https://gitlab.winehq.org/wine/wine/-/blob/master/loader/wine.man.in#L204

It's also a neat trick that it can mix 32-bit and 64-bit libraries in the same string.

pchome:
> Should you use WINEDLLPATH instead for the above?

Last time I checked WINEDLLPATH only handle builtin dlls (path to wine itself).

ltsdw:
@shmerl as important as installing, is how one would remove dxvk then?

from the scrip it used to rename the original dlls to *.old before replacing them, and renaming them back when unistalling, perhaps this also should be mentioned.

I think it would also be good mention how one does that entirely from the command line without the need of winecfg (or maybe not?).

for example:

```
wine reg add 'HKEY_CURRENT_USER\Software\Wine\DllOverrides' /v d3d9 /d native /f
```

or

```
wine reg delete 'HKEY_CURRENT_USER\Software\Wine\DllOverrides' /v d3d9 /f
```

shmerl:
I think using .old files isn't a good approach, since by the time you might want to un-install things, they'd be outdated. I think the method with deleting and using `wineboot -u` that readme alraedy mentions is a better method.

PR #3100, opened 2022-11-26
README: fix wine dll path

wael444:
`$WINEPREFIX/drive_c/system32` does not exist for me on wine 7.20

`drive_c/`
```
./windows
./users
./Program Files
./ProgramData
./Program Files (x86)
```


wael444:
unless this means that older or newer Wine versions have different paths..

PR #3101, opened 2022-11-27
Clarify installation for 64-bit and 32-bit prefixes.

shmerl:
More details for manual installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,30 @@ The most recent development builds can be found [here](https://github.com/doitsu
 Release builds can be found [here](https://github.com/doitsujin/dxvk/releases).
 
 ## How to use
-In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, copy or symlink the DLLs into the following directories as follows, then open `winecfg` and manually add DLL overrides for `d3d11`, `d3d10core`, `dxgi`, and `d3d9`:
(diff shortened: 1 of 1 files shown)


wu-yue:
Maybe it would be better to explicitly point out that the syswow64 and system32 would swap under different arch. Took me sometime to realize it.

wael444:
plz merge because the one in README doesnt work but this clarifies what to actually do

README.md
```diff
@@ -9,14 +9,25 @@ The most recent development builds can be found [here](https://github.com/doitsu
 Release builds can be found [here](https://github.com/doitsujin/dxvk/releases).
 
 ## How to use
-In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, copy or symlink the DLLs into the following directories as follows, then open `winecfg` and manually add DLL overrides for `d3d11`, `d3d10core`, `dxgi`, and `d3d9`:
+In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, copy or symlink the DLLs into the Wine prefix system directories as shown below, then open `winecfg` and manually add DLL overrides for `d3d11`, `d3d10core`, `dxgi`, and `d3d9`.
```
wael444, line 12: it's not exactly clarified if native or builtin overrides are supposed to be added, setup_dxvk does native.

shmerl: OK, may be I can simply add an actual command that does the override to simplify this and clarify it in the text as well.

PR #3109, opened 2022-12-01
[wsi] Fix crash when parsing EDID fails

doitsujin:
If info is null then we shouldn't pass it around.

Also improve formatting a bit.

Found by @K0bin .


K0bin (approved):
@Joshua-Ashton should probably still look into replacing `tmpfile()` in libdisplay-info to avoid the error in the first place.

Joshua-Ashton:
Yes, I will do that. Annoying that the MinGW implementation is broken :/

PR #3114, opened 2022-12-04
[util] Set Secret World Legends launcher to SM 2

Blisto91:
With default shader model 3 the launcher UI elements are invisible.

<details>
  <summary>Screenshots without and with the option</summary>
  
Screenshot_20221126_214912
Screenshot_20221126_214838
</details>

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,5 +1,5 @@
-                  Copyright (c) 2017-2021 Philip Rebohle
-                  Copyright (c) 2019-2021 Joshua Ashton
+                  Copyright (c) 2017 Philip Rebohle
+                  Copyright (c) 2019 Joshua Ashton
 
                           zlib/libpng license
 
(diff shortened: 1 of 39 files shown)


K0bin:
Do we know what the problem is? If it's partial presentation, I'd be fine with merging this as a workaround. If not, I'd rather investigate the root cause.

Blisto91:
No idea. I asked a couple of times if anyone had any ideas on discord since it doesn't seem like a partial presentation issue, but didn't get any feedback and just opted for the PR for now. Can also make a issue if you think so.
Sadly apitrace has issues replaying traces made for the launcher. Even captured on windows and replayed natively again there. So will  have to make a issue at the apitrace repo also i guess.

K0bin:
Please open an issue.

Blisto91:
https://github.com/doitsujin/dxvk/issues/3116

Blisto91:
Reopenes since Partial Presentation is suspected anyway

PR #3139, opened 2022-12-18
[d3d11] Implement HDR swapchain features

Joshua-Ashton:
Implements CheckColorSpaceSupport, SetColorSpace and SetHDRMetadata.

### DIFF
--- a/src/vulkan/vulkan_presenter.h
+++ b/src/vulkan/vulkan_presenter.h
@@ -49,6 +49,7 @@ namespace dxvk::vk {
    */
   struct PresenterFeatures {
     bool                fullScreenExclusive : 1;
+    bool                hdrMetadata : 1;
   };
   
   /**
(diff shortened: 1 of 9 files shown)


src/vulkan/vulkan_presenter.h
```diff
@@ -183,6 +183,17 @@ namespace dxvk::vk {
...
     }
 
+    /**
+     * \brief Gets list of supported surface formats
+     *
+     * \param [out] formats List of supported surface formats
+     * \param [in] fullScreenExclusive Fullscreen exclusive state
+     * \returns Status of the operation
+     */
+    VkResult getSupportedFormats(
```
doitsujin, line 194: Can we expose a `supportsColorSpace` kind of function instead? Having to pass somewhat unrelated data around and also receiving info that we don't need for that purpose (the exact VkFormat) is kind of bad design.

doitsujin (changes requested):
One thing I'd like to see addressed, looks good otherwise.

src/vulkan/vulkan_presenter.h
Joshua-Ashton: Sure

Joshua-Ashton: Done.

PR #3163, opened 2023-01-02
[dxvk] Use bottom-of-pipe stage for writeTimestamp

ishitatsuyuki:
The Vulkan specifications says:
> VUID-vkCmdWriteTimestamp2-stage-03859
> stage must only include a single pipeline stage

`VK_PIPELINE_STAGE_ALL_COMMANDS_BIT` is a union of stages, so use `VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT` which is logically equivalent instead.

### DIFF
--- a/src/dxvk/dxvk_gpu_query.cpp
+++ b/src/dxvk/dxvk_gpu_query.cpp
@@ -306,7 +306,7 @@ namespace dxvk {
       handle.queryId);
     
     cmd->cmdWriteTimestamp(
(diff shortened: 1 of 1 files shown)


doitsujin:
No objection to merging this, but is this still true? `BOTTOM_OF_PIPE` is supposed to be deprecated as of KHR_synchronization2.

ishitatsuyuki:
Hmm, yeah. Looks like I just encountered some inconsistency in the spec. Since `ALL_COMMANDS` works on all driver anyway, I'll just close this.

PR #3173, opened 2023-01-07
[dxvk] Remove SDL dummy window

hkva:
SDL >= 2.0.9 permits passing a nullptr window to SDL_Vulkan_GetInstanceExtensions, so there's no point in going though the work of creating a window just to call this function.


doitsujin:
Not sure what SDL we need to stay compatible with so just gonna ping @Joshua-Ashton

Joshua-Ashton:
The Steam Runtime ships 2.26.1 so it should be fine.

PR #3176, opened 2023-01-08
Add options to enforce sample rate shading

doitsujin:
Or suplersampling in simple terms.

Main motivation was to fix the annoing shimmering in Cold Steel 3 and 4 that's caused by normal mapping. Obviously the GPU performance impact is a bit on the absurd side of things, but MSAA is only really supported in old/simple games that run on a toaster anyway.

Regular 4xMSAA:
off

4xMSAA with sample rate shading:
on

Haven't really tested the D3D9 side of things much.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -201,6 +201,15 @@
 # d3d9.invariantPosition = True
 
 
+# Forces per-sample rate shading when MSAA is enabled, rather than per-pixel
+# shading. May improve visual clarity at a significant performance cost, but
+# may also introduce visual issues in some games.
+#
+# Supported values: True, False
+
+# d3d11.forceSampleRateShading = False
+
+
 # Forces the sample count of all textures to 1, and performs
 # the needed fixups in resolve operations and shaders.
 #
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -208,6 +208,7 @@
 # Supported values: True, False
 
 # d3d11.forceSampleRateShading = False
+# d3d9.forceSampleRateShading = False
 
 
 # Forces the sample count of all textures to 1, and performs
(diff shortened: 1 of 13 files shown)


doitsujin:
Merging manually since github doesn't allow me to for some reason.

OCRBonk:
This is beautiful! Sample-Rate shading is the same as Nvidia [SGSSAA](forums.guru3d.com/showthread.php?t=357956) (Forceable in DX8/9/Old OGL. Enhanceable with variable results in DX11) and AMD's [SRSSAA](https://github.com/GPUOpen-LibrariesAndSDKs/SSAA11)
Seeing this supported here is fantastic(And vendor agnostic!). 
If only you could implement sort of compatibility flags heuristics to force it in games that don't support MSAA by default. (like what Nvidia driver does for DX9, which works even on games that don't support MSAA) Sure the performance impact immediately is large, but it's a future proof setting that allows extra scalability on future GPUs to improve rendering quality in a better way than regular OGSSAA. I can remember 10 years ago, GPUs only being able to hit 4xSGSSAA at 1080p30 in many games.  On current GPUs you can get 8x Sample Shading SSAA at 4k60 in a lot of older DX9 games and boy does it look good. Is it overkill? 8x maybe, 4x not at all. Specular/shading aliasing is just awful.

Blisto91:
If you have feature suggestions it's better if you make a separate issue or it might get lost.

K0bin:
> If only you could implement sort of compatibility flags heuristics to force it in games that don't support MSAA by default.

This only works for games that do support MSAA and we're not interested in adding broken heuristics and hacks for games that don't.

doitsujin:
If you want MSAA in games that don't support MSAA, make a game-specific mod that hooks the D3D9 or D3D11 library. It's not possible to implement this on our end in any useful way.

PR #3178, opened 2023-01-09
[util] remove maxFeatureLevel for EVE Online

Blisto91:
Current version of dxvk should allow for this without an override.

But should be verified anyway before merge.


Blisto91:
CC @Leopard1907

Blisto91:
Got feedback that this is not a great idea if a device only reports 11_1.

Issue #2613, opened 2022-04-15
The shadows in alien rage are broken


barfin:
Please describe your issue as accurately as possible. If you run into a problem with a binary release, make sure to test with latest `master` as well.

**Important:** When reporting an issue with a specific game or application, such as crashes or rendering issues, please include log files and, if possible, a D3D11/D3D9 Apitrace (see https://github.com/apitrace/apitrace) so that the issue can be reproduced.
In order to create a trace for **D3D11/D3D10**: Run `wine apitrace.exe trace -a dxgi YOURGAME.exe`.
In order to create a trace for **D3D9**: Follow https://github.com/doitsujin/dxvk/wiki/Making-a-Trace.
Preferably record the trace on Windows, or wined3d if possible.

**Reports with no log files will be ignored.**

### Software information
[Alien Rage - Unlimited](https://store.steampowered.com/app/217920/Alien_Rage__Unlimited/)

### System information
- GPU: GTX 970
- Driver: 510.60.02
- Wine version: 7.6 Staging
- DXVK version: dxvk-mingw-git 1.10.1.r13.g3672375b-1

### Apitrace file(s)
- Put a link here

### Log files
- d3d9.log:[ShippingPC-AFEARGame_d3d9.log](https://github.com/doitsujin/dxvk/files/8497528/ShippingPC-AFEARGame_d3d9.log)
- d3d11.log:
- dxgi.log:[ShippingPC-AFEARGame_dxgi.log](https://github.com/doitsujin/dxvk/files/8497529/ShippingPC-AFEARGame_dxgi.log)
- wine log: [alienrage.txt](https://github.com/doitsujin/dxvk/files/8497576/alienrage.txt)

![image](https://user-images.githubusercontent.com/37030489/163620837-36269780-5d5d-42b6-887f-e8682003ba4b.png)

K0bin:
Please add an apitrace so we can look into this.

barfin:
> Please add an apitrace so we can look into this.

I tried and i can't, the final size is way too big for my internet speed to upload however there is a free demo for this game in steam with very low download size, you can see for yourself

Blisto91:
Cannot reproduce in the demo on R9 380 mesa git. (proton)
Will try on Nvidia setup when i have time.

Does it appear with all ingame settings?

barfin:
> Cannot reproduce in the demo on R9 380 mesa git. (proton) Will try on Nvidia setup when i have time.
> 
[...]

PR #3183, opened 2023-01-11
[util] Add config for Alien Rage

Blisto91:
Reporting a GTX 295 puts us into the highest behind the scenes graphics settings preset bucket and prevents the shadow issue with `FloatingPointRenderTargets` set to false in the lower preset bucket. 
NvapiHack needs to be disabled on Linux since the game pings it in the Windows system folder and want's the VendorId to match your GPU. Else the issue will still happen on Nvidia Linux

This will take effect on fresh new installs as the file responsible for the above is generated on first launch.
On already installed games or prefix with left over config files one will have to delete `AFEAREngine.ini` in `C:/users/USERNAME/Documents/My Games/AlienRage/SinglePlayer/AFEARGame/Config/` so the game can generate it again.
If one instead want's to edit the file one will need [a special tool](https://www.pcgamingwiki.com/wiki/Alien_Rage_-_Unlimited#Encrypted_configuration_files) as they are encrypted.

Fixes https://github.com/doitsujin/dxvk/issues/2613

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -696,6 +696,13 @@ namespace dxvk {
     { R"(\\Secret World Legends\\ClientPatcher\.exe$)", {{
       { "d3d9.shaderModel",                 "2" },
     }} },
+    /* Alien Rage                              *
+     * GTX 295 & disable Hack to fix shadows   */
+    { R"(\\(ShippingPC-AFEARGame|ARageMP)\.exe$)", {{
+      { "d3d9.customVendorId",              "10de" },
+      { "d3d9.customDeviceId",              "05E0" },
+      { "dxgi.nvapiHack",                   "False" },
+    }} },
   }};
 
 


doitsujin:
i hate games.

doitsujin:
also is this actually a game bug that also happens on native d3d9/wined3d or just something we need to look at?

Blisto91:
It is a game bug and can also happen natively, tho rarely it seems as i didn't find alot of mentions online. The issue linked is mainly Nvidia users so the NvapiHack part is probably the primary thing here but there was also 1 amd user.

I have not been able to reproduce this on my AMD setup myself without also messing with our config options so I'm not sure how that would happen besides the user having a very old card (non vulkan) from the internal GPU list. The game seems to default to "tier 4" (if my memory doesn't fail) on my setup when i guess it doesn't recognize my card. But apparently it is a thing that can happen so telling it we are on a GTX 295 should hopefully cover our bases.

PR #3187, opened 2023-01-14
D3D11 Flush heuristic rework

doitsujin:
Aims to fix inconsistent resource readback latencies, and in general just brings flushing behaviour more in line with what a D3D11 driver would do.

The old code was really designed for a world in which interrupting a render pass was a really bad idea due to image layout transitions, but we're no longer doing those when a render pass instance is merely suspended anyway, so flushing in the middle of a draw sequence isn't a problem anymore.

This is kind of scary anyway though so any testing is highly appreciated.

### DIFF
--- a/src/d3d11/d3d11_cmdlist.cpp
+++ b/src/d3d11/d3d11_cmdlist.cpp
@@ -41,44 +41,64 @@ namespace dxvk {
   }
   
   
-  void D3D11CommandList::AddChunk(DxvkCsChunkRef&& Chunk) {
-    m_chunks.push_back(std::move(Chunk));
-  }
-  
-  
   void D3D11CommandList::AddQuery(D3D11Query* pQuery) {
     m_queries.emplace_back(pQuery);
   }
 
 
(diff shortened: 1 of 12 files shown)


ViNi-Arco:
> This is kind of scary anyway though so any testing is highly appreciated.

What's Up **Mr. Philip**.

I tested it on some games, in:
**AMD RADV 22.3.3**
**Proton 7.0-5**
**Wine-Staging 8.0rc3**

**GTA Vice City The Definitive Edition** **(** Rockstar Launcher, Wine-Staging 8.0rc3, futex_waitv **)**
**Far Cry 4** **(** Ubisoft Connect, Wine-Staging 8.0rc3, futex_waitv **)**
**Dark Souls Remastered** **(** Steam, Proton 7.0-5, futex_waitv **)**
**The Evil Within 2** **(** Steam, Proton 7.0-5, futex_waitv **)**

I tested them all for about 13~15 minutes, and didn't notice any anomalies.

PR #3193, opened 2023-01-16
Support HDR color spaces for HUD

doitsujin:
Not really tested in an actual HDR scenario.


Joshua-Ashton:
Looks fine, probably want the HUD to be brighter than 80 nits though, 203 is probably a better starting value...

PR #3216, opened 2023-01-27
 [meta] Improve How to use section in README

### LINKED ISSUE #3211: [RFE] create a script to automatically install DXVK into a wine prefix
Having to copy dlls into wine prefix and then starting `winecfg` manually and adding overrides is tedious and error-prone. Would be great if there was some script that automate that effort.

Release 2.1 [mentions at the bottom](https://github.com/doitsujin/dxvk/releases/tag/v2.1)

> setup script `setup_dxvk.sh` was no longer deemed useful and got removed

Sounds like a usecase was found! ☺

### System information
- GPU: Intel HD 620
- Driver: Mesa
- Wine version: 7.22
- DXVK version: 2.1

Riesi:
Lets improve the README a bit to help new users to find good tools to install dxvk.

I would like to not immediately merge this to allow for a bit of discussion to cover most concerns from #3211.

fixes #3211

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ In order to remove DXVK from a prefix, remove the DLLs and DLL overrides, and ru
 On Linux there exists a wide ecosystem of programs to manage your Wine game installs. These tools can also install or even auto-update DXVK. To name a few tools with different capabilities
 [Bottles](https://github.com/bottlesdevs/Bottles), 
 [Heroic](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher), 
-[Lutris](https://github.com/lutris/lutris) and [Winetricks](https://github.com/Winetricks/winetricks). There is also a simple [setup_script.sh](https://gist.github.com/doitsujin/1652e0e3382f0e0ff611e70142684d01) that might work for your use case. This is clearly not an exhaustive list nor a recommendation to use any of said tools. Use the method that you best see fit!
+[Lutris](https://github.com/lutris/lutris) and [Winetricks](https://github.com/Winetricks/winetricks). There is also a simple [setup_script.sh](https://gist.github.com/doitsujin/1652e0e3382f0e0ff611e70142684d01), purposed specifically to install/remove DXVK from a WINEPREFIX, that might work for your use case. This is clearly not an exhaustive list nor a recommendation to use any of said tools. Use the method that you best see fit!
 
 Here an example on how to install DXVK from the command line via Winetricks:
 
--- a/README.md
+++ b/README.md
@@ -9,7 +9,10 @@ The most recent development builds can be found [here](https://github.com/doitsu
 Release builds can be found [here](https://github.com/doitsujin/dxvk/releases).
 
 ## How to use
+
+### Manual install
 In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, copy or symlink the DLLs into the following directories as follows, then open `winecfg` and manually add DLL overrides for `d3d11`, `d3d10core`, `dxgi`, and `d3d9`:
+
 ```
 WINEPREFIX=/path/to/wineprefix
 cp x64/*.dll $WINEPREFIX/drive_c/windows/system32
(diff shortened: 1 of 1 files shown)


Hi-Angel:
Thanks!

You forgot to mention `setup_dxvk.sh` from older branches though, which we agreed to mention [per this comment and the following discussion](https://github.com/doitsujin/dxvk/issues/3211#issuecomment-1404705684)

Hi-Angel:
Actually, that's the main reason I posteponed writing such a PR on my own, because I have no idea what to write regarding the reasons the script was rejected. Just "maintainance burden", or more specific examples? Idk.

Riesi:
I dont think new users will care for reasons  they just want a tool that works for them and I think thats for them to figure out in the end.

Hi-Angel:
I agree! Mentioning reasons the script was removed is not for users, but rather for DXVK maintainers. In absence of such entry people will continue to create bug reports and pull requests that create such script, thus wasting time both yours and theirs.

README.md
```diff
@@ -21,6 +24,20 @@ Verify that your application uses DXVK instead of wined3d by checking for the pr
 
 In order to remove DXVK from a prefix, remove the DLLs and DLL overrides, and run `wineboot -u` to restore the original DLL files.
 
+### External tools
+
+On Linux there exists a wide ecosystem of programs to automate setting up DXVK or even let it auto-update. To name a few tools
```
Blisto91, line 29: > On Linux there exists a wide ecosystem of programs to automate setting up DXVK or even let it auto-update

Might be better to reword this a little bit so it doesn't sound like these tools are just to manage DXVK. They are really wine managers.

Riesi:
I think explaining why it was removed is a waste of space especially when people will simply not read the README in the end.

I also dont think we need to explain every tool or how it works thats for the user to figure out.

README.md
```diff
@@ -21,6 +24,20 @@ Verify that your application uses DXVK instead of wined3d by checking for the pr
 
 In order to remove DXVK from a prefix, remove the DLLs and DLL overrides, and run `wineboot -u` to restore the original DLL files.
 
+### External tools
+
+On Linux there exists a wide ecosystem of programs to manage your Wine game installs. These tools can also install or even auto-update DXVK. To name a few tools with different capabilities
+[Bottles](https://github.com/bottlesdevs/Bottles), 
+[Heroic](https://github.com/Heroic-Games-Launcher/HeroicGamesLauncher), 
+[Lutris](https://github.com/lutris/lutris) and [Winetricks](https://github.com/Winetricks/winetricks). There is also a simple [setup_script.sh](https://gist.github.com/doitsujin/1652e0e3382f0e0ff611e70142684d01) that might work for your use case. This is clearly not an exhaustive list nor a recommendation to use any of said tools. Use the method that you best see fit!
```
Hi-Angel, line 32: Nice! Tho, it's not really clear what's the difference between the 3 earlier tools and the script *(and they are clearly completely different)*. How about changing the sentence

> There is also a simple [setup_script.sh](https://gist.github.com/doitsujin/1652e0e3382f0e0ff611e70142684d01) that might work for your use case

to

> There is also a simple [setup_script.sh](https://gist.github.com/doitsujin/1652e0e3382f0e0ff611e70142684d01), purposed specifically to install/remove DXVK from a WINEPREFIX, that might work for your use case

Hi-Angel: *(I'm not a native english speaker btw, so pardon me if that may be further improved)*

Riesi: yeah :+1:

Hi-Angel:
Thanks, LGTM! If you wish, you might add a Fixes tag that refers to #3211 btw as that is basically the compromise we came to in the report.

Blisto91 (approved):

pchome:
Do not advertise third-party tools, their users already know if it can install/setup DXVK, imho.
_I guess tomorrow there will be tons of requests to add an "LaunchernameIUse" to this list._

Typical INSTALL section in README usually have several examples, in order from easy to hard, and in this case could be something like
```md
### Install DXVK from the command line via Winetricks
WINEPREFIX=/path/to/wineprefix winetricks dxvk

### Install DXVK using legacy [setup_script.sh](https://gist.github.com/doitsujin/1652e0e3382f0e0ff611e70142684d01)
WINEPREFIX=/path/to/wineprefix ./setup_script.sh

### Manual install
...
```

Hi-Angel:
> Do not advertise third-party tools, their users already know if it can install/setup DXVK, imho.

Sounds like a contradiction: if users of those tools already know how to use them to set up DXVK, why would they need to advertise the tools to them. The point of this section is make life easier for users of DXVK but not those other tools.

> _I guess tomorrow there will be tons of requests to add an "LaunchernameIUse" to this list._

Well, we could create a rule like a "only add a tool to the list, whose repository has XXX or more amount of stars". Overall, I don't see nothing bad here: people are striving to improve Linux gaming ecosystem, that's great! Just create some limitation to avoid spam of PRs from every newbie who decided to code their own script, and that should work just fine.

Hi-Angel:
Oh, the PR has been for two months here, any further plans?

PR #3225, opened 2023-01-29
[util] Age of Empires 2 DE: Relaxed barriers

sulix:
Age of Empires II DE generates an absurd number of barriers in the Co-op campaign screen (~1000/frame vs ~40/frame during gameplay and ~20/frame during other menus).

Using d3d11.relaxedBarries improves performance from ~3 FPS → ~70–80 FPS on my system.

There are no noticeable changes (performance or correctness) in other parts of the game. The same applies (on my nVidia system, at least) for the existing d3d11.cachedDynamicResources setting.


Blisto91:
Apitraces (one with coop campaign menu and one going ingame to a battle) since the devs want to take a look first.

https://mega.nz/file/10pQhIgD#HRQw8JmorAl9SRVviIgCagkrO1RysKjZx-JKjmRheW4

doitsujin:
I finally got around to looking at this.

Based on the trace, I can't confirm that enabling the option does anything at all, the problem is quite simply that the game uses 8 render passes for every single UI element on screen, until it finally composites everything into one image. It's the most inefficient way of rendering a UI that I've ever seen in my life, but it's not something that any DXVK workaround will improve.

sulix:
Hmm… there was definitely a major performance difference which relaxedBarriers fixed when I filed this bug, but I can't reproduce the poor performance locally anymore. There are still ~1000 barriers, but it's no-longer affecting performance the way it used to.

Maybe an update in the game, dxvk, or the nVidia drivers have fixed this since.

There's still noticably bad performance in those menus on lower-end Intel iGPUs, but enabling relaxedBarriers doesn't help.

So I think this is probably obsolete.

Thanks for looking into it, though!

PR #3243, opened 2023-02-09
Update README.md

srasteiro:
Edited README.md to include a short note about performance increase on Windows for AMD GPU's and lower end CPU's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 A Vulkan-based translation layer for Direct3D 9/10/11 which allows running 3D applications on Linux using Wine.
 
+Also improves performance for AMD GPU's and\or lower end CPU's on Windows.
+
 For the current status of the project, please refer to the [project wiki](https://github.com/doitsujin/dxvk/wiki).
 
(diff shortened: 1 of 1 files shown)


srasteiro:
New joiners wanting to use it may be confused and think this is only for Linux when that is not the case.

Blisto91:
> New joiners wanting to use it may be confused and think this is only for Linux when that is not the case.

It is only meant to be used for Linux as that is the only use case that this repo only really have official support for.
As the [wiki says](https://github.com/doitsujin/dxvk/wiki/Windows#common-issues) it may in general work on Windows, and it's cool when it does (and sometimes we do assist Windows users in issues), but we don't want to officially encourage it as it is outside our scope.

Both Nvidia and Intel use or will use dxvk in some capacity on Windows, but that is separate projects from what this repository offers.

K0bin:
> New joiners wanting to use it may be confused and think this is only for Linux when that is not the case.

It is only for Linux. It working on Windows is a byproduct and not officially supported.

https://github.com/doitsujin/dxvk/wiki/Windows

doitsujin:
Spelling aside, this isn't even true anymore, AMD's driver outperforms DXVK in most games on RDNA2 onwards.

PR #3248, opened 2023-02-11
[d3d9] Refactor state blocks to allocate dynamically.

Joshua-Ashton:


K0bin (commented):
Looks fine to me except for the mess with `TextureChangePrivate_`.

K0bin:
One thing my branch also does, which this one doesn't do, is only allocate the HWVP constant set if SWVP is not enabled.

I think that's absolutely worthwhile considering that most games don't use SWVP and the SWVP constant set is huge.

Joshua-Ashton:
> One thing my branch also does, which this one doesn't do, is only allocate the HWVP constant set if SWVP is not enabled.

Let's come back to this, fwiu that space is always allocated in DX and games can write up there and probably expect to read them back so we'd need to be careful.

> Looks fine to me except for the mess with TextureChangePrivate_.

I solved that by just moving it to the CPP file and doing an explicit template instantiation with:
```cpp
    template struct D3D9State<dynamic_item>;
    template struct D3D9State<static_item>;
```

PR #3270, opened 2023-03-01
[util] Reduce maximum chunk size for WILD HEARTS.

Guy1524:


src/util/config/config.cpp
```diff
@@ -721,6 +721,14 @@ namespace dxvk {
     { R"(\\bf10\.exe$)", {{
       { "d3d9.maxFrameRate",                "60" },
     }} },
+    /* WILD HEARTS™️                            *
+     * D3D12 title using D3D11 device for      *
+     * media texture creation, whereby a large *
+     * chunk size only slows down media        *
+     * initialization                          */
+    { R"(\\WILD HEARTS\.exe$)", {{
```
Blisto91, line 729: Might as well also do the same for the trail exe

```suggestion
    { R"(\\WILD HEARTS(_Trial)?\.exe$)", {{
```

Blisto91 (approved):

PR #3276, opened 2023-03-03
Update README.md

locriacyber:


K0bin:
This is wrong.

SysWow64 is short for "System Windows [32] on Windows 64", so that's where the 32 bit DLLs go. 64 bit DLLs go to System32 (presumably because of backwards compatibility reasons).

locriacyber:
wait really

Blisto91:
Yes 🙂

PR #3278, opened 2023-03-03
Rework available query processing

doitsujin:
Supersedes #3273.

@CFSworks please let me know if this looks alright to you. I made the proposed changes yesterday and put you as a co-author since it's the same idea.

### DIFF
--- a/src/dxvk/dxvk_gpu_query.cpp
+++ b/src/dxvk/dxvk_gpu_query.cpp
@@ -28,7 +28,7 @@ namespace dxvk {
   }
 
 
-  DxvkGpuQueryStatus DxvkGpuQuery::getData(DxvkQueryData& queryData) const {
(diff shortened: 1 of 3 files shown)


CFSworks (commented):
This approach is a lot cleaner than mine, and thank you for the co-author tag!

I have had a very busy day today so I'll have to test this change sometime this weekend, but it does look like it does the job quite nicely.

My only code comment on this PR is related to whether `m_ended` really needs to be atomic after all: It may even be good to do away with it, or use it only in debug builds for an assert. But either way, I don't feel very strongly about it. Do as you like.

CFSworks:
Just tested and confirmed that this PR does indeed fix the problem in Halo: MCC (both tests done on the pause menu):
- Commit `11f76306`: memory usage stays fairly constant
- Reverting to commit `c5b7a669`: At 60 FPS, RSS increases at ~123 MiB/min. with query allocation at ~36 pools/min.

PR #3282, opened 2023-03-09
[util] Add config for Tom Clancy's The Division.

giomasce:


mbriar:
Why does it need to be limited to 60 fps? E.g. [PCGamingWiki](https://www.pcgamingwiki.com/wiki/Tom_Clancy%27s_The_Division) doesn't mention any issues with higher frame rates.

nutta-git:
I play this game and haven't noticed any massive performance issues above 60 fps. I play at 200+ frames.

giomasce:
Apologies for the confusion, the frame limitation was indeed a mistake on my part. However, the `syncInterval` setting really fixes some stuttering problems, at least on my setup, so I opened #3420 hoping that it looks better justified.

PR #3288, opened 2023-03-14
[d3d9] remove support for the A1/X1R5G5B5 formats

### LINKED ISSUE #1545: [d3d9] Beyond Good & Evil
Only remaining issue is the Settings Application crashing when running through dxvk, galliumNine, or dgVoodoo.
Not sure if this is really a dxvk issue or a wine interaction issue though

~~With dxvk the game stays in a really tiny window with 320x200. This does not happen with wined3d, which renders the game at full 1080p. Apitraces with dxvk and wined3d attached.~~
^This has been fixed

### Software information
Beyond Good & Evil CD Version

### System information
- GPU: GTX970
- Driver: 440.66.07
- Wine version: 5.5
- DXVK version: 904d3e6c9014bff7aa0843630b009c51879a1d06

### Apitrace file(s)
[BGE_dxvk.trace.zip](https://github.com/doitsujin/dxvk/files/4423266/BGE_dxvk.trace.zip)

[BGE_wined3d.trace.tar.gz](https://github.com/doitsujin/dxvk/files/4423264/BGE_wined3d.trace.tar.gz)

### Log files
- d3d9.log: [BGE_d3d9.log](https://github.com/doitsujin/dxvk/files/4423278/BGE_d3d9.log)

WinterSnowfall:
Fixes #1545 , namely the SettingsApplication crash. I don't think there are any other pending issues with BGE, and even if there were, they're probably best tracked separately by now.

Regarding the formats, @Riesi has done some testing on Windows and neither Intel nor Nvidia list any adapter modes for D3DFMT_X1R5G5B5. I've only commented them out since they're technically part of the spec (something we should still be aware of), but can also outright remove them if that's preferred.

X1R5G5B5 is a 16-bit format, so I don't expect it would see a lot of use anyway (hopefully), but more testing is welcome.

### DIFF
--- a/src/d3d9/d3d9_monitor.cpp
+++ b/src/d3d9/d3d9_monitor.cpp
@@ -27,9 +27,10 @@ namespace dxvk {
 
   bool IsSupportedAdapterFormat(
           D3D9Format Format) {
+    // D3D9Format::X1R5G5B5 is unsupported by native drivers and some apps, 
+    // such as the BGE SettingsApplication, rely on it not being exposed.
     return Format == D3D9Format::A2R10G10B10
         || Format == D3D9Format::X8R8G8B8
-        || Format == D3D9Format::X1R5G5B5
         || Format == D3D9Format::R5G6B5;
   }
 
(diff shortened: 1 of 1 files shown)


doitsujin:
Probably better to just remove these and derp a comment somewhere why we don't expose X1R5G5B5.

PR #3301, opened 2023-03-22
[util] Enable strict float emulation for NFS Carbon

TheComputerGuy96:
This fixes a random segfault with NVIDIA drivers (not sure if NVK is affected though because I didn't test it with NVK yet 😄)

I've been meaning to push this for a while but I didn't do it until now 🐸

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -739,6 +739,10 @@ namespace dxvk {
     { R"(\\(PANZERS|PANZERS_Phase_2)\.exe$)", {{
       { "d3d9.enableDialogMode",         "True"   },
     }} },
+    /* Need for Speed Carbon                   */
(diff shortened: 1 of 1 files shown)


doitsujin:
If we segfault then we should figure out *why* we're segfaulting rather than just put a band-aid on it that doesn't actually fix the root cause.

Do you have a trace?

TheComputerGuy96:
> Do you have a trace?
I should really make one once I get out of nouveau driver 🐸

Joshua-Ashton:
Are you sure it's a segfault and not a fp exception? Did you actually see it was SIGSEGV?

TheComputerGuy96:
Let's double-check if I didn't mis-remember that it was segfault once I leave nouveau

K0bin:
Nvidia is looking into the compiler crash. So hopefully we won't need this workaround.

PR #3307, opened 2023-03-23
Add few defines and types in windows.h for DXVK Native

q4a:
Add few defines and types in windows.h. It will be useful projects that will build native Linux version of DXVK


include/native/windows/windows_base.h
```diff
@@ -35,7 +38,7 @@ typedef uint16_t UINT16;
 typedef uint32_t UINT32;
 typedef uint64_t UINT64;
 typedef void VOID;
-typedef void* LPVOID;
+typedef void* PVOID, LPVOID;
```
turol, line 41: This is wrong, it changes `LPVOID` to non-pointer.

q4a: Sorry, my bad. Fixed.

q4a:
Anyone can check and merge it?

PR #3310, opened 2023-03-24
Update dxvk.conf

VincentProuchet:
Added hud config valuesalong with their definition the goal is to eliminate the necesity to rely on project readme for configuration

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -339,6 +339,24 @@
 # environment variable is not set, otherwise it will be
 # ignored. The syntax is identical.
 
+# a valid value is : devinfo,fps,frametimes,pipelines,memory,version,api,compiler,
+
+#  here the different elements you can use 
+#    devinfo	: Displays the name of the GPU and the driver version.
+#    fps		: Shows the current frame rate.
+#    frametimes	: Shows a frame time graph.
+#    submissions: Shows the number of command buffers submitted per frame.
+#    drawcalls	: Shows the number of draw calls and render passes per frame.
+#    pipelines	: Shows the total number of graphics and compute pipelines.
+#    descriptors: Shows the number of descriptor pools and descriptor sets.
+#    memory		: Shows the amount of device memory allocated and used.
+#    gpuload	: Shows estimated GPU load. May be inaccurate.
+#    version	: Shows DXVK version.
+#    api		: Shows the D3D feature level used by the application.
+#    cs			: Shows worker thread statistics.
+#    compiler	: Shows shader compiler activity
+#    samplers	: Shows the current number of sampler pairs used [D3D9 Only]
+#    scale=x	: Scales the HUD by a factor of x (e.g. 1.5)
 # dxvk.hud = 
 
 


dxvk.conf
```diff
@@ -339,6 +339,24 @@
 # environment variable is not set, otherwise it will be
 # ignored. The syntax is identical.
```
Blisto91, line 340: > The syntax is identical.

This still hints you to look at the readme, so i'd just note here that the values are comma separated.

dxvk.conf
```diff
@@ -339,6 +339,24 @@
...
 
+# a valid value is : devinfo,fps,frametimes,pipelines,memory,version,api,compiler,
```
Blisto91, line 342: If the above feedback is addressed then this is probably redundant and can be left out.

dxvk.conf
```diff
@@ -339,6 +339,24 @@
...
+
+#  here the different elements you can use
```
Blisto91, line 344: Capitalize the first letter

To make it consistent with the readme i'd probably change the wording to "options"
e.g "The following options are supported" or similar

dxvk.conf
```diff
@@ -339,6 +339,24 @@
...
+#    pipelines	: Shows the total number of graphics and compute pipelines.
+#    descriptors: Shows the number of descriptor pools and descriptor sets.
+#    memory		: Shows the amount of device memory allocated and used.
+#    gpuload	: Shows estimated GPU load. May be inaccurate.
+#    version	: Shows DXVK version.
+#    api		: Shows the D3D feature level used by the application.
+#    cs			: Shows worker thread statistics.
+#    compiler	: Shows shader compiler activity
+#    samplers	: Shows the current number of sampler pairs used [D3D9 Only]
+#    scale=x	: Scales the HUD by a factor of x (e.g. 1.5)
```
Blisto91, line 359:
```suggestion
# - devinfo: Displays the name of the GPU and the driver version.
# - fps: Shows the current frame rate.
# - frametimes: Shows a frame time graph.
# - submissions: Shows the number of command buffers submitted per frame.
# - drawcalls: Shows the number of draw calls and render passes per frame.
# - pipelines: Shows the total number of graphics and compute pipelines.
# - descriptors: Shows the number of descriptor pools and descriptor sets.
# - memory: Shows the amount of device memory allocated and used.
# - gpuload: Shows estimated GPU load. May be inaccurate.
# - version: Shows DXVK version.
# - api: Shows the D3D feature level used by the application.
# - cs: Shows worker thread statistics.
# - compiler: Shows shader compiler activity
# - samplers: Shows the current number of sampler pairs used [D3D9 Only]
# - scale=x: Scales the HUD by a factor of x (e.g. 1.5)
```

dxvk.conf
```diff
@@ -339,6 +339,24 @@
...
 # dxvk.hud =
```
Blisto91, line 360: Add a empty line before this

Blisto91 (commented):
Hello and thank you for the PR.
I've noted some styling nits

Blisto91:
Also try to make the commit name a bit informing so it can be seen just from your name what the commit does.

The common style would be something like
`[meta] Add dxvk.hud options to example config`

dxvk.conf
VincentProuchet: I just wanted to avoid opening the config file and the github readme
I'm lazy

dxvk.conf
VincentProuchet: A sample to guide out low level like me

dxvk.conf
VincentProuchet: I will integrate the presentation to the next commit 
do you want another pull request ? with your requested modification ?

VincentProuchet:
thanks for the code reviews and all the hints

PR #3315, opened 2023-03-27
[util] Add new "Zusi 3" exe filename

### LINKED ISSUE #3250: Zusi 3 (1040730) has broken cab images
Rendering of cabs in the simulator is broken on DXVK (works in WineD3D) .

### Software information
Zusi 3 (1040730) - defaults

### System information
- GPU: Nvidia 1660ti
- Driver: 528.85
- Wine version: 7 and 8 (Proton 7 and 8-alpha)
- DXVK version: 2.1

### Apitrace file(s)
https://drive.google.com/file/d/1aApV8a9l5xFtqHEe8EyGdkAJvLOrxDhN/view?usp=share_link

### Log files
- d3d9.log: [ZusiSim.64_d3d9.log](https://github.com/doitsujin/dxvk/files/10716697/ZusiSim.64_d3d9.log)

w-flo:
Hopefully the regex is correct. It works for ZusiSim.64.exe (which is the executable launched by steam), but I'm not sure how to launch the legacy 32 bit "ZusiSim.exe" in steam, so I have not tested if that still works correctly.

Commit message:
---
In addition to the legacy "ZusiSim.exe" executable, the recent 3.5 release added a 64-bit "ZusiSim.64.exe" executable. This is now the default executable when launching the game in Steam. The game is unplayable without this option, so update the regex for Zusi 3.

Fixes #3250.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -457,7 +457,7 @@ namespace dxvk {
       { "d3d9.allowDiscard",                "False" },
     }} },
     /* ZUSI 3 - Aerosoft Edition                  */
-    { R"(\\ZusiSim\.exe$)", {{
+    { R"(\\ZusiSim(\.64)?\.exe$)", {{
       { "d3d9.noExplicitFrontBuffer",       "True" },
     }} },
     /* GTA IV (NVAPI)                             */


w-flo:
Updated commit message to add "Fixes #3250.".

Also, I have now verified that the regex still works for the 32-bit exe filename "ZusiSim.exe" after updating 32 bit dxvk dlls in the compatdata wine prefix (launching it from the command line using Proton's wine executable and the compatdata wine prefix simply works as expected, I thought there would be some DRM/Steam issues).

w-flo:
That's a great question!

I've never used these editors and don't really know how to use them (they seem complicated), but giving them a very superficial test:
- ZusiFtdEditor.64.exe: This one appears broken with dxvk, with and without "noExplicitFrontBuffer". Instead of the cab view image, there is just a gray area with the cab view flickering for ~1 frame when clicking the gray area. This is not an issue when using wine d3d9 apparently.
- Zusi3DEditor.64.exe: I haven't noticed any obvious issues when not using the config.
- ZusiFzgEditor.64.exe, ZusiDateiverwaltung.exe: It seems like these don't use D3D
- ZusiGleisplanEditor.exe: Not sure how to launch this one (it attempts to launch using my wine-mono installation, even though it should probably use MS dotnet from the game's Proton wineprefix instead)

Blisto91:
Thanks for the extra testing. I have asked besentv if he has any input on the last exe.

w-flo:
> Edit: Would you be able to make a issue for the `ZusiFtdEditor.64.exe` problem?

Sure, I'll try to find out what is going on, and then open an issue (if I can't find an existing issue for this problem).

Thanks for the review :-)

PR #3316, opened 2023-03-27
[d3d9] Implement rudimentary device loss

### LINKED ISSUE #3314: DC Universe Online alt tab render freeze
When you alt tab out and in in DC Universe Online with dxvk the renderer will freeze (or jiggle back and forth between last couple of frames) with the game still going on behind. This doesnt happen with wined3d but does happen when using dxvk on Windows.

### Software information
DC Universe Online (fullscreen, haven't tried windowed yet)

### System information
- GPU: 7900xtx
- Driver: mesa-git
- Wine version: Proton bleeding edge
- DXVK version: Every version that can launch the game.

### Apitrace file(s)
Two traces, both made on Windows. One with dxvk and one with native.
When replaying both afterwards on Windows without dxvk it can be seen that the games renderer freezes when alt tabbed out of focus (intended i guess,) but in the one that was recorded with dxvk it never resumes the rendering after going in again, unlike native recorded version.
https://mega.nz/file/4oAV3RpJ#ddYibN84ACeM-j753TLRTtfYCvDClllY8FV4972z7f8

### Log files
[DCGAME_d3d9.log](https://github.com/doitsujin/dxvk/files/11079882/DCGAME_d3d9.log)

K0bin:
I hate that we have to do this but:

Fixes #3314

It's hidden behind a config option because pretty much all games work fine without it and enabling it would just cause unnecessary work for those games.

Aside from DC Online, I enabled it for GTA IV because that game also has a ton of issues with alt-tab. Maybe we'll get lucky and these changes finally fix it.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -211,8 +211,16 @@ namespace dxvk {
 
 
   HRESULT STDMETHODCALLTYPE D3D9DeviceEx::TestCooperativeLevel() {
+    D3D9DeviceLock lock = LockDevice();
+
     // Equivelant of D3D11/DXGI present tests. We can always present.
-    return D3D_OK;
+    if (likely(m_deviceLostState == D3D9DeviceLostState::Ok)) {
+      return D3D_OK;
+    } else if (m_deviceLostState == D3D9DeviceLostState::NotReset) {
+      return D3DERR_DEVICENOTRESET;
+    } else {
+      return D3DERR_DEVICELOST;
+    }
   }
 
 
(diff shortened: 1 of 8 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -382,13 +390,19 @@ namespace dxvk {
...
+    m_deviceLostState = D3D9DeviceLostState::Ok;
+
     HRESULT hr = ResetSwapChain(pPresentationParameters, nullptr);
     if (FAILED(hr))
       return hr;
 
     hr = ResetState(pPresentationParameters);
-    if (FAILED(hr))
+    if (FAILED(hr)) {
+      Logger::warn("Reset failed: Device lost");
```
Joshua-Ashton, line 402: ResetState always returns D3D_OK, I'd juts make it `void` at this point.

Joshua-Ashton (changes requested):
Minor nit.

Not as bad as I thought it'd be...

Probably worth also hard disabling this for 9Ex devices as we may have games with a launcher that mix and match here or whatever.

PR #3317, opened 2023-03-27
[d3d9] add a A2R10G10B10 adapter format workaround for AquaNox 2

### LINKED ISSUE #3255: AquaNox 2 crash on resolution change

Screenshot_20230215_201542

<details>

Screenshot_20230215_200657
</details>

### Software information
AquaNox 2

### System information
- GPU: 7900xtx
- Driver: mesa-git
- DXVK version: master

### Apitrace file(s)

### Log files

WinterSnowfall:
Gets the game going on ANV and it should also now work with Wayland according to @Blisto91. The crashing that's observed on resolution change is only partially resolved, and seems to happen with WineD3D as well in my case, so not entirely sure what the problem is or how to fix it. The game's probably just very picky when it comes to supporting certain resolutions. I'll see if I can do more, but it will come in a separate PR.

The background of this PR is that A2R10G10B10 is supported at least by AMD on Windows, therefore there's no reason to outright remove it, so I've only included it as a config option and disabled it specifically for AquaNox 2.

Please do let me know if there's a better way of getting the option struct where I need it to be other than passing it around like a dog chew toy. As you may know, my C++-fu is mostly white belt level really :sweat_smile:. I'm also open to any naming changes if I'm braking any conventions I'm not aware of.

Fixes #3255. The game still hates ANV at times, but should be fine otherwise.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -475,6 +475,15 @@
 
 # d3d9.noExplicitFrontBuffer = False
 
+# Support the A2R10G10B10 adapter format
+#
+# Usually supported by native drivers, but may cause issues with some games
+#
+# Supported values:
+# - True/False
+
+# d3d9.supportAFA2R10G10B10 = True
+
 # Support DF formats
 #
 # Support the vendor extension DF floating point depth formats
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -98,7 +98,7 @@ namespace dxvk {
           D3D9Format AdapterFormat,
           D3D9Format BackBufferFormat,
           BOOL       bWindowed) {
-    if (!IsSupportedAdapterFormat(AdapterFormat))
+    if (!IsSupportedAdapterFormat(AdapterFormat, m_parent->GetOptions()))
       return D3DERR_NOTAVAILABLE;
 
     if (!IsSupportedBackBufferFormat(AdapterFormat, BackBufferFormat, bWindowed))
@@ -114,7 +114,7 @@ namespace dxvk {
           DWORD           Usage,
           D3DRESOURCETYPE RType,
           D3D9Format      CheckFormat) {
-    if (!IsSupportedAdapterFormat(AdapterFormat))
+    if (!IsSupportedAdapterFormat(AdapterFormat, m_parent->GetOptions()))
       return D3DERR_NOTAVAILABLE;
 
     const bool dmap = Usage & D3DUSAGE_DMAP;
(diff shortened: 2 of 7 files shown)


pchome:
Instead of redefining `IsSupported*Format()` maybe better add additional conditions inline, with comments. Something like 
```
bool sourceSupported = SourceFormat != D3D9Format::Unknown
                        && IsSupportedBackBufferFormat(SourceFormat)
                        && (SourceFormat == D3D9Format::A2R10G10B10 ? m_parent->GetOptions().supportAFA2R10G10B10 : true);
```
and early returns before `if (!IsSupportedAdapterFormat(AdapterFormat))` checks,
and maybe separate log message where `"D3D9DeviceEx::ResetSwapChain: Unsupported backbuffer format: "` is (supported but disabled).

This will require less changes, but can lead to bugs in future if new code will rely only on IsSupported*Format() functions somewhere, so ...

WinterSnowfall:
> separate log message where `"D3D9DeviceEx::ResetSwapChain: Unsupported backbuffer format: "` is (supported but disabled).

I don't think it will ever be shown with AquaNox 2 at least, since the game won't use the format at all if it doesn't detect any modes for it. That is the point of the PR, but I've disabled it globally, for good measure. And for all intents and purposes the config option makes the format unsupported, so that logger is technically correct (the best kind of correct :P).

> This will require less changes, but can lead to bugs in future if new code will rely only on IsSupported*Format() functions somewhere, so ...

Appreciate the suggestion, but indeed it looks more convoluted to me. It's probably better to keep all the AF checks in their consecrated spots (who knows what other format hacks we may need to do in the future...).

K0bin:
Why don't we just remove support for that format altogether instead of adding yet another config option?

WinterSnowfall:
Because it seems to be supported on native and it's only this game in particular that takes issues with it. I'm not shy of outright removing formats, but there has to be a good reason for doing that globally IMHO.

Joshua-Ashton:
@K0bin Everyone exposes this format, and a bunch of AMD demos even rely on it.

K0bin:
Okay, we should clearly keep it then.

WinterSnowfall:
I've simplified things a bit and dropped the checks against the back-buffer because they weren't really needed. The game will happily ignore the format altogether is the adapter format is unsupported and we don't report any display modes for it.

I've found no other way to make the game work on ANV unfortunately, and it may very well be ANV/Wayland take issues with this format for some reason, since Nvidia doesn't seem to care. This isn't a problem with WineD3D, because WineD3D globally disables A2R10G10B10 for some reason (which is dumb, if you ask me, but no actual games seem to care, apparently, since it gets away with it).

WinterSnowfall:
Back to the drawing board, since this isn't A2R10G10B10 related after all - the game actually has an issue with the total number of reported display modes. If that's too high all manner of jankiness ensues. Not reporting A2R10G10B10 got the mode count below the critical threshold, I guess, but it's not a problem in itself. I'll PR something else once I come up with a better solution.

Issue #3321, opened 2023-03-30
Native builds with SDL and GLFW are incompatible, so should have different SONAMEs


smcv:
As far as I can tell, each of the available WSIs for DXVK Native has a different ABI: in the SDL2 WSI, whenever a function in the DirectX API returns or receives a `HWND` it's really a `SDL_Window *`, but in the GLFW WSI, whenever a function returns or receives a `HWND` it's really a `GLFWwindow *`.

When SDL 3 becomes stable, presumably there will be SDL3 games that want to use DXVK Native, at which point that would become a third incompatible WSI?

If that's the case, then they should have different SONAMEs. One option would be to append the WSI to the `dxvk_name_prefix`, making them `libdxvk_sdl2_d3d9.so`, `libdxvk_glfw_d3d9.so` and so on. `libdxvk_d3d9.so` could be a symlink to `libdxvk_sdl2_d3d9.so`, and so on, for backwards compatibility.

Or, the default `sdl2` WSI could use an unadorned name because it got there first, and all other WSIs (including SDL3, GLFW, and any other WSIs that are invented in future) could append their name to the `dxvk_name_prefix`?

smcv:
cc @Joshua-Ashton

Joshua-Ashton:
I think the true game plan to solve this is simply to add an interface that lets the dev pick the WSI at runtime.

In the meantime though, I am happy to append the other WSI such as GLFW to the name prefix.
We don't provide builds with the glfw WSI in releases right now either fwiw.

smcv:
> In the meantime though, I am happy to append the other WSI such as GLFW to the name prefix

OK, I'll do a PR to make it `libdxvk_d3d9.so` (implicitly SDL2) and `libdxvk_glfw_d3d9.so`.

doitsujin:
yeah i can live with that as a temporary solution, but as josh said, we should rework this at some point.

Beyley:
> I think the true game plan to solve this is simply to add an interface that lets the dev pick the WSI at runtime.
> 
> In the meantime though, I am happy to append the other WSI such as GLFW to the name prefix. We don't provide builds with the glfw WSI in releases right now either fwiw.

Different so names is something we do downstream at Silk.NET to handle this problem and it works great, but when the runtime WSI selection does happen, could you drop me a ping if you remember so i can update our native lib handling?

PR #3324, opened 2023-03-31
build: Use a different SONAME for the glfw WSI

smcv:
Selecting the GLFW WSI gives DXVK an incompatible ABI: for example, CreateDevice takes a `SDL_Window *` parameter in the default SDL WSI, but takes a `GLFWwindow *` instead in the GLFW WSI. This means it should get a different name, so that binaries expecting one WSI don't unexpectedly load the other.

Resolves: https://github.com/doitsujin/dxvk/issues/3321

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -121,8 +121,12 @@ else
     lib_glfw = cpp.find_library('glfw')
     compiler_args += ['-DDXVK_WSI_GLFW']
   endif
-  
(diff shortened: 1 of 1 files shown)


expipiplus1:
This will interact with https://github.com/doitsujin/dxvk/pull/3365

flibitijibibo:
This got obsoleted by #3738

PR #3339, opened 2023-04-07
[dxgi] Leave current fullscreen mode when SetFullscreenState changes output

rbernon:
This fixes RiME not moving its window when output display is changed in the game settings.

### DIFF
--- a/src/dxgi/dxgi_swapchain.cpp
+++ b/src/dxgi/dxgi_swapchain.cpp
@@ -425,8 +425,18 @@ namespace dxvk {
 
     Com<IDXGIOutput1> target;
 
-    if (pTarget)
+    if (pTarget) {
+      DXGI_OUTPUT_DESC desc;
+
       pTarget->QueryInterface(IID_PPV_ARGS(&target));
+      target->GetDesc(&desc);
+
+      if (!m_descFs.Windowed && Fullscreen && m_monitor != desc.Monitor) {
+        HRESULT hr = this->LeaveFullscreenMode();
+        if (FAILED(hr))
+          return hr;
+      }
+    }
 
     if (m_descFs.Windowed && Fullscreen)
       return this->EnterFullscreenMode(target.ptr());


src/dxgi/dxgi_swapchain.cpp
```diff
@@ -425,8 +425,17 @@ namespace dxvk {
...
+    if (pTarget) {
+      DXGI_OUTPUT_DESC desc;
+
+      if (FAILED(pTarget->QueryInterface(IID_PPV_ARGS(&target))))
+        desc.Monitor = m_monitor;
+      else
+        target->GetDesc(&desc);
+
+      if (!m_descFs.Windowed && m_monitor != desc.Monitor && FAILED(this->LeaveFullscreenMode()))
+        Logger::err("DXGI: SetFullscreenState: Failed to leave previous fullscreen mode");
```
doitsujin, line 437: `LeaveFullscreenMode` already logs errors on its own and we don't do this down below, maybe this can be removed?

Also it's probably better if we got rid of the second `LeaveFullscreenMode` call as well, I know that this currently works but the only reason why it works is because `LeaveFullscreenMode` modifies global state, which makes this more than a little bit confusing.

src/dxgi/dxgi_swapchain.cpp
```diff
@@ -425,8 +425,17 @@ namespace dxvk {
 
     Com<IDXGIOutput1> target;
 
-    if (pTarget)
-      pTarget->QueryInterface(IID_PPV_ARGS(&target));
+    if (pTarget) {
+      DXGI_OUTPUT_DESC desc;
+
+      if (FAILED(pTarget->QueryInterface(IID_PPV_ARGS(&target))))
```
doitsujin, line 431: When is this ever expected to fail? We should probably just error out if the app passes an invalid pointer here.

doitsujin (changes requested):

src/dxgi/dxgi_swapchain.cpp
rbernon: No idea, I was just being defensive.

src/dxgi/dxgi_swapchain.cpp
rbernon: Something like that?

```cpp
    if (pTarget) {
      DXGI_OUTPUT_DESC desc;

      pTarget->QueryInterface(IID_PPV_ARGS(&target));
      target->GetDesc(&desc);

      if (!m_descFs.Windowed && Fullscreen && m_monitor != desc.Monitor) {
        this->LeaveFullscreenMode();
        return this->EnterFullscreenMode(target.ptr());
      }
    }

    if (m_descFs.Windowed && Fullscreen)
      return this->EnterFullscreenMode(target.ptr());
    else if (!m_descFs.Windowed && !Fullscreen)
      return this->LeaveFullscreenMode();
```

doitsujin: On second thought, maybe just keep it as-is, just get rid of the error log and return the HRESULT from `LeaveFullscreenMode` properly if that fails for whatever reason.

PR #3347, opened 2023-04-12
simplify parsing of the configuration options

oltolm:
This trivial patch simplifies option parsing by using `std::getline`.


pchome:
imho:
1. amend commit message to `[hud] Simplify parsing of the configuration options`
2. variable name `eq` is a bit misleading, because `=` character here is a parameter delimiter.
  Something like `if (hasParam) {m_options.insert({...});} else {m_enabled.insert(...);}` would improve readability.

doitsujin:
does this solve any problem in practice other than us now potentially having to worry about locale issues?

oltolm:
No feel free to close.

PR #3349, opened 2023-04-13
[d3d9] Fall back to GDI blit for partial presents

### LINKED ISSUE #2749: D3D9 Partial Presentation Mega-issue
Made to consolidate the other issues which concern the lack of partial presentation in dxvk.

This issue most commonly presents itself as the application being mostly black except for any element that changes look or redraws when hovered over with the mouse. Like a button that has a highlight when the mouse cursor is on it

<details>
  <summary>Example screenshots</summary>

Return of Reckoning launcher  

Screenshot_20220723_193300
Screenshot_20220723_193614
</details>

It is most common in various game launchers and wpf apps. A few games might use it too.
It can usually be worked around in dxvk by setting `d3d9.shaderModel = 1` which have been added in the past into dxvk for a few launchers.
If set in a local config usually you'll only want this to affect the launcher itself. Which can be done like below for a specific exe in `dxvk.conf`
```
[RoRLauncher.exe]
d3d9.shaderModel = 1
```

Tho setting `d3d9.shaderModel = 1` is not always a desirable solution since in some cases the app might become very sluggish if it falls back on software rendering or if it's unpacked into a temporary folder with a generic name where a config option can't reliably catch it.
Setting it for every app thrown this way is probably also not a welcome solution.

Concerned issues:
https://github.com/doitsujin/dxvk/issues/2732
https://github.com/doitsujin/dxvk/issues/2240
https://github.com/doitsujin/dxvk/issues/2568 (Part of issue)
https://github.com/doitsujin/dxvk/issues/2278
https://github.com/doitsujin/dxvk/issues/1476
https://github.com/doitsujin/dxvk/issues/2334
https://github.com/doitsujin/dxvk/issues/1481 (Maybe?)
https://github.com/doitsujin/dxvk/issues/2456
https://github.com/doitsujin/dxvk/issues/2542
https://github.com/doitsujin/dxvk/issues/2860 (Likely)
https://github.com/doitsujin/dxvk/issues/3005
https://github.com/doitsujin/dxvk/issues/3116 (Likely)
https://github.com/doitsujin/dxvk/issues/3208

K0bin:
Hopefully fixes #2749.

The idea is from Wine. When the present effect is COPY and the dst rect doesn't cover the entire screen, we read back the contents of the backbuffer and copy it to the screen using GDI.

It's slow but hopefully this only impacts launchers and weird japanese visual novels, where this isn't a problem. Slow is also much better than a black window.

The HUD is missing because I grab the contents from the D3D9Surface (because that already has the `GetDC` function). It's obviously possible to get the contents off of the DxvkImage after the HUD is rendered but the code is quite neat and tidy the way it is right now and I don't think it's a big deal.

It seems to work for DNSSpy (WPF) at least:
Before:
image
After:
image

ESO Launcher:
Before:
image

After:
image

### DIFF
--- a/src/d3d9/d3d9_surface.cpp
+++ b/src/d3d9/d3d9_surface.cpp
@@ -165,7 +165,9 @@ namespace dxvk {
     createInfo.hBitmap     = nullptr;
     createInfo.hDc         = nullptr;
 
-    D3DKMTCreateDCFromMemory(&createInfo);
+    if (D3DKMTCreateDCFromMemory(&createInfo))
+      Logger::err("D3D9: Failed to create GDI DC");
+
     DeleteDC(createInfo.hDeviceDc);
 
     // These should now be set...
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -139,6 +139,12 @@ namespace dxvk {
 
     m_lastDialog = m_dialog;
 
+#ifdef _WIN32
+    const bool useGDIFallback = m_partialCopy && m_presentParams.SwapEffect == D3DSWAPEFFECT_COPY;
+    if (useGDIFallback)
+      return BlitGDI(window);
+#endif
+
     try {
       if (recreate)
         CreatePresenter();
@@ -156,10 +162,47 @@ namespace dxvk {
       return D3D_OK;
     } catch (const DxvkError& e) {
       Logger::err(e.message());
+#ifdef _WIN32
+      return BlitGDI(window);
+#else
       return D3DERR_DEVICEREMOVED;
+#endif
     }
   }
 
+#ifdef _WIN32
+  HRESULT D3D9SwapChainEx::BlitGDI(HWND Window) {
+    if (!std::exchange(m_warnedAboutFallback, true))
+      Logger::warn("Using GDI for swapchain presentation. This will impact performance.");
+
+    HDC hDC;
+    HRESULT result = m_backBuffers[0]->GetDC(&hDC);
+    if (result) {
+      Logger::err("D3D9SwapChainEx::BlitGDI Surface GetDC failed");
+      return D3DERR_DEVICEREMOVED;
+    }
+
+    HDC dstDC = GetDCEx(Window, 0, DCX_CACHE);
+    if (!dstDC) {
+      Logger::err("D3D9SwapChainEx::BlitGDI: GetDCEx failed");
+      m_backBuffers[0]->ReleaseDC(hDC);
+      return D3DERR_DEVICEREMOVED;
+    }
+
+    bool success = StretchBlt(dstDC, m_dstRect.left, m_dstRect.top, m_dstRect.right - m_dstRect.left,
+            m_dstRect.bottom - m_dstRect.top, hDC, m_srcRect.left, m_srcRect.top,
+            m_srcRect.right - m_srcRect.left, m_srcRect.bottom - m_srcRect.top, SRCCOPY);
+
+    if (!success) {
+      Logger::err("D3D9SwapChainEx::BlitGDI: StretchBlt failed");
+      m_backBuffers[0]->ReleaseDC(hDC);
+      return D3DERR_DEVICEREMOVED;
+    }
+
+    m_backBuffers[0]->ReleaseDC(hDC);
+    return S_OK;
+  }
+#endif
 
   HRESULT STDMETHODCALLTYPE D3D9SwapChainEx::GetFrontBufferData(IDirect3DSurface9* pDestSurface) {
     D3D9DeviceLock lock = m_parent->LockDevice();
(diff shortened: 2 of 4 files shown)


K0bin:
@Joshua-Ashton Do we just `#ifdef WIN32` this to fix the native build? Doesn't make a lot of sense to support this in DXVK Native IMO.

zfigura:
How was this written?

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -156,10 +162,44 @@ namespace dxvk {
...
+      m_backBuffers[0]->ReleaseDC(hDC);
+      return D3DERR_DEVICEREMOVED;
+    }
+
+    bool success = StretchBlt(dstDC, m_dstRect.left, m_dstRect.top, m_dstRect.right - m_dstRect.left,
+            m_dstRect.bottom - m_dstRect.top, hDC, m_srcRect.left, m_srcRect.top,
+            m_srcRect.right - m_srcRect.left, m_srcRect.bottom - m_srcRect.top, SRCCOPY);
+
+    if (!success) {
+      Logger::err(str::format("D3D9SwapChainEx::BlitGDI: StretchBlt failed"));
```
jsimmons, line 195: unnecessary `str::format` I think?

K0bin:
@zfigura Like I said, I got the idea from the Wine code. I did get the GetDCEx call and the idea to use StretchBlt from there. I'll leave it to you whether that's already problematic when it comes to licenses or not.

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -156,10 +162,47 @@ namespace dxvk {
       return D3D_OK;
     } catch (const DxvkError& e) {
       Logger::err(e.message());
-      return D3DERR_DEVICEREMOVED;
+#ifdef WIN32
```
jsimmons, line 165: missing return value here if it's compiled out and throws I believe?

src/d3d9/d3d9_surface.cpp
```diff
@@ -165,7 +165,9 @@ namespace dxvk {
     createInfo.hBitmap     = nullptr;
     createInfo.hDc         = nullptr;
 
-    D3DKMTCreateDCFromMemory(&createInfo);
+    if (D3DKMTCreateDCFromMemory(&createInfo))
+      Logger::err(str::format("D3D9: Failed to create GDI DC"));
```
jsimmons, line 169: and another one :)

Joshua-Ashton:
I would definitely appreciate a kill switch option so we can disable this for games that accidentally hit this but don't need that.

Seeing TES: IV Oblivion in both windowed and fullscreem not hit this path would be nice. I don't know if it hits this specifically,, but I know its got some funk going on with child windows and stuff.

zfigura:
> @zfigura Like I said, I got the idea from the Wine code. I did get the GetDCEx call and the idea to use StretchBlt from there. I'll leave it to you whether that's already problematic when it comes to licenses or not.

I'm not a lawyer, but that sounds like it makes this a derivative work, and therefore puts this code under the LGPL 2.1 license.

jsimmons:
> > @zfigura Like I said, I got the idea from the Wine code. I did get the GetDCEx call and the idea to use StretchBlt from there. I'll leave it to you whether that's already problematic when it comes to licenses or not.
> 
> I'm not a lawyer, but that sounds like it makes this a derivative work, and therefore puts this code under the LGPL 2.1 license.

a "derived work" has a specific legal meaning (notably the copyrightable original work has to still exist in the derived work, and the derived work has to be independently copyrightable, neither of which are really true here it seems), it doesn't mean that any ideas you take from a project are automatically derived works.

K0bin:
I really thought this wouldn't be an issue with 2 lines of code.

uczekalla-cw:
@K0bin In reviewing, I don't think there is a violate here. It's safe to reopen and commit.

I think @zfigura really just wanted point out the licensing incompatibilities and in the future if there was any significant pieces of code from Wine that you wanted to adapt it can't simply be copied over. However, in these cases, we would be happy to have a conversation on how we could help develop a fix.

K0bin:
Great, thanks for looking into it!

zfigura:
At this point I've heard multiple different opinions as to whether this is copyrightable, and I don't have the legal knowledge to judge myself.

For what it's worth (and I don't hold copyright on the relevant code anyway, so my opinion is not worth much), I'm not trying to stand in opposition to DXVK qua DXVK, or DXVK qua a "competitor" to Wine, and I certainly don't want to contribute to any hostility between the projects. However, licensing is important to me, as I gather that it is important to the copyright owner, and I do take issue with copyleft code being copied into a project which seeks to undermine copyleft.

K0bin:
@zfigura Just to be clear, I never intended to violate any license here either. Sowing hostility between DXVK and Wine is the last thing I want.

src/d3d9/d3d9_swapchain.cpp
```diff
@@ -156,10 +162,49 @@ namespace dxvk {
...
+#endif
     }
   }
 
+#ifdef _WIN32
+  HRESULT D3D9SwapChainEx::BlitGDI(HWND Window) {
+    if (!m_warnedAboutFallback || true) {
+      Logger::warn("Using GDI for swapchain presentation. This will impact performance.");
+      m_warnedAboutFallback = true;
+    }
```
pchome, line 178: DXVK uses 
```cpp
    static bool s_errorShown = false;

    if (!std::exchange(s_errorShown, true))
      Logger::warn(...);
```
pretty much everywhere. Or I missed something and this case is special?

K0bin: You didn't miss anything. Good point.

xpander69:
Not sure if i should report it here? Warhammer Online Launcher is still broken with it. Allthough it kinda renders the main launcher window now, with few glitches. second settings window is still black:
launcher can be downloaded from here: http://launcher.returnofreckoning.com/RoRLauncher.exe (be aware that it will start downloading the data when you open it, allthough quite slowly due to bandwidth limitations)

ss_30042023_12 52 18

Joshua-Ashton (approved):

PR #3352, opened 2023-04-15
[dxgi] Implement support for more formats in GetMonitorFormatBpp

Etaash-mathamsetty:
Reduces log spam in 3dmark fire strike
I don't know if it is any more useful than that lol

### DIFF
--- a/src/dxgi/dxgi_monitor.cpp
+++ b/src/dxgi/dxgi_monitor.cpp
@@ -97,7 +97,46 @@ namespace dxvk {
 
   uint32_t GetMonitorFormatBpp(DXGI_FORMAT Format) {
     switch (Format) {
+      case DXGI_FORMAT_R1_UNORM:
+        return 1;
+
+      case DXGI_FORMAT_R8_UNORM:
+      case DXGI_FORMAT_R8_UINT:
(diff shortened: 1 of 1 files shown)


doitsujin:
What is Firestrike doing exactly? None of this should be necessary since we already handle all legal swap chain formats.

doitsujin:
Actually we're missing the RGB10A2 ones, but would still be good to know what the benchmark needs.

Etaash-mathamsetty:
> What is Firestrike doing exactly? None of this should be necessary since we already handle all legal swap chain formats.

Not sure, it just seems to be calling this function for the hell of it, it doesn't change anything related to how it runs or performs

doitsujin:
can you post a log?

Etaash-mathamsetty:
without patch:
[steam-231350.log](https://github.com/doitsujin/dxvk/files/11240837/steam-231350.log)

with patch:
[steam-231350.log](https://github.com/doitsujin/dxvk/files/11240834/steam-231350.log)

Etaash-mathamsetty:
@doitsujin I posted the logs, could you look at them?

doitsujin:
I did look at them, for some reason it's querying *everything* somehow. Need to think about a more elegant way to improve this.

PR #3361, opened 2023-04-18
Implement some small utilities in native headers

expipiplus1:
DEFINE_ENUM_FLAG_OPERATORS is used by WIDL 8.0 it seems, so this is
handy to have around if using more up to date headers.


expipiplus1:
Thanks!

PR #3363, opened 2023-04-18
AC Origins Bug

MCChristor:
In AC Origins is a bug where all textures are white and extreme bright after 20-40sec.


K0bin:
Again, you're trying to merge old branches that have already been merged.

This should be an issue. We do not support Mac OS though, so take this to the MoltenVK developers.

PR #3365, opened 2023-04-19
Add None WSI

expipiplus1:
This, of course, is unable to create windows.

TODO:

- [x] VK_EXT_headless surface support? (I think this could be a separate PR if desired)
- [ ] Testing 
- [x] Does this violate assumptions elsewhere? (Nothing that doesn't require `createSurface` succeeding first)

### DIFF
--- a/src/wsi/meson.build
+++ b/src/wsi/meson.build
@@ -20,6 +20,7 @@ wsi_glfw_src = [
 wsi_none_src = [
   'none/wsi_monitor_none.cpp',
   'none/wsi_window_none.cpp',
+  'none/wsi_edid_none.cpp',
 ]
 
 if dxvk_wsi == 'win32'
@@ -32,8 +33,8 @@ elif dxvk_wsi == 'glfw'
   wsi_src  = wsi_common_src + wsi_glfw_src
   wsi_deps = [ dep_displayinfo, lib_glfw ]
 elif dxvk_wsi == 'none'
-  wsi_src  = wsi_common_src + wsi_none_src
-  wsi_deps = [ dep_displayinfo ]
+  wsi_src  = wsi_none_src
+  wsi_deps = []
 else
   error('Unknown wsi')
 endif
--- a/src/wsi/meson.build
+++ b/src/wsi/meson.build
@@ -17,6 +17,11 @@ wsi_glfw_src = [
   'glfw/wsi_window_glfw.cpp',
 ]
 
+wsi_none_src = [
+  'none/wsi_monitor_none.cpp',
+  'none/wsi_window_none.cpp',
+]
+
 if dxvk_wsi == 'win32'
   wsi_src  = wsi_common_src + wsi_win32_src
   wsi_deps = [ dep_displayinfo ]
@@ -26,6 +31,9 @@ elif dxvk_wsi == 'sdl2'
 elif dxvk_wsi == 'glfw'
   wsi_src  = wsi_common_src + wsi_glfw_src
   wsi_deps = [ dep_displayinfo, lib_glfw ]
+elif dxvk_wsi == 'none'
+  wsi_src  = wsi_common_src + wsi_none_src
+  wsi_deps = [ dep_displayinfo ]
 else
   error('Unknown wsi')
 endif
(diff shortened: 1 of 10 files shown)


src/wsi/meson.build
```diff
@@ -26,6 +31,9 @@ elif dxvk_wsi == 'sdl2'
 elif dxvk_wsi == 'glfw'
   wsi_src  = wsi_common_src + wsi_glfw_src
   wsi_deps = [ dep_displayinfo, lib_glfw ]
+elif dxvk_wsi == 'none'
+  wsi_src  = wsi_common_src + wsi_none_src
+  wsi_deps = [ dep_displayinfo ]
```
pchome, line 36: I suppose `dep_displayinfo` and `wsi_common_src` (`wsi_edid.cpp`) are not required for this.
+1 TODO if yes

expipiplus1: ah, thanks!

expipiplus1: I added `wdi_edid_none` to replace `parseColorimetryInfo`.

expipiplus1:
I couldn't find any tests or builds for the non-windows wsi. Do these exist somewhere I've not checked?

pchome:
idk,
```cpp
#define DXVK_WSI_NONE 1
#include <d3d11.h>
#include <cstdio>

int main(int argc, char** argv) {
  ID3D11Device* device;
  ID3D11DeviceContext* ctx;
  D3D_FEATURE_LEVEL level = D3D_FEATURE_LEVEL_11_1;
  D3D_FEATURE_LEVEL level2 = D3D_FEATURE_LEVEL_11_1;
  HRESULT hr = D3D11CreateDevice(nullptr, D3D_DRIVER_TYPE_HARDWARE, nullptr, 0, &level, 1, D3D11_SDK_VERSION, &device, &level2, &ctx);

  if (FAILED(hr)) {
    printf("ko\n");
    return 1;
  }

  // TODO: do something
  // ...
  ctx->Flush();

  printf("done\n");
  return 0;
}
```
Run 
`DXVK_LOG_LEVEL=debug DXVK_LOG_PATH=none ./native_test_wsi_none`

Output
```
info:  Game: native_test_wsi_none
info:  DXVK: v2.1
info:  Vulkan: Found vkGetInstanceProcAddr in libvulkan.so @ 0x7f009c8bb920
info:  Built-in extension providers:
info:    No WSI
...
```

expipiplus1:
@pchome sure, writing such a test isn't hard, but there doesn't seem to be any infrastructure from which to call the test.

Joshua-Ashton:
I did have some tests at one point -- but currently unmerged. I think they are in the `sample` branch.

Joshua-Ashton:
Yeah sure.

Should probably hook up headless surface though.

expipiplus1:
ok, I've put a call to `vkCreateHeadlessSurfaceEXT` in there.

expipiplus1:
@Joshua-Ashton is this good to merge?

expipiplus1:
FWIW in terms of "testing", I've been using this since april without issue.

flibitijibibo:
@expipiplus1, do you have time to review the conversation in #3451 to see how this MR would best fit into an "official" WSI system for dxvk? Once we have a working system it should be easy to include a None implementation as part of that work.

expipiplus1:
@flibitijibibo It looks as though it would fit in with the API @Joshua-Ashton proposed there. As long as nothing *requires* getting a window or monitor then it's all fine. `CreateSurface` here just ignores the window parameter.

I would be keen to get this merged, and can commit to porting it to the `IDXVKWsi` when it lands.

flibitijibibo:
That makes sense to me - the absolute earliest I'll be able to look at DxvkWsi is December (if that; SDL_ActionSet work starts January so it's a tiny window), so this should probably go in first so that the implementer knows for sure that all backends still work correctly.

flibitijibibo:
The dynamic WSI patch series can now be found at #3738. Definitely give this a try whenever you have time; it's mostly the same at the backend level but with a C++ interface to allow for multiple simultaneous implementations in the official binary.

flibitijibibo:
This got stepped on by #3738, but the good news is that this should now be possible to support in the same binary as SDL/GLFW.

PR #3366, opened 2023-04-19
Update some app configs for Tecmo Koei games

mbriar:
I suspect the change for Nioh 1 should also be applied to Nioh 2, but I don't have this game to test.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -68,9 +68,9 @@ namespace dxvk {
       { "dxgi.deferSurfaceCreation",        "True" },
       { "d3d11.cachedDynamicResources",     "c" },
     }} },
-    /* Nioh: See Frostpunk, apparently?           */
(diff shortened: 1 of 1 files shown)


Blisto91:
Does dxgi not work for Nioh? 

> but I don't have this game to test.

I can give it a check when i get home today. If they are on steam

mbriar:
> Does dxgi not work for Nioh?

dxgi.deferSurfaceCreation, or even nothing at all (i.e. overriding the default config with `dxgi.deferSurfaceCreation = False`), works for Nioh nowadays. However, not deferring surface creation for d3d9 breaks VRR on X11 due to probably some invisible d3d9 window being created as soon as it plays videos.

Blisto91:
> I can give it a check when i get home today. If they are on steam

I realized i probably can't check this as i don't have VRR 🐸

Issue #3293, opened 2023-03-19
Cold fear dont render correctly (light issue?)


mrdeathjr28:
Cold fear dont show any texture but when use a lantern some things appear but as transparent in my case (change light settings dont affect)

![cldfra](https://user-images.githubusercontent.com/6601480/226146223-0e1783cf-c71b-4bbc-b24b-647905d29d61.png)

![cldfrb](https://user-images.githubusercontent.com/6601480/226146227-4bbebcd2-3490-4c00-9f8f-ba8b30619bc9.png)

- GPU: Intel UHD 630 CFL-GT2
- Driver: Mesa Oibaf
- Wine version:  Staging 8.4
- DXVK version:  2.1-git

Log files

[coldfear_retail_d3d9.log](https://github.com/doitsujin/dxvk/files/11009907/coldfear_retail_d3d9.log)

K0bin:
Please provide an apitrace.
https://github.com/doitsujin/dxvk/wiki/Using-Apitrace

mrdeathjr28:
Added trace

https://drive.google.com/file/d/12dzV9tBALOR5yrTmJAjAYMUW_VsWsuh3/view?usp=share_link

CME42:
FF_FS_ed95124a53eb473131462fe45cb6b82357b29a05.spv has the following validation error:
```
error: line 291: Vector access is out of bounds, vector size is 4, but access index is 6
  %108 = OpCompositeExtract %float %49 6
```
[FF_FS_ed95124a53eb473131462fe45cb6b82357b29a05.spv.zip](https://github.com/doitsujin/dxvk/files/11023480/FF_FS_ed95124a53eb473131462fe45cb6b82357b29a05.spv.zip)

mrdeathjr28:
Many thanks issue fixed

![cldfr](https://github.com/doitsujin/dxvk/assets/6601480/171525b1-7d7d-4b31-8644-25e792fa9217)

Joshua-Ashton:
Thanks for testing, good to see you still around @mrdeathjr28 :smile:

mrdeathjr28:
> Thanks for testing, good to see you still around @mrdeathjr28 smile

![pacino-al](https://github.com/doitsujin/dxvk/assets/6601480/11831ee4-f28f-4cc6-9c0f-a74d4ec2778c)

PR #3390, opened 2023-05-03
[d3d9] fix opCompositeExtract out of bound

AlexBozhenk:
Closes: https://github.com/doitsujin/dxvk/issues/3293

Signed-by: Oleksii Bozhenko <oleksii.bozhenko@globallogic.com>

### DIFF
--- a/src/d3d9/d3d9_fixed_function.cpp
+++ b/src/d3d9/d3d9_fixed_function.cpp
@@ -1102,19 +1102,30 @@ namespace dxvk {
     m_module.opStore(m_vs.out.NORMAL, outNrm);
 
     for (uint32_t i = 0; i < caps::TextureStageCount; i++) {
-      uint32_t inputIndex = (m_vsKey.Data.Contents.TexcoordIndices >> (i * 3)) & 0b111;
-      uint32_t inputFlags = (m_vsKey.Data.Contents.TexcoordFlags   >> (i * 3)) & 0b111;
+      uint32_t inputIndex = (m_vsKey.Data.Contents.TexcoordIndices     >> (i * 3)) & 0b111;
+      uint32_t inputFlags = (m_vsKey.Data.Contents.TexcoordFlags       >> (i * 3)) & 0b111;
+      uint32_t texcoordCount = (m_vsKey.Data.Contents.TexcoordDeclMask >> (i * 3)) & 0b111;
 
       uint32_t transformed;
 
       const uint32_t wIndex = 3;
 
       uint32_t flags = (m_vsKey.Data.Contents.TransformFlags >> (i * 3)) & 0b111;
-      uint32_t count = flags;
+      uint32_t count;
       switch (inputFlags) {
         default:
         case (DXVK_TSS_TCI_PASSTHRU >> TCIOffset):
           transformed = m_vs.in.TEXCOORD[inputIndex & 0xFF];
+          // flags is actually the number of elements that get passed
+          // to the rasterizer.
+          count = flags;
+          if (texcoordCount) {
+            // Clamp by the number of elements in the texcoord input.
+            if (!count || count > texcoordCount)
+              count = texcoordCount;
+          }
+          else
+            flags = D3DTTFF_DISABLE;
           break;
 
         case (DXVK_TSS_TCI_CAMERASPACENORMAL >> TCIOffset):
@@ -1798,10 +1809,9 @@ namespace dxvk {
             texcoord, texcoord, texcoordCnt, indices.data());
 
           uint32_t projIdx = m_fsKey.Stages[i].Contents.ProjectedCount;
-          if (projIdx == 0)
+          if (projIdx == 0 || projIdx > texcoordCnt)
             projIdx = texcoordCnt;
-          else
-            projIdx--;
+          --projIdx;
 
           uint32_t projValue = 0;
 


DadSchoorse:
Can you explain why this is nessecary? Fixed function code isn't easy to understand.

The commit message should be `[d3d9] ...` instead of `spirv: ...`, by the way.

AlexBozhenk:
@DadSchoorse Yep! As mentioned in https://github.com/doitsujin/dxvk/issues/3293#issuecomment-1477018851 we have out of bound during compilation to spirv. Disassembling the shader shows that the var is float[4]

```
%26 = OpCompositeConstruct %v4float %23 %24 %float_0 %float_0
...
%48 = OpLoad %v4float %in_Texcoord0
%49 = OpSelect %v4float %45 %26 %48
...
%108 = OpCompositeExtract %float %49 6
```

src/d3d9/d3d9_fixed_function.cpp
```diff
@@ -1115,6 +1115,7 @@ namespace dxvk {
         default:
         case (DXVK_TSS_TCI_PASSTHRU >> TCIOffset):
           transformed = m_vs.in.TEXCOORD[inputIndex & 0xFF];
+          count = 4;
```
Joshua-Ashton, line 1118: This is wrong, `TCI_PASSTHRU` doesn't set `count = 4` because it's meant to use the size of the texcoord in the vertex decl.

Is this required to close the other issue? If so I imagine there is something else going on, but this is definitely not the right solution!

https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dtexturetransformflags

src/d3d9/d3d9_fixed_function.cpp
```diff
@@ -1798,10 +1799,9 @@ namespace dxvk {
             texcoord, texcoord, texcoordCnt, indices.data());
 
           uint32_t projIdx = m_fsKey.Stages[i].Contents.ProjectedCount;
-          if (projIdx == 0)
+          if (projIdx == 0 || projIdx > texcoordCnt)
             projIdx = texcoordCnt;
-          else
-            projIdx--;
+          --projIdx;
```
Joshua-Ashton, line 1804: This part looks fine. Good catch.

src/d3d9/d3d9_fixed_function.cpp
Joshua-Ashton: Are you hitting a case here where `count == 0`? Curious... That might be the `D3DTTFF_DISABLE` case that is missing some handling.

Joshua-Ashton:
Once we solve what is going on with that random `count = 4`, I am happy to merge this.

src/d3d9/d3d9_fixed_function.cpp
AlexBozhenk: Not really. I'm hitting `count == 7.` Which is more than texcoord size

AlexBozhenk: For case if we have `count` smaller than 4 can suggest this:
```suggestion
          count = std::min(count, static_cast<uint32_t>(4));
```

AlexBozhenk: one more way it to prevent `D3DTTFF_DISABLE` if we have `count > 4` [here](https://github.com/doitsujin/dxvk/blob/101617753e9d4279877b9f7e0fa0721ffd6b8083/src/d3d9/d3d9_fixed_function.cpp#L1173). Obviously, it doesn't make sense to do `opCompositeExtract` if we don't `opCompositeInsert`(which won't do if count > 4)

AlexBozhenk: BTW, I'm not pretty sure it's a good way to get `count` assigning it to bit mask
https://github.com/doitsujin/dxvk/blob/101617753e9d4279877b9f7e0fa0721ffd6b8083/src/d3d9/d3d9_fixed_function.cpp#L1113
It looks for me that correct way to get count from bit mask is sth like this:
```
count = 0;
for (int i = 0; i<3; ++i)
  if ((flags >> i) & 1)
  	++count;
```

Joshua-Ashton: It's not like that though, the 'flags' are actually just a regular value.

```cpp
typedef enum _D3DTEXTURETRANSFORMFLAGS {
    D3DTTFF_DISABLE         = 0,    // texture coordinates are passed directly
    D3DTTFF_COUNT1          = 1,    // rasterizer should expect 1-D texture coords
    D3DTTFF_COUNT2          = 2,    // rasterizer should expect 2-D texture coords
    D3DTTFF_COUNT3          = 3,    // rasterizer should expect 3-D texture coords
    D3DTTFF_COUNT4          = 4,    // rasterizer should expect 4-D texture coords
    D3DTTFF_PROJECTED       = 256,  // texcoords to be divided by COUNTth element
    D3DTTFF_FORCE_DWORD     = 0x7fffffff,
} D3DTEXTURETRANSFORMFLAGS;
```

Joshua-Ashton: What game are you seeing this with btw? Is there an apitrace available?

Maybe we need to consider the vertex declaration's TEXCOORD stuff or something too, or just clamp it... Not sure.

AlexBozhenk: https://github.com/doitsujin/dxvk/issues/3293#issuecomment-1475057852

Joshua-Ashton: Ya I see this in the trace:
`2031784 @0 IDirect3DDevice9::SetTextureStageState(this = 0xca51c8, Stage = 3, Type = D3DTSS_TEXTURETRANSFORMFLAGS, Value = -1) = D3D_OK`

Joshua-Ashton: (Obviously you are seeing 7 because of the & with 0b111 -> 7, just to be clear :p)

Joshua-Ashton: We already handle the vertex decl thing later on with:

```cpp
          for (uint32_t j = count; j < 4; j++) {
            // If we're outside the component count of the vertex decl for this texcoord then we pad with zeroes.
            // Otherwise, pad with ones.

            // Very weird quirk in order to get texcoord transforms to work like they do in native.
            // In future, maybe we could sort this out properly by chopping matrices of different sizes, but thats
            // a project for another day.
            uint32_t texcoordCount = (m_vsKey.Data.Contents.TexcoordDeclMask >> (3 * inputIndex)) & 0x7;
            uint32_t value = j > texcoordCount ? m_module.constf32(0) : m_module.constf32(1);
            transformed = m_module.opCompositeInsert(m_vec4Type, value, transformed, 1, &j);
          }
```

So I am happy for us to just do a `count = min(count, 4)` here for the solution.

Joshua-Ashton: Once you have done that I will merge it.

Joshua-Ashton: Hmmm now I'm wondering if that should really be `count = min(count, whatever to read from texcoord deck mask)` actually given how that for loop works.

AlexBozhenk: I suggest  https://github.com/doitsujin/dxvk/blob/101617753e9d4279877b9f7e0fa0721ffd6b8083/src/d3d9/d3d9_fixed_function.cpp#L1173
`if (type != D3DTTFF_DISABLE && count < 4) {`
the routine below doesn't make any sense when count is 4 or bigger. the only thing we do is `opCompositeExtract` (in the way that makes spirv invalid)and don't use the value

Joshua-Ashton: We still need to do the matrix tho

AlexBozhenk: then, we should rewrite for cicles to

Joshua-Ashton: This is what I mean https://github.com/doitsujin/dxvk/tree/d3d9_fixed_function_tci

AlexBozhenk: Okay. Do you want me to grab the commit?
https://github.com/doitsujin/dxvk/commit/9d88e38c016aaac4c84882c208d1457884c63b52I 

With the `for` routine I meant we construct  transform only with `std::array<uint32_t, 4>`.
There is no sense in setting indeces from 4th.

Joshua-Ashton: Ya. Can you make sure it fixes the issue for you as well?

Joshua-Ashton: How can we have `texcoordCount` > 4 though? Isn't it literally from the decl. What's going on there? I am travelling now for a bit so won't get a chance to look myself.

AlexBozhenk: `count == 7 && texcoordCount == 0`
Good for you ;)

Joshua-Ashton: If texcoordCount == 0, we probably want to treat it as `NINED3DTSS_TCI_DISABLE` right?

AlexBozhenk: Sry, what is `NINED3DTSS_TCI_DISABLE `?
are you talking about [this ](https://github.com/doitsujin/dxvk/pull/3390/files#diff-f8e71b8e89bd998acbde4d82f201fa4f9cd73dfc7432eb2d905525df7bf476abR1182)?

Joshua-Ashton: Yeah, I had a look at what Gallium Nine (see link above) did and it disabled it in that case. So I imagine that's probably what we should do.

AlexBozhenk: Updated. Pls, check if it's what you meant :)

PR #3397, opened 2023-05-06
[d3d9] Keep reference to D3D9 module around

Joshua-Ashton:
Some games check if d3d9.dll is loaded in their WindowProc and use that to make decisions.

Some of these apps FreeLibrary after calling Direct3DCreate9 which is bad to start with... but ends up with them getting very confused and they constantly remake the device in the window proc we hooked causing a stack overflow.

I wonder if D3D11/DXGI should do the same here, needs testing.


Joshua-Ashton:
Seems like this solution is a red herring.

PR #3409, opened 2023-05-09
[dxgi] Enumerate all monitors if only one unique GPU is present

doitsujin:
TL;DR if users set something like `DXVK_DEVICE_FILTER` or various Vulkan-related environment variables to select a specific device, our current logic may end up reporting zero monitors for the DXGI adapter, which may cause all sorts of problems such as Assassin's Creed Valhalla being forced into a resolution of 1280x720. This should (hopefully) work around that.

@gofman

### DIFF
--- a/src/dxgi/dxgi_adapter.cpp
+++ b/src/dxgi/dxgi_adapter.cpp
@@ -162,7 +162,7 @@ namespace dxvk {
 
     auto linkedAdapter = m_adapter->linkedIGPUAdapter();
 
-    /* If either LUID is not valid enumerate all monitors. */
+    // If either LUID is not valid, enumerate all monitors.
     if (numLUIDs && linkedAdapter != nullptr) {
       const auto& deviceId = linkedAdapter->devicePropertiesExt().vk11;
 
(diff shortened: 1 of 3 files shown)


gofman:
This looks to me like a good thing to do to increase the overall robustness even regardless of DXVK_DEVICE_FILTER used. But I can suggest a way to make it more generic. instead of relying on seeing one GPU we can instead check if we end up with a full monitor set  when enumerating by adapter's LUIDs. That is probably a more thorough config sanity validation. Not entirely sure if that worth the complication over this patch (but also maybe it isn't much more convoluted).

gofman:
Thinking a bit more about it, it seems that "full monitor set check" is really preferable as it will catch the cases which might happen without any user special config and even with a single GPU when GDI adapter is not linked to Vulkan GPU at all (like, e. g., with broken Xrandr on XWayland with upstream Wine).

doitsujin:
Good point actually, the previous workaround would also not have worked if users forced Lavapipe etc. I updated the branch with the new approach, now we basically check if there are any orphaned monitors.

PR #3415, opened 2023-05-12
[d3d11] Fix min luminance of HDR Metadata

Joshua-Ashton:
This should be a multiply instead of a divide. Values are 1/10000th of a nit (0.0001 nit).


doitsujin:
How did we mange to fuck this up?

doitsujin (approved):

PR #3420, opened 2023-05-16
[util] Add config for Tom Clancy's The Division.

giomasce:
During some cinematics the game calls Present() frequently switching between SyncInterval 0 and 1. At each change the Vulkan swapchain must be rebuilt. This can cause some stuttering, which is fixed by overriding SyncInterval to 1.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -127,6 +127,10 @@ namespace dxvk {
       { "dxgi.maxFrameRate",                "60"   },
       { "d3d11.cachedDynamicResources",     "vi"   },
     }} },
+    /* Tom Clancy's The Division                  */
+    { R"(\\TheDivision\.exe)", {{
+      { "dxgi.syncInterval",                "1"    },
+    }} },
     /* SteamVR performance test                   */
     { R"(\\vr\.exe$)", {{
       { "d3d11.dcSingleUseMode",            "False" },


doitsujin:
I **really** don't like force-enabling vsync unconditionally, especially since this introduces different behaviour between D3D11 and D3D12.

I think the correct fix will be to support dynamic present modes from `VK_EXT_swapchain_maintenance1`, but driver support for that hasn't been stellar so far so I haven't really bothered with that until now.

giomasce:
As far as I can see the game requests `SyncInterval = 0` only for a few dozens (non consecutive) frames during the initial video. Once you reach main game, it seems to be `SyncInterval = 1` forever (though, to be fair, I didn't play more than a few minutes). In this situation you would be overriding the game's request only for a few seconds, preventing some stuttering there.

That said, it's of course your call. I thought overriding `SyncInterval` was a pretty common thing to do, given that many other games have that config enables, but no hard feelings.

mbriar:
> Once you reach main game, it seems to be SyncInterval = 1 forever

but probably only if you enable vsync in the game settings, overriding the sync interval in DXVK would mean that the game's vsync setting doesn't work anymore.

doitsujin:
That, and also all games that currently have that setting also have a hardcoded 60 FPS lock.

EDF5 also constantly changes SyncInterval during gameplay so we currently force it into the Mailbox present mode, but that's also something we can get rid of by implementing the new Vulkan swapchain stuff.

mbriar:
> I think the correct fix will be to support dynamic present modes from VK_EXT_swapchain_maintenance1

You are probably already aware of this, but dynamic present mode toggling is still not supported for the x11 wsi in Mesa (only wayland wsi FIFO <> MAILBOX), so it won't work for any game on Proton.

doitsujin:
Do we know if there's anything preventing it from ever being supported or is this just a case of nobody having worked on it yet? Given that even GL can toggle present modes dynamically I'm hoping for the latter.

mbriar:
From https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/20235:

> Toggling of presentation modes on the fly. I implemented this for Wayland since we can trivially toggle between MAILBOX and FIFO this way. On X11, we could do this too, but the existing implementation has acquire/present queue threads that complicate things. We'd need to rewrite it to always use threads regardless of present mode. Then we could just push the present mode into wsi_acquire_queue instead, but that has to be revisited later.

Doesn't sound impossible, just not done yet.

doitsujin:
Got a branch with swapchain_maintenance1 work: https://github.com/doitsujin/dxvk/tree/swapchain-rework

Currently only works when using Gamescope and setting `ENABLE_GAMESCOPE_WSI=1` for the reasons mentioned above.

oscarbg:
seems Nvidia 545.37 Windows "insider driver" just gained VK_EXT_swapchain_maintenance1 support..

PR #3428, opened 2023-05-18
[dxvk] Add configurable HUD opacity

### LINKED ISSUE #3371: [Feature Request] DXVK_HUD Opacity
Would like to be able to set either max nits of overlay or overlay transparency/opacity to 50% to avoid burning my oled if possible?

WinterSnowfall:
Adds an extra parameter to the DXVL_HUD env variable and config option, which lets users configure HUD opacity, similarly to how HUD scaling can currently be adjusted. It can help prevent burn-in on some OLED panels and will help hardcore testers such as @Blisto91 and myself make sense of what is hiding behind the HUD when dealing with games stuck in low resolutions.

Closes #3371.


src/dxvk/hud/shaders/hud_graph_frag.frag
```diff
@@ -39,7 +40,7 @@ void main() {
 
   float value = mix(p0.value, p1.value, fx);
   float alpha = value + v_coord.y - 1.0f;
-  alpha = min(alpha / dFdy(v_coord.y), 1.0f);
+  alpha = min(alpha / dFdy(v_coord.y), opacity);
```
doitsujin, line 43: Shouldn't we be multiplying the final alpha with `opacity`?

doitsujin, line 43: Shouldn't we be multiplying the final alpha value with `opacity`?

doitsujin (changes requested):

src/dxvk/hud/shaders/hud_graph_frag.frag
WinterSnowfall: I'm rather new to shaders, but I've observed it has the same end effect here. If it helps with clarity, I'll do the multiplication at the end, sure.

pchome:
What if it will be 50% (or so) by default?
With the switch option "bright" for 100%.

WinterSnowfall:
Some may not like "50% (or so)". I think the fully opaque default is fine and have no intention of changing it. The config option or env variable makes it easy enough to use whatever one finds best suited.

PR #3441, opened 2023-05-22
[dxvk] fix copy mistake in dxvk_adapter + add macro to avoid the issue in the future

EndlesslyFlowering:
titles says it all


EndlesslyFlowering:
I added a macro as per @DadSchoorse 's suggestion to avoid this isssue in the future :)

doitsujin:
not a big fan of macros but whatever i guess

Issue #2375, opened 2021-11-22
Flammable Freddy - Bad performance


K0bin:
Flammable Freddy, a game made by a Youtuber I regularly watch, runs very poorly with DXVK and runs out of address space. The game is made with a game engine called Clickteam Fusion, I assume other games using it have the same issue.

The problem is that the game heavily relies on D3DLOCK_NOOVERWRITE when uploading data before every draw. The buffer is a vertex buffer in the systemmem pool with the usages D3DUSAGE_WRITEONLY | D3DUSAGE_DYNAMIC. DXVK ignores D3DLOCK_NOOVERWRITE for everything that isn't in D3DPOOL_DEFAULT, so we do a ton of copying and eventually run out of address space.

The game itself isn't really important but it highlights that we still havent gotten the D3DLOCK flags right.

### Software information
Flammable Freddy

### System information
- GPU: GTX 1070
- Driver: 495.44
- Wine version: 6.21
- DXVK version: 1.9.2 or PR 2364.

### Log files
- d3d9.log: Not useful.

K0bin:
The problematic buffer is ~300kb. It doesn't crash anymore because DXVK limits the amount of staging memory that can be used for buffer uploads but performance is atrocious with 30+ gpu syncs per frame taking a combined 30+ms.

Unfortunately the game uses SizeToLock = 0, so we end up uploading the entire 300kb buffer over and over and over again.

The solution to this would be to only upload the parts of dynamic sysmem buffers that are used for drawing.
Nine does this now:
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/9451

K0bin:
https://github.com/K0bin/dxvk/tree/dynamic-smol-uploads fixes the issue but I'm not sure if I want to land that.

PR #3454, opened 2023-05-26
Buffer locking cleanup

K0bin:
Feels good to remove hacks.

Thanks to Axel Davy of Gallium Nine fame for telling me about the SWVP edge case here.

8bf11fd helps Injustice: Gods Among Us.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -2463,6 +2463,9 @@ namespace dxvk {
     if (bSoftware && !CanSWVP())
       return D3DERR_INVALIDCALL;
 
+    if (!bSoftware && (m_behaviorFlags & D3DCREATE_SOFTWARE_VERTEXPROCESSING))
+      return D3DERR_INVALIDCALL;
+
     m_isSWVP = bSoftware;
 
     return D3D_OK;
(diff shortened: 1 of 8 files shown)


K0bin:
I decided to keep the "allowDirectBufferMapping" option after all and enable it for Injustice.

WinterSnowfall:
Does this also fix https://github.com/doitsujin/dxvk/issues/2375 mayhaps?

K0bin:
@WinterSnowfall It doesn't, see my comment in the issue.

src/d3d9/d3d9_device.cpp
```diff
@@ -4808,6 +4808,10 @@ namespace dxvk {
     if (desc.Usage & D3DUSAGE_DYNAMIC)
       Flags &= ~D3DLOCK_DONOTWAIT;
 
+    // Tests show that D3D9 drivers ignore DISCARD when the device is lost.
+    if (unlikely(m_deviceLostState != D3D9DeviceLostState::Ok))
```
Joshua-Ashton, line 4809: What about for images?

Joshua-Ashton (commented):
I am sus about this commit but I guess we can see how it plays out...

PR #3470, opened 2023-06-02
[util] Set maxChunkSize to 1 for the Rockstar launcher and social club

esullivan-nvidia:
Currently when Red Dead Redemption 2 is running the Rockstar launcher, and social club each use over 600MB of VRAM. The root cause of this is DXVK creating two memory pools of 256MB for read, and read/write resources. Reducing the max chunk size to 1MB reduces the memory usage of each to around 40MB.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -364,6 +364,14 @@ namespace dxvk {
     { R"(\\SnowRunner\.exe$)", {{
       { "d3d11.dcSingleUseMode",        "False"   },
     }} },
+    /* Rockstar Games Launcher                    */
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
This makes sense to me, it would be nice if we didnt have to have hacks like this, but that would probably require a large scale rewrite of the memory allocator to be smarter.

doitsujin:
I guess one thing we could do is gradually increase the chunk size up to a certain threshold as apps create more and more resources, but doing so **will** negatively affect apps that do use larger amounts of memory. I guess the question is whether or not it'll be bad enough to actually matter though.

doitsujin:
Going to merge this for now but this is an increasingly common issue and we should at least try to cater to these use cases to some degree.

PR #3477, opened 2023-06-04
[util] Add app profile for Fallout 76

Joshua-Ashton:


doitsujin:
Should we do something along the lines of forcing the game into Mailbox mode and enabling our own 60 FPS lock perhaps?

Also I'm not really sure about what to do with these workarounds in situations where dynamic present modes actually work now that support for those is merged.

Joshua-Ashton:
Well the game actually runs fine at 144Hz, etc unlike other Bethesda games... It will also target that in fullscreen mode but not in windowed, the game is just being weird and it works fine by forcing it here.

The present modes aren't really desirable as even on Windows the game can get stuck targeting a lower refresh rate than it could for the rest of the session...

PR #3480, opened 2023-06-06
Add some WAs for UE4 games

Samsuper12:
These games use UE4 and, without customVendorId, they don't run on Intel GPUs due to NVAPI.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -303,9 +303,11 @@ namespace dxvk {
     }} },
     /* Stray - writes to the same UAV every draw, *
      * presumably for culling, which doesn't play *
-     * nicely with D3D11 without vendor libraries */
+     * nicely with D3D11 without vendor libraries 
+     * Also, need the NVAPI hack for Intel GPUs   */
     { R"(\\Stray-Win64-Shipping\.exe$)", {{
       { "d3d11.ignoreGraphicsBarriers",     "True" },
+      { "dxgi.customVendorId",              "1002" },
     }} },
     /* Metal Gear Solid V: Ground Zeroes          *
      * Texture quality can break at high vram     */
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -309,6 +309,11 @@ namespace dxvk {
       { "d3d11.ignoreGraphicsBarriers",     "True" },
       { "dxgi.customVendorId",              "1002" },
     }} },
+    /* System Shock (2023) - NVAPI hack for
+     * Intel GPUs                                 */
+    { R"(\\SystemReShock-Win64-Shipping\.exe$)", {{
+      { "dxgi.customVendorId",              "1002" },
+    }} },
     /* Metal Gear Solid V: Ground Zeroes          *
      * Texture quality can break at high vram     */
     { R"(\\MgsGroundZeroes\.exe$)", {{


Blisto91:
The game won't use Nvapi if it sees a Intel vendorId. 

While i am familiar with the theory of UE games doing a driver check there are a couple of things i'm confused about.
Why does Stray need this? I know from [the mesa issue](https://gitlab.freedesktop.org/mesa/mesa/-/issues/6924#note_1485522) that it at least used to work on some Intel gens.
Why if it's a driver version check issue does it work when people launch these games with `-dx12`?  From what i've gathered

Also settings System Shock permanently to a AMD id would break Nvidia being able to use DLSS out of the box with Nvapi enabled i'm guessing so imo a proper workaround like the current nvapiHack that would only apply to Intel would be needed.

mbriar:
> Why if it's a driver version check issue does it work when people launch these games with -dx12

As I've posted [here](https://gitlab.freedesktop.org/mesa/mesa/-/issues/7928#note_1695389), the problem seems to be a driver version check that is specific to the DX11 RHI that will affect all games uses those versions of UE4 on DX11. Which is also why I think that a better workaround, potentially on the [proton side](https://github.com/ValveSoftware/Proton/issues/6402), is needed. Working around it per game will miss a bunch of stuff and working around it like this on the dxvk side will cause problems for non-intel GPUs, as @Blisto91 already said. If a fix on the proton side is not possible, then maybe anv should add a dri.conf workaround for all games which have an executable name ending in `-Shipping.exe` (I think the only way to detect ue4 dx11 games).

Samsuper12:
Ok. Thanks all for the feedback. Let's finish this bug.

I'll think about a better option and I'll come back.

Samsuper12:
> While i am familiar with the theory of UE games doing a driver check there are a couple of things i'm confused about. Why does Stray need this? I know from [the mesa issue](https://gitlab.freedesktop.org/mesa/mesa/-/issues/6924#note_1485522) that it at least used to work on some Intel gens. Why if it's a driver version check issue does it work when people launch these games with `-dx12`? From what i've gathered

Hm, and I forgot something important. Stray can run on proton 6.3 without any issues (remove an old wine prefix before doing it)

Samsuper12:
>  If a fix on the proton side is not possible, then maybe anv should add a dri.conf workaround for all games which have an executable name ending in -Shipping.exe (I think the only way to detect ue4 dx11 games.

Even Dead By Daylight has that ending: `DeadByDaylight-Win64-Shipping.exe`

So, yep, it could be a marker.

mbriar:
> Hm, and I forgot something important. Stray can run on proton 6.3 without any issues

I haven't checked yet, but it's possible that proton 6.3 still defaults to wine's opengl-based dxgi implementation or that something else w.r.t. vendorId reporting has changed.

mbriar:
> The UE code link doesn't work for me

You need to link a github account to an Epic Games account to get source access (freely available for anyone to do).

> I'm confused why the dx12 path would do things differently

The problematic check causing crashes on Intel that is hit at least [here](https://gitlab.freedesktop.org/mesa/mesa/-/issues/7928) only exists in the DX11 RHI code.

Blisto91:
Would be nice to figure the specific reason why there possibly was a behavior change in Proton that made it work before but not now. If that even is the case.

Anyway hopefully it gets some attention either way.

doitsujin:
So what is it that's actually being proposed here? Should we extend the already existing Nvidia workaround to Intel or just come up with something UE4-specific?

Blisto91:
Since there are also issues with the Intel version of XeSS in some games like Cyberpunk if a vendorId hack were implemented in dxvk it would make sense to do it all around and not just UE4 imo

Samsuper12:
This PR might be closed. These games fixed with my [PR](https://github.com/ValveSoftware/wine/pull/191) in ValveSoftware/Proton

PR #3492, opened 2023-06-14
[dxvk] Always use LoadLibrary in OpenXR extension provider

gofman:
Fixes GTA V randomly crashing on start after recent update.

The crash is due to DxvkXrProvider::queryDeviceExtensions() calls wineopenxr functions while openxr is already unloaded. The logic with GetModuleHandle() can only work correctly if DxvkXrProvider class is a process global singleton. This is not the case during GTA V startup as different instances of the object are present in both dxgi.dll and d3d9.dll and that races resulting in premature library unloading.

### DIFF
--- a/src/dxvk/dxvk_openvr.cpp
+++ b/src/dxvk/dxvk_openvr.cpp
@@ -305,11 +305,14 @@ namespace dxvk {
 
 
   HMODULE VrInstance::loadLibrary() {
-    HMODULE handle = nullptr;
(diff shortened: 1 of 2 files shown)


Joshua-Ashton (approved):
Makes sense to me. Any other places we are doing this?

gofman:
OpenVR also has that as a fallback path (which I think is never taken now). I will fix there as well just in case.

PR #3508, opened 2023-06-17
[util] Limit STEINS;GATE ELITE to 60 fps

tgurr:
Intros, menu and ui animations are playing way too fast, also causing input issues.

---
Executable name `Game.exe` is unfortunately a bit vague but:

image

Tested with

```
[Game.exe]
d3d9.maxFrameRate = 60
```
and running the game via steam with `DXVK_CONFIG_FILE=/mnt/games/dxvk-temp/dxvk-steinsgate.conf %command%`.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -778,6 +778,10 @@ namespace dxvk {
     { R"(\\injustice\.exe$)", {{
       { "d3d9.allowDirectBufferMapping",    "False" },
     }} },
+    /* STEINS;GATE ELITE                       */
+    { R"(\\SG_ELITE\\Game\.exe$)", {{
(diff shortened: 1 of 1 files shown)


Blisto91:
Verified in the game. In Windowed mode it will already cap at 62.5fps, but in fullscreen it will just up to what ever your refresh rate is.

As Bitwolfies said when the exe name is this generic we should include the game default directory name so we don't risk crossfire with other applications.

tgurr:
@Bitwolfies, @Blisto91 thanks I didn't know that I'll work this way, I thought that would just work for subdirectories, I've updated the PR to include the games default main directory.

Blisto91 (approved):

PR #3509, opened 2023-06-18
[util] Report NVIDIA for Star Wars Battlefront as it errors out on AMD

Trevonn:
The message you get before the change

image

https://www.protondb.com/app/1237980

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -119,7 +119,7 @@ namespace dxvk {
     }} },
     /* Star Wars Battlefront (2015)               */
     { R"(\\starwarsbattlefront(trial)?\.exe$)", {{
-      { "dxgi.nvapiHack",                   "False" },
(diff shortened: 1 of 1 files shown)


src/util/config/config.cpp
```diff
@@ -120,6 +120,7 @@ namespace dxvk {
     /* Star Wars Battlefront (2015)               */
     { R"(\\starwarsbattlefront(trial)?\.exe$)", {{
       { "dxgi.nvapiHack",                   "False" },
```
Blisto91, line 122: The change will make the nvapiHack line redundant so might as well also remove that.

Trevonn: That makes sense thanks. I'll keep that in mind for the future.

Should be fixed now

Blisto91:
Verified the game doesn't complain in either Proton or regular upstream Wine with a AMD GPU.

PR #3411, opened 2023-05-10
Implement Direct3D 8 Frontend

AlpyneDreams:
D8VK is far more feature complete at this point than I believe D9VK was when it was merged into DXVK in 2019. There are still maybe a few dozen edge case games that haven't been fixed yet but [every game in this list](https://github.com/AlpyneDreams/d8vk/wiki/Supported-Games-List) plus many more do work at this point.

Issues that would need to be closed before merging: (a lot of this is just admin)
- AlpyneDreams/d8vk#15

Things to do before merge:
- [x] Separate PR's for changes to common code and D3D9
  - [x] #3408 
  - [x] #3510 - b2b89ee0b8f99ed35c063ad3c69649f91a66f70f
  - [x] #3511 - 4a7c627e3dc7d706916b870cd1e15f510c3951fd
  - [x] #3514 - 0dbc8763e9ab971829cff46ea5ecfc2c82e3c1d1
  - [x] #3515 - 125cb1fefc927d07de92925a60a901d889721ca6
  - [x] #3517
  - [x] #3519
  - ~~#3516~~
  - [x] #3518 - DxvkD3D8Bridge 
  - [x] #3547 
  - ~~The warning disable in 7d45fb6f1a62995bd198acfe03da043fdac97d7c~~
- [x] Remove setup_d3d8.sh and build_d3d8.sh?
- [x] Rebase everything now and sort out things like build system changes
- [x] Squash into one mega commit with co-authors - GitHub will do this for us

Things that can be done after the merge:
- #3565 
- [Remove any upstream d3d8 workarounds in Proton](https://github.com/doitsujin/dxvk/pull/3411#issuecomment-1547302863)
- CopyRects edge cases and optimizations
- [All the issues](https://github.com/AlpyneDreams/d8vk/issues)
  - AlpyneDreams/d8vk#68
- Plus these fun but minor things:
  - AlpyneDreams/d8vk#148
  - AlpyneDreams/d8vk#52
- AlpyneDreams/d8vk#45
  - #1498
  - #1828  (Apparently works?)
  - #2605  (Apparently works?) 
  - #1798
- AlpyneDreams/d8vk#44
  - f974535b7058ada89cb3de886ef9687de73610e0 - [d3d8] EndScene on Reset (maybe?)
  - AlpyneDreams/d8vk#43
- AlpyneDreams/d8vk#40

**Note about static linking**: We are moving back to dynamic linking. The following games were fixed specifically when we added static linking, but currently work with the latest main even with dynamic linking:
[...]

PR #3510, opened 2023-06-19
[d3d9] Initialize m_mapFlags to 0

AlpyneDreams:
Part of #3411

I have found in debugging games with d8vk you could sometimes get weird default values for m_mapFlags


K0bin (approved):
Nice find.

PR #3516, opened 2023-06-19
[util] Allow switching log file at runtime

AlpyneDreams:
Part of #3411

Required since d3d8 is a single DLL, but links d3d9 as a static library, so we need a way to override the initialization-time log filename.

### DIFF
--- a/src/util/log/log.cpp
+++ b/src/util/log/log.cpp
@@ -6,15 +6,54 @@
 
 namespace dxvk {
   
-  Logger::Logger(const std::string& fileName)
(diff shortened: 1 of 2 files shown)


Joshua-Ashton:
Is there a reason to static link d3d9?

Joshua-Ashton:
I do not really see the point in overcomplicating this, let's just dynamically link and keep the log files the same.

WinterSnowfall:
@Joshua-Ashton Dynamically linking caused a lot of crashing on load issues on Windows. Not entirely sure why, but Windows compatibility drastically increased when d8vk switched to static linking. Maybe it's something we need to look deeper into.

PR #3527, opened 2023-06-22
Minor hazard fixes + opts

Joshua-Ashton:


src/d3d9/d3d9_device.cpp
```diff
@@ -7309,7 +7314,7 @@ namespace dxvk {
   void D3D9DeviceEx::ResetState(D3DPRESENT_PARAMETERS* pPresentationParameters) {
     SetDepthStencilSurface(nullptr);
 
-    for (uint32_t i = 0; i < caps::MaxSimultaneousRenderTargets; i++)
+    for (uint32_t i = 1; i < caps::MaxSimultaneousRenderTargets; i++)
```
K0bin, line 7317: This will break Reset.

Reset checks for alive D3DPOOL_DEFAULT textures after calling ResetState, so we cannot hold a reference to RT0 during Reset.

src/d3d9/d3d9_device.cpp
```diff
@@ -443,7 +442,7 @@ namespace dxvk {
       m_autoDepthStencil = nullptr;
     } else {
       // Extended devices only reset the bound render targets
-      for (uint32_t i = 0; i < caps::MaxSimultaneousRenderTargets; i++) {
+      for (uint32_t i = 1; i < caps::MaxSimultaneousRenderTargets; i++) {
```
K0bin, line 445: Changing it to 1 here is fine because we don't check for D3DPOOL_DEFAULT resources on extended devices.

src/d3d9/d3d9_device.cpp
```diff
@@ -5592,18 +5594,24 @@ namespace dxvk {
 
 
   void D3D9DeviceEx::MarkRenderHazards() {
+    struct {
+      uint8_t RT : 1;
+      uint8_t DS : 1;
+    } hazardState;
```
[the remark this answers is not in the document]
Joshua-Ashton: Indeed, but better than wasting space on something that can be 2 bits.

K0bin (approved):

PR #3529, opened 2023-06-22
D3D9: Port flush heuristic from D3D11

doitsujin:
Needs testing.

The flush heuristic itself already works well for D3D11 in practice, with major benefits being that:
- There is no race condition with the pending submission count, unlike in the old code
- We can flush at CS chunk boundaries, even during extremely long command sequences where we otherwise wouldn't
- CPU timing is no longer used as a source of information, which makes this a lot more consistent
- Resource readback latency in general should be **much** lower.


Joshua-Ashton:
We probably don't want to flush halfway through some commands like StretchRect, ProcessVertices, LockImage and Clear which have a couple of EmitCs'es...

A few of these can probably be changed to be a lot smarter/less shit than the code I wrote like 3 years ago to avoid that too.

Riesi:
This currently breaks Gothic 3, which wont continue after finishing loading a game. Not sure if Josh's remarks would fix it.

doitsujin:
Gothic should be fixed.

Riesi:
yep works now :beers:

Joshua-Ashton:
Looks fine but I still want to merge some of these EmitCses at some point..,

PR #3532, opened 2023-06-24
[d3d9] Keep 1 presenter per swapchain window

Joshua-Ashton:
Some apps such as level editors such as Hammer World Editor, some GUI apps/launchers etc use window overrides in presentation.

Previously we'd remake a new surface every time, which was incredibly slow making these apps basically unusable.

Now we keep one surface + swapchain + image views around per window/window override we have, along with the frame latency objs + frame counter. (Obviously an app may present to multiple windows in a frame, so for frame latency purposes we track that per-window.


K0bin (approved):

K0bin:
I'm not a huge fan of `m_window` and `m_wctx` though. They change every time present is called, so IMO they should ideally be passed through as arguments from Present().

PR #3536, opened 2023-06-24
[d3d8] Add NVIDIA shadow filtering vendor hack for d3d8

AlpyneDreams:
Part of #3411, but not a requirement for #3411 to merge.

Features:
- Some early D3D8 games expect Dref to be on the range of [0..2^bitDepth - 1], so this adds a config option for DXSO and D3D9 fixed function to scale it back down to [0..1]. [Source](http://www.jiri-dvorak.cz/scellpt/)
- Certain D3D8 games expect hardware to perform shadow filtering when they sample depth textures. This PR adds another config option to enable a basic 2x2 PCF to emulate hardware shadow filtering.

More info:
- http://www.jiri-dvorak.cz/scellpt/
- AlpyneDreams/d8vk#65
- AlpyneDreams/d8vk#69

### DIFF
--- a/src/dxso/dxso_compiler.cpp
+++ b/src/dxso/dxso_compiler.cpp
@@ -2929,6 +2929,12 @@ void DxsoCompiler::emitControlFlowGenericLoop(
         fetch4,
         imageOperands);
 
+      // [D3D8] Emulate hardware shadow filtering for 2D depth texture lookups.
+      if (depth && m_moduleInfo.options.shadowFilter && samplerType == SamplerTypeTexture2D) {
+        const uint32_t sampledImage = m_module.opLoad(sampler.typeId, sampler.varId);
+        result.id = DoFixedFunctionShadowFilter(m_module, result.id, sampledImage, texcoordVar.id, reference, imageOperands);
+      }
+
       // If we are sampling depth we've already specc'ed this!
       // This path is always size 4 because it only hits on color.
       if (isNull != 0) {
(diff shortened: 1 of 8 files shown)


AlpyneDreams:
Goals:
- Re-implement for DXSO ideally without a spec constant.
- Re-implement in fixed function renderer ideally without extending the PS shader key (want to avoid variants).
  - Use `D3D9FixedFunctionOptions` if necessary to configure behavior
- Try to avoid conditionals in the pixel shader too
- Should also work with Splinter Cell: Pandora Tomorrow as well as Splinter Cell
  - And be applicable for other games

Options:
- Best case: entirely hard-coded based on texture type and config settings for both FF and DXSO
- Push constants have more room than spec constants
- Extending D3D9SharedPS is also an option

Questions:
- Can the Dref scale factor be hard-coded per-game (e.g. if game only uses D24S8) or per-shader (most likely case assuming games never bind more than one depth texture in a shader anyway)?

src/dxso/dxso_compiler.cpp
```diff
@@ -2896,10 +2896,28 @@ void DxsoCompiler::emitControlFlowGenericLoop(
...
         reference = m_module.opCompositeExtract(
           fType, texcoordVar.id, 1, &component);
+
+        // Clamp Dref to [0..1] for D32F emulating UNORM textures 
         uint32_t clampDref = m_spec.get(m_module, m_specUbo, SpecDrefClamp, samplerIdx, 1);
         clampDref = m_module.opINotEqual(bool_t, clampDref, m_module.constu32(0));
         uint32_t clampedDref = m_module.opFClamp(fType, reference, m_module.constf32(0.0f), m_module.constf32(1.0f));
         reference = m_module.opSelect(fType, clampDref, clampedDref, reference);
+
+        // Scale Dref to [0..(2^N - 1)] for D24S8 and D16
```
DadSchoorse, line 2906: Shouldn't this be done before the clamp?

PR #3539, opened 2023-06-25
[util] BLADESTORM: Nightmare - Game speed increases when above 60 FPS

Trevonn:
The game has 3 v-sync options but doesn't explain what they do. 
0 = 60 FPS
1 = Monitor Refresh Rate
2 = 30 FPS

Framerate is capped at 60 in missions and then up to monitor refresh in the main menu and tavern area

This PR would provide a better default experience for people using option 1 with high refresh displays


Blisto91:
Verified ingame

PR #3549, opened 2023-07-05
[util] Limit fps in The Incredibles

Blisto91:
The higher uneven framerates [can bug the game out](https://www.pcgamingwiki.com/wiki/The_Incredibles#Dash_Glitch). The horrible frame timings is allegedly because even though it appears uncapped besides vsync it will try to reach for 60fps. Even then it has some frame time issues, so opted to cap it just below.

<details>
  
uncapped

60

59

</details>

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -788,6 +788,10 @@ namespace dxvk {
     { R"(\\SG_ELITE\\Game\.exe$)", {{
       { "d3d9.maxFrameRate",                "60" },
     }} },
+    /* The Incredibles                         */
+    { R"(\\IncPC\.exe$)", {{
+      { "d3d9.maxFrameRate",                "59" },
+    }} },
     
     /**********************************************/
     /* D3D12 GAMES (vkd3d-proton with dxvk dxgi)  */


K0bin:
I'm not entirely against this but do we really want to keep workarounds for every last shovelware game?

Blisto91:
Dunno. Up to you guys how much and what you wanna accept ofc.

On a semi related note i don't know if it's been considered to move the built in configs in to another more organized file format or split them up? Just if it starts getting too cluttered or packed in a regular code file.

jams3223:
> I'm not entirely against this but do we really want to keep workarounds for every last shovelware game?

I agree, developers need to learn on how to fix their games.

K0bin:
@jams3223 That's not gonna happen, especially not for 20 year old games. I'm definitely in favor of working around bugs so our users have a good out of the box experience in popular games. I just don't think we should do it for every game.

jams3223:
> @jams3223 That's not gonna happen, especially not for 20 year old games. I'm definitely in favor of working around bugs so our users have a good out of the box experience in popular games. I just don't think we should do it for every game.

These developers could remaster their game and fix the issues, it would be a win win as they could increase the pricing but anyways developers nowadays are only looking for a cash grab that's why i mostly but innovative or indie games, there's nothing out of the box that triple A studio tries to do anymore other than Nintendo and Valve when they finally choose to release a game.

jams3223:
@Joshua-Ashton can we drop the workaround for Resident Evil 2 & 3 as they've removed support for DX11, the game only use DX12 for now on.

Blisto91:
The old d3d11 version is still available as a beta option in Steam and used by a bunch of people I believe.

jams3223:
> The old d3d11 version is still available as a beta option in Steam and used by a bunch of people I believe. Just as a extra note.

Yeah you're right maybe we should wait a couple of months before we do that since some people are starting to complain that their losing compatibility with the game or have issues that the dev won't ever fix since the dx11 version is abandonware.

K0bin:
> These developers could remaster their game and fix the issues, it would be a win win as they could increase the pricing but anyways developers nowadays are only looking for a cash grab that's why i mostly but innovative or indie games, there's nothing out of the box that triple A studio tries to do anymore other than Indies, Nintendo and Valve when they finally choose to release a game.

That's often not possible because of licensing and it would require people to buy the game again.

> Yeah you're right maybe we should wait a couple of months before we do that

We're not gonna drop the RE2 or RE3 config. There's simply no reason to do that.

PR #3550, opened 2023-07-06
[d3d11] Shared texture flags handling fixes

gofman:
Fixes Night Security (steam id 2367230) crash upon reaching 2nd floor during the first mission (soon after beginning of the game, at "1F" to "2F" level transition). The crash is due to shared texture import failure which is due to the mess between handle types (wrongly determined from flags). The texture is created with (D3D11_RESOURCE_MISC_SHARED | D3D11_RESOURCE_MISC_SHARED_NTHANDLE) which looks confusing from the point of view of currently present logic but actually this is the only way one can create a texture with NT shared handle without using D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX (see below).

Only patch 2 is strictly need for the game, while patch 1 fixes handle (and the shared memory under it) possible leak and the rest are straightening the thing up by adding validations (also covered by the tests) and giving some sense to the changed flag handling.

Here is the test I made which covers all of this behaviour: https://gitlab.winehq.org/wine/wine/-/merge_requests/3247
Attached is the diff on top of the referenced MR with the test which removes fixed todo's (and adds one for NT kernel object name), as well as leaving only the related test to run.

To summarize the test essentials (for some reason not clearly covered in MS docs) for easier understanding what the patches do:
- D3D11_RESOURCE_MISC_SHARED_NTHANDLE can't be used alone, it is actually modifier to D3D11_RESOURCE_MISC_SHARED or D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX;
- D3D11_RESOURCE_MISC_SHARED and D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX can't be used together;
- D3D11_RESOURCE_MISC_SHARED or D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX alone means KMT handle is used, D3D11_RESOURCE_MISC_SHARED_NTHANDLE modifier means NT handle is used;
- IDXGIResource::OpenSharedHandle can only open KMT handle, IDXGIResource1::CreateSharedHandle can only return NT handle (using those with texture with non-matching MiscFlags results in a error);
- Similarly, ID3D11Device_OpenSharedResource can only open KMT handle, ID3D11Device1::OpenSharedResource1 can only open NT handle (this part was already working correctly).

There are some returned value fixes which are not strictly related, but since I tested that anyway...

[test.txt](https://github.com/doitsujin/dxvk/files/11962977/test.txt)

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -2309,6 +2309,10 @@ namespace dxvk {
     d3d11Desc.CPUAccessFlags = metadata.CPUAccessFlags;
     d3d11Desc.MiscFlags      = metadata.MiscFlags;
     d3d11Desc.TextureLayout  = metadata.TextureLayout;
+    if ((d3d11Desc.MiscFlags & D3D11_RESOURCE_MISC_SHARED_NTHANDLE) && !(d3d11Desc.MiscFlags & (D3D11_RESOURCE_MISC_SHARED | D3D11_RESOURCE_MISC_SHARED_KEYEDMUTEX))) {
+      Logger::warn("Fixing up wrong MiscFlags");
+      d3d11Desc.MiscFlags |= D3D11_RESOURCE_MISC_SHARED;
+    }
 
     // Only 2D textures may be shared
     try {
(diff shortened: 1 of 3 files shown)


gofman:
I also sent vkd3d-proton PR: https://github.com/HansKristian-Work/vkd3d-proton/pull/1625 which should ideally be synced in some way (as currently vkd3d-proton sets MiscFlags incorrectly in the view of corrected behaviour after the present PR.

gofman:
I added one more commit to make version dependency with vkd3d-proton one way: so dxvk after this PR should be fine with old vkd3d-proton (while vkd3d-proton related change still depends on this dxvk PR)>

gofman:
v2:
 - formatting fixes.

Issue #3116, opened 2022-12-04
[d3d9] Secret World Legends launcher


Blisto91:
Launcher UI elements are invisible when used together with dxvk both on Linux and Windows. This doesn't happen with wined3d.
The issue can be worked around by setting `d3d9.shaderModel = 2` tho i don't think it is a partial presentation issue as it doesn't behave how they usually do and there we set SM to 1.

<details>
  <summary>Screenshots without and with the option</summary>
  
![Screenshot_20221126_214912](https://user-images.githubusercontent.com/47954800/205483708-3326a012-9ae2-4bca-b6b2-3733c82cf370.png)
![Screenshot_20221126_214838](https://user-images.githubusercontent.com/47954800/205483712-35d32ffb-58e2-4e84-8f7d-957920fc3bec.png)
</details>

### Software information
Secret World Legends

### System information
- GPU: R9 380
- Driver: 22.2.3
- Wine version: Proton 7.0-5, experimental and Windows
- DXVK version: 2.0, 1.10.3 etc

### Apitrace file(s)
https://mega.nz/file/AkRSiLQa#48-9ywH-yrTIKHVHUchH8weThS6Fs4bWWSP6X5ek_Ow
Trace is made on Windows

~apitrace seems to have issues actually replaying the trace as even on native Windows it will render 0 frames.
Will make a issue at the apitrace repo later if this a issue with apitrace itself.~

See first comment 

### Log files
[ClientPatcher_d3d9.log](https://github.com/doitsujin/dxvk/files/10148371/ClientPatcher_d3d9.log)
[ClientPatcher_d3d11.log](https://github.com/doitsujin/dxvk/files/10148372/ClientPatcher_d3d11.log)
[ClientPatcher_dxgi.log](https://github.com/doitsujin/dxvk/files/10148373/ClientPatcher_dxgi.log)

Blisto91:
Forgot to update the issue.
The recent partial presentation work sadly didn't help the issue so the `d3d9.shaderModel = 2` workaround have been kept for the game. This issue will remain open tho so it isn't forgotten.

PR #3490, opened 2023-06-13
Reset cleanup & bugfix

K0bin:
Fixes #3489

I didn't intend #3486 to get merged yet. The plan was to get a review (which is why it wasn't a draft) and do more testing. Anyway, shit happens, lets fix it.

PR #3486, opened 2023-06-09
[d3d9] Reject Reset if there's any remaining DEFAULT resources

K0bin:
Fixes #3472

PR #3564, opened 2023-07-15
[util] Remove Secret World Legends config

Blisto91:
Works out of the box now since either https://github.com/doitsujin/dxvk/pull/3486 or https://github.com/doitsujin/dxvk/pull/3490

Closes https://github.com/doitsujin/dxvk/issues/3116


K0bin:
Nice, I'm always happy to remove app-hacks. :)

PR #3565, opened 2023-07-16
[d3d8] Implement Dref scaling and fixed-function depth textures

AlpyneDreams:
Part of #3411, but not a requirement for #3411 to merge.

- Some early D3D8 games expect Dref to be on the range of [0..2^bitDepth - 1], so this adds a config option for DXSO and D3D9 fixed function to scale it back down to [0..1]. [Source](http://www.jiri-dvorak.cz/scellpt/)
- D3D9 tests confirm that you can bind a depth texture to fixed-function, so this PR also enables that

More info:
- http://www.jiri-dvorak.cz/scellpt/
- AlpyneDreams/d8vk#65
- AlpyneDreams/d8vk#69

Simplified from #3536 since we probably don't need additional shadow filtering based on videos I can find of period hardware. Only dgVoodoo2 does extra shadow filtering as far as I'm aware.


Blisto91:
The build error is known and have been fixed with this PR https://github.com/doitsujin/dxvk/pull/3556

src/d3d9/d3d9_device.cpp
```diff
@@ -7145,10 +7145,7 @@ namespace dxvk {
         stage.Projected      = (ttff & D3DTTFF_PROJECTED) ? 1      : 0;
         stage.ProjectedCount = (ttff & D3DTTFF_PROJECTED) ? count  : 0;
 
-        // [D3D8] Only use Dref sampling for games that could
-        // bind depth textures to fixed function shaders.
-        if (m_dxsoOptions.drefScaling)
-          stage.SampleDref = (m_depthTextures & (1 << idx)) != 0;
+        stage.SampleDref = (m_depthTextures & (1 << idx)) != 0;
```
doitsujin, line 7148: is there a reason why this is a separate commit and not just done by default?

PR #3570, opened 2023-07-20
Make driver requirements slightly more visible

Blisto91:
Being more direct here might lead to less confusion and questions. Also moved driver note in readme up into the How to use section so it's near the top and more visible.


pchome:
idk, add some badges at the top
[c++17][vulkan 1.3][linux][mingw 10][...] only!

Issue #2894, opened 2022-09-07
[feature] able to define configuration properties through an environment variable


ranplayer:
It would be useful if we were able to define the configuration properties through an environment variable (e.g: `DXVK_CONFIG=dxgi.syncInterval=1,d3d11.samplerAnisotropy=16`). 

It would be easier and faster for testing and customizing than creating a `dxvk.conf` file for every game. Also for integrating with GUIs (Steam, Lutris, etc...)

pchome:
`alias dxvk='echo -e ${DXVK_CONFIG//,/\\n} > /tmp/dxvk.conf; $@'`

```
export DXVK_CONFIG=dxgi.syncInterval=1,d3d11.samplerAnisotropy=16
export DXVK_CONFIG_FILE=/tmp/dxvk.conf
dxvk less ${DXVK_CONFIG_FILE}
```
:man_shrugging:

doitsujin:
This was actually planned and I just kind of forgot about it.

PR #3581, opened 2023-07-25
[util] Add DXVK_CONFIG to define additional options

pchome:
Lazy-ass implementation

* <s>`#` used as separator</s>
* `;` used as separator
* options will override/extend those in dxvk.conf file
* <s>`DXVK_CONFIG="dxvk.hud=full#dxvk.blah = blah;blah:blah,#"`</s>
* `DXVK_CONFIG="dxvk.hud = full; dxvk.blah=blah"`

Note:
* `DXVK_HUD` variable still have higher priority than `dxvk.conf` and `DXVK_CONFIG`
* `#` act as comment now (same as in `dxvk.conf`)
     e.g. `DXVK_CONFIG="#Show compiler; dxvk.hud = compiler; #dxvk.blah=blah"`
    Lets call it feature.

Closes: #2894

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1090,6 +1090,7 @@ namespace dxvk {
 
     // Load either $DXVK_CONFIG_FILE or $PWD/dxvk.conf
     std::string filePath = env::getEnvVar("DXVK_CONFIG_FILE");
+    std::string confLine = env::getEnvVar("DXVK_CONFIG");
 
     if (filePath == "")
       filePath = "dxvk.conf";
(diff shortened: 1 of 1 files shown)


src/util/config/config.cpp
```diff
@@ -1090,30 +1090,42 @@ namespace dxvk {
...
+      while (std::getline(stream, line))
+        parseUserConfigLine(config, ctx, line);
+    }
+
+    if (!confLine.empty()) {
+      // Inform the user that we parsing config from environment, might
+      // help when debugging configuration issues
+      Logger::info(str::format("Found config env: ", confLine));
+
+      for(auto l : str::split(confLine, "#"))
```
doitsujin, line 1125: why '#' specifically? that's such an odd choice when everything else either uses `;` or `:`, and much like `;` it hard-requires using quotes on the string as well so i don't really see the benefit here.

pchome: Most standard separators already taken by some options, I saw there was `,` and maybe `;` in upcoming dx8 implementation. `:` already used by `d3d9.forceAspectRatio`.

Could be whitespace, but copy-pasting  something from dxvk.conf or web will not work, e.g `DXVK_CONFIG="dxvk.hud = full"` "user friendly" _vs_ `DXVK_CONFIG="dxvk.hud=full"` with strict (whitespace separator) variant.

Just random choice, to be honest.

doitsujin: I think `;` should be reserved as a separator and not be allowed in option values then. I'd rather change this on the d3d8 side than introduce a weird separator here.

pchome: fine by me, changed it to `;`

WinterSnowfall: d8vk only uses `,` and `:` AFAIK, so `; ` should be perfectly safe to use here.

PR #3585, opened 2023-07-27
Add CHAR and PCSTR to windows_base.h

expipiplus1:


include/native/windows/windows_base.h
```diff
@@ -31,6 +31,9 @@ typedef wchar_t WCHAR;
 typedef WCHAR *NWPSTR, *LPWSTR, *PWSTR;
 typedef unsigned char UCHAR, *PUCHAR;
 
+typedef char CHAR;
+typedef const CHAR *PCSTR;
```
doitsujin, line 35: do we need LPCSTR or is that already defined somewhere?

expipiplus1: I've added it :)

PR #3590, opened 2023-07-27
[d3d11] 32-bit DXGI stdcall fixup

Kaldaien:
After long deliberation, I believe that loading DXVK's DXGI DLL manually is the only practical solution for 32-bit MinGW-built versions of DXVK.

Given that Proton ships DLLs built with MinGW, addressing this is critical for software compatibility. I have explained what is going on in comments and how mods can make themselves compatible with older versions of Proton/DXVK prior to these changes being integrated.

I think the case where "CreateDXGIFactory1" is not exported by a mod DLL is unimportant because Wine would normally crash in that situation anyway. Nonetheless, I opted to return `E_NOTIMPL` for completeness sake.

### DIFF
--- a/src/d3d11/d3d11_main.cpp
+++ b/src/d3d11/d3d11_main.cpp
@@ -14,7 +14,39 @@ namespace dxvk {
   
 extern "C" {
   using namespace dxvk;
-  
+
+  //
+  // Workaround MinGW symbol name issues that prevent using DXVK in Wine with a 32-bit
+  // mod that works by proxying dxgi.dll.
+  //
+  // If a linker-dependency on DXVK's dxgi.dll were used, then it would require DXGI.dll
+  // to export "CreateDXGIFactory1@8" and Windows mod DLLs do not do this.
+  //
+  //  Mods can make themselves compatible with old versions of DXVK by adding this code:
+  // 
+  //   #ifdef _M_IX86
+  //   #pragma comment (linker, "/export:CreateDXGIFactory1@8=_CreateDXGIFactory1@8")
+  //   #endif
+  //
+  // To support unaltered 32-bit game mods, DXVK must load the DLL manually...
+  //
+  HRESULT __stdcall CreateDXGIFactory1_Import(REFIID riid, void **ppFactory) {
+#ifdef _M_IX86
+    using CreateDXGIFactory1_pfn = HRESULT (WINAPI*)(REFIID,void**);
+          CreateDXGIFactory1_pfn
+       DllCreateDxgiFactory1 =
+         (CreateDXGIFactory1_pfn)GetProcAddress(LoadLibraryA("dxgi.dll"),
+         "CreateDXGIFactory1");
+
+    if (DllCreateDxgiFactory1 != nullptr)
+      return DllCreateDxgiFactory1(riid, ppFactory);
+
+    return E_NOTIMPL;
+#else
+    return CreateDXGIFactory1(riid,ppFactory);
+#endif
+  }
+
   DLLEXPORT HRESULT __stdcall D3D11CoreCreateDevice(
           IDXGIFactory*       pFactory,
           IDXGIAdapter*       pAdapter,
@@ -143,7 +175,7 @@ extern "C" {
         Logger::warn("D3D11CreateDevice: Unsupported driver type");
       
       // We'll use the first adapter returned by a DXGI factory
-      hr = CreateDXGIFactory1(__uuidof(IDXGIFactory), reinterpret_cast<void**>(&dxgiFactory));
+      hr = CreateDXGIFactory1_Import(__uuidof(IDXGIFactory), reinterpret_cast<void**>(&dxgiFactory));
 
       if (FAILED(hr)) {
         Logger::err("D3D11CreateDevice: Failed to create a DXGI factory");


Joshua-Ashton:
Does the native d3d11 DLL link against dxgi in the import DLL or LoadLibrary?

pchome:
Maybe enable by default and drop `dxgi_dep` from `d3d11/meson.build`?

Did a quick test on `d3d11-triangle.exe` with this patch and dropped `dxgi_dep` and seems build/work ok.
Also e.g. https://github.com/fholger/vrperfkit previously failed with error
`wine: Call from 7EF6B971 to unimplemented function DXGI.DLL.CreateDXGIFactory1@8, aborting`
now it have something related in the log file
```
11:59:58 [296] d3d11.dll is loaded in the process, installing hooks...
11:59:58 [296] Installing hook for D3D11CreateDevice from 6A568B9F to 0173B960
11:59:58 [296] Installing hook for D3D11CreateDeviceAndSwapChain from 6A568C03 to 0173BD80
11:59:58 [296] (DEBUG) Redirecting CreateDXGIFactory1 to system
11:59:58 [296] Loading DLL at "C:\\windows\\system32\\dxgi.dll"
11:59:58 [296] (DEBUG) LoadLibraryW(C:\windows\system32\dxgi.dll)
11:59:58 [296] (DEBUG) LoadLibraryA(winevulkan.dll)
12:00:11 [296] Shutting down
```
_(tested DLL loading only)_

doitsujin:
> Does the native d3d11 DLL link against dxgi in the import DLL or LoadLibrary?

Yeah we should definitely test this.

Even if DXGI does get statically loaded however, we can still just do GetModuleHandle+GetProcAddress to load the correct symbol.

src/d3d11/d3d11_main.cpp
```diff
@@ -14,7 +14,39 @@ namespace dxvk {
...
+  //   #endif
+  //
+  // To support unaltered 32-bit game mods, DXVK must load the DLL manually...
+  //
+  HRESULT __stdcall CreateDXGIFactory1_Import(REFIID riid, void **ppFactory) {
+#ifdef _M_IX86
+    using CreateDXGIFactory1_pfn = HRESULT (WINAPI*)(REFIID,void**);
+          CreateDXGIFactory1_pfn
+       DllCreateDxgiFactory1 =
+         (CreateDXGIFactory1_pfn)GetProcAddress(LoadLibraryA("dxgi.dll"),
```
doitsujin, line 38: Could we just do GetModuleHandle to not leak a reference?

Probably not a big deal since ideally this should be cleaned up when unloading d3d11.dll anyway, but you never know.

src/d3d11/d3d11_main.cpp
```diff
@@ -14,7 +14,39 @@ namespace dxvk {
...
+  //  Mods can make themselves compatible with old versions of DXVK by adding this code:
+  // 
+  //   #ifdef _M_IX86
+  //   #pragma comment (linker, "/export:CreateDXGIFactory1@8=_CreateDXGIFactory1@8")
+  //   #endif
+  //
+  // To support unaltered 32-bit game mods, DXVK must load the DLL manually...
+  //
+  HRESULT __stdcall CreateDXGIFactory1_Import(REFIID riid, void **ppFactory) {
+#ifdef _M_IX86
```
doitsujin, line 34: I'd rather just have `#ifdef _WIN32` here and only use the direct call for native builds. There shouldn't be any harm in doing this on 64-bit Windows builds, but if for whatever reason we ever run into an issue here i'd rather not have two different code paths.

doitsujin (changes requested):
I'm pretty much fine with this, just some small nits.

Just confirmed that loading d3d11.dll pulls in dxgi.dll on Windows as well so @pchome 's suggestion won't work.

src/d3d11/d3d11_main.cpp
Kaldaien: GetModuleHandle would not handle the case where `d3d11.dll` is manually loaded by an application and the application has not loaded `dxgi.dll` yet.

`GetModuleHandle (...)` would only work if D3D11.dll has a linker dependency on dxgi.dll, and that's what's trying to be avoided here. `LoadLibrary` is the only way to replicate the normal DLL loader behavior where DXVK's d3d11.dll loads dxgi.dll automatically.

I'm not sure how this would constitute a reference leak? If we went the linker dependency route, d3d11.dll would be holding a reference to dxgi.dll the entire time it's loaded, so kind of status quo. I guess if d3d11.dll is expected to be unloaded while the program continues running, then it could be a problem, but does that ever actually happen in the real-world?

Kaldaien:
> Does the native d3d11 DLL link against dxgi in the import DLL or LoadLibrary?

It imports dxgi.dll

image

src/d3d11/d3d11_main.cpp
Kaldaien: Fair. I was more interested in just applying the fix to 32-bit builds, since they're the ones affected... but that wasn't accounting for native builds that are also 32-bit but not Windows.

src/d3d11/d3d11_main.cpp
doitsujin: Hm, yeah, if we don't get the DLL loaded at all then that won't work.

src/d3d11/d3d11_main.cpp
doitsujin: nevermind, see above.

pchome:
I don't remember how exactly this linking is different in meson for native/winelib/mingw but maybe it has something to do with `dxgi_dep` (`link_with`) vs `lib_dxgi` (-ldxgi)?
https://github.com/doitsujin/dxvk/commit/baba2e3c093553797688eb2aea98d934d6e0a60a (#2853)
https://github.com/doitsujin/dxvk/commit/88cda2074630ca2705ac9ccd419455e46a2b26f7 (#501)

Just guessing, but maybe some combination of `lib_dxgi` and this PR will work as expected.

PR #3591, opened 2023-07-29
[build] Build using SSSE3/SSE4/4.1/4.2

WinterSnowfall:
Would like to start by saying I don't have a horse in the race and mostly want to raise this PR as a discussion point, however:
- We've reached the (in)glorious year 2023 AD now
- It might not impact DXVK performance by a lot, but it's free real estate that might make a difference on some newer "potato" CPUs (low end mobile and the like)
- DXVK isn't shy about bumping Vulkan requirements, so should we be lenient on (now ancient) CPUs?
- No offense, but people with a Vulkan 1.3 capable GPU and a Phenom can probably compile dxvk themselves without these flags and let the rest of us live in the present

Also, I think @Joshua-Ashton wanted to do this at some point last year but there was some objection...

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -11,6 +11,10 @@ compiler_args = [
   '-msse',
   '-msse2',
   '-msse3',
+  '-mssse3',
+  '-msse4',
+  '-msse4.1',
+  '-msse4.2',
   '-mfpmath=sse',
   '-Wimplicit-fallthrough',
   # gcc


pchome:
This will affect a lot of people.
IIRC the game about potato now can't be played on potato because of this (Portal, dxvk-native branch).
TF2 team was asked to revert similar changes and reverted ...

Now it will be default for all proton/dxvk-native users in Steam. Will break for a lot of people, for what?
Any benchmarks?

_I always have mixed fillings about such "dropping support" things, especially in projects where actually potatoes are ones who really benefit._
_Also nothing stops you from building with `-march=native` compiler flag in this case, or this involves YOU in "compile dxvk themselves"?_

WinterSnowfall:
> This will affect a lot of people. 

You are asking for performance numbers (which, admittedly, I don't have), but aren't giving any numbers yourself. How many people do you wager would be affected by this? Maybe 1-2% of the overall Steam Linux userbase? Why can't people simply stick to older versions of dxvk on older hardware?

> I always have mixed fillings about such "dropping support" things, especially in projects where actually potatoes are ones who really benefit.

We're not discussing potatoes here, but rather the age of said potatoes. Reasonably new potatoes will work just fine, mind you. 

This change will ultimately come, either now or in a few years. You can't expect older hardware to be supported indefinitely, and if you do, I don't think it's too much to ask to go the extra mile and make modern stuff backward compatible by compiling it yourself, should you want to keep your hardware around forever (it's your choice after all).

The flags that I've added are supported even on Bulldozer (2011) processors. I'm not a decision factor in this change, of course, but I don't recall the support for aging hardware ever being an aim of dxvk.

And yes, anyone can use `-march=native` to compile it, but that's not the point here. The point is IMHO these changes are long overdue in this day and age. The vast majority of people won't be affected, only a small niche with very old hardware, but they will of course be the most vocal about it.

I'm fine with this being rejected (yet again) now, but it shouldn't stay like this forever. A modern codebase isn't a museum. My 2 cents.

pchome:
> I'm fine with this being rejected (yet again) now, but it shouldn't stay like this forever. A modern codebase isn't a museum. My 2 cents.

I'm fine with this being merged, go ahead.

_P.S. Why stop with sse4.2? Go avx ;)_
https://github.com/ValveSoftware/Dota-2/issues/2003#issuecomment-997492117

doitsujin:
We've talked about this in the past.

I strongly agree with the point that dropping support for *any* kind of hardware should only be done if there are tangible benefits to doing so. This was the case when we dropped support for Kepler / GCN 1.0 GPUs with the 2.0 release since moving to Vulkan 1.3 (specifically, dynamic rendering) was essential to implement GPL.

In this case however, I don't see the point. The percentage of users affected might be small, but the benefits of building DXVK even with `-march=native` are virtually nonexistent. Even more so if we're limiting ourselves to SSE 4.2 - i could see *some* arguments being made for AVX since we have a custom memcmp sitting around somewhere that is used in some potentially hot code paths, but even then we're talking about code that maybe eats 1% of the execution time on a thread that may not even be the primary bottleneck.

WinterSnowfall:
> We've talked about this in the past.

Ok, guess I wasn't around when this was discussed. If it really brings no benefit then I agree there's no point. What IS then the point of stopping at SSE3 though? Does it help with anything at all and shouldn't SSE2 be enough (vkd3d-proton only uses up to SSE2 apparently).

pchome:
@WinterSnowfall 
> What IS then the point of stopping at SSE3 though?

Using github blame feature I dug a little history:
https://github.com/doitsujin/dxvk/pull/676
https://github.com/doitsujin/dxvk/commit/6fb09cb9fceaf9890ec83ded507ddae22c3b6a40
https://github.com/doitsujin/dxvk/commit/7d673063f5323f359cc2bbd639a8019fe38d3820
https://github.com/doitsujin/dxvk/commit/f8c7e29d6fdd06073be25bffca60464ce0e9cf06
https://github.com/doitsujin/dxvk/commit/403af8820bf00d3f4372c8db8a34fe85186c6f5a

it was 2021, so ...

If steam/proton explicitly or implicitly require SSE2 minimum then it's better to follow the requirement (for consistency), IMHO.
_Yes, some games in steam have higher requirements, AFAIK._

WinterSnowfall:
> Any benchmarks?

Not trying to beat a dead horse, but I was curios to see for myself if there's any benefit, so here's some "Ashes of the Benchmarkality" results below:

<details>
  <summary>Everything up to SSE3 (dxvk default)</summary>

SSE3

</details>
<details>
  <summary>Everything up to SSE4.2</summary>

SSE42

</details>
<details>
  <summary>Everything up to AVX2</summary>

AVX2

</details>

There are _some_ minor performance deltas that pretty much stay the same (done 3 sets of tests, the above is the final run), but yeah, nothing to write home about.

Issue #3485, opened 2023-06-09
Bug in Conflict: Vietnam


HQSantosilva:
I've tried to use the software in this game to check if it can solve the shaking camera problem this game has with directx 11. It worked fine, but every time you try to walk uphill or down through a slope, the character gets stuck in the air, or simply can't move
![bug](https://github.com/doitsujin/dxvk/assets/110571779/bb2947a0-c78c-4ee1-a164-dad3090ec7a0)

Blisto91:
> directx 11.

The game looks to be DirectX 9 looking at gaming wiki?

We would really like some additional info as this isnt alot.
I'm guessing this doesn't without dxvk? Is the fps unlocked or locked?

K0bin:
> the character gets stuck in the air, or simply can't move

That's almost certainly not a DXVK bug. DXVK has nothing to do with game physics or movement.

HQSantosilva:
> > directx 11.
> 
> The game looks to be DirectX 9 looking at gaming wiki?
> 
> We would really like some additional info as this isnt alot. Usually as a minimum also include dxvk logs. I'm guessing this doesn't happen without dxvk? Is the fps unlocked or locked?

The installer had a directx9 file with it, so its like the old cd installation games

HQSantosilva:
> 



> > the character gets stuck in the air, or simply can't move
> 
> That's almost certainly not a DXVK bug. DXVK has nothing to do with game physics or movement.

It's wierd because this bug only happened when i tried to run with DXVK, i played the game for a while and it never happened before.

Blisto91:
The physics can bug out at high fps.

PR #3606, opened 2023-08-08
[util] Limit Conflict Vietnam to 60fps

Blisto91:
Physics can bug out at high fps making the character fly or get stuck when running up or down slopes.

Closes https://github.com/doitsujin/dxvk/issues/3485

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -795,6 +795,10 @@ namespace dxvk {
     { R"(\\IncPC\.exe$)", {{
       { "d3d9.maxFrameRate",                "59" },
     }} },
+    /* Conflict Vietnam                        */
+    { R"(\\Vietnam\.exe$)", {{
+      { "d3d9.maxFrameRate",                "60" },
+    }} },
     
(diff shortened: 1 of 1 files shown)


SuperSamus:
According to [the PCGamingWiki article](https://www.pcgamingwiki.com/wiki/Conflict:_Vietnam), the game is already capped at 60 FPS. Is the wiki wrong?

Blisto91:
Wasn't capped for me when testing at least

SuperSamus:
Edited the article.

> I think it was vsync capped by default maybe

The wiki already said so. I guess whoever wrote the article was testing on a 60 Hz monitor and didn't try forcing a different monitor refresh rate.

Blisto91:
Note I can't remember if it specifically bugged out at 120fps, might have been upwards of 140-150 to get really noticable when I turned off vsync and messed with fps caps. But I choose 60fps as that is what is usually the case with these types of games, I also mostly messed around with it in the beginning so can't tell what above 60 behaves like in other parts of the game.

PR #3607, opened 2023-08-09
[dxgi] Fix behavior of GetWindowAssociation

Etaash-mathamsetty:


doitsujin:
looks good, does anything rely on this?

PR #3612, opened 2023-08-11
[d3d11] Implement synchronization on keyed mutexes

gofman:
That apparently relies on custom winevulkan functions which are currently implemented in Proton Experimental [bleeding-edge].


doitsujin:
Merged Josh's cleanup version of this.

PR #3619, opened 2023-08-14
[meta] Document DXVK_CONFIG in readme

Blisto91:
dxvk.:frog: = True


doitsujin:
I'd drop the note about # and put an actual example with two options in so you can see the `;` separator in action.

Blisto91:
Hope it's fine i also moved the two lines down since the example with two configs might look a bit unwieldy in the middle when it's so long compared to the others.

PR #3625, opened 2023-08-17
[dxgi] Add global HDR interop interface for NVAPI/AGS

Joshua-Ashton:


src/dxgi/dxgi_factory.cpp
```diff
@@ -519,4 +549,15 @@ namespace dxvk {
   }
 
 
+  DXVK_VK_GLOBAL_HDR_STATE DxgiFactory::GlobalHDRState() {
+    std::unique_lock lock(s_globalHDRStateMutex);
+    return s_globalHDRState;
+  }
+
+
+  uint32_t DxgiFactory::GlobalHDRSerial() {
```
doitsujin, line 558: Can we get rid of this function and just work with the struct?

Calling this separately works due to all the locking we have in place, but it a) forces us to lock twice and b) encourages patterns that can easily turn into a race condition if we ever change anything.

doitsujin (changes requested):
Nit.

src/dxgi/dxgi_factory.cpp
Joshua-Ashton: Done.

PR #3630, opened 2023-08-20
[util] Limit fps in Project: Snowblind

Blisto91:
Player movement and animation can bug out at high fps like issues moving around objects and the players head detaching slightly when moving backwards. 
Seems like it also helps some crash issues.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -799,6 +799,10 @@ namespace dxvk {
     { R"(\\Vietnam\.exe$)", {{
       { "d3d9.maxFrameRate",                "60" },
     }} },
+    /* Project: Snowblind                      */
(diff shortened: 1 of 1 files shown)


pchome:
Weird regex. What are actual extensions of the files?
`.SP`, `.MP` and `.exe` ? Or `.SP.exe` etc.?

Not tested this but could be `\\Snowblind(\.SP|\.MP)?\.exe$`  for the later case.

Blisto91:
> Weird regex. What are actual extensions of the files? `.SP`, `.MP` and `.exe` ? Or `.SP.exe` etc.?

The former. The multiplayer and single player executable have `.MP` and `.SP` respectively and the launcher `.exe`.
The launcher could probably have been left out but kept it just so the fps didn't go crazy there for no reason when vsync is off.

PR #3637, opened 2023-08-24
[dxgi] Allow HDR on UE4/D3D11 when NVAPI is enabled

jp7677:
DXVK-NVAPI has the HDR bits since https://github.com/jp7677/dxvk-nvapi/commit/e59101009bf51fcc5ecf159d97788ab0546ac4b3

Pseudo-tested with Assetto Corsa Competizione (UE4) on a non HDR setup (NVIDIA GPU and non-HDR monitor), of course no real HDR, but also no crashes and enabling HDR in-game does have effect on colors.

### DIFF
--- a/src/dxgi/dxgi_options.cpp
+++ b/src/dxgi/dxgi_options.cpp
@@ -27,6 +27,11 @@ namespace dxvk {
   }
 
 
+  static bool isNvapiEnabled() {
+    return env::getEnvVar("DXVK_ENABLE_NVAPI") == "1";
+  }
+
+
   static bool isHDRDisallowed() {
 #ifdef _WIN32
     // Unreal Engine 4 titles use AGS/NVAPI to try and enable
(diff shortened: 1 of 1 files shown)


src/dxgi/dxgi_options.cpp
```diff
@@ -92,7 +93,7 @@ namespace dxvk {
     this->hideIntelGpu = config.getOption<Tristate>("dxgi.hideIntelGpu", Tristate::Auto) == Tristate::True;
 
     this->enableHDR = config.getOption<bool>("dxgi.enableHDR", env::getEnvVar("DXVK_HDR") == "1");
-    if (this->enableHDR && isHDRDisallowed()) {
+    if (this->enableHDR && isHDRDisallowed() && !nvapiEnabled) {
```
jp7677, line 96: Extracting ` env::getEnvVar("DXVK_ENABLE_NVAPI") == "1"` into a function and already checking within `isHDRDisallowed()` might improve readability. Please let me know what you prefer.

PR #3648, opened 2023-08-30
[d3d9] Adjust the MaxActiveLights capability

WinterSnowfall:
Probably won't affect much, but in practice the number of SWVP active lights shouldn't be limited to 8, and native drivers report either 255 or UINT_MAX here. Also bumped the HWVP value to 10 while at it, as it's what both native Nvidia and AMD report.

Native Nvidia:
`  ~ MaxActiveLights: 8 (I*), 255 (SWVP), 10 (HWVP)`

Native AMD:
`  ~ MaxActiveLights: 8 (I*), 4294967295 (SWVP), 10 (HWVP)`

\* I is what IDirect3D->GetDeviceCaps() reports, while SWVP/HWVP values are reported on IDirect3DDevice->GetDeviceCaps()

There might be some games that end up using less lights then they would on native because of the current status quo, but that's really hard to pin down (dxvk isn't actively enforcing any of these limits anyway, so it's up to game code). 

Also, d8vk is far more likely to be impacted due to the increased prevalence of SWVP.

### DIFF
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -523,7 +523,7 @@ namespace dxvk {
                                      | D3DVTXPCAPS_TEXGEN_SPHEREMAP
                                   /* | D3DVTXPCAPS_NO_TEXGEN_NONLOCALVIEWER*/;
     // Max Active Lights
-    pCaps->MaxActiveLights           = caps::MaxEnabledLights;
+    pCaps->MaxActiveLights           = 8;
     // Max User Clip Planes
     pCaps->MaxUserClipPlanes         = MaxClipPlanes;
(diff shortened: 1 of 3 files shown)


WinterSnowfall:
In retrospect, we're using the caps::MaxEnabledLights limit to allocate the SWVP constants set. Not sure if it makes any sense to bump it up to 255 just for allocation?

I believe there was some discussion around not allocating the entire SWVP constant set when in HWVP mode anyway, which sounds like a good idea for the above.

WinterSnowfall:
Nevermind. dxvk can't cater for more than 8 active lights in d3d9 at the moment and there's no reason to worry about SWVP too much, I guess. Definitely no sense in reporting what isn't actually supported, so closing.

PR #3652, opened 2023-09-03
[util] Escape .

Margen67:


doitsujin:
what's the purpose of these changes?

I see like one line where we have an actual bug but other than that these seem equivalent to what we already have.,

Margen67:
> what's the purpose of these changes?
> 
> I see like one line where we have an actual bug but other than that these seem equivalent to what we already have.,

I was trying to shorten them.
I've pushed only bugfixes (aside from whitespace removal) so this can be merged.

PR #3656, opened 2023-09-07
A few more WinDef types in windows_base.h

expipiplus1:
These specific ones are used in MS's d3dx12 headers


expipiplus1:
Thanks :)

@doitsujin, any chance you could take a look at https://github.com/doitsujin/dxvk/pull/3365 this also?

PR #3663, opened 2023-09-17
[d3d9] Mark presenter for recreation on device reset with deferSurfaceCreation

WinterSnowfall:
I don't particularly like having to do this either, so I've hid it under deferSurfaceCreation for now, as it seems to work around the same sort of problem, albeit at later stages in a swapchain's lifecycle.

Fixes Scrapland (Remastered), which apart from the intro video plays videos using ddraw (or maybe d3d7, haven't *really* checked, but it's not d3d8) when starting a new game or in between some levels, which will end up in a black screen with d8vk/dxvk unless we mark the presenter for recreation. Latest WineD3D somehow handles this sort of multiple swapchains on the same window situation just fine, although Scrapland was broken for a while there too with the same behavior (black screen after video playblack apart from the intro).

The first d3d8 thing the game calls after it's done with playing videos is a device Reset(), which is what gave me this rather awful, but working idea.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -482,6 +482,9 @@ namespace dxvk {
     Flush();
     SynchronizeCsThread(DxvkCsThread::SynchronizeAll);
 
+    if (m_d3d9Options.deferSurfaceCreation)
+      m_deviceHasBeenReset = true;
+
     return D3D_OK;
   }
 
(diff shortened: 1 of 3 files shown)


WinterSnowfall:
In theory waiting for more testing from @Blisto91, but unless there are any objections I think it's good enough.

There's no point in enabling it globally, and I don't see much of a point in creating a separate config option since at the moment there's only one known game to be affected by it, namely Scrapland Remastered, which needs deferSurfaceCreation anyway.

There could potentially be other games that juggle with ddraw and d3d8/9 swapchains on the same window during gameplay and in consequence hit such issues, but we can always reuse the deferSurfaceCreation option in these cases as well.

Blisto91:
d3d9 games the config affects have been found to still act fine with this PR.

PR #3664, opened 2023-09-18
[d3d9] Only set initial NeedsUpload for D3DPOOL_MANAGED textures

K0bin:
IMO this is the better fix for #3662.

Right now, binding a system mem / scratch texture will make us call FlushImage for that texture which crashes because there is no image. On top of that, we may end up uploading recently created D3DPOOL_DEFAULT textures which means we basically clear it again.


Joshua-Ashton:
Do we need to do this for SYSTEMMEM what is the behaviour there?

PR #3690, opened 2023-10-14
Add NV low latency support

esullivan-nvidia:
This pull request implements a new device interface called ID3DLowLatencyDevice using the VK_NV_low_latency2 extension. The purpose of this interface, is to give dxvk-nvapi a way to implement the nvapi Reflex interface.

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -15,6 +15,7 @@
 #include "d3d11_device.h"
 #include "d3d11_fence.h"
 #include "d3d11_input_layout.h"
+#include "d3d11_interfaces.h"
 #include "d3d11_interop.h"
 #include "d3d11_query.h"
 #include "d3d11_resource.h"
@@ -2469,12 +2470,14 @@ namespace dxvk {
         return deviceFeatures.nvxBinaryImport
             && deviceFeatures.vk12.bufferDeviceAddress;
 
+      case D3D11_VK_NV_LOW_LATENCY_2:
+        return deviceFeatures.nvLowLatency2;
+
       default:
         return false;
     }
   }
-  
-  
+
   bool STDMETHODCALLTYPE D3D11DeviceExt::GetCudaTextureObjectNVX(uint32_t srvDriverHandle, uint32_t samplerDriverHandle, uint32_t* pCudaTextureHandle) {
     ID3D11ShaderResourceView* srv = HandleToSrvNVX(srvDriverHandle);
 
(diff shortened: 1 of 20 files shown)


adamnejm:
Great contribution.
Please also consider distributing the **Nvidia Reflex SDK** for **Linux**, so native titles such as **Counter-Strike 2** can implement it :)

src/d3d11/d3d11_swapchain.cpp
```diff
@@ -410,9 +438,11 @@ namespace dxvk {
           uint32_t                Repeat) {
     auto lock = pContext->LockContext();
 
-    // Bump frame ID as necessary
-    if (!Repeat)
-      m_frameId += 1;
+    if (!Repeat) {
+      m_frameId = m_presenter->lowLatencyEnabled() ?
+        m_device->getLatencyMarkers().present :
```
ishitatsuyuki, line 443: Would this work with games that are not using latency markers, but only the sleep API? (e.g. Apex Legends)

adamnv:
Support for the low-latency extensions just got released in the [545.23.06 NVIDIA driver](https://www.nvidia.com/download/driverResults.aspx/213007/en-uk/) so this is usable/testable now.  (Propagating this note to all three project PRs, apologies for any duplication! :smile_cat:  )

src/dxvk/dxvk_device.cpp
```diff
@@ -274,6 +275,7 @@ namespace dxvk {
           DxvkSubmitStatus*         status) {
     DxvkSubmitInfo submitInfo = { };
     submitInfo.cmdList = commandList;
+    submitInfo.frameId = m_latencyMarkers.render;
```
ishitatsuyuki, line 278: Note that resource initialization (zeroing) submissions can happen independently from the immediate context or CS thread. `submitCommandList` is used by every submission including init submissions, and this can cause slightly awkward consequences such as submission tied with a particular frame ID after present.

It might be better to exclude InitContext from low latency tracking completely to avoid edge cases.

src/d3d11/d3d11_interfaces.h
```diff
@@ -114,6 +142,33 @@ ID3D11VkExtDevice1 : public ID3D11VkExtDevice {
...
 
+/**
+ * \brief Extended extended D3D11 device
+ * 
+ * Introduces methods to get virtual addresses and driver
+ * handles for resources, and create and destroy objects
+ * for D3D11-Cuda interop.
+ */
+MIDL_INTERFACE("f3112584-41f9-348d-a59b-00b7e1d285d6")
+ID3DLowLatencyDevice : public IUnknown {
```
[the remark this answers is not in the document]
Saancreed: If it did, it would be much harder to [reuse the same interface in vkd3d-proton](https://github.com/HansKristian-Work/vkd3d-proton/pull/1739#discussion_r1373304094). This way we can query for only one interface in dxvk-nvapi without checking if we were passed D3D11 or D3D12 device.

But the comment above this inteface seems incorrect / copy-pasted from existing DeviceExt and should probably be changed.

src/dxvk/dxvk_presenter.cpp
```diff
@@ -48,6 +57,7 @@ namespace dxvk {
 
 
   VkResult Presenter::acquireNextImage(PresenterSync& sync, uint32_t& index) {
+    std::lock_guard<dxvk::mutex> lock(m_lowLatencyMutex);
```
ishitatsuyuki, line 60: AcquireNextImage can block. I think holding a mutex for the entire duration of acquire can cause potential trouble.

src/dxvk/dxvk_presenter.cpp
```diff
@@ -68,11 +78,13 @@ namespace dxvk {
   VkResult Presenter::presentImage(
           VkPresentModeKHR  mode,
           uint64_t          frameId) {
+    std::lock_guard<dxvk::mutex> lock(m_lowLatencyMutex);
```
ishitatsuyuki, line 81: QueuePresent is also allowed to block, although this might be less of a concern if the driver is implemented in a way that doesn't block.

src/dxvk/dxvk_presenter.cpp
ishitatsuyuki: I see the same concern was also raised in vkd3d-proton review, so I suppose this can be addressed in a similar way (only hold the lock in case of recreation).

ishitatsuyuki: With the locking in acquire and present removed my layer-based LL2 implementation works with significantly less stutters — so this is probably that has to be fixed. I think the locks aren't really necessary here anyway since access to swapchain is externally synchronized (which implies it can't be destroyed for the duration of call).

src/dxvk/dxvk_presenter.cpp
```diff
@@ -151,6 +163,8 @@ namespace dxvk {
 
 
   VkResult Presenter::recreateSwapChain(const PresenterDesc& desc) {
+    std::lock_guard<dxvk::mutex> lock(m_lowLatencyMutex);
```
ishitatsuyuki, line 166: `recreateSurface` can also result in the swapchain getting destroyed. I think we currently lack locking there (and this might be a tricky one since `recreateSurface` will keep `m_swapchain` NULL until the next call to `recreateSwapchain`).

src/dxvk/dxvk_presenter.cpp
```diff
@@ -422,6 +452,69 @@ namespace dxvk {
...
+    uint32_t timingCount = 0;
+
+    std::lock_guard<dxvk::mutex> lock(m_lowLatencyMutex);
+    m_vkd->vkGetLatencyTimingsNV(m_vkd->device(), m_swapchain, &timingCount, &markerInfo);
+
+    if (timingCount != 0) {
+      frameReports.resize(timingCount, { VK_STRUCTURE_TYPE_GET_LATENCY_MARKER_INFO_NV });
+      markerInfo.pTimings = frameReports.data();
+
+      m_vkd->vkGetLatencyTimingsNV(m_vkd->device(), m_swapchain, &timingCount, &markerInfo);
```
Saancreed, line 513: I think `VkStructureType` passed to `resize` is wrong here (`VK_STRUCTURE_TYPE_GET_LATENCY_MARKER_INFO_NV` → `VK_STRUCTURE_TYPE_LATENCY_TIMINGS_FRAME_REPORT_NV`)? So together with updated Vulkan headers for spec 2 version of NV_ll2 this should look more like:

```suggestion

    std::lock_guard<dxvk::mutex> lock(m_lowLatencyMutex);
    m_vkd->vkGetLatencyTimingsNV(m_vkd->device(), m_swapchain, &markerInfo);

    if (markerInfo.timingCount != 0) {
      frameReports.resize(markerInfo.timingCount, { VK_STRUCTURE_TYPE_LATENCY_TIMINGS_FRAME_REPORT_NV });
      markerInfo.pTimings = frameReports.data();

      m_vkd->vkGetLatencyTimingsNV(m_vkd->device(), m_swapchain, &markerInfo);
```

src/d3d11/d3d11_device.h
```diff
@@ -428,6 +429,22 @@ namespace dxvk {
 
     bool Is11on12Device() const;
 
+    void AddSwapchain(D3D11SwapChain* swapchain) {
+      m_swapchains.push_back(swapchain);
+    }
+
+    void RemoveSwapchain(D3D11SwapChain* swapchain) {
+      std::remove(m_swapchains.begin(), m_swapchains.end(), swapchain);
```
ishitatsuyuki, line 437:
```suggestion
      m_swapchains.erase(std::remove(m_swapchains.begin(), m_swapchains.end(), swapchain));
```

Found by clang-cl.

src/d3d11/d3d11_device.cpp
```diff
@@ -3021,7 +3149,11 @@ namespace dxvk {
 
       Com<D3D11SwapChain> presenter = new D3D11SwapChain(
         m_container, m_device, pSurfaceFactory, pDesc);
-      
+
+      if (m_device->GetDXVKDevice()->features().nvLowLatency2) {
+        m_device->AddSwapchain(presenter.ref());
```
ishitatsuyuki, line 3154: This seems to miss a matching `RemoveSwapchain` call.

ishitatsuyuki (changes requested):

src/dxvk/dxvk_device.cpp
esullivan-nvidia: Updated to disable passing frame ids at vkQueueSubmit time for work submitted from the initializer context.

src/d3d11/d3d11_device.h
```diff
@@ -428,6 +429,22 @@ namespace dxvk {
...
+    }
+
+    UINT GetSwapchainCount() {
```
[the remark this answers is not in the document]
esullivan-nvidia: Fixed.

src/dxvk/dxvk_presenter.cpp
esullivan-nvidia: The lock should only be taken when the swapchain is destroyed now. So I think this issue can be considered resolved.

src/dxvk/dxvk_presenter.cpp
esullivan-nvidia: Updated this to only take the lock when destroying the swapchain.

src/dxvk/dxvk_presenter.cpp
esullivan-nvidia: Good catch, this should be fixed now.

src/d3d11/d3d11_device.h
esullivan-nvidia: Fixed, thanks for running this change through clang-cl!

src/d3d11/d3d11_swapchain.cpp
esullivan-nvidia: This was fixed in a previous update to this PR to only use the present latency marker if it is valid.

src/dxvk/dxvk_presenter.cpp
esullivan-nvidia: Updated this to only take the lock when destroying the swapchain. Whenever dxvk calls one of the other LL2 entry points it holds the low latency lock, and confirms the swapchain handle is valid. This approach should prevent any usage of the destroyed swapchain handle.

esullivan-nvidia:
Just like I mentioned in the vkd3d-proton PR, thank you for your patience in regards to the slow updates. Feel free to provide any additional feedback, or concerns and I will address them as quickly as possible. I am aware there are a couple of outstanding issues. I will get to those tomorrow. I think most of the major concerns have been resolved though.

src/d3d11/d3d11_swapchain.cpp
```diff
@@ -410,9 +451,11 @@ namespace dxvk {
           uint32_t                Repeat) {
     auto lock = pContext->LockContext();
 
-    // Bump frame ID as necessary
-    if (!Repeat)
-      m_frameId += 1;
+    if (!Repeat) {
+      m_frameId = (m_presenter->lowLatencyEnabled() && m_device->getLatencyMarkers().present) ?
```
ishitatsuyuki, line 455: Two things mandates that m_frameId is monotonically increasing:
- **VK_KHR_present_id**: A non-zero presentId must be greater than any non-zero presentId passed previously by the application for the same swapchain.
- **CallbackFence**: The helper works similarly to a timeline semaphore so the timeline value must be increasing.

I think we need to at least ensure that the value doesn't decrease here. For present IDs, recreating the swapchain will do the job. CallbackFence will need some more modification: I'll describe my idea below.

- Make D3D9Swapchain use CallbackFence and remove Fence; the latter seems to be a superset of the former.
- Move CallbackFence into Presenter, so that it can be easily recreated by the Presenter. Change accesses to e.g. `D3D11Swapchain::m_frameLatencySignal` to something like `m_presenter->frameLatencySignal()`.
- Make `Presenter::destroySwapchain` destroy the signal too, then also recreate the signal in `Presenter::recreateSwapchain`. (`destroySwapchain` waits for the current frame to be signaled.) Add a new parameter to `Presenter::recreateSwapchain` for the frame ID the signal should be initialized with. Also reset `m_lastFrameId` to the specified value.

cc @doitsujin in case they have feedback for the proposed refactor.

ishitatsuyuki:
> We have updated on proton-cachyos the nvidia-reflex patches. One user is following a issue in apex legends. The issue happens as soon Reflex + Boost gets enabled. Reflex only is okay. The error message is: `There was a problem processing game logic `
> 
> I have attached a Screenshot.
> 
> Reflex without boost is working without problems. 545 drivers.

Toggling Reflex causing game logic crash feels pretty unlikely so it might be good to test this multiple times to confirm that it's not a fluke.

If it's reproducible, I think attaching logs with `PROTON_LOG=1` would be useful for confirming or ruling out exceptions within the driver.

ishitatsuyuki:
Please don't make "state of merging" type of comments. The PR will be merged once all the concerns are addressed. Currently, the frame ID interaction with swapchain and CallbackFence needs to be fixed.

ptr1337:
> Please don't make "state of merging" type of comments. The PR will be merged once all the concerns are addressed. Currently, the frame ID interaction with swapchain and CallbackFence needs to be fixed.

We have also a user, which reports problems with this MR, if VRR is used together with Reflex in Overwatch. i can provide later the day some logs from him.

doitsujin:
FWIW one motivation for #4609 is to eventually support this. I haven't looked into supporting low_latency_2 specifically yet, and some problems remain that need to be solved (we'll need to virtualize our internal frame IDs so we can pass LL2 frame IDs to the vulkan swapchain etc; threading is also still a thing but should be much more logical now), but an implementation should be much more viable in the future.

doitsujin:
This is implemented in master now and tested in a few games. If you have any comments, questions or suggestions about the implementation, please file an issue.

PR #3710, opened 2023-11-04
Add config for C&C Generals ZH (solve crashing + performance increase )

anoosi:
Solved the problem of the game crashing when it starts loading saved missions or skirmish, especially for Mods
+ Huge performance increase.
+ Solve Alt tab crashing.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -837,7 +837,6 @@ namespace dxvk {
      * solve loading saved and replay missions  *
      * crash due to limited memory              */
     { R"(\\generals\.exe$)" , {{
-      { "d3d9.maxAvailableMemory",          "6144" },
       { "d3d9.memoryTrackTest",             "True" },
       { "d3d9.textureMemory",               "2048" },
     }} },
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -833,6 +833,14 @@ namespace dxvk {
     { R"(\\SkyDrift\.exe$)" , {{
       { "d3d9.allowDirectBufferMapping",    "False" },
     }} },
+    /* C&C Generals                             *
+     * solve loading saved and replay missions  *
+     * crash due to limited memory              */
+    { R"(\\generals\.exe$)" , {{
+      { "d3d9.maxAvailableMemory",          "6144" },
+      { "d3d9.memoryTrackTest",             "True" },
+      { "d3d9.textureMemory",               "2048" },
+    }} },
 
     /**********************************************/
     /* D3D12 GAMES (vkd3d-proton with dxvk dxgi)  */


WinterSnowfall:
For anyone who might not be aware, this is actually a d3d8 game.

@anoosi `"d3d9.textureMemory", "2048"` seems a bit excessive to say the least... and so does the value for `d3d9.maxAvailableMemory`. How did you come up with these numbers?

anoosi:
> For anyone who might not be aware, this is actually a d3d8 game.
thanks for mention that.
+using d3d8to9
> 
> @anoosi `"d3d9.textureMemory", "2048"` seems a bit excessive to say the least... and so does the value for `d3d9.maxAvailableMemory`. How did you come up with these numbers?
> 
> P.S.: You can use `"d3d9.textureMemory", "0"` to disable the limit entirely, if that's what you're after.
If you play with ai like map 8 players, there will be a large amount of texture that must be loaded, especially if it is a replay file or a saved skirmish, missions,
some mod, contains high resolution textures and the size of files much bigger , also some mods contains high poly  3d objects,
i tested so many times d3d9.maxAvailableMemory = 4GB sometimes crashing if you play mods like CWC with a long session playing.

anoosi:
> For anyone who might not be aware, this is actually a d3d8 game.
> 
> @anoosi `"d3d9.textureMemory", "2048"` seems a bit excessive to say the least... and so does the value for `d3d9.maxAvailableMemory`. How did you come up with these numbers?
> 
> P.S.: You can use `"d3d9.textureMemory", "0"` to disable the limit entirely, if that's what you're after.


If you play with ai like map 8 players, there will be a large amount of texture that must be loaded, especially if it is a replay file or a saved skirmish, missions,
some mod, contains high resolution textures and the size of files much bigger , also some mods contains high poly  3d objects,
i tested so many times d3d9.maxAvailableMemory = 4GB sometimes crashing if you play mods like CWC with a long session playing.

Blisto91:
Which of these variables are you saying help with crashing?
`d3d9.maxAvailableMemory` caps at 4GB which is already default so the value you set doesn't do anything (the value is in megabytes).
If not raising `d3d9.textureMemory` makes the game crash then I'd like you to make an issue.

I assume you are using d8vk or d3d8to9 with it for it to work with dxvk?

anoosi:
> Which of these variables are you saying help with crashing? `d3d9.maxAvailableMemory` caps at 4GB which is already default so the value you set doesn't do anything (the value is in megabytes). If not raising `d3d9.textureMemory` makes the game crash then I'd like you to make an issue.
> 
> I assume you are using d8vk or d3d8to9 with it for it to work with dxvk?

maybe the problem with d3d9.textureMemory only
d8vk and d3d8to9 same problem.

WinterSnowfall:
> `d3d9.maxAvailableMemory` caps at 4GB which is already default so the value 

True, but note it still needs a `"d3d9.memoryTrackTest", "True"` to actually enforce that limit IIRC.

I only have the base game (or the addon, rather) and haven't noticed any problems even on the largest maps, however I don't use any mods, so can't comment on that front.

anoosi:
> > `d3d9.maxAvailableMemory` caps at 4GB which is already default so the value
> 
> True, but note it still needs a `"d3d9.memoryTrackTest", "True"` to actually enforce that limit IIRC.
> 
> I only have the base game (or the addon, rather) and haven't noticed any problems even on the largest maps, however I don't use any mods, so can't comment on that front.

if you start the game like skirmish there is no issue,  but, if you load a replay or saved skirmish or mission it will get crash!
i faced this in the game without mods in sometimes, but it's happen always with mods that contains 3d objects with high resolution textures.
+ i update the configuration (Remove d3d9.maxAvailableMemory)

WinterSnowfall:
> if you start the game like skirmish there is no issue, but, if you load a replay or saved skirmish or mission it will get crash! i faced this in the game without mods in sometimes, but it's happen always with mods that contains 3d objects with high resolution textures.

FWIW, I've tried my best to make the vanilla game (& addon) crash with loading games/replays from large maps, but it never did in my case. Alt+tabbing seems to cause a crash on refocus, but none of the suggesting config options fix it (not that I expected them to).

I also get no performance benefit with any of these options. Again, I've only tested things unmodded, but it would be rather strange if the game's engine would behave differently when modded. The only thing I can think of is that the game's hard limited to 31.2 fps by default, so if any mods change that perhaps some engine bottlenecks become more apparent.

If you mention the game's crashing during long play sessions, are you sure you're using a LAA executable (or Proton, which enables LAA globally)? Issues solved by tweaking d3d9.maxAvailableMemory / d3d9.memoryTrackTest don't usually manifest in time, but are rather more abrupt.

PR #3717, opened 2023-11-15
[d3d11] Fix race in WaitForResource.

yshui:


yshui:
from: https://github.com/ValveSoftware/Proton/issues/1199#issuecomment-1790053082

user observed vrchat crash in d3d11 when system is under load

K0bin (changes requested):
I'd prefer the `LockContext` in DXGIKeyedMutex::ReleaseSync.
But this only works if the game enables D3D11Multithread.

PR #3718, opened 2023-11-15
[d3d11] Lock context in KeyedMutex::ReleaseSync

K0bin:
WaitForResource is supposed to be externally synchronized.

Supersedes #3717

@yshui checked and the VRChat does enable D3D11Multithread. So this is an easy fix.
*If* it's legal to access the KeyedMutex from other threads without the context lock we can worry about that once we hit that case.

### DIFF
--- a/src/d3d11/d3d11_resource.cpp
+++ b/src/d3d11/d3d11_resource.cpp
@@ -111,7 +111,17 @@ namespace dxvk {
     D3D11CommonTexture* texture = GetCommonTexture(m_resource);
     Rc<DxvkDevice> dxvkDevice = m_device->GetDXVKDevice();
 
-    m_device->GetContext()->WaitForResource(texture->GetImage(), DxvkCsThread::SynchronizeAll, D3D11_MAP_READ_WRITE, 0);
+    {
+      D3D11ImmediateContext* context = m_device->GetContext();
+      D3D10Multithread& multithread = context->GetMultithread();
+      static bool s_errorShown = false;
+
+      if (!multithread.GetMultithreadProtected() && !std::exchange(s_errorShown, true))
+        Logger::warn("D3D11DXGIKeyedMutex::ReleaseSync: Called without context locking enabled.");
+
+      D3D10DeviceLock lock = context->LockContext();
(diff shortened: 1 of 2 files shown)


Joshua-Ashton (approved):

digitalf0x:
From what I've gathered, this is on track to being merged and *should* fix [a known issue with VRChat and AVPro](https://github.com/ValveSoftware/Proton/issues/1199#issuecomment-1790053082 ).  Thank you!

If you do still need more testing or any other help moving this forward, let me know.  (That said, I haven't attempted building Proton and dependencies before, so don't wait for me to figure all that out if this is already good to go.)

digitalf0x:
Quick follow-up good news - I've built Proton locally from [`bleeding-edge` commit `0a87cdc`](https://github.com/ValveSoftware/Proton/commit/0a87cdc677534bc0be26c963eaea926dee2fe87c ), with this (and only this) patch applied to `dxvk`, and [it resolves the crashing I experienced with VRChat AVPro media players](https://github.com/ValveSoftware/Proton/issues/1199#issuecomment-1813844933 ).

src/d3d11/d3d11_resource.cpp
```diff
@@ -111,7 +111,17 @@ namespace dxvk {
...
     Rc<DxvkDevice> dxvkDevice = m_device->GetDXVKDevice();
 
-    m_device->GetContext()->WaitForResource(texture->GetImage(), DxvkCsThread::SynchronizeAll, D3D11_MAP_READ_WRITE, 0);
+    {
+      D3D11ImmediateContext* context = m_device->GetContext();
+      D3D10Multithread& multithread = context->GetMultithread();
+      static bool s_errorShown = false;
+
+      if (!multithread.GetMultithreadProtected() && !std::exchange(s_errorShown, true))
+        Logger::warn("D3D11DXGIKeyedMutex::ReleaseSync: Called without context locking enabled.");
```
doitsujin, line 120: Does this imply that it works on Windows without implicit context locking enabled?

K0bin: Paul Gofman said it most likely does. He suggested the logging.

PR #3734, opened 2023-12-05
Fix crash when srv is submitted to ClearUnorderedAccessViewUint

canadacow:
The Settlers: New Allies, crashes on launch due to the fact that it tries to submit an SRV to clearance via ClearUnorderedAccessViewUint. 

The native behavior is to ignore the errant resource type (see attached repro of the behavior) but dxvk, due to a static cast, crashes when it references the mis-cast object. This change does a QI and confirms the correct object before use, returning immediately if it is not.

[D3D11SrvClearTest.cpp](https://github.com/doitsujin/dxvk/files/13553138/D3D11SrvClearTest.txt)

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -386,10 +386,17 @@ namespace dxvk {
     const UINT                              Values[4]) {
     D3D10DeviceLock lock = LockContext();
 
-    auto uav = static_cast<D3D11UnorderedAccessView*>(pUnorderedAccessView);
-
-    if (!uav)
+    if (!pUnorderedAccessView)
       return;
+
+    Com<ID3D11UnorderedAccessView> qiUav;
+
+    if (FAILED(pUnorderedAccessView->QueryInterface(IID_PPV_ARGS(&qiUav))))
+    {
+      return;
+    }
+
+    auto uav = static_cast<D3D11UnorderedAccessView*>(qiUav.ptr());
 
     // Gather UAV format info. We'll use this to determine
     // whether we need to create a temporary view or not.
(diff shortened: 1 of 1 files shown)


src/d3d11/d3d11_context.cpp
```diff
@@ -386,10 +386,17 @@ namespace dxvk {
...
+    if (!pUnorderedAccessView)
+      return;
 
-    if (!uav)
+    Com<ID3D11UnorderedAccessView> qiUav;
+
+    if (FAILED(pUnorderedAccessView->QueryInterface(IID_PPV_ARGS(&qiUav))))
+    {
       return;
+    }
```
doitsujin, line 397: formatting nit (no reason to have {} here, and we never put { on separate lines)

Joshua-Ashton:
Can we do something other than QueryInterface here which involves useless atomic ref-ing? Do we have a base class for views we can grab some type enum from?

canadacow:
@Joshua-Ashton , I don't know dxvk well enough to suggest something that would pass muster here. Frankly though, seems like premature optimization when QI is _the_ correct and expected behavior in all other such instances.

doitsujin:
QI is fine, it's not like games call ClearUAV more than a dozen times or so per frame.

My real issue here is that passing an SRV by accident shouldn't even be possible, the parameter is typed after all and only accepts a UAV pointer. The only way this can even happen is if the game **deliberately and knowingly** casts a non-UAV pointer to UAV.

canadacow:
@doitsujin , completely agreed, which is why I provided the native sample showing the behavior Microsoft has implemented, no doubt earned in blood. :-)

Issue #3451, opened 2023-05-25
[native] Establish a libdxvk ABI


flibitijibibo:
This is mostly a follow-up for #3321 and #3323. It's also an attempt to document the occasional one-off discussions Joshua and I have had re: making WSI work long-term. That said, this should _not_ be considered a definitive document for how this will be implemented until all maintainers have formally approved it!

## Where we're at

Currently, dxvk-native acts pretty much like Win32 DXGI would, except we just treat the HWND window handle as an SDL_Window* instead and implement the rest with a custom WSI backend in dxvk. This has worked fine up until now, but will fall apart pretty soon:

- GLFW and None WSI implementations cannot use the same fields with the same runtime library
- SDL3 will eventually replace SDL2, and will hit the same problems as GLFW and None

## What we can't do

My first gut reaction would have been to turn the HWND into a public structure, similar to that of SDL_SysWMInfo, where applications could inform dxvk of which WSI driver to use...

```
/* Don't get too excited, remember, this is the BAD example */

#define DXVK_WSI_VERSION 0

typedef enum
{
    DXVK_WSI_SDL2, /* Default */
    DXVK_WSI_SDL3,
    DXVK_WSI_GLFW,
    DXVK_WSI_NONE
} DXVKWSIPlatform;

typedef struct DXVKWindow
{
    uint32_t version; /* Application sets this to DXVK_WSI_VERSION */
    DXVKWSIPlatform platform;

    union
    {
#if DXVK_HAS_WSI_SDL2
        struct
        {
            SDL_Window *window;
        } sdl2;
#endif /* DXVK_HAS_WSI_SDL2 */
#if DXVK_HAS_WSI_SDL3
        struct
        {
            SDL_Window *window;
            /* Maybe something else...? */
        } sdl3;
#endif
        /* etc... */
        void* padding[16]; /* Up to 16 pointers are supported */
    };
} DXVKWindow;
```

[...]

PR #3736, opened 2023-12-06
[DONOTMERGE] [native] Load SDL/glfw via dlopen instead of linking at build time

flibitijibibo:
This modifies DXVK's SDL2 linkage to be dlopened rather than referenced at build time. This is an important part of #3451 which needs all WSI implementations to coexist while also not requiring hard dependencies on every possible WSI library in existence.

~~Notably absent from this draft is any sort of boilerplate to simplify loading SDL symbols: this does every declaration, every load, and every cast manually. Nobody in their right mind would want to maintain this, but I didn't want to start getting clever until there was a clear path forward for what DXVK's preferred style is here. Personally I like doing inline files that macro everything out (example: [part 1](https://github.com/FNA-XNA/FNA3D/blob/master/src/FNA3D_Driver_Vulkan_vkfuncs.h), [part 2](https://github.com/FNA-XNA/FNA3D/blob/master/src/FNA3D_Driver_Vulkan.c#L2666-L2672)), but that usually assumes all the API access comes from one place, which is not the case here.~~

Major TODO items aside from the horrendous style:
- Does Meson define a DLL name somewhere? For testing I just hardcoded `libSDL2-2.0.so.0` but it'd be great to replace that with something like `SDL2_LIBRARY_NAME`.
- Because the WSI namespace is currently all static, the DLL never gets unloaded when we're all finished... I'm not sure how this might be addressed, even in the case where the WSI is selected at runtime.
- ~~I'm not familiar with the dependency tree for DXVK internals, so if there's a way to just have one WsiLibrary instead of the weird one-off case I have in dxvk_sdl2_exts.cpp that would probably clean things up a lot!~~


flibitijibibo:
Updated to depend on #3737, with that fix I was able to do the quick and simple boilerplate to simplify loading the SDL functions. Points 1 and 2 in the OP still remain!

flibitijibibo:
Threw together a way to unload the DLLs, though I'm not 100% sure if this covers all WSI cases... looking around this _should_ be safe as long as wsi is always called by objects created by Direct3DCreate9 and CreateDXGIFactory. If this idea seems okay I'll fold this back into the patch series properly!

flibitijibibo:
#3738 rewrites this series!

PR #3738, opened 2023-12-08
[wsi] Refactor platform system to support multiple WSI implementations

### LINKED ISSUE #3321: Native builds with SDL and GLFW are incompatible, so should have different SONAMEs

flibitijibibo:
This reorganizes the WSI backend system to support multiple backends in a single binary - this is important for dxvk-native which can support either SDL2 or GLFW. We achieve this by doing the following:

- Add a refcounted init/quit system to WSI, which is used by DxvkInstance
- Move the WSI-specific code from dxvk_platform_exts to WSI
- SDL2 and GLFW are now dynamically _loaded_, rather than dynamically _linked_, to avoid hard dependencies that could otherwise be optional
- The WSI backends are now implementations of a common interface, which effectively implements the same API that was exposed to the rest of DXVK. WSI init will iterate through each compiled-in backend and continue until one of them successfully initializes. This can be overridden with the `DXVK_WSIDRIVER` environment variable.

Supercedes #3736

Part of #3451

Fixes #3321

Obsoletes #3324


flibitijibibo:
Latest push is another rebase - I've got some spare cycles this week so if there's anything that needs to be addressed in this (or the other two) PRs let me know and I'll get it done ASAP!

src/wsi/wsi_platform.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+  void init() {
+    if (s_refcount++ > 0)
+      return;
+
+    bool success = false;
+    const std::string hint = dxvk::env::getEnvVar("DXVK_WSIDRIVER");
+    for (const WsiBootstrap *b : wsiBootstrap) {
+      if ((hint != "") && (hint != b->name))
+          continue;
```
doitsujin, line 31: Talking to @Joshua-Ashton, I think a potential issue here is that if multiple WSIs are available we might initialize a random one that the app might not be able to deal with if the env var isn't set. Maybe we should outright fail if we have no env var, or just default to the first WSI in the list?

doitsujin (commented):
Looking into (finally) getting this merged for the 2.4 release among other stuff, looks good overall, just one concern.

src/wsi/wsi_platform.cpp
flibitijibibo: I'd be okay with that; it's something I've run into particularly with SDL3 migration - will make this change later this morning!

flibitijibibo:
Latest push adjusts the behavior of DXVK_WSIDRIVER, particularly when the variable is unset: When Win32 is available we assume that is the default, when it is _not_ available we throw an explicit error requesting that the variable be set.

doitsujin (approved):

PR #3743, opened 2023-12-12
[build] Add soversion to dxvk-native binaries

flibitijibibo:
Part of #3451

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -54,7 +54,7 @@ jobs:
       shell: bash
       run: |
         export VERSION_NAME="${GITHUB_REF##*/}-${GITHUB_SHA##*/}"
-        ./package-native.sh ${VERSION_NAME} build --no-package
+        ./package-native.sh ${VERSION_NAME} build
         echo "VERSION_NAME=${VERSION_NAME}" >> $GITHUB_ENV
 
     - name: Upload artifacts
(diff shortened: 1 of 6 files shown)


flibitijibibo:
Quick update to use meson.project_version() to get the major/minor version of the soname, to avoid adding new constants that need bumping for every release.

Joshua-Ashton:
You need to update `package-native.sh` to handle symlinks or something, it's actually extracting like 3 copies of the .so rn.

I am also not completely confident we need soversioning? We should never be breaking ABI.

flibitijibibo:
Will look into the symlinks soon, I don't think the artifact zip files will be able to preserve them so this would only help the tar releases.

@smcv may have a better explanation but I think this is needed in case a breakage is necessary; the unversioned .so should only be present as part of the development package.

flibitijibibo:
Confirmed that artifactsv3 is what's causing the symlinks to get resolved, so the latest push includes a commit to tar the steamrt-sniper build. I could have sworn that the action supports using tar as the format, but even their own README suggests using tar inside of the zip to fix this, so here we are...

flibitijibibo:
Looks like the added `.1` caused some problems with dxvk_so_version - any preferences on how to deal with this? Currently thinking of just making another string at the top that ignores any numbers after the first two, but maybe there's a better way to do this.

PR #3759, opened 2023-12-21
Add Depandabot, Update Submodules and CI Dependencies

CanerKaraca23:


doitsujin:
What does all of this do and which problems is it supposed to solve?

CanerKaraca23:
> What does all of this do and which problems is it supposed to solve?

It's workflow steps update, not solving any issue. Dependabot automatically detects these updates and creates a pull request.

LingMan:
Having added Dependabot to a few repos myself, I'm going to chime in here.

Dependabot is developed by Github and you can easily enable it for a repo by adding a `.github/dependabot.yml` file. Its task is to check for dependency updates and create a PR with a bump when it finds something outdated.

The idea is that most updates do not require any further changes to most projects, even if they might contain breaking changes for some projects. CI is supposed to catch those cases.

While outdated dependencies often don't require immediate attention, not modernizing once in a while can cause issues down the line. E.g. Github Actions dropped old versions of NodeJS this year. Old versions of the `checkout` action depended on them and I believe stopped working.

Instructing Dependabot to watch for updates to the actions used in `.github/workflows` is a good idea, but there's no point wasting cycles on a daily check. Weekly at most and even monthly would be plenty. The actions used by dxvk rarely get updates, so the PRs shouldn't get annoying in any case.

Looking at the commit histories of the submodules, daily or weekly updates would likely turn into a chore for no gain. The vulkan headers do get manual updates every few months. Your call, @doitsujin, if you'd like to receive an automatic monthly bump PR. Dependabot sadly doesn't offer less frequent update intervals.
The spirv headers haven't seen an update in well over a year. Not sure what's with those.

(I'd also limit this PR to the creation of the config file and have Dependabot open PRs with the actual dependency updates directly against this repo. Your call though. It doesn't really matter.)

doitsujin:
@LingMan thanks for the detailed answer.

Both in vkd3d-proton and dxvk we generally only update headers when needed to use new Vulkan extensions, since we generally just don't need to otherwise. This also generally only happens once every couple of months in either project, and there's no real benefit to updating them more frequently since the existing API doesn't change.

TL;DR I can see this being useful for projects with actual third-party libraries as dependencies, but here it sounds like this would create more work than it would end up saving.

Blisto91:
@CanerKaraca23 If you still want to contribute then i think if you drop Dependabot and the submodule updates and just keep the CI action bump then i think this would get accepted.

CanerKaraca23:
> @CanerKaraca23 If you still want to contribute then i think if you drop Dependabot and the submodule updates and just keep the CI actions version bump then i think this would get accepted.

Thanks, no need if dev not want to. He can bump CI action versions himself and submodules if he wants.

PR #3764, opened 2023-12-25
[d3d9] Fix off-by-one when copying shader defined constants

K0bin:
Fixes: https://gitlab.freedesktop.org/mesa/mesa/-/issues/9126


Joshua-Ashton:
You also need to fix this in UploadSoftwareConstantSet

Joshua-Ashton (approved):

PR #3791, opened 2024-01-11
Micro-optimize locking in fences

ryao:
I found a possible micro-optimization opportunity in the fence code. When a fence has been missed, we can avoid locking *most* of the time via the double-checked locking pattern. We still lock before a second check in case the scheduler caused us to miss the fence. If the scheduler did cause us to miss the fence, we can drop the lock prior to executing the callback function, as a second micro-optimization.

It compiles. I have not tried running it. I figured I would post this here to get feedback before trying to figure out what a sane runtime test would be.


doitsujin:
I mean, fair enough, but also this code is called ~once per frame in a very small number of games.

PR #3800, opened 2024-01-17
[d3d9] workaround for EDG frontend based compilers

r-a-sattarov:
EDG frontend based compilers can't find suitable constructor to convert from "const int *"

```
lcc: "../src/d3d9/d3d9_state.h", line 306: error #415: no suitable constructor
          exists to convert from "const int *" to "dxvk::Vector4Base<float>"
              set->fConsts[StartRegister + i] = replaceNaN(pConstantData + (i * 4));
                                                           ^
          detected during:
            instantiation of
                      "HRESULT dxvk::UpdateStateConstants<ProgramType,ConstantType,T,StateType>(StateType *, UINT, const T *, UINT, bool) [with ProgramType=dxvk::DxsoProgramTypes::VertexShader, ConstantType=dxvk::D3D9ConstantType::Int, T=int, StateType=dxvk::D3D9CapturableState]"
                      at line 374 of "../src/d3d9/d3d9_stateblock.h"
            instantiation of
                      "HRESULT dxvk::D3D9StateBlock::SetShaderConstants<ProgramType,ConstantType,T>(UINT, const T *, UINT) [with ProgramType=dxvk::DxsoProgramTypes::VertexShader, ConstantType=dxvk::D3D9ConstantType::Int, T=int]"
                      at line 6871 of "../src/d3d9/d3d9_device.cpp"
            instantiation of
                      "HRESULT dxvk::D3D9DeviceEx::SetShaderConstants<ProgramType,ConstantType,T>(UINT, const T *, UINT) [with ProgramType=dxvk::DxsoProgramTypes::VertexShader, ConstantType=dxvk::D3D9ConstantType::Int, T=int]"
                      at line 3143 of "../src/d3d9/d3d9_device.cpp"

1 error detected in the compilation of "../src/d3d9/d3d9_device.cpp".
```

Added workaround for this case.

### DIFF
--- a/src/d3d9/d3d9_state.h
+++ b/src/d3d9/d3d9_state.h
@@ -303,7 +303,8 @@ namespace dxvk {
         }
         else {
           for (UINT i = 0; i < Count; i++)
-            set->fConsts[StartRegister + i] = replaceNaN(pConstantData + (i * 4));
+            // The (float*) cast is a workaround for a bug in the EDG compiler frontend
+            set->fConsts[StartRegister + i] = replaceNaN((float*)pConstantData + (i * 4));
         }
       }
       else if constexpr (ConstantType == D3D9ConstantType::Int) {


r-a-sattarov:
@daveedvdv
Hello Daveed. Sorry for mentioning you, but could you help us with EDG frontend bug (possible EDGcpfe/24622)?

Triang3l:
Although relying on the order of passes within the compiler definitely sounds a lot like depending on undefined behavior, I think this specific workaround would introduce a caller/callee responsibility reversal.

According to [this example](https://stackoverflow.com/a/47668210/962670), it should be possible to explicitly specialize a template based on enum values. So, it's possible to implement "callbacks" for the three types either by specializing a function (however, it appears that functions may be specialized only fully — and that would restrict the type `T` of `pConstantData` to something exact, but here that may even be desirable for additional validation to catch issues like this earlier), or by partially specializing a class template (somewhat like that `std::vector<bool>` special case in the standard library) with one method.

Another option would be to replace the `D3D9ConstantType` enumeration with three separate empty classes, and using function overloading for the selection everywhere, but that'd require much wider code changes.

Toxblh (approved):

r-a-sattarov:
> Although relying on the order of passes within the compiler definitely sounds a bit like depending on undefined behavior (though I'm not sure, maybe the expectations from `if constexpr` are actually well-defined and it's a compiler bug), I think this specific workaround would introduce a caller/callee responsibility reversal.

I redid my patch.

Toxblh (approved):
LGTM

ilyakurdyukov (approved):

K0bin:
Sorry, but I gotta ask: Why are you even using that compiler? Why not just use GCC, MSVC or Clang?

r-a-sattarov:
> Sorry, but I gotta ask: Why are you even using that compiler? Why not just use GCC, MSVC or Clang?

Hallo Robin! Sorry for long reply.

On my hardware platform (Elbrus-2000) only one compiler is available so far and it uses the EDG frontend.

PR #3828, opened 2024-02-01
Add a D3D9 option to allow enlarge hardware cursor into 2X

ghost:
When run a D3D9 game on a modern HiDPI display, the hardware cursor would be tiny.

This is due to good old days display and OS were built with 96 DPI in mind.
I play World of Warcraft on Turtle server (https://turtle-wow.org). The game client running on DX 9 with a hard-coded 32x32 cursor size.
While this was good during the game's original time between 2004 - 2006, modern 4K display would show a tiny cute cursor which my eyes often lose track of it. Modern World of Warcraft client already switched to DX11 or DX12 with the capability to set the cursor size bigger. 

So I feel it would be better to add an additional capability for DX9 games to enlarge the cursor. BTW, full screen upscaling from Proton fshack do not work on hardware cursor.

I thought about where to implement such thing:
- The original game: Not possible without source code nor commercial passion.
- Operating system: in this case it would be WINE or Windows. However after reading MS documentation, it looks like HiDPI support was added around Windows 8.1, so elder platform has no such API to support HiDPI.
- D3D9 in DXVK: So here is the patch.

This patch adds a new D3D9 option d3d9.enlargeHardwareCursor which is a boolean. When set True, D3D9 hardware cursor would be enlarged by Nearest-neighbor algorithm into 2X. Should be enough for a 4K display.

About the upscaling algorithm, I wished using hqx (https://en.wikipedia.org/wiki/Hqx). However it comes with LGPL while DXVK is Zlib license. I m not sure about the license thing.
And hqx's implementation comes with a huge look up table (https://github.com/grom358/hqx). I m not sure DXVK would accept such code.
So while I used hqx in the past and it works great, this patch still using Nearest-neighbor.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -370,8 +370,33 @@ namespace dxvk {
         HardwareCursorHeight,
         inputHeight);
 
-      for (uint32_t h = 0; h < copyHeight; h++)
-        std::memcpy(&bitmap[h * HardwareCursorPitch], &data[h * lockedBox.RowPitch], copyPitch);
+      size_t copyWidth = std::min<size_t>(
+        HardwareCursorWidth,
+        inputWidth);
+
+      if (m_d3d9Options.enlargeHardwareCursor &&
+          copyHeight * 2 <= HardwareCursorHeight &&
+          copyWidth * 2 <= HardwareCursorWidth) {
+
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Not many games set the cursor with DX9. This seems like a job for your compositor more than anything. Gamescope and SteamOS support cursor scaling already...

PR #3834, opened 2024-02-04
Install headers and pkg-config files, MkII

### LINKED ISSUE #3323: Install headers and pkg-config metadata for DXVK Native
When building a native library for Linux, it's conventional to install header files to a subdirectory of /usr/include, and install pkg-config metadata describing how a dependent project can make use of the library.

In the case of DXVK, the current Fedora packaging https://src.fedoraproject.org/rpms/dxvk-native/tree/rawhide (which is somewhat outdated, currently still on 1.9) has added `dxvk-native-d3d9.pc` and `dxvk-native-d3d11.pc`, which lets a dependent project use dependencies like `dependency('dxvk-native-d3d9 >= 1.9')` or their build system's equivalent to learn the right include directories, libraries and recursive dependencies.

Newer Fedora packaging under review in https://bugzilla.redhat.com/show_bug.cgi?id=2152207, which unifies the Wine and native packaging for DXVK, seems to be using `dxvk-d3d9.pc` and `dxvk-d3d10.pc` instead, so the dependency would be `dependency('dxvk-d3d9 >= 2.1')`. Is that an intentional or unintentional change?

DXVK Native isn't currently packaged in Debian at all (Debian only builds the Wine-compatible version), but I'm looking into updating the packaging to also build DXVK Native, similar to what's planned in Fedora. I'd like to make the Debian packaging consistent with what's in Fedora so that the same source code can work in both places, ideally via the upstream project installing the necessary headers and pkg-config itself.

Does the upstream project have a preference for what its "official" name is, when used as a Linux library by games like Portal 2?

cc @Conan-Kudo @flibitijibibo @Joshua-Ashton

flibitijibibo:
This reworks #3325 to apply to the latest dxvk revision, and is based on #3738 and #3743. As a result, only the last 3 commits are actually for this specific PR.

CC @smcv, I had to adjust a few of the patches, so we'll need to re-confirm the signoffs for those.

Fixes #3323

Part of #3451

Supercedes #3325

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,5 +1,6 @@
 project('dxvk', ['c', 'cpp'], version : 'v2.3.1', meson_version : '>= 0.58', default_options : [ 'cpp_std=c++17', 'warning_level=2' ])
 
+pkg = import('pkgconfig')
 cpu_family = target_machine.cpu_family()
 platform   = target_machine.system()
 fs = import('fs')
@@ -146,6 +147,7 @@ else
   endif
   
   dxvk_name_prefix = 'dxvk_'
+  dxvk_pkg_prefix = 'dxvk-'
 
   link_args += [
     '-static-libgcc',
--- a/meson.build
+++ b/meson.build
@@ -54,7 +54,7 @@ dep_displayinfo = dependency(
 )
 
 if platform == 'windows'
-  dxvk_so_version = {}
+  dxvk_so_version = {'name_prefix': ''}
 
   compiler_args += [
     '-DNOMINMAX',
@@ -145,7 +145,7 @@ else
     error('SDL2 or GLFW are required to build dxvk-native')
   endif
   
-  dxvk_name_prefix = 'libdxvk_'
+  dxvk_name_prefix = 'dxvk_'
 
   link_args += [
     '-static-libgcc',
@@ -161,7 +161,6 @@ add_project_link_arguments(cpp.get_supported_link_arguments(link_args), language
 add_project_link_arguments(cc.get_supported_link_arguments(link_args), language: 'c')
 
 exe_ext = ''
-dll_ext = ''
 def_spec_ext = '.def'
 
 glsl_compiler = find_program('glslang', 'glslangValidator')
--- a/meson.build
+++ b/meson.build
@@ -185,4 +185,8 @@ dxvk_version = vcs_tag(
   output: 'version.h',
 )
 
+if platform != 'windows'
+  subdir('include/native')
+endif
+
 subdir('src')
--- a/meson.build
+++ b/meson.build
@@ -54,6 +54,8 @@ dep_displayinfo = dependency(
 )
 
 if platform == 'windows'
+  dxvk_so_version = {}
+
   compiler_args += [
     '-DNOMINMAX',
     '-D_WIN32_WINNT=0xa00',
@@ -118,6 +120,10 @@ if platform == 'windows'
   dxvk_name_prefix = ''
   compiler_args += ['-DDXVK_WSI_WIN32']
 else
+  dxvk_abi_version = '0'
+  dxvk_version = meson.project_version().strip('v').split('.')
+  dxvk_so_version = {'version': dxvk_abi_version + '.' + dxvk_version[0] + dxvk_version[1] + dxvk_version[2]}
+
   wrc           = find_program('touch')
   wrc_generator = generator(wrc, output : [ '@BASENAME@_ignored.h' ], arguments : [ '@OUTPUT@' ] )
 
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -54,7 +54,7 @@ jobs:
       shell: bash
       run: |
         export VERSION_NAME="${GITHUB_REF##*/}-${GITHUB_SHA##*/}"
-        ./package-native.sh ${VERSION_NAME} build --no-package
+        ./package-native.sh ${VERSION_NAME} build
         echo "VERSION_NAME=${VERSION_NAME}" >> $GITHUB_ENV
 
     - name: Upload artifacts
@@ -62,5 +62,5 @@ jobs:
       uses: actions/upload-artifact@v3
       with:
         name: dxvk-${{ env.VERSION_NAME }}
-        path: build/dxvk-native-${{ env.VERSION_NAME }}
+        path: build/dxvk-native-${{ env.VERSION_NAME }}.tar.gz
         if-no-files-found: error
--- a/include/native/meson.build
+++ b/include/native/meson.build
@@ -1,0 +1,19 @@
+install_subdir(
+  'directx',
+  install_dir: get_option('includedir') / 'dxvk',
+  strip_directory: true,
+  exclude_files: '.git'
+)
+
+install_subdir(
+  'windows',
+  install_dir: get_option('includedir') / 'dxvk',
+  strip_directory: true,
+)
+
+install_headers(
+  'wsi/native_wsi.h',
+  'wsi/native_sdl2.h',
+  'wsi/native_glfw.h',
+  subdir: 'dxvk/wsi',
+)
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -28,3 +28,10 @@ d3d10_core_dll = shared_library(dxvk_name_prefix+'d3d10core', d3d10_core_src, d3
 d3d10_core_dep = declare_dependency(
   link_with           : [ d3d10_core_dll ],
 )
+
+if platform != 'windows'
+  pkg.generate(d3d10_core_dll,
+    filebase: dxvk_pkg_prefix + 'd3d10core',
+    subdirs:  'dxvk',
+  )
+endif
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -15,8 +15,7 @@ else
   d3d10_d3d11_dep = d3d11_dep
 endif
 
-d3d10_core_dll = shared_library('d3d10core'+dll_ext, d3d10_core_src, d3d10_core_res,
-  name_prefix         : dxvk_name_prefix,
+d3d10_core_dll = shared_library(dxvk_name_prefix+'d3d10core', d3d10_core_src, d3d10_core_res,
   dependencies        : [ d3d10_d3d11_dep ],
   include_directories : dxvk_include_path,
   install             : true,
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -23,6 +23,7 @@ d3d10_core_dll = shared_library('d3d10core'+dll_ext, d3d10_core_src, d3d10_core_
   vs_module_defs      : 'd3d10core'+def_spec_ext,
   link_args           : d3d10_core_ld_args,
   link_depends        : [ d3d10_core_link_depends ],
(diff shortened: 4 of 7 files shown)


smcv:
I have no problem with the versions of my commits that appear here. Feel free to add a `Co-authored-by`, or not, as you prefer.

smcv:
On #3325 I wrote:

> It's a bit unfortunate that [the name prefix for `*.so`] is `dxvk_` and [the name prefix for `*.pc`] is `dxvk-`. I've kept the naming from the Fedora packaging for now, but I'd be very tempted to change it to `dxvk_d3d9.pc`, `dxvk_glfw_dxgi.pc` and so on, to be consistent with the library names and avoid having two subtly different name-prefixes.

What do maintainers think about that? This PR is changing the pkg-config API *anyway*, so if there was a time to make that change, now would probably be that time.

flibitijibibo:
I'd be okay with doing _ over -, I'll check with Neal to see if this is an okay change for Fedora.

src/wsi/sdl2/wsi_platform_sdl2.cpp
```diff
@@ -0,0 +1,69 @@
+#if defined(DXVK_WSI_SDL2)
+
+#include "wsi_platform_sdl2.h"
+#include "../../util/util_error.h"
+#include "../../util/util_string.h"
+#include "../../util/util_win32_compat.h"
+
+#include <SDL2/SDL_vulkan.h>
```
smcv, line 8: The canonical way to include SDL2 headers is `#include "SDL.h"` (as per https://wiki.libsdl.org/SDL2/FAQDevelopment). This will require using the `-I` flags from `sdl2-config --cflags` or `pkg-config --cflags sdl2`, which in Meson means compiling the SDL2 WSI files with `dependency('sdl2')` or at least a `partial_dependency`. In this case, because of the use of `dlopen()`-style dynamic loading, you'd presumably want to use a `partial_dependency`.

This is not just a theoretical thing: in Debian, we have had some trouble in the past with interoperability with the `SDL2/SDL_whatever.h` style. It works most of the time, except for when it doesn't, and can have weird quirks dependent on whether SDL was compiled with Autotools or CMake.

(If you're planning to add a SDL3 WSI in future when SDL3 has stabilized, note that this is different in SDL3: the recommendation in https://wiki.libsdl.org/SDL3/FAQDevelopment is `<SDL3/SDL.h>`. This is an intentional interface change, and all packaged versions of SDL3 are expected to make it work.)

src/wsi/sdl2/wsi_platform_sdl2.h
```diff
@@ -2,20 +2,107 @@
 
 #include <SDL2/SDL.h>
```
smcv, line 3: Similarly this is canonically

```suggestion
#include "SDL.h"
```

in SDL2.

src/wsi/sdl2/wsi_platform_sdl2.cpp
flibitijibibo: That all makes sense to me - incidentally, is it possible to get the soname via meson? We'd be mimicking what SDL does for its dynamic loading support: https://github.com/libsdl-org/SDL/blob/main/cmake/sdlchecks.cmake#L546

smcv: No, I don't know of a trivial way to get the SONAME.

Honestly, my answer would be to keep hard-coding the expected SONAME. If the ABI of SDL or GLFW changes incompatibly, then you already need to know about that and react to it, because the symbols that you list in [wsi_platform_sdl2_funcs.h](https://github.com/doitsujin/dxvk/pull/3834/files#diff-4ccb978d31117e5c2b73e6f2f3766f412b897cec609c1ad010950e38481cb0fe) and [wsi_platform_glfw_funcs.h](https://github.com/doitsujin/dxvk/pull/3834/files#diff-e109f7a3c9d6cbe0419f2dc9fce561cc2cf2cc81b814827bab5748ae73cee968) will need reviewing anyway. That's because one possible reason for an ABI break would be an incompatible change to the signature of one of those functions, for example if `GLFWmonitor **glfwGetMonitors(int *)` changed to `GLFWmonitor **glfwGetMonitors(size_t *)` in a newer version of GLFW.

The traditional build-system-developer answer, e.g. in `libtool`, would be "no you can't learn the SONAME, and it's wrong to want to", because shared libraries and `dlopen()`'able modules happen to be the same thing on popular platforms like Windows and Linux but have historically been implemented differently in some OSs. In that view of the world, each shared object is either correct to link like `-lSDL3` or correct to dlopen like `dlopen("/path/to/myplugin.so")`, but never both; and the pedantically correct approach would be for DXVK to have in-tree plugins for SDL and GLFW (no need for a stable API/ABI between library and plugin), similar to how libdecor has in-tree plugins for Cairo and GTK. In that design, the plugin would be the only thing that is dlopened, and the plugin would have a normal DT_NEEDED dependency (hard dependency) on the library it uses.

But, assuming that you have no interest in making DXVK run on AIX or HPUX or whatever other museum piece has this limitation, I can see why you would prefer to cut out the middle-man and dlopen the library itself!

It looks as though SDL's `FindLibraryAndSONAME` assumes that if you take the `basename` of the `realpath` of the shared library, and then truncate `libfoo.so.1.2.3.4` to `libfoo.so.1`, the result is the SONAME. This is not strictly correct - some libraries, confusingly, have SONAME `libfoo.so.0` and a symlink `libfoo.so.0 -> libfoo.so.1.2.3.4` - but it "often" works correctly for "normal" libraries.

I'm guessing that compiling a version of https://salsa.debian.org/gnome-team/gobject-introspection/-/blob/debian/1.78.1-14/debian/elf-get-needed.c?ref_type=tags (except with `DT_SONAME` instead of `DT_NEEDED`) for the `native` (build) architecture, and then running it against the discovered host-architecture library, is not the answer you're looking for either - but that would be the most "correct" way to do it.

flibitijibibo: Fair enough, thanks for the info - maybe this is something we can punt off to the other SDL maintainers to solve For Real :P

Should have fixes for all of the other notes at some point this week!

Conan-Kudo:
> I'd be okay with doing _ over -, I'll check with Neal to see if this is an okay change for Fedora.

Does Meson's pkgconfig generator not support setting `Provides` names? It's documented in [`pc(5)`](https://www.mankier.com/5/pc).

smcv:
> Does Meson's pkgconfig generator not support setting `Provides` names? It's documented in [`pc(5)`](https://www.mankier.com/5/pc).

Not as far as I can see in <https://mesonbuild.com/Pkgconfig-module.html>.

Does original freedesktop.org pkg-config support `Provides`, or are they a pkgconf-ism?

smcv:
I believe the way the equivalent of Provides was traditionally implemented is as a symlink `oldname.pc -> newname.pc`, as is done in the Debian packaging of `sdl12-compat` and `sdl2-compat`.

Conan-Kudo:
It was always part of the spec, but not implemented fully in `pkg-config`. `pkgconf` actually implemented it to work.

meson.build
```diff
@@ -121,17 +126,20 @@ else
     './include/native/directx'
   ]
 
-  dxvk_wsi = get_option('dxvk_native_wsi')
-  
-  if dxvk_wsi == 'sdl2'
-    lib_sdl2 = cpp.find_library('SDL2')
+  lib_sdl2 = cpp.find_library('SDL2', required: false)
+  lib_glfw = cpp.find_library('glfw', required: false)
```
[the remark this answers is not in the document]
flibitijibibo: I've pushed this change to the WSI PR: #3738

src/wsi/sdl2/wsi_platform_sdl2.cpp
flibitijibibo: The includes have been fixed in the WSI PR as well.

flibitijibibo:
Rebased and updated with the latest changes from the other two PRs.

Did we ever make a call with the `-` v. `_` change? Left that part alone for now.

Conan-Kudo:
Well, I'm shipping `dxvk-native-2.3` now with `-`, so... 🤷🏾‍♂️

flibitijibibo:
Latest push is just a rebase, this PR now only depends on #3743

meson.build
```diff
@@ -1,5 +1,6 @@
-project('dxvk', ['c', 'cpp'], version : 'v2.3.1', meson_version : '>= 0.58', default_options : [ 'cpp_std=c++17', 'warning_level=2' ])
+project('dxvk', ['c', 'cpp'], version : '2.3.1', meson_version : '>= 0.58', default_options : [ 'cpp_std=c++17', 'warning_level=2' ])
```
Joshua-Ashton, line 1: You already .strip the 'v' later on. This shouldn't need to change.

flibitijibibo: Good catch, dropped that commit from the changeset.

meson.build
```diff
@@ -118,6 +121,10 @@ if platform == 'windows'
   dxvk_name_prefix = ''
   compiler_args += ['-DDXVK_WSI_WIN32']
 else
+  dxvk_abi_version = '0'
+  dxvk_version = meson.project_version().strip('v').split('.')
+  dxvk_so_version = {'version': dxvk_abi_version + '.' + dxvk_version[0] + '.' + dxvk_version[1]}
```
Joshua-Ashton, line 126: probably wants `+ '.' + dxvk_version[2]` as well?

Joshua-Ashton (changes requested):

meson.build
Joshua-Ashton: Alternatively make it `.0.231`

Joshua-Ashton: Sure.

Joshua-Ashton: I still think it would be nice to know the full version. We don't really do versioning other than "feels" in DXVK.

So let's have the third digit.

Joshua-Ashton:
If you fix that comment and the commit commit names, I will merge it.

meson.build
flibitijibibo: Added in the latest revision!

flibitijibibo:
Looks like Meson made the choice for us:

```
src/dxgi/meson.build:24:0: ERROR: Invalid Shared library version "0.2.3.1". Must be of the form X.Y.Z where all three are numbers. Y and Z are optional.
```

So I guess I need to revert that...? Also, what were the commit/comment changes I needed to make, I couldn't see them in this thread.

Joshua-Ashton:
Just make the commit names match up with the style of the rest of the repo.

Joshua-Ashton:
> Looks like Meson made the choice for us:
> 
> ```
> src/dxgi/meson.build:24:0: ERROR: Invalid Shared library version "0.2.3.1". Must be of the form X.Y.Z where all three are numbers. Y and Z are optional.
> ```
> 
> So I guess I need to revert that...? Also, what were the commit/comment changes I needed to make, I couldn't see them in this thread.

lol

flibitijibibo:
Latest changes to to be the '231' style, and adjusts the commit tag to use [].

Joshua-Ashton:
One final nit, the tarball for Steam Runtime binaries contains a .git file inside of the include folder :P

flibitijibibo:
Added an exclude_files from the directx install block to ignore a possible .git file when packaging a repo clone (shouldn't happen with source tarballs).

PR #3836, opened 2024-02-05
meson: Check for bundled Vulkan/SPIR-V headers

### LINKED ISSUE #3322: Use system SPIRV and Vulkan headers if new enough
When packaging a library like DXVK Native, Linux distributions generally prefer to use system dependencies rather than vendoring a copy into dependent projects, so that if there's a bug in the dependency, it can be fixed in one place. This would also make it easier to build from the source tarballs provided by Github, which don't include git submodules, or from a simple git clone without submodules.

Would it be possible to use something like this (slightly pseudocode, I haven't tested this):

```
vulkan_dep = dependency('vulkan', '>=1.3.239', required: false)

if vulkan_dep.found()
  vulkan_headers = vulkan_dep.partial_dependency(includes: true)
else
  vulkan_headers = declare_dependency(include_directories: ['./include/vulkan/include'])
endif

spirv_dep = dependency('SPIRV-Headers', '>=1.5.5', required: false)

if not spirv_dep.found()
  spirv_dep = declare_dependency(include_directories: ['./include/spirv/include'])
endif
```

and then use `vulkan_headers` and `spirv_dep` in the dependencies of each target, instead of adding these two directories to `dxvk_include_dirs`?

If that seems like an OK approach, I'll prepare a merge request.

(Unfortunately, checking for SPIRV-Headers newer than 1.5.5 doesn't currently work, as a result of https://github.com/KhronosGroup/SPIRV-Headers/issues/334.)

flibitijibibo:
This is at least enough to use the system headers when building from the tarball, but doesn't check for any version requirements. While imperfect, it's an improvement on the current solution which is to `sed -i` meson.build, which is sort of overkill...

Partially fixes #3322

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,8 @@
-project('dxvk', ['c', 'cpp'], version : 'v2.3', meson_version : '>= 0.49', default_options : [ 'cpp_std=c++17', 'warning_level=2' ])
+project('dxvk', ['c', 'cpp'], version : 'v2.3', meson_version : '>= 0.58', default_options : [ 'cpp_std=c++17', 'warning_level=2' ])
 
 cpu_family = target_machine.cpu_family()
 platform   = target_machine.system()
+fs = import('fs')
 
 cpp = meson.get_compiler('cpp')
 cc = meson.get_compiler('c')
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
There is no reason to ever use system headers for DXVK

flibitijibibo:
For completion's sake, this is the FNA "archivinate.sh" script:

```bash
#!/bin/bash

set -ex

cd "`dirname "$0"`"

git clone --recursive https://github.com/FNA-XNA/FNA
rm -rfv FNA/.git* FNA/lib/*/.git* FNA/lib/FNA3D/*/.git*
```

This could _totally_ be made into a generic script by reading .gitmodules files recursively, but for projects that know their submodule tree a quick script like this does the job!

PR #3841, opened 2024-02-05
[d3d9] Add a software cursor emulation config option

### LINKED ISSUE #3020: Dungeon Siege 2 invisible mouse cursor

### Software information
Dungeon Siege 2

### System information
- GPU: R9 380
- Driver: days old mesa-git
- Wine version: Wine-GE 7-32

### Apitrace file(s)
Reproduces in the trace

### Log files

WinterSnowfall:
In relation to #3020. The game tries to create a 64 x 64 cursor, which wasn't supported prior to this PR, but is otherwise happy with our existing code path catering for hardware cursors. It even happily switches cursor icons in game on hover over certain objects.

Since this isn't exactly a software cursor implementation, rather a workaround, I've created a config option for it.

As far as I know this is the only d3d9 game that actually depended on software cursor support and this PR is enough to get it working properly. I can't guarantee other games will work just as well, since I haven't done a lot of validations and exotic bitmap sizes may cause issues. But we can improve the logic a bit if that ever happens to be a problem.

I welcome any testing, but this shouldn't affect the previous hardware cursor logic (unless I messed up badly somewhere, which doesn't seem to be the case based on my tests at least :sweat_smile:).

Would also appreciate very critical pairs of eyes on my memory allocations and use of pointers, since it's not something I get to do every day (and frankly I'm getting too old for it). Or I'm open to better alternatives, if any exist.

### DIFF
--- a/src/d3d9/d3d9_cursor.cpp
+++ b/src/d3d9/d3d9_cursor.cpp
@@ -25,24 +25,52 @@ namespace dxvk {
   }
 
 
-  HRESULT D3D9Cursor::SetHardwareCursor(UINT XHotSpot, UINT YHotSpot, const CursorBitmap& bitmap) {
-    DWORD mask[32];
-    std::memset(mask, ~0, sizeof(mask));
+  HRESULT D3D9Cursor::SetHardwareCursor(
+          UINT                   XHotSpot,
+          UINT                   YHotSpot,
+    const std::vector<uint8_t>&  bitmap,
+          bool                   cursorEmulation,
+          UINT                   width,
+          UINT                   height,
+          HWND                   window) {
+    bool noPreviousHardwareCursor = false;
+
+    uint32_t cursorWidth = cursorEmulation ? width : HardwareCursorWidth;
+    uint32_t cursorHeight = cursorEmulation ? height : HardwareCursorHeight;
+
+    // For color icons, the hbmMask and hbmColor bitmaps
+    // are the same size, each of which is the size of the icon.
+    std::vector<DWORD> mask(cursorWidth * cursorHeight, ~0);
 
     ICONINFO info;
     info.fIcon    = FALSE;
     info.xHotspot = XHotSpot;
     info.yHotspot = YHotSpot;
-    info.hbmMask  = ::CreateBitmap(HardwareCursorWidth, HardwareCursorHeight, 1, 1,  mask);
-    info.hbmColor = ::CreateBitmap(HardwareCursorWidth, HardwareCursorHeight, 1, 32, &bitmap[0]);
+    info.hbmMask  = ::CreateBitmap(cursorWidth, cursorHeight, 1, 1,  &mask[0]);
+    info.hbmColor = ::CreateBitmap(cursorWidth, cursorHeight, 1, 32, &bitmap[0]);
 
     if (m_hCursor != nullptr)
       ::DestroyCursor(m_hCursor);
+    else
+      noPreviousHardwareCursor = true;
 
     m_hCursor = ::CreateIconIndirect(&info);
 
     ::DeleteObject(info.hbmMask);
     ::DeleteObject(info.hbmColor);
+
+    if (cursorEmulation) {
+      ::SetClassLongPtr(window, GCLP_HCURSOR, reinterpret_cast<LONG_PTR>(m_hCursor));
+      
+      CURSORINFO ci;
+      while (::GetCursorInfo(&ci) && ci.flags == 0u)
+        ::ShowCursor(TRUE);
+
+      // Castle Strike needs one extra initial increment
+      // to display the emulated cursor on its menu
+      if (noPreviousHardwareCursor)
+        ::ShowCursor(TRUE);
+    }
 
     ShowCursor(m_visible);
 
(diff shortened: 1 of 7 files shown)


WinterSnowfall:
> I happened to make a very close pull to yours recently: #3828 #3827

Indeed, saw those. Together with @Blisto91's regular mentions about implementing a software cursor (which, ironically, I still haven't), they reminded me there's a game that actually cares about these things. To be fair, my main objective here is to fix Dungeon Siege 2, so it's far less broad in scope.

> I vote for this pull that it would be better to support variable cursor size rather than hard-code as 32x32. I did some test on Windows 11 and Xfce + X11 both of them being able to display a 64x64 cursor.

I'm on MATE + X11 and it also works fine.

> And FYI, from what I read from [MS document](https://learn.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3ddevice9-setcursorproperties), cursor size API should have a fallback capability scaling the size back to 32x32 when the underlying platform can not support different size. I didn't write such code.

A good point, but that's still not being done. The hardware cursor path always assumes a 32 x 32 cursor regardless of bitmap size and does no scaling, which has never been a problem really, so it's probably good enough. Getting Dungeon Siege 2 to display a cursor was actually a lot less work initially, however the cursor was getting clipped to the top left 32 x 32, which I also had to  address.

Blisto91:
Confirmed on Linux and Windows in Dungeon Siege II from GoG

WinterSnowfall:
> Confirmed on Linux and Windows in Dungeon Siege II from GoG

Thanks for testing and confirming, especially on Windows!

WinterSnowfall:
I've added one more... well, I'd like to say "elegant fix", but we only have "cursed hack" on the menu today... in order to make the cursor emulation work properly with Castle Strike, a d3d8 game that also tries to create a 64 x 64 hardware cursor, but then also tries to control it with win32 calls (don't ask). Retested everything with Dungeon Siege II, which doesn't seem to mind any of the changes.

I'll PR the config option for Castle Strike against d8vk, since it's a d3d8 game.

WinterSnowfall:
There's a bit of a hitch on Windows, since the cursor won't show initially, rather only after alt+tabbing out and back in, but that bit is probably due to some events we're not handling. It works fine in Wine/Proton at least.

There's some minor differences when compared to WineD3D's software cursors (namely the cursor stay on screen during loading screens), but a somewhat janky cursor is better than no cursor. It also seems to hide/show just fine in-game, so that's good.

I've run out of ideas how to make this any better without actually implementing a software cursor, which would be a major pain I'm not sure anyone is willing to pursue, so I'll leave it as is.

src/d3d9/d3d9_cursor.cpp
```diff
@@ -25,24 +25,55 @@ namespace dxvk {
...
+          bool               cursorEmulation,
+          UINT               width,
+          UINT               height,
+          HWND               window) {
+    bool noPreviousHardwareCursor = false;
+
+    // For color icons, the hbmMask and hbmColor bitmaps
+    // are the same size, each of which is the size of the icon.
+    uint32_t maskSize = cursorEmulation ? width * height : HardwareCursorWidth * HardwareCursorHeight;
+    DWORD* mask = new DWORD[maskSize];
```
Joshua-Ashton, line 41: Just use a std::vector.

src/d3d9/d3d9_device.cpp
```diff
@@ -359,20 +359,36 @@ namespace dxvk {
...
+      uint32_t emulatedCursorPitch = inputWidth * HardwareCursorFormatSize;
+      
+      uint32_t cursorHeight = m_d3d9Options.softwareCursorEmulation ? inputHeight : HardwareCursorHeight;
+      uint32_t cursorPitch = m_d3d9Options.softwareCursorEmulation ? emulatedCursorPitch : HardwareCursorPitch;
+
       // Windows works with a stride of 128, lets respect that.
       // Copy data to the bitmap...
-      CursorBitmap bitmap = { 0 };
+      // Format Size of 4 bytes (ARGB)
+      uint8_t* bitmap = new uint8_t[cursorHeight * cursorPitch];
```
Joshua-Ashton, line 370: Same here

src/d3d9/d3d9_cursor.cpp
WinterSnowfall: @Joshua-Ashton Done, hopefully in a right manner. Let me know if I've initialized it correctly :sweat_smile:. I was half expecting MSVC to complain about something, because why not.

src/d3d9/d3d9_device.cpp
WinterSnowfall: Ditto

K0bin:
I don't quite get why this is a config option. Is there any downside to always enabling/exposing this?

WinterSnowfall:
> I don't quite get why this is a config option. Is there any downside to always enabling/exposing this?

There's two reasons for that mostly:
- Hardware cursors are only ever supposed to be 32 x 32 pixels, especially in windowed mode. I've done some tests out of curiosity and it looks like WineD3D validates this, so we're going out of spec to make some games happy without actually implementing software cursors. This path is essentially a workaround for the absence 64 x 64 pixels icon support and mouse input message handling in the case of hardware cursors. Only 2 games out there that we know of would need it to work properly.
- [This bit](https://github.com/doitsujin/dxvk/blob/1299ad050fd8ebcc2b2c3842785532c36290e454/src/d3d9/d3d9_cursor.cpp#L62) is a rather cursed hack done exclusively for Castle Strike, which relies on some cursed form of win32 message handling to show the cursor on its main menu (then it show/hides it properly with d3d8 calls during gameplay). I would not hijack the window cursor pointer for every well behaving game out there.

K0bin:
Honestly, why don't we just implement software cursor?

WinterSnowfall:
> Honestly, why don't we just implement software cursor?

... go ahead :). This PR is mostly a stopgap until we get there and this path should be removed once we do. That being said, the window message handling bit sounds like a potential major pain to get right (and a bit beyond what I'd venture to implement).

PR #3843, opened 2024-02-06
D3D9: Make windowed GetFrontBufferData more accurate

### LINKED ISSUE #3724: [d3d9] Medieval 2 black loading screen in windowed mode with dxvk
The game can still have black loading screens with dxvk when it is run in [windowed mode](https://www.pcgamingwiki.com/wiki/Medieval_II:_Total_War#Video). Doesn't happen without dxvk (tested on Windows).

### Software information
Total War Medieval 2.
[windowed mode](https://www.pcgamingwiki.com/wiki/Medieval_II:_Total_War#Video)

### System information
- GPU: RX 6800
- Driver: Adrenalin 23.11.1. Haven't tested Linux yet.
- DXVK version: master

### Apitrace file(s)
https://drive.proton.me/urls/613B5Y9KWG#53STJVAgRxn6
One made with Windows native and one with Windows dxvk as it is baked in

### Log files
[medieval2_d3d9.log](https://github.com/doitsujin/dxvk/files/13400452/medieval2_d3d9.log)

K0bin:
Fixes #3724

Total War Medieval 2 uses GetFrontBufferData to take a screenshot and then copies that screenshot to a texture which is used to render the background of the loading screen.

In windowed mode, GetFrontbufferData will in fact take a screenshot of the entire screen. We don't implement it that way in DXVK, we always blit the most recent front buffer.

There's two problems with our implementations in Medieval 2:
- The image is positioned incorrectly. The game retrieves the window client rect and copies those parts of the GetFrontBufferData destination surface. DXVK always puts the screenshot at 0,0.
- The game uses additional swapchains (on the same window no less) and presents using one of those. DXVK always blits the implicit swapchain.

So to solve this without having to implement GetFrontBufferData properly (taking a screenshot of the entire screen), I propose that we:
- Query the position of the window and blit the screenshot to that location. (easy)
- Track the swapchain that was the most recently used for presenting and forward GetFrontBufferData calls to that swapchain instead of the implicit one. That's obviously a hack but it works for Medieval 2 and should be fine for most other games, as very few games use additional swapchains and even fewer of those use them together with GetFrontBufferData.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1083,7 +1083,12 @@ namespace dxvk {
     if (unlikely(iSwapChain != 0))
       return D3DERR_INVALIDCALL;
 
-    return m_implicitSwapchain->GetFrontBufferData(pDestSurface);
+    D3D9DeviceLock lock = LockDevice();
+
+    // In windowed mode, GetFrontBufferData takes a screenshot of the entire screen.
+    // We use the last used swapchain as a workaround.
+    // Total War: Medieval 2 relies on this.
+    return m_mostRecentlyUsedSwapchain->GetFrontBufferData(pDestSurface);
   }
 
 
@@ -7813,6 +7818,7 @@ namespace dxvk {
     }
     else
       m_implicitSwapchain = new D3D9SwapChainEx(this, pPresentationParameters, pFullscreenDisplayMode);
+      m_mostRecentlyUsedSwapchain = m_implicitSwapchain.ptr();
 
     if (pPresentationParameters->EnableAutoDepthStencil) {
       D3D9_COMMON_TEXTURE_DESC desc;
--- a/src/d3d9/d3d9_device.h
+++ b/src/d3d9/d3d9_device.h
@@ -1241,6 +1241,34 @@ namespace dxvk {
 
     uint64_t GetCurrentSequenceNumber();
 
+    /**
+     * @brief Get the swapchain that was used the most recently for presenting
+     * Has to be externally synchronized.
+     * 
+     * @return D3D9SwapChainEx* Swapchain
+     */
+    D3D9SwapChainEx* GetMostRecentlyUsedSwapchain() {
+      return m_mostRecentlyUsedSwapchain;
+    }
+
+    /**
+     * @brief Set the swapchain that was used the most recently for presenting
+     * Has to be externally synchronized.
+     * 
+     * @param swapchain Swapchain
+     */
+    void SetMostRecentlyUsedSwapchain(D3D9SwapChainEx* swapchain) {
+      m_mostRecentlyUsedSwapchain = swapchain;
+    }
+
+    /**
+     * @brief Reset the most recently swapchain back to the implicit one
+     * Has to be externally synchronized.
+     */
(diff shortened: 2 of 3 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -7813,6 +7816,7 @@ namespace dxvk {
     }
     else
       m_implicitSwapchain = new D3D9SwapChainEx(this, pPresentationParameters, pFullscreenDisplayMode);
+      m_mostRecentlyUsedSwapchain = m_implicitSwapchain.ptr();
```
PatrickvL, line 7819: There's `ResetLastUsedSwapchain()` for this now.

src/d3d9/d3d9_device.h
```diff
@@ -1241,6 +1241,18 @@ namespace dxvk {
...
 
+    D3D9SwapChainEx* GetLastUsedSwapchain() {
+      return m_mostRecentlyUsedSwapchain;
+    }
+
+    void SetLastUsedSwapchain(D3D9SwapChainEx* swapchain) {
+      m_mostRecentlyUsedSwapchain = swapchain;
+    }
+
+    void ResetLastUsedSwapchain() {
```
PatrickvL, line 1252: Consider renaming this to `RefreshLastUsedSwapchain()`

PatrickvL (changes requested):
Two remarks

src/d3d9/d3d9_device.h
K0bin: But it doesn't refresh it, it resets it to the built-in swapchain.

src/d3d9/d3d9_device.cpp
K0bin: I'm not a fan of calling that function for the initial initialization.

src/d3d9/d3d9_device.cpp
```diff
@@ -1083,7 +1083,10 @@ namespace dxvk {
     if (unlikely(iSwapChain != 0))
       return D3DERR_INVALIDCALL;
 
-    return m_implicitSwapchain->GetFrontBufferData(pDestSurface);
+    // In windowed mode, GetFrontBufferData takes a screenshot of the entire screen.
+    // We use the last used swapchain as a workaround.
+    // Total War: Medieval 2 relies on this.
+    return m_mostRecentlyUsedSwapchain->GetFrontBufferData(pDestSurface);
```
doitsujin, line 1089: Potential use-after-free here, we could grab the pointer to a swap chain that's currently being destroyed, swap chain destructor completes, and *then* the code inside `GetFrontBufferData` grabs the device lock.

I think a reasonable fix would be to unconditionally lock the device in `Swapchain::Release` if called on a secondary (i.e. non-implicit) swap chain, reset the most recently used swap chain there if 0, then also lock here and grab a proper reference to the most recently used swap chain before calling `GetFrontBufferData`.

doitsujin (changes requested):

src/d3d9/d3d9_device.cpp
K0bin: Fixed by adding a DeviceLock to Device::GetFrontBufferData. Our DeviceLock is re-entrant, so that works.

PR #3870, opened 2024-02-24
[util] Disable hideNvidiaGpu for The Finals

Blisto91:
Prevents a crash on Nvidia when XeSS is selected in the graphics menu.
Probably some kind of mismatch when vendor Id's doesn't match causing trouble as it could also be worked around in Proton Wine with `WINE_HIDE_NVIDIA_GPU=1`. Besides just enabling Nvapi that is.


Blisto91:
This randomly stopped crashing all of a sudden so not completely sure what is up with it now. Same for another user i wrote a bit with.

PR #3877, opened 2024-03-01
Initialize heap budget

talkingerbil:
Currently the heap budget is 0 at startup and it never receives a non-zero value.

### DIFF
--- a/src/dxvk/dxvk_memory.cpp
+++ b/src/dxvk/dxvk_memory.cpp
@@ -188,7 +188,7 @@ namespace dxvk {
     for (uint32_t i = 0; i < m_memProps.memoryHeapCount; i++) {
       m_memHeaps[i].properties = m_memProps.memoryHeaps[i];
       m_memHeaps[i].stats      = DxvkMemoryStats { 0, 0 };
-      m_memHeaps[i].budget     = 0;
+      m_memHeaps[i].budget     = m_memHeaps[i].properties.size;
     }
     
     for (uint32_t i = 0; i < m_memProps.memoryTypeCount; i++) {
@@ -547,12 +547,8 @@ namespace dxvk {
   bool DxvkMemoryAllocator::shouldFreeEmptyChunks(
     const DxvkMemoryHeap*       heap,
           VkDeviceSize          allocationSize) const {
-    VkDeviceSize budget = heap->budget;
 
-    if (!budget)
-      budget = (heap->properties.size * 4) / 5;
-
-    return heap->stats.memoryAllocated + allocationSize > budget;
+    return heap->stats.memoryAllocated + allocationSize > heap->budget * 4 / 5;
   }
 
 


talkingerbil:
> Hm, I think the correct fix here is to remove the budget-related code. 

Please don't. Indeed the ultimate intention is to prevent bad-behaving ("over-specifying") code from encroaching on other apps when VRAM is limited. It is specifically **because** of this bad behavior that this was investigated. If app stability is a problem, can that not be up to the user to decide? My admittedly limited testing has been on AMD only and this has worked flawlessly. 

If FPS is the only concern, allowing allocs to end up using sysram should be up to the user. If they can deal with the FPS hit, so be it. It's a success if the overall system doesn't suffer too.

The few apps I've looked at all treat the DXGI limits as guidelines but do cut back their usage even though they do continue to over-specify. 

> IIRC the intention here was to have a hard upper limit for Nvidia HVV allocations since there used to be all sorts of bugs with it, 

I totally get your point here (and am somehow not surprised). Might future NVK be better behaved? As far as I've been able to see thus far, "upper limit" right now is just system max.

> don't want to enforce this based on some arbitrary number that can change dynamically

These heap sizes don't appear to be changing dynamically?? As far as device heap goes, when RADV is driving the budget stays fixed at max available unless altered by `override_vram_size`. Neither heap type seems to change during runtime.

The single budget check in that cpp file currently returns an empty allocation if budget is exceeded. I've assumed this is equivalent to what would be returned in an actual hard VRAM exhaustion situation (if budget != 0, obviously). But even if it's not, apps gracefully deal with this all the time, no? Regardless of their bad behavior? So is the problem actually in what's being returned?

doitsujin:
> I totally get your point here (and am somehow not surprised). Might future NVK be better behaved? As far as I've been able to see thus far, "upper limit" right now is just system max.

What I mean is that there used to be a workaround which set the budget for HVV heaps on Nvidia to 32MB or something. We've since removed that code since it got fixed in the driver, but didn't remove the budget checks.

> It is specifically because of this bad behavior that this was investigated.

So what exactly is the problem you're trying to solve?

> These heap sizes don't appear to be changing dynamically?? 

I misread the code and it's using the size rather than budget, so yeah they won't change dynamically, but I'm still not sure what you're trying to solve here.

talkingerbil:
> So what exactly is the problem you're trying to solve?

Correcting what looks to be incomplete code, which makes sense given your explanation for an NV workaround that isn't needed anymore.

End goal? Enforce a per-app hard VRAM limit. As written now, this works very very well on my admittedly limited batch of tests with RADV.

If the app requirements necessitate bleeding into system ram, so be it. FPS and app stability is up to the user. If this works for a given app, great. If not, the user has to deal with the consequences of letting DXVK+app run amok. 

DXVK taking a bite of the VRAM budget is merely 'taxes' associated with running Windows apps in Linux.

talkingerbil:
Ah, I see what's happening now. DXVK never rejects alloc requests by the app, correct?

Well, for what it's worth, modern high end systems are still perfectly usable when sysram starts being used. I drop into the 30 FPS range at the worst. 

The ultimate point here is that DXVK is ignoring the GPU driver. 

Mid/low systems are already suffering with games that over-specify. This will at least protect the entire system GUI from lagging because of it.

PR #3895, opened 2024-03-06
[d3d9] Fix a bunch of Wine test failures

K0bin:
Fixes the following Wine test failures:

StretchRect, lighting, specular_lighting
```
visual.c:4451: Test failed: Test 6, got unexpected hr 0.
visual.c:4451: Test failed: Test 7, got unexpected hr 0.
visual.c:4451: Test failed: Test 8, got unexpected hr 0.
visual.c:4451: Test failed: Test 9, got unexpected hr 0.
visual.c:4451: Test failed: Test 14, got unexpected hr 0.
visual.c:4451: Test failed: Test 15, got unexpected hr 0.
visual.c:4451: Test failed: Test 16, got unexpected hr 0.
visual.c:4451: Test failed: Test 17, got unexpected hr 0.
visual.c:4451: Test failed: Test 22, got unexpected hr 0.
visual.c:4451: Test failed: Test 23, got unexpected hr 0.
visual.c:4451: Test failed: Test 24, got unexpected hr 0.
visual.c:4451: Test failed: Test 25, got unexpected hr 0.
visual.c:4451: Test failed: Test 30, got unexpected hr 0.
visual.c:4451: Test failed: Test 31, got unexpected hr 0.
visual.c:4451: Test failed: Test 32, got unexpected hr 0.
visual.c:4451: Test failed: Test 36, got unexpected hr 0.
visual.c:4451: Test failed: Test 38, got unexpected hr 0.
visual.c:4451: Test failed: Test 39, got unexpected hr 0.
visual.c:4451: Test failed: Test 40, got unexpected hr 0.
visual.c:4451: Test failed: Test 41, got unexpected hr 0.
visual.c:4451: Test failed: Test 42, got unexpected hr 0.
visual.c:4451: Test failed: Test 43, got unexpected hr 0.
visual.c:4451: Test failed: Test 44, got unexpected hr 0.
visual.c:4451: Test failed: Test 46, got unexpected hr 0.
visual.c:4451: Test failed: Test 47, got unexpected hr 0.
visual.c:4451: Test failed: Test 49, got unexpected hr 0.
visual.c:4451: Test failed: Test 51, got unexpected hr 0.
visual.c:4451: Test failed: Test 53, got unexpected hr 0.

visual.c:15446: Test failed: Got hr 0.
visual.c:15462: Test failed: Got hr 0.
visual.c:15466: Test failed: Got hr 0.
visual.c:15468: Test failed: Got hr 0.

visual.c:1126: Test failed: Expected color 0x00000000 at location (160, 120), got 0x00ffffff, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (320, 120), got 0x00adadad, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (480, 120), got 0x005a5a5a, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (160, 240), got 0x00ffffff, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (320, 240), got 0x00d1d1d1, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (480, 240), got 0x00636363, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (160, 360), got 0x00ffffff, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (320, 360), got 0x00adadad, case 7.
visual.c:1126: Test failed: Expected color 0x00000000 at location (480, 360), got 0x005a5a5a, case 7.
```

Actually matches the documentation here: https://learn.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3ddevice9-stretchrect

### DIFF
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -278,6 +278,8 @@ namespace dxvk {
 
     auto& options = m_parent->GetOptions();
 
+    const VkPhysicalDeviceLimits& limits = m_adapter->deviceProperties().limits;
+
     // TODO: Actually care about what the adapter supports here.
     // ^ For Intel and older cards most likely here.
 
@@ -538,7 +540,7 @@ namespace dxvk {
     // Max Vertex Blend Matrix Index
     pCaps->MaxVertexBlendMatrixIndex = 0;
     // Max Point Size
-    pCaps->MaxPointSize              = 256.0f;
+    pCaps->MaxPointSize              = limits.pointSizeRange[1];
     // Max Primitive Count
     pCaps->MaxPrimitiveCount         = 0x00555555;
     // Max Vertex Index
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -7767,6 +7767,8 @@ namespace dxvk {
     rs[D3DRS_CLIPPLANEENABLE] = 0;
     m_flags.set(D3D9DeviceFlag::DirtyClipPlanes);
 
+    const VkPhysicalDeviceLimits& limits = m_dxvkDevice->adapter()->deviceProperties().limits;
+
     rs[D3DRS_POINTSPRITEENABLE]          = FALSE;
     rs[D3DRS_POINTSCALEENABLE]           = FALSE;
     rs[D3DRS_POINTSCALE_A]               = bit::cast<DWORD>(1.0f);
@@ -7774,7 +7776,7 @@ namespace dxvk {
     rs[D3DRS_POINTSCALE_C]               = bit::cast<DWORD>(0.0f);
     rs[D3DRS_POINTSIZE]                  = bit::cast<DWORD>(1.0f);
     rs[D3DRS_POINTSIZE_MIN]              = bit::cast<DWORD>(1.0f);
-    rs[D3DRS_POINTSIZE_MAX]              = bit::cast<DWORD>(64.0f);
+    rs[D3DRS_POINTSIZE_MAX]              = bit::cast<DWORD>(limits.pointSizeRange[1]);
     UpdatePushConstant<D3D9RenderStateItem::PointSize>();
     UpdatePushConstant<D3D9RenderStateItem::PointSizeMin>();
     UpdatePushConstant<D3D9RenderStateItem::PointSizeMax>();
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -2953,14 +2953,19 @@ namespace dxvk {
     auto slice = dst->GetBufferSlice<D3D9_COMMON_BUFFER_TYPE_REAL>();
          slice = slice.subSlice(offset, slice.length() - offset);
 
+    D3D9CompactVertexElements elements;
+    for (const D3DVERTEXELEMENT9& element : decl->GetElements()) {
+      elements.emplace_back(element);
+    }
+
     EmitCs([this,
-      cVertexElements = decl->GetElements(),
+      cVertexElements = std::move(elements),
       cVertexCount    = VertexCount,
       cStartIndex     = SrcStartIndex,
       cInstanceCount  = GetInstanceCount(),
       cBufferSlice    = slice
     ](DxvkContext* ctx) mutable {
-      Rc<DxvkShader> shader = m_swvpEmulator.GetShaderModule(this, cVertexElements);
+      Rc<DxvkShader> shader = m_swvpEmulator.GetShaderModule(this, std::move(cVertexElements));
 
       auto drawInfo = GenerateDrawInfo(D3DPT_POINTLIST, cVertexCount, cInstanceCount);
 
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -2954,13 +2954,13 @@ namespace dxvk {
          slice = slice.subSlice(offset, slice.length() - offset);
 
     EmitCs([this,
-      cDecl          = ref(decl),
-      cVertexCount   = VertexCount,
-      cStartIndex    = SrcStartIndex,
-      cInstanceCount = GetInstanceCount(),
-      cBufferSlice   = slice
+      cVertexElements = decl->GetElements(),
+      cVertexCount    = VertexCount,
+      cStartIndex     = SrcStartIndex,
+      cInstanceCount  = GetInstanceCount(),
+      cBufferSlice    = slice
     ](DxvkContext* ctx) mutable {
-      Rc<DxvkShader> shader = m_swvpEmulator.GetShaderModule(this, cDecl);
+      Rc<DxvkShader> shader = m_swvpEmulator.GetShaderModule(this, cVertexElements);
 
       auto drawInfo = GenerateDrawInfo(D3DPT_POINTLIST, cVertexCount, cInstanceCount);
 
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1144,6 +1144,9 @@ namespace dxvk {
     if (unlikely((srcSubresource.aspectMask & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT)) && m_flags.test(D3D9DeviceFlag::InScene)))
       return D3DERR_INVALIDCALL;
 
+    if (unlikely(Filter != D3DTEXF_NONE && Filter != D3DTEXF_LINEAR && Filter != D3DTEXF_POINT))
+      return D3DERR_INVALIDCALL;
+
     VkExtent3D srcExtent = srcImage->mipLevelExtent(srcSubresource.mipLevel);
     VkExtent3D dstExtent = dstImage->mipLevelExtent(dstSubresource.mipLevel);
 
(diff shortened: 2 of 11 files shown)


mrdeathjr28:
hi

which apps are affected with this?

src/d3d9/d3d9_device.cpp
```diff
@@ -2945,13 +2945,13 @@ namespace dxvk {
          slice = slice.subSlice(offset, slice.length() - offset);
 
     EmitCs([this,
-      cDecl          = ref(decl),
-      cVertexCount   = VertexCount,
-      cStartIndex    = SrcStartIndex,
-      cInstanceCount = GetInstanceCount(),
-      cBufferSlice   = slice
+      cVertexElements = decl->GetElements(),
```
doitsujin, line 2948: This is kind of not ideal since `D3D9VertexElements` is a vector, but unless games spam this thousands of times per frame it *should* be fine?

K0bin: I know but the only game I know off the top of my head that uses ProcessVertices and that's Sims 2. So performance wasn't the greatest priority.

WinterSnowfall: FWIW, it's a bit more common in d3d8. I've seen it show up in quite a few traces, not quite at the levels of thousands per frame, but there's definitely some occasional heavy use of it out there as well.

K0bin:
229 failures left.

K0bin:
> 82d8:stateblock: 14738 tests executed (0 marked as todo, 0 as flaky, 0 failures), 0 skipped.

We now pass all state block tests.

src/d3d9/d3d9_device.cpp
```diff
@@ -7711,7 +7711,7 @@ namespace dxvk {
     rs[D3DRS_POINTSCALE_C]               = bit::cast<DWORD>(0.0f);
     rs[D3DRS_POINTSIZE]                  = bit::cast<DWORD>(1.0f);
     rs[D3DRS_POINTSIZE_MIN]              = bit::cast<DWORD>(1.0f);
-    rs[D3DRS_POINTSIZE_MAX]              = bit::cast<DWORD>(64.0f);
+    rs[D3DRS_POINTSIZE_MAX]              = bit::cast<DWORD>(256.0f);
```
DadSchoorse, line 7714: Shouldn't you query this from vk device properties? Not all devices support 256.

K0bin: Good point.

src/d3d9/d3d9_device.cpp
```diff
@@ -1221,8 +1224,41 @@ namespace dxvk {
...
+
     // Copies would only work if the extents match. (ie. no stretching)
     bool stretch = srcCopyExtent != dstCopyExtent;
+
+    bool dstHasRTUsage = (dstTextureInfo->Desc()->Usage & (D3DUSAGE_RENDERTARGET | D3DUSAGE_DEPTHSTENCIL)) != 0;
+    if (stretch) {
+      if (unlikely(pSourceSurface == pDestSurface))
+        return D3DERR_INVALIDCALL;
+
+      if (unlikely(IsDepthFormat(dstFormat)))
```
WinterSnowfall, line 1251: Small nit, but might as well `dstIsDepth` directly here.

K0bin: Done

src/d3d9/d3d9_device.cpp
K0bin: Done

K0bin:
I think I'll try to land this first batch before looking into more of the tests.

src/d3d9/d3d9_device.cpp
```diff
@@ -1221,8 +1224,41 @@ namespace dxvk {
...
+        return D3DERR_INVALIDCALL;
+
+      if (unlikely(dstIsDepth))
+        return D3DERR_INVALIDCALL;
+
+      if (unlikely(!dstHasRTUsage))
+        return D3DERR_INVALIDCALL;
+    } else {
+      bool srcHasRTUsage = (srcTextureInfo->Desc()->Usage & (D3DUSAGE_RENDERTARGET | D3DUSAGE_DEPTHSTENCIL)) != 0;
+      if (unlikely(!dstHasRTUsage && (dstTextureInfo->GetType() != D3DRTYPE_SURFACE || srcTextureInfo->GetType() != D3DRTYPE_SURFACE || srcHasRTUsage)))
```
Joshua-Ashton, line 1258: This is a long if to do similar stuff that is more segmented above.

src/util/util_matrix.cpp
```diff
@@ -205,6 +205,10 @@ namespace dxvk {
     Vector4 dot0    = { m[0] * row0 };
     float dot1      = (dot0.x + dot0.y) + (dot0.z + dot0.w);
 
+    if (dot1 == 0.0f) {
```
Joshua-Ashton, line 208: unlikely

src/d3d9/d3d9_device.cpp
```diff
@@ -7704,14 +7704,16 @@ namespace dxvk {
...
+
     rs[D3DRS_POINTSPRITEENABLE]          = FALSE;
     rs[D3DRS_POINTSCALEENABLE]           = FALSE;
     rs[D3DRS_POINTSCALE_A]               = bit::cast<DWORD>(1.0f);
     rs[D3DRS_POINTSCALE_B]               = bit::cast<DWORD>(0.0f);
     rs[D3DRS_POINTSCALE_C]               = bit::cast<DWORD>(0.0f);
     rs[D3DRS_POINTSIZE]                  = bit::cast<DWORD>(1.0f);
     rs[D3DRS_POINTSIZE_MIN]              = bit::cast<DWORD>(1.0f);
-    rs[D3DRS_POINTSIZE_MAX]              = bit::cast<DWORD>(64.0f);
+    rs[D3DRS_POINTSIZE_MAX]              = bit::cast<DWORD>(limits.pointSizeRange[1]);
```
Joshua-Ashton, line 7721: "The default render state value has to match what we report in the device caps."

And does that match what any DX9 implementation exposes? We should check that.

Joshua-Ashton (changes requested):

src/d3d9/d3d9_device.cpp
K0bin: MaxPointSize is 8192 in the D3D8 dump that Winter did. (With a NVIDIA GeForce4 Ti 4800 SE).

The Vulkan max point size range is generally much lower unless you're using an AMD GPU.
https://vulkan.gpuinfo.org/displaydevicelimit.php?name=pointSizeRange[1]&platform=linux

WinterSnowfall: For what it's worth that still is the case for Nvidia, namely MaxPointSize is 8192 even with modern drivers/cards. AMD sets it somewhat lower, e.g. 256 on @Blisto91's 7900XTX.

src/d3d9/d3d9_stateblock.h
```diff
@@ -211,11 +216,19 @@ namespace dxvk {
...
-            vbo.stride);
+          if constexpr (!IgnoreStreamOffset) {
+            dst->SetStreamSource(
+              idx,
+              vbo.vertexBuffer.ptr(),
+              vbo.offset,
+              vbo.stride);
+          } else {
+            // For whatever reason, D3D9 doesn't capture the stream offset
+            dst->SetStreamSourceWithoutOffset(
```
doitsujin, line 227: Where does the offset we apply to the context come from if it is never captured?

src/d3d9/d3d9_swvp_emu.h
```diff
@@ -11,26 +11,41 @@ namespace dxvk {
   class D3D9VertexDecl;
   class D3D9DeviceEx;
 
+  struct D3D9CompactVertexElement {
+      uint8_t Stream : 4;
```
doitsujin, line 15: the bitfield needs to use `uint16_t`, otherwise elements straddling byte boundaries get weird (and we already have a uint16 in this anyway).

src/d3d9/d3d9_stateblock.h
K0bin: Probably whatever offset was set on the device at the time.

K0bin: Okay it's pretty weird:

It **only** ignores the offset when calling `IDirect3DStateBlock9::Capture`.
That means, that both the initial capture when creating a state block using `IDirect3DDevice9::CreateStateBlock` as well as capturing the stream source by calling `IDirect3DDevice9::SetStreamSource` in between `IDirect3DDevice9::BeginStateBlock` and `IDirect3DDevice9::EndStateBlock` will capture the offset as one would normally expect.

`IDirect3DStateBlock9::Apply` only touches the stream sources if the state block has been created using `IDirect3DDevice9::CreateStateBlock(D3DSBT_VERTEXSTATE)` or by calling `IDirect3DDevice9::SetStreamSource` in between `IDirect3DDevice9::BeginStateBlock` and `IDirect3DDevice9::EndStateBlock`. Both of those capture the offset just fine.

So the only thing that does not work is **updating** the offset of a stream source that is capturing in a state block by calling `Capture`. So to answer your question: it's not strictly correct that the offset is "never captured" and the offset we apply comes from capturing it earlier.

I hope that was somewhat understandable. 😅

src/d3d9/d3d9_adapter.cpp
```diff
@@ -531,7 +533,7 @@ namespace dxvk {
     // Max Vertex Blend Matrix Index
     pCaps->MaxVertexBlendMatrixIndex = 0;
     // Max Point Size
-    pCaps->MaxPointSize              = 256.0f;
+    pCaps->MaxPointSize              = limits.pointSizeRange[1];
```
doitsujin, line 536: Are games going to be fine with this not being an integer value / power-of-two? Vulkan drivers will report the upper limit of their fixed-point representation here, which is generally something of the format `(2^n - 1)/(2^m)`.

src/util/util_matrix.cpp
```diff
@@ -205,6 +205,10 @@ namespace dxvk {
     Vector4 dot0    = { m[0] * row0 };
     float dot1      = (dot0.x + dot0.y) + (dot0.z + dot0.w);
 
+    if (unlikely(std::abs(dot1) <= 0.000001f)) {
```
doitsujin, line 208: which part of the API relies on this and do we have a way to verify native behaviour?

src/d3d9/d3d9_device.cpp
```diff
@@ -1221,8 +1224,43 @@ namespace dxvk {
...
+      if (unlikely(dstIsDepth))
+        return D3DERR_INVALIDCALL;
+
+      if (unlikely(!dstHasRTUsage))
+        return D3DERR_INVALIDCALL;
+    } else {
+      bool srcIsSurface = srcTextureInfo->GetType() == D3DRTYPE_SURFACE;
+      bool dstIsSurface = dstTextureInfo->GetType() == D3DRTYPE_SURFACE;
+      bool srcHasRTUsage = (srcTextureInfo->Desc()->Usage & (D3DUSAGE_RENDERTARGET | D3DUSAGE_DEPTHSTENCIL)) != 0;
+      if (unlikely(!dstHasRTUsage && (!dstIsSurface || !srcIsSurface || srcHasRTUsage)))
```
doitsujin, line 1260: What exactly does the `D3DRTYPE_SURFACE` requirement accomplish here? What about textures with only one mip?

doitsujin (commented):
Couple questions.

In general I'd prefer to have unrelated fixes to be PR'd separately though, even if it means a fair bit of PR spam. Makes it easier to review and most of the commits in here would be good to go as-is anyway.

src/d3d9/d3d9_adapter.cpp
K0bin: pCaps->MaxPointSize is a float, so you'd hope that games are fine if that's not an integer value. At least I don't have anything indicating they wouldn't be. We could ofc round it down to the nearest integer or power of two but right now I don't see any reason to do that so I'd prefer to wait until we hit a game that actually needs it.

src/util/util_matrix.cpp
K0bin: It's specifically about [calculating the normal matrix from the world matrix](https://github.com/doitsujin/dxvk/blob/master/src/d3d9/d3d9_device.cpp#L7334).

Wine tests what D3D9 does with a singular matrix [here](https://github.com/wine-mirror/wine/blob/master/dlls/d3d9/tests/visual.c#L604).

Would you prefer it if I check the determinant separately even if that means calculating it twice?

src/d3d9/d3d9_device.cpp
K0bin: This is implementing the no stretching table: 

table
[Coming from the docs](https://learn.microsoft.com/en-us/windows/win32/api/d3d9/nf-d3d9-idirect3ddevice9-stretchrect)

K0bin: I also added a  comment explaining it.

PR #3904, opened 2024-03-11
[util] Set cachedDynamicResources for Guild Wars 2

Blisto91:
Helps the games CPU bound performance on some setups so it gets closer to Windows.

See example comparison without and with below. Screenshots are taken by @LankyFranky89
<details>
  
image
image
</details>


Blisto91:
Actually this may help the specific setup above because they seem to have performance issues with rebar on their laptop in general. I'll leave it up to you whether that invalidates a general config here. I don't personally see any gains from it.

Blisto91:
Then I'll close it as I now believe this is on specific setups

PR #3909, opened 2024-03-14
Implement D3D11DXGIDevice::EnqueueSetEvent

BunniKaitlyn:
Not sure if this is entirely accurate, but works fine in my limited testing. Please let me know if this is correct or if there is any changes I should make.

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -3411,8 +3411,9 @@ namespace dxvk {
 
 
   HRESULT STDMETHODCALLTYPE D3D11DXGIDevice::EnqueueSetEvent(HANDLE hEvent) {
-    Logger::err("D3D11DXGIDevice::EnqueueSetEvent: Not implemented");
-    return DXGI_ERROR_UNSUPPORTED;           
+    auto immediateContext = m_d3d11Device.GetContext();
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Can you let us know what you are testing with if possible? What is using/needs this?

Joshua-Ashton:
I am guessing the Oculus runtime in Wine + DXVK you mentioned on the other MR.

Joshua-Ashton (approved):
Looking at https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgidevice2-enqueuesetevent 

The impl makes sense to me, but seems weird that there are two ways of doing this :frog: sus

BunniKaitlyn:
> I am guessing the Oculus runtime in Wine + DXVK you mentioned on the other MR.

Yeah my bad, all testing was done on the Oculus runtime, which requires both of these PRs. Works fine as seen here: https://twitter.com/BunniKaitlyn/status/1768130826506531034

ViNi-Arco:
> Works fine as seen here: https://twitter.com/BunniKaitlyn/status/1768130826506531034

That's interesting, you could pass env WINEDEBUG=fixme-all to suppress all those fixme and gain a bit of performance..

Issue #3910, opened 2024-03-15
scRGB HDR seems generally broken on the OLED Steam Deck


PeterTh:
When using a `R16G16B16A16_FLOAT` swapchain buffer for HDR, in Windows, the compositor expects **scRGB** color space (`DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709`).
I.e. (1,1,1) should be about 80 nits, and 1000 nits is something like (12,12,12).

When presenting the same frame on the OLED Steam Deck, it is completely blown out (and the gamma is also way off).

Note: I'm not sure if this is actually a DXVK bug or a bug elsewhere in the display chain -- that's hard for me to tell. If I should report it elsewhere please let me know.

### Software information
Unreleased game. But I really don't think this is a game-specific problem.

### System information
Steam Deck OLED

I haven't included an apitrace or log, since I don't think it makes much sense or helps for this type of problem.

doitsujin:
Do we actually get an `R16G16B16A16_SFLOAT` swap chain on the Vulkan side or is it `R10G10B10A2`? The D3D11 log should include the swapchain format.

Either way this is probably something for @Joshua-Ashton to look at. scRGB is more or less untested and known to be broken with Special K, but I don't have an HDR-capable setup to test things with.

PeterTh:
Thanks for the fast replies!

> Are you calling SetColorSpace1, or relying on some implicit behaviour we do not implement?

We are not (at least not right now, I'll implement that to check to see if that fixes it).

We're currently relying on the ([documented](https://learn.microsoft.com/en-us/windows/win32/direct3darticles/high-dynamic-range#option-1-use-fp16-pixel-format-and-scrgb-color-space)) behaviour that `R16G16B16A16_FLOAT` swapchains are automatically `DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709`.

If setting the color space manually does not solve the issue I'll send you an email.

Joshua-Ashton:
Yeah we are not doing that rn. We should, I am going to sleep now so I'll do that tomorrow.

Joshua-Ashton:
I was beaten to it =)

PeterTh:
That also was the issue: manually setting `DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709` solved the problem.

doitsujin:
Should be fixed in current master, let me know if there are issues with your use case.

PeterTh:
[...]

PR #3911, opened 2024-03-15
Implicitly set HDR color space for RGBA16_FLOAT swap chains

doitsujin:
See #3910.

### DIFF
--- a/src/dxgi/dxgi_swapchain.cpp
+++ b/src/dxgi/dxgi_swapchain.cpp
@@ -32,6 +32,9 @@ namespace dxvk {
     // Apply initial window mode and fullscreen state
     if (!m_descFs.Windowed && FAILED(EnterFullscreenMode(nullptr)))
       throw DxvkError("DXGI: Failed to set initial fullscreen state");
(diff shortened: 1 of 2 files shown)


Joshua-Ashton:
We should test how this interacts with ResizeBuffers

doitsujin:
Oh right, that does allow changing the format on the fly for some stupid reason. Great design, thanks Microsoft.

PeterTh:
FWIW, we do exactly that (when someone toggles HDR in-game).

doitsujin:
Did some testing on Windows, looks like RGBA16 **only** exposes scRGB support (even on SDR systems), and any explicitly set color space persists even through incompatible format changes if we change back to that format (e.g. when enabling HDR10 on RGB10A2).

It seems to be rather unreliable though, at least on my AMD setup, I've seen it break completely a couple of times to the point where it's just hard-stuck in SDR mode, or use the wrong color space for exactly one frame before figuring out the correct one.

Either way, updated PR to hopefully behave as expected.

PR #3921, opened 2024-03-21
[d3d9] Handle mismatching vertex layout and vbo strides

### LINKED ISSUE #3919: [d3d9] Final Fantasy XIII (dxvk 2.3.1 regression?)
Going from DXVK version 2.3 to 2.3.1 artifacts / vertex explosion appeared. I think it's from 2d (font, hud,...) elements.

v.2.3
ff13-dxvk-v 2 3

v.2.3.1
ff13-dxvk-v 2 3 1

v.2.3
ff13-dxvk-v 2 3-ingame

v.2.3.1
ff13-dxvk-v 2 3 1-ingame

P.S.: Yes, DXVK HUD appears with `DXVK_HUD=full` , so it's using DXVK. In screenshots I only was using MANGOHUD.

### Software information
Final Fantasy XIII

### System information
- GPU: RTX 4080
- Driver: 550
- Wine version:  proton-8.0-5, proton-9-exp 
- DXVK version:  2.3 - 2.3.1

### Apitrace file(s)
https://drive.google.com/file/d/1kkeu1T3xLKVRgudiKBhJW1WZKkYNgBfg/view?usp=sharing
(made in linux/wine, I hope it's OK)

### Log files
[game.log](https://github.com/doitsujin/dxvk/files/14693907/game.log)

K0bin:
Fixes #3919

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5158,7 +5158,7 @@ namespace dxvk {
     return D3D_OK;
   }
 
-  
+
 
   void D3D9DeviceEx::UploadDynamicSysmemBuffers(
           UINT&                   FirstVertexIndex,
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -5253,11 +5266,22 @@ namespace dxvk {
...
+        uint32_t offset = (BaseVertexIndex + FirstVertexIndex) * vboStride + m_state.vertexBuffers[i].offset;
 
         uint8_t* data = reinterpret_cast<uint8_t*>(upSlice.mapPtr) + vboUPBufferOffsets[i];
         uint8_t* src = reinterpret_cast<uint8_t*>(vbo->GetMappedSlice().mapPtr) + offset;
-        std::memcpy(data, src, vboUPBufferSizes[i]);
+        if (vertexStride == vboStride) {
+          std::memcpy(data, src, vboUPBufferSizes[i]);
+        } else {
+          const uint32_t vertexBufferSize = vbo->Desc()->Size;
+          const uint32_t boundVertexBufferRange = vertexBufferSize - offset;
```
Joshua-Ashton, line 5278: What if offset is > size?

src/d3d9/d3d9_device.cpp
```diff
@@ -5202,13 +5202,26 @@ namespace dxvk {
...
-        vboUPBufferSizes[i] = vertexDataSize;
-        upBufferSize += vertexDataSize;
+
+      uint32_t vertexDataSize;
+      if (unlikely(vboOffset + vertexOffset > vertexBufferSize)) {
+        // All vertices are out of bounds
+        vertexDataSize = 0;
+      } else if (unlikely(vboOffset + vertexOffset + NumVertices * vboStride > vertexBufferSize)) {
+        // Some vertices are (partially) out of bounds
+        uint32_t boundVertexBufferRange = vertexBufferSize - vboOffset;
```
Joshua-Ashton, line 5216: Same here

Joshua-Ashton:
This seems like a mess

K0bin:
I pushed an updated version of this PR that should hopefully be a bit easier to read. It's still quite complex because we have to be robust against going out of bounds but it's a bit less messy now in my opinion.

src/d3d9/d3d9_device.cpp
```diff
@@ -5240,30 +5273,34 @@ namespace dxvk {
...
-          ](DxvkContext* ctx) {
-            ctx->bindVertexBuffer(cStream, DxvkBufferSlice(), 0);
-          });
-          m_flags.set(D3D9DeviceFlag::DirtyVertexBuffers);
-          continue;
-        }
-
-        auto* vbo = GetCommonBuffer(m_state.vertexBuffers[i].vertexBuffer);
+        const VBOCopy& copy = vboCopies[i];
+        const auto* vbo = GetCommonBuffer(m_state.vertexBuffers[i].vertexBuffer);
```
PatrickvL, line 5277: This `vbo` variable could be moved inside below if block, to narrow its scope.

src/d3d9/d3d9_device.cpp
```diff
@@ -5190,25 +5190,58 @@ namespace dxvk {
...
 
+    struct VBOCopy {
+      uint32_t srcOffset;
+      uint32_t dstOffset;
+      uint32_t copyBufferLength;
+      uint32_t copyElementCount;
+      uint32_t copyElementSize;
+      uint32_t copyElementStride;
+    };
     uint32_t upBufferSize = 0;
```
PatrickvL, line 5201: Usage of this variable implies it's a rolling value. Perhaps renaming it into `upBufferRollingSize` (or something comparable) would make this fact more clear. (Or perhaps there already is a naming convention in similar code that could be applied here too?)

src/d3d9/d3d9_device.cpp
```diff
@@ -5190,25 +5190,58 @@ namespace dxvk {
...
-    std::array<uint32_t, caps::MaxStreams> vboUPBufferOffsets = {};
-    std::array<uint32_t, caps::MaxStreams> vboUPBufferSizes = {};
-    for (uint32_t i = 0; i < caps::MaxStreams && dynamicSysmemVBOs; i++) {
-      vboUPBufferOffsets[i] = upBufferSize;
+    std::array<VBOCopy, caps::MaxStreams> vboCopies = {};
 
+    for (uint32_t i = 0; i < caps::MaxStreams && dynamicSysmemVBOs; i++) {
```
PatrickvL, line 5204: In the other for loop over `caps.MaxStreams` below, the check on `dynamicSysmemVBOs` is done in a preceding `if`, not inside the if condition as happens here. 
Best to be consistent, so either move this loop inside an if block, or update below loop to follow the same approach as here.

src/d3d9/d3d9_device.cpp
K0bin: What's a rolling value?

`upBufferSize` just sums up the size of the vertex data for all vertex buffers and the index buffer so we can then allocate a single UpBuffer slice rather than needing multiple slices.

PatrickvL: Indeed. It starts at zero and is incremented in each for loop iteration. 
Of course, to you it is probably obvious this is the way this works, but this usage can be seen as a pattern that happens elsewhere as well, and capturing that fact via a naming convention would make this clearer to a reader/maintainer of the code base.
Maybe "rolling" is not the most accurate terminology, perhaps something like "incremental" is the better word to describe this usage, but that's longer.
Feel free to ignore this reasoning, it's just a suggestion.

K0bin: I could call it something like `totalUpBufferSize`. How about that?

src/d3d9/d3d9_device.cpp
```diff
@@ -5190,25 +5190,58 @@ namespace dxvk {
...
-        upBufferSize += vertexDataSize;
+      const uint32_t srcOffset = vboOffset + vertexOffset;
+
+      if (unlikely(srcOffset > vertexBufferSize)) {
+        // All vertices are out of bounds
+        vboCopies[i].copyBufferLength = 0;
+      } else if (unlikely(srcOffset + elementCount * srcStride > vertexBufferSize)) {
+        // Some vertices are (partially) out of bounds
+        uint32_t boundVertexBufferRange = vertexBufferSize - vboOffset;
+        elementCount = boundVertexBufferRange / srcStride;
```
doitsujin, line 5229: what about the case where `stride > size` and the vbo size is something like `(n - 1) * stride + size`? All vertex data would be valid in that case.

doitsujin (approved):

PR #3924, opened 2024-03-22
[dxgi] Force vendor ID change when XeSS is detected on an Intel GPU

cgbowman:
Games using libxess.dll or wrapper modules will crash. To work around this, the vendor ID is set to -1 to avoid using the XeSS module.

This change was tested with various XeSS titles, including Shadow of the Tomb Raider, Hitman 3, Dying Light 2, and Diablo IV (I ensured the relevant driconf changes were made so that the mesa driver isn't the one setting the vendor ID to '-1').

Note that Dying Light 2 created its own DX11 wrapper module, libxess_dx11.dll, which uses some functionality from the libxess.dll module and consequently causes a hang. I don't anticipate many titles doing this, so it's included in the module check.

### DIFF
--- a/src/dxgi/dxgi_options.cpp
+++ b/src/dxgi/dxgi_options.cpp
@@ -26,6 +26,21 @@ namespace dxvk {
     return id;
   }
 
+  /* First generation XeSS causes crash on proton for Intel due to missing
+   * Intel interface. Avoid crash by pretending to be non-Intel if the
+   * libxess.dll module is loaded by an application.
+   */
+  static bool isXessUsed() {
+#ifdef _WIN32
+      if (GetModuleHandleA("libxess") != nullptr ||
+          GetModuleHandleA("libxess_dx11") != nullptr)
+        return true;
+      else
+        return false;
+#else
+      return false;
+#endif
+  }
 
   static bool isNvapiEnabled() {
     return env::getEnvVar("DXVK_ENABLE_NVAPI") == "1";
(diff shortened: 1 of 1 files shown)


doitsujin:
Could we do this by overriding the default behaviour for the `hideIntelGpu` option (in `dxgi_options.cpp`) instead, like we do for Nvidia? The mechanism is there for exactly this purpose and I'd like these things to stay in one place if possible.

cgbowman:
I'll do some testing with overriding the default behavior in `dxgi_options.cpp`. Thanks for the pointer, I'll try to keep the changes in line with the existing mechanisms.

So far, during my testing with this patch across a few titles, I've not encountered a case where the XeSS library was loaded after the DXGI adapters, but as you say, I suppose it shouldn't be ruled out. As is with the case of `libxess_dx11.dll` for Dying Light 2, it could be possible that another game does something different.

Would you feel more comfortable with a larger test size of XeSS titles?

cgbowman:
With this update, I feel it's a bit more in line with using the options, but I'm still not sure how I feel about the additional check for an Intel vendor ID. 

I didn't want to overstep and forcefully change the vendor ID when there's a case of another GPU vendor being used and the XeSS module is detected, so I felt the extra check was necessary to ensure we only enter the flow for fallback vendor/device IDs when it's an Intel device and the XeSS module is detected.

Sometime later, we'll likely want to introduce a version check of the XeSS lib to distinguish between the initial, crashing XeSS modules that games currently have and later versions that are expected to have working implementations in the Mesa driver. So likely we can use the ```isXessUsed()``` check for that later.

Let me know if this update aligns more with your expectations, else we'll continue to rework it until it does.

src/dxgi/dxgi_adapter.cpp
```diff
@@ -285,7 +285,8 @@ namespace dxvk {
       ? std::string(deviceProp.deviceName)
       : options->customDeviceDesc;
 
-    if (options->customVendorId < 0) {
+    if (options->customVendorId < 0
+    || (deviceProp.vendorID == uint16_t(DxvkGpuVendor::Intel) && options->hasXess)) {
```
doitsujin, line 289: This check seems redundant unless the user *explicitly* overrides the vendor ID? I don't think we should try to be smart and work around broken user configs, if anything that makes it harder to test things on the fly.

doitsujin (commented):
Just a nit.

src/dxgi/dxgi_adapter.cpp
cgbowman: Ah, I didn't realize it defaulted to -1, that works perfect then! I'll remove & re-upload.

cgbowman:
Hi @doitsujin, anything else you'd like to change here about the latest updates or does it look good to you?

doitsujin:
Sorry, I was away over the easter weekend and haven't gotten around to this yet.

cgbowman:
No worries, just checking in is all :)

PR #3942, opened 2024-04-06
[util] Remove framerate limiter for Nier Replicant

SuperSamus:
### Context

Without mods, Nier Replicant runs faster when going above 60 FPS.
The game had an official patch that implemented a framerate limiter to prevent this. This limiter is terrible, because it's not a stable 60 FPS, but a weird 57-58 FPS. There's no way to disable this in-game, it has to be done by editing a config file or through a mod.

### So, why remove the default frame limiter from DXVK?

- In the default case (a user plays the game as it is), it does nothing, since 57-58 is lower than 60.
- If a user is going out of their way to edit the config file, why would they assume that DXVK already provides a frame limiter? [Guides](https://www.pcgamingwiki.com/wiki/NieR_Replicant#Framerate_limited_to_57.7E58_FPS) already tell the user to set up a frame limiter.
- They are using [Special K in order to use a mod to play at high refresh rates at normal speed](https://wiki.special-k.info/SpecialK/Custom/Replicant). In this case, DXVK's default limiter is harmful, since it is not documented that it's there by default.

Since this default limiter is useless in the first two cases and harmful in the third, I think it should be removed.
The alternative would be to document this (e.g. in PCGamingWiki), but this would bloat the instructions to the point that nothing is being simplified...
- "If you want a stable 60 FPS, [...], then set up a frame limiter, but don't do the last step if you are using DXVK/Proton, because in that case it's already there"
- "To play at higher framerates at normal speed, install Special K [...], and if are using DXVK/Proton, also do these things to disable its default 60 FPS cap: [...]"

Especially because that the game isn't broken in the default case, I don't think DXVK should tamper with these things in a way that requires documentation to revert: default settings should be there only when they are only an advantage, not when they are opinionated.
I say opinionated because this is basically an undocumented half-mod M that works only if a user does A (with guides already telling the user to do M), and causes conflicts when a user does B.

And considering that the only advantage to having this default in DXVK is not having to set manually `DXVK_FRAME_RATE=60`, I think the cons outweigh the pros.

Tested Special K's mod to play at higher refresh rates on Linux.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -122,8 +122,6 @@ namespace dxvk {
     }} },
     /* NieR Replicant                             */
     { R"(\\NieR Replicant ver\.1\.22474487139\.exe)", {{
-      { "dxgi.syncInterval",                "1"    },
-      { "dxgi.maxFrameRate",                "60"   },
       { "d3d11.cachedDynamicResources",     "vi"   },
     }} },
     /* SteamVR performance test                   */


mbriar:
Doesn't the "dxgi.syncInterval=1 part of the current config make it so that the default does actually work properly at 60 fps?

SuperSamus:
I've been negligent with my testing...

> Doesn't the "dxgi.syncInterval=1 part of the current config make it so that the default does actually work properly at 60 fps?
With just that (and no DXVK frame limiter), on vanilla you get 61~62 FPS (on my 144 Hz screen), which is an improvement.
And you don't get conflicts with Special K's high framerate mod.

Shall I just remove the 60 FPS cap, then?

doitsujin:
We've had this discussion before, the reason why we decided to enable the 60 lock for this game is to reduce the amount of user configuration necessary to even run this game in an acceptable way at all. Once you start throwing mods at it, you can probably also set an appropriate DXVK config.

SuperSamus:
> the reason [...] is to reduce the amount of user configuration necessary to even run this game in an acceptable way at all

I understand the intention, but this is hidden behavior, which is made worse by the fact that it only applies to users who are already tampering with the game, and that it saves only a very simple step.

I've mentioned the problems of having hidden behavior (like bloating the documentation in the guides), which haven't been addressed yet in the discussion.

> I just don't really see a good reason to fuck over everyone else several years after the game's release

The high framerate mod [has been recently made easier to use](https://github.com/SpecialKO/SpecialK/pull/130/commits/b6e93635a1d354ecde823c9b977375c26f60e932). (You previously had to downgrade the game to use it)

> there's realistically like three people using S-K on Proton

I doubt it's going to screw anyone's configuration[^1], especially someone who isn't able to fix it, considering that they already tampered with the game.

I suspect there are more users who attempted to utilize the high framerate mod, but encountered DXVK's 60 FPS cap without understanding why.

Also, mods are important in PC gaming. Don't say it doesn't matter just because not many people mod games.

[^1]: If we take all the users that are interested in putting effort for a stable frame pacing, [followed a guide](https://www.pcgamingwiki.com/wiki/NieR_Replicant#Framerate_limited_to_57.7E58_FPS), went to the game's configuration file, and set `FpsControlDisable = 1`, I think there are less than three people who skipped the last step (setting a frame limiter) because they already knew that DXVK already does this by default.

doitsujin:
I'm not convinced but have it your way I guess.

mirh:
Gamebryo games should definitively be capped. I can only assume they aren't yet just because they use vsync as a framerate limiter, and so you don't notice the gameplay exploding everywhere. 

Then, I get that the least stuff is done under the nose the better, but it seems obvious what you should do if you can even marginally fix the game experience for the 90% of very dumb clueless users that won't ever touch it (half of them would probably rather play with a broken game than handle it, even if informed and given the choice). 
The only puzzling thing is how syncinterval could automagically smooth the bad frame pacing while also simultaneously avoiding breakage on higher refresh rates, but if the facts are these they seem pretty straightforward.

PR #3948, opened 2024-04-10
[d3d9] Keep front buffer around when we don't swap with it

K0bin:
Fixes #3943

I made it a config option because modded Silent Hill 2 is the only game we've ever seen that actually needs the extra blit.

PR #3949, opened 2024-04-11
[d3d9] Implement GetFrontBufferData using GDI

### LINKED ISSUE #3943: SH2ee: incorrect directx front buffer
Hello,
[SH2ee](https://enhanced.townofsilenthill.com/SH2/) is an open source project to play Silent Hill 2 on modern computer, the devs maintain a Wine compatibility.
Since DXVK 2.2, an issue appears: when you want to quit the game, quit menu superimposed with main menu, main menu does not disappear to let the place to quit menu :

image

I opened an issue on SH2ee Github here:
https://github.com/elishacloud/Silent-Hill-2-Enhancements/issues/942

The SH2ee devs think it is related to directx front buffer

I make several tests and can not reproduce this issue with DXVK =< 2.1
- try different Wine versions: issue still present
- DXVK off: no issue
- DXVK 2.3.1: issue still present
- DXVK 2.3: issue still present
- DXVK 2.2: issue still present
- DXVK 2.1 and less: no issue

Thank you for your help.

my config:
Manjaro 23.1.3 Vulcan
Kernel: x86_64 Linux 6.6.19-1-MANJARO
Proc: AMD Ryzen 9 5900HX
GPU: AMD Radeon RX 6600M
Drivers: Mesa 24.0.2
Resolution: 1920x1080
32 Go RAM
DE: Xfce4

K0bin:
Fixes #3943

Alternative to #3948

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1093,6 +1093,10 @@ namespace dxvk {
     if (unlikely(iSwapChain != 0))
       return D3DERR_INVALIDCALL;
 
+    #ifdef _WIN32
+    return D3D9SwapChainEx::GetFrontBufferDataGDI(pDestSurface);
+    #endif
+
     D3D9DeviceLock lock = LockDevice();
 
     // In windowed mode, GetFrontBufferData takes a screenshot of the entire screen.
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -183,6 +183,8 @@ namespace dxvk {
     m_activeRTsWhichAreTextures = 0;
     m_alphaSwizzleRTs = 0;
     m_lastHazardsRT = 0;
+
+    m_gamescopeWSI = dxvk::env::getEnvVar("ENABLE_GAMESCOPE_WSI") == "1";
   }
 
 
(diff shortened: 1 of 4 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -1093,6 +1095,12 @@ namespace dxvk {
     if (unlikely(iSwapChain != 0))
       return D3DERR_INVALIDCALL;
 
+    #ifdef _WIN32
+    if (!IsGamescopeWSIEnabled()) {
+      return D3D9SwapChainEx::GetFrontBufferDataGDI(pDestSurface);
```
doitsujin, line 1100: do we need to be robust against the GDI call failing (e.g. due to a non-rgba8 swapchain) or is this only supported for specific formats?

K0bin: That should work fine.

`GetFrontBufferData` requires the target surface to be `A8R8G8B8`, so that one isn't gonna be an issue.
On the GDI side, we're capturing the contents by blitting into a compatible bitmap which can have a device specific format. After that we're using `GetDIBits`. DI stands for Device Independent and that function handles the conversion to the specified format for us.

> The GetDIBits function retrieves the bits of the specified compatible bitmap and copies them into a buffer as a DIB using the specified format.

Blisto91:
Tested this PR and it sadly doesn't fix the issue when using gamescope

Blisto91:
Sadly doesn't work with KDE Wayland either

K0bin:
This doesn't work well and the less GDI in DXVK, the better.

PR #3956, opened 2024-04-13
Add FreeBSD support

er2off:
This pull request adds initial FreeBSD operating system support.

To build: `env CC=cc CXX=c++ ./package_native.sh <other args here> --64-only` (for 64-bit system)
~~(gcc13 is currently not working)~~

### DIFF
--- a/package-native.sh
+++ b/package-native.sh
@@ -25,6 +25,8 @@ shift 2
 opt_nopackage=0
 opt_devbuild=0
 opt_buildid=false
+opt_64_only=0
+opt_32_only=0
 
 CC=${CC:="gcc"}
 CXX=${CXX:="g++"}
@@ -40,6 +42,12 @@ while [ $# -gt 0 ]; do
     ;;
   "--build-id")
     opt_buildid=true
+    ;;
+  "--64-only")
+    opt_64_only=1
+    ;;
+  "--32-only")
+    opt_32_only=1
     ;;
   *)
     echo "Unrecognized option: $1" >&2
@@ -81,8 +89,12 @@ function package {
   rm -R "dxvk-native-$DXVK_VERSION"
 }
 
-build_arch 64 lib
-build_arch 32 lib32
+if [ $opt_32_only -eq 0 ]; then
+  build_arch 64 lib
+fi
+if [ $opt_64_only -eq 0 ]; then
+  build_arch 32 lib32
+fi
 
 if [ $opt_nopackage -eq 0 ]; then
   package
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -6,6 +6,10 @@
 #ifdef __linux__
 #include <unistd.h>
 #include <limits.h>
+#elif defined(__FreeBSD__)
+#include <sys/sysctl.h>
+#include <unistd.h>
+#include <limits.h>
 #endif
 
 #include "util_env.h"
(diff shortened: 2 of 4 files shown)


meson.build
```diff
@@ -131,7 +131,13 @@ else
   dxvk_wsi = get_option('dxvk_native_wsi')
   
   if dxvk_wsi == 'sdl2'
-    lib_sdl2 = cpp.find_library('SDL2')
+    if platform == 'freebsd'
+      # kinda hacky
+      dxvk_include_dirs += ['/usr/local/include']
```
Joshua-Ashton, line 136: You should remove your local hack for this

package-native.sh
```diff
@@ -41,6 +43,12 @@ while [ $# -gt 0 ]; do
   "--build-id")
     opt_buildid=true
     ;;
+  "--64-only")
+    opt_64_only=1
```
Joshua-Ashton, line 47: Why?

meson.build
er2off: FreeBSD installs all packages to /usr/local. DXVK includes `<SDL2/SDL.h>` which is located in /usr/local/include. Also as I know, SDL2 recommends to use include `<SDL.h>` instead, luckily it will be changed to `<SDL3/SDL.h>` with SDL3 release.

package-native.sh
er2off: To build 64-bit only version of library of course. If build 32 bit version too, library can be built but can't be linked with SDL because FreeBSD package manager can install only 64 or 32 bit, so need to build SDL2 for 32 bit too.

meson.build
Joshua-Ashton: How do other meson projects targeting freebsd handle this? We don't have anything like this in gamescope and we support FreeBSD there... at least it builds anyway?

cc: @jbeich

jbeich: See https://github.com/mesonbuild/meson/issues/4468. Besides hardcoding `/usr/local/include` in the source or CI config the value can be derived via  ` get_option('prefix') / 'include'`  or  `dep_displayinfo.get_variable(pkgconfig: 'includedir'))` a la https://github.com/st3r4g/wlanthy/commit/89f2a6a4d0ca.

Note, gamescope uses `dependency('SDL2')` instead of `compiler.find_library('SDL2')` thus ex plicitly passing `/usr/local/include` isn't necessary.

package-native.sh
Joshua-Ashton: Let's use dependency(...) then for Linux/BSD paths.

I am okay with not caring about 32-bit BSDs, it's 2024.

meson.build
er2off: Meson have weird dependency handling, if provide `dependency('sdl2')`, it will compile wsi normally but not other libs that depends on it because include path isn't added.

er2off:
> does rebasing this on top of those changes help make the meson changes unnecessary?

@flibitijibibo yes, it works perfect. Also gcc build is workable if add limits.h include to src/util/util_env.cpp (why it worked with clang without this include?)

flibitijibibo:
The relevant meson changes were merged, so this should be able to rebase without the meson.build changes and work as intended!

src/util/util_env.cpp
```diff
@@ -85,6 +89,17 @@ namespace dxvk::env {
     size_t count = readlink("/proc/self/exe", exePath.data(), exePath.size());
 
     return std::string(exePath.begin(), exePath.begin() + count);
+#elif defined(__FreeBSD__)
+    int mib[4] = {CTL_KERN, KERN_PROC, KERN_PROC_PATHNAME, getpid()};
+    std::string exePath(PATH_MAX, '\0');
+    size_t size = exePath.size();
+
+    if (sysctl(mib, 4, exePath.data(), &size, NULL, 0) != 0) {
```
Joshua-Ashton, line 97: It would be cool, but you can't use std::string like this, sorry!

Make a char array on the stack then put that into a std::string.

Joshua-Ashton (changes requested):
Will merge after nit addressed

PR #3978, opened 2024-04-30
[util] Add configs for Delta Force Xtreme 1 & 2

Blisto91:
Prevents the games from black screening on Alt-Tab


WinterSnowfall:
> Added extra config to prevent large performance dips in some areas.

FYI, Delta Force: Black Hawk Down (d3d8) also reads back from write-only dynamic buffers, so apparently it's a trend with these games...

Blisto91:
Ye i saw that one was d3d8, but didn't want to check it out until the merge since d8vk is kinda in hiatus until then anyway.

PR #3984, opened 2024-05-03
[d3d11] Implement better filtering when blitting video content

doitsujin:
@rbernon This is mostly intended to replace the `NEAREST` sampler so that we can still have proper interpolation when upscaling. Please let me know if there are issues with your test case.


rbernon:
Seems to be working as well, thanks!

PR #4001, opened 2024-05-15
[dxvk] Replace deprecated Vulkan macros

Blisto91:
Replace deprecated Vulkan macros with their new equivalents

### DIFF
--- a/src/dxvk/dxvk_adapter.cpp
+++ b/src/dxvk/dxvk_adapter.cpp
@@ -415,7 +415,7 @@ namespace dxvk {
       m_deviceFeatures.khrPresentWait.presentWait;
 
     // Unless we're on an Nvidia driver where these extensions are known to be broken
-    if (matchesDriver(VK_DRIVER_ID_NVIDIA_PROPRIETARY, 0, VK_MAKE_VERSION(535, 0, 0))) {
+    if (matchesDriver(VK_DRIVER_ID_NVIDIA_PROPRIETARY, 0, VK_MAKE_API_VERSION(0, 535, 0, 0))) {
       enabledFeatures.khrPresentId.presentId = VK_FALSE;
       enabledFeatures.khrPresentWait.presentWait = VK_FALSE;
     }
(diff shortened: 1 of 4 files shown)


src/dxvk/dxvk_adapter.cpp
```diff
@@ -431,9 +431,9 @@ namespace dxvk {
     Logger::info(str::format("Device properties:"
       "\n  Device : ", m_deviceInfo.core.properties.deviceName,
       "\n  Driver : ", m_deviceInfo.vk12.driverName, " ",
-      VK_VERSION_MAJOR(m_deviceInfo.core.properties.driverVersion), ".",
-      VK_VERSION_MINOR(m_deviceInfo.core.properties.driverVersion), ".",
-      VK_VERSION_PATCH(m_deviceInfo.core.properties.driverVersion)));
+      VK_API_VERSION_MAJOR(m_deviceInfo.core.properties.driverVersion), ".",
```
jp7677, line 434: Unfortunately this will truncate the major version for NVIDIA. 

The NVIDIA driver uses 9 bits for the major version (in DXVK: `(m_deviceInfo.core.properties.driverVersion >> 22) & 0x3ff`) whereas `VK_API_VERSION_MAJOR` truncates to 7 bits (`#define VK_API_VERSION_MAJOR(version) (((uint32_t)(version) >> 22U) & 0x7FU)`). 

The former `VK_VERSION_MAJOR` macro didn't truncate (`#define VK_VERSION_MAJOR(version) ((uint32_t)(version) >> 22U)`), that's why it worked with that macro, even if the Vulkan specs also mentioned 7 bits back then.

cdavis5e: It's worse than that. Some drivers use a totally different scheme for encoding the driver version number, [and this is permitted by Vulkan](https://registry.khronos.org/vulkan/specs/1.3-extensions/html/chap5.html#VkPhysicalDeviceProperties). [MoltenVK is one of them](https://github.com/KhronosGroup/MoltenVK/blob/6b9e55983a6d22eaf96003297119f25256fdf3ba/MoltenVK/MoltenVK/API/mvk_private_api.h#L53), to name just one example.

Blisto91: :panik:

I'll have to let someone with actual programming skills deal with this now 😆

Blisto91:
Superseded by https://github.com/doitsujin/dxvk/pull/4005

PR #4003, opened 2024-05-16
[d3d11,d3d9,util] Add a config option for reproducible VK output

werman:
It ensures that for the same D3D commands the output VK commands don't change between runs.

Useful for comparative benchmarking, can negatively affect performance.

---

To fully test it I'd need to disable all threading and compare runs with api dump layer. For now, what was compared are renderpasses after d3d11 translation - they are the same between runs.

When `reproducibleCommandStream` is not set the number of draw calls in renderpass and even the number of renderpasses may vary, which makes hard to make comparisons between runs, especially on GPUs where renderpasses do matter.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -312,6 +312,20 @@
 # d3d11.exposeDriverCommandLists = True
 
 
+# Reproducible Command Stream
+#
+# Ensure that for the same D3D commands the output VK commands
+# don't change between runs. Useful for comparative benchmarking,
+# can negatively affect performance and can break some games
+# that don't use queries correctly.
+#
+# Supported values:
+# - True/False
+
+# d3d11.reproducibleCommandStream = False
+# d3d9.reproducibleCommandStream = False
+
+
 # Sets number of pipeline compiler threads.
 # 
 # If the graphics pipeline library feature is enabled, the given
--- a/src/d3d11/d3d11_context_imm.cpp
+++ b/src/d3d11/d3d11_context_imm.cpp
@@ -19,6 +19,7 @@ namespace dxvk {
     m_csThread(Device, Device->createContext(DxvkContextType::Primary)),
     m_maxImplicitDiscardSize(pParent->GetOptions()->maxImplicitDiscardSize),
     m_submissionFence(new sync::CallbackFence()),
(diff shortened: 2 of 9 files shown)


dxvk.conf
```diff
@@ -312,6 +312,18 @@
...
+# Reproducible Command Stream
+#
+# Ensure that for the same D3D commands the output VK commands
+# don't change between runs. Useful for comparative benchmarking,
+# can negatively affect performance.
+#
+# Supported values:
+# - True/False
+
+# d3d11.reproducibleCommandStream = False
```
Blisto91, line 324: Since both the d3d9 and d3d11 option does the same and have the same description it is fine to just list both in one place

```

# d3d11.reproducibleCommandStream = False
# d3d9.reproducibleCommandStream = False

```

doitsujin:
Why are the thread syncs in `ConsiderFlush` necessary? We should get consistent behaviour simpy by adjusting the heuristic.

The specific implementation also massively inflates the actual submission count compared to a GPU-bound scneario; if anything I feel like we'd want to do the opposite and *minimize* submission counts to get more realistic performance characteristics, i.e. something like

```c++
if (m_ensureReproducibleHeuristic && flushType != GpuFlushType::ExplicitFlush)
  return false;

switch (flushType) {
  ...
```

This **will** break real games that don't use queries correctly (Fallout 4), but should generally be fine for replaying traces and captures since readback isn't really a thing there anyway.

werman:
> I feel like we'd want to do the opposite and minimize submission counts to get more realistic performance characteristics

You are right, we should expect GPU bound scenario and that means there is no pressure to submit earlier, meaning most likely submit would happen only at `GpuFlushType::ExplicitFlush`.

> Why are the thread syncs in ConsiderFlush necessary? We should get consistent behaviour simpy by adjusting the heuristic.

Without it there were differences in VK commands when looping the same frame, probably caused by different current chunk id. However, if we flush only on `ExplicitFlush` - there is no need to sync.

dxvk.conf
werman: Makes sense

doitsujin (approved):

Issue #4008, opened 2024-05-18
PROTOTYPE ™ missing shadows and weak other graphical effects


Mrblubb:
When using dxvk shadows, bloom and other graphical effects are missing/weakened/not rendered correctly

![Untitled-1](https://github.com/doitsujin/dxvk/assets/88938655/7c82c3e4-0c85-4537-a0df-bfca397b663d)
![Untitled-2](https://github.com/doitsujin/dxvk/assets/88938655/4917a0e8-0b7e-437b-9e35-de71541a9608)
![Untitled-3](https://github.com/doitsujin/dxvk/assets/88938655/95346009-da7f-41f9-b5b5-ff73f5dc22c2)
![Untitled-4](https://github.com/doitsujin/dxvk/assets/88938655/acb1d2f9-dba8-4ccb-86ab-1f74839764df)
![Untitled-5](https://github.com/doitsujin/dxvk/assets/88938655/2b0b51a1-d43c-468a-ab0c-7ed7ad5d5643)
![Untitled-6](https://github.com/doitsujin/dxvk/assets/88938655/eb88edc3-50b1-4378-9948-192da61b15e3)

### Software information
PROTOTYPE PC Max settings with emoose patch

### System information
- GPU: GTX 1060 6GB stock
- CPU: FX 8350 stock
- Driver: 546.17
- Windows: Win 10 x64 
- DXVK version: 2.3.1 

### Apitrace file(s)
https://we.tl/t-Bu8NKoCMzb

### Log files
[prototypef_d3d9.log](https://github.com/doitsujin/dxvk/files/15362704/prototypef_d3d9.log)

K0bin:
This might just be a game bug:
https://www.pcgamingwiki.com/wiki/Prototype#No_shadows_with_AMD.2FATI_graphics_cards

WinterSnowfall:
@Mrblubb Out of curiosity, does `dxgi.nvapiHack = False` do anything for the missing shadows?

Mrblubb:
That happens on AMD cards and the fix provided in that makes shadows and some effects pixelated and not exactly accurate. With dxvk on nvidia it should render everything correctly

K0bin:
![image](https://github.com/doitsujin/dxvk/assets/1131720/a085e734-a881-49de-9baf-d18cf3e10b13)

Works fine when replaying the apitrace.

Mrblubb:
[...]

PR #4010, opened 2024-05-18
[d3d9] Do not enable support for DF formats on Nvidia

WinterSnowfall:
Should get #4008 shadows working properly on Nvidia and provide a saner default behavior for some games. Since native Nvidia d3d9 drivers don't support DF16 or DF24 it makes little sense to expose them IMHO.

The game in question would still need a config option for AMD and Intel, since it seems to be broken in regards to shadows in general.

Alternatively, we could allow it only for AMD and Intel (more historically accurate), rather than disabling it only on Nvidia, since this would also disable it for more exotic vendors. Let me know what you think.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -515,7 +515,9 @@
 
 # Support DF formats
 #
-# Support the vendor extension DF floating point depth formats
+# Support the vendor extension DF floating point depth formats on AMD and Intel.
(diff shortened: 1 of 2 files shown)


Blisto91 (approved):

Blisto91:
Verified that https://github.com/doitsujin/dxvk/issues/4008 does look correct when spoofing Nvidia vendor id

PR #4018, opened 2024-05-21
[d3d9] Options parser vendorID fix and DF format support validation rework

WinterSnowfall:
I'm guessing the use of customDeviceId for vendorID was unintentional, right? :sweat_smile: Not that it affected much beyond VQuery support on the few titles that have Nvidia vendorID (& deviceID) overrides set in place (even then, only on AMD and Intel).

Also moved the DF Format support disabling on Nvidia during options parsing, sorry for not doing that earlier (didn't really think a vendorID check was already done there, so didn't bother to check).

### DIFF
--- a/src/d3d9/d3d9_options.cpp
+++ b/src/d3d9/d3d9_options.cpp
@@ -37,7 +37,7 @@ namespace dxvk {
     this->customDeviceId        = parsePciId(config.getOption<std::string>("d3d9.customDeviceId"));
     this->customDeviceDesc      =            config.getOption<std::string>("d3d9.customDeviceDesc");
 
-    const int32_t vendorId = this->customVendorId != -1
+    const uint32_t vendorId = this->customVendorId != -1
       ? this->customVendorId
       : (adapter != nullptr ? adapter->deviceProperties().vendorID : 0);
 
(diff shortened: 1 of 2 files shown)


doitsujin (approved):

src/d3d9/d3d9_options.cpp
```diff
@@ -53,7 +53,7 @@ namespace dxvk {
     this->deferSurfaceCreation          = config.getOption<bool>        ("d3d9.deferSurfaceCreation",          false);
     this->samplerAnisotropy             = config.getOption<int32_t>     ("d3d9.samplerAnisotropy",             -1);
     this->maxAvailableMemory            = config.getOption<int32_t>     ("d3d9.maxAvailableMemory",            4096);
-    this->supportDFFormats              = config.getOption<bool>        ("d3d9.supportDFFormats",              true);
+    this->supportDFFormats              = config.getOption<bool>        ("d3d9.supportDFFormats",              vendorId != 0x10de);
```
K0bin, line 56: `DxvkGpuVendor::Nvidia`

WinterSnowfall: Done

PR #4015, opened 2024-05-20
[d3d9] Fixed function fix a bunch of wine tests with FF texture coord…

K0bin:
Fixes #3117

Master:

> visual.c:28422: Driver string: "nvd3dum.dll"
visual.c:28423: Description string: "NVIDIA GeForce RTX 3090"
visual.c:28426: Device name string: "\\.\DISPLAY1"
visual.c:28428: Driver version 32767.65535.65535.65535
visual.c:5709: Test failed: quad 4 has color 00000000, expected 0x0033cc00
visual.c:5375: Test failed: D3DTTFF_COUNT3 | D3DTTFF_PROJECTED (draws non-projected) - top right: pixel (341, 211) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_COUNT3 | D3DTTFF_PROJECTED (draws non-projected) - top right: pixel (341, 224) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_COUNT3 | D3DTTFF_PROJECTED (draws non-projected) - top right: pixel (359, 211) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_COUNT3 | D3DTTFF_PROJECTED (draws non-projected) - top right: pixel (359, 224) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_PROJECTED (like COUNT3 | PROJECTED, texcoord has only 3 components) - bottom right: pixel (401, 361) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_PROJECTED (like COUNT3 | PROJECTED, texcoord has only 3 components) - bottom right: pixel (401, 419) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_PROJECTED (like COUNT3 | PROJECTED, texcoord has only 3 components) - bottom right: pixel (479, 361) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_PROJECTED (like COUNT3 | PROJECTED, texcoord has only 3 components) - bottom right: pixel (479, 419) has color ff000000, expected ffffffff.
visual.c:5375: Test failed: D3DTTFF_COUNT1 (draws non-projected) - top right: pixel (341, 209) has color ffffffff, expected ff000000.
[...]

PR #4026, opened 2024-05-25
[d3d9] FF: Apply transform flags count to generated texture coords

K0bin:
Fixes the bug in Phantasy Star Online that was reported after merging #4015.

I've tested this PR with both Phantasy Star Online apitraces, the Cold Fear apitrace and the Snowblind apitrace. Those were the games that I know were broken at some point because of fixed function texture coordinate processing. On top of that, I ran the Wine FF texcoord tests again and those were fine too.

______

Tests show that the count out of `D3DTSS_TEXTURETRANSFORMFLAGS` gets applied to generated texcoords like `D3DTSS_TCI_CAMERASPACEPOSITION` too.

At the same time, projection only gets applied (or renders something sensible) when used with `FVF_XYZ` and `D3DTTFF_COUNT3` or higher. With lower values it either produces garbage like this:

image
(FVF_TEXCOORDSIZE1 + FLAGS_COUNT1)

or this:
image
(FVF_TEXCOORDSIZE2 + FLAGS_COUNT1/2)

https://docs.google.com/spreadsheets/d/1XfSsxl1qMMY0mmpn02l3Wph-6CyXiugcR9wkb0VwJg0/edit?usp=sharing

By writing a test app that uses fixed function vertex processing and a shader that does nothing except output the texcoord, I've now verified that native D3D9 also writes the component that gets used for projection into the w component just like we're doing here.

### DIFF
--- a/src/d3d9/d3d9_fixed_function.cpp
+++ b/src/d3d9/d3d9_fixed_function.cpp
@@ -1139,12 +1139,12 @@ namespace dxvk {
 
         case (DXVK_TSS_TCI_CAMERASPACENORMAL >> TCIOffset):
           transformed = outNrm;
-          count = 4;
+          count = std::min(flags, 4u);
           break;
 
         case (DXVK_TSS_TCI_CAMERASPACEPOSITION >> TCIOffset):
           transformed = m_module.opCompositeInsert(m_vec4Type, m_module.constf32(1.0f), vtx, 1, &wIndex);
-          count = 4;
+          count = std::min(flags, 4u);
           break;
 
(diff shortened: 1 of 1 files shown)


K0bin:
@nolrinale This PR fixes the ripples.

nolrinale:
> @nolrinale This PR fixes the ripples.

I just checked it, all working as intended, even the monsters with refraction cloaking too its amazing K0bin thank you so much for your work! now the game is working 100% under Linux/Deck without any graphic bugs left.

Happy 20th Anniversary PSOBB!~

K0bin:
Awesome, thanks for the good bug reports!

doitsujin (approved):

Joshua-Ashton:
Thanks for doing all this :D

PR #4028, opened 2024-05-25
[CI] Update CI actions

Blisto91:
Silences the CI deprecation warnings

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -9,13 +9,13 @@ jobs:
     steps:
     - name: Checkout code
       id: checkout-code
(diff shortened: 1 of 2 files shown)


Blisto91:
Sorry for the force push spam. 
Checked a bit late that upload-artifact v4 had some big changes that made it break backwards compatibility. This meant that separate artifacts could no longer be merged in to one by just giving them the same name. Messed around with it for a bit to more or less get the same output as before.

See https://github.com/actions/upload-artifact/blob/00e36f94d817ea235422592a23d468b262071bf4/docs/MIGRATION.md for more info.

PR #4047, opened 2024-06-02
[util] Add dxvk.maxChunkSize 1 for Origin Web Helper Service and Ubisoft Connect (UPlay)

Tiagoquix:
A follow-up to #4037.

Tested by exiting Steam completely, then launching with `DXVK_CONFIG="dxvk.maxChunkSize = 1" steam` and playing Watch Dogs 1 for about 5 minutes. Haven't experienced any problems.

Below are the processes that stay open with the game in the background:
ubisoft

`UbisoftConnect.exe` is the main process, which doesn't stay open, but I'm adding it for extra safety. `upc.exe` seems to be what really matters here (I guess it means "UPlay Core").

We also have these ones in the Ubisoft Game Launcher installation folder:

- `UbisoftGameLauncher64.exe` and `UbisoftGameLauncher.exe` (opens temporarily during Steam game launch; maybe a proxy service for Steam-Ubisoft?)
- `UplayCrashReporter.exe` (I don't think this one needs maxchunksize)
- `UplayWebCore.exe` (probably related to the overlay, but the UPlay overlay causes issues on Linux and Lutris even disables it by editing a config. file)
- `UbisoftExtension.exe` and `UplayService.exe` (I have no idea, but they do not stay open with the game)

Feel free to add maxchunksize for any process mentioned before that I haven't added if you have a better reasoning.

To be clear, Ubisoft does not support its legacy launcher anymore, and forces the new one onto its users, so there's no point to add UPlay-exclusive-related process names here.

Thank you for your attention.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -359,11 +359,11 @@ namespace dxvk {
       { "d3d11.dcSingleUseMode",            "False" },
     }} },
     /* Rockstar Games Launcher                    */
-    { R"(\\Rockstar Games\\Launcher\.exe$)", {{
+    { R"(\\Rockstar Games\\Launcher\\Launcher\.exe$)", {{
       { "dxvk.maxChunkSize",                "1"   },
     }} },
     /* Rockstar Social Club                       */
-    { R"(\\Rockstar Games\\SocialClub\\SocialClubHelper\.exe$)", {{
+    { R"(\\Rockstar Games\\Social Club\\SocialClubHelper\.exe$)", {{
       { "dxvk.maxChunkSize",                "1"   },
     }} },
     /* EA Desktop App                             */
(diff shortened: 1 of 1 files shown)


src/util/config/config.cpp
```diff
@@ -359,19 +359,23 @@ namespace dxvk {
...
+    { R"(\\Rockstar Games\\SocialClub\\SocialClubHelper\.exe$)", {{
       { "dxvk.maxChunkSize",                "1"   },
     }} },
     /* EA Desktop App                             */
     { R"(\\EADesktop\.exe$)", {{
       { "dxvk.maxChunkSize",                "1"   },
     }} },
     /* Origin app (legacy EA Desktop)             */
-    { R"(\\Origin\.exe$)", {{
+    { R"(\\Origin\\OriginWebHelperService\.exe$)", {{
```
Blisto91, line 374: Did you mean to remove the Origin exe itself?

Blisto91: `R"(\\(Origin|OriginWebHelperService)\.exe$)"` or `R"(\\Origin(WebHelperService)?\.exe$)"` would be if you want both exe's.
`R"(\\Origin\\OriginWebHelperService\.exe$)"` is looking for `OriginWebHelperService.exe` inside a folder called `Origin`

src/util/config/config.cpp
```diff
@@ -359,19 +359,23 @@ namespace dxvk {
...
+      { "dxvk.maxChunkSize",                "1"   },
+    }} },
+    /* Ubisoft Connect (UPlay)                    */
+    { R"(\\UbisoftConnect\\upc\.exe$)", {{
```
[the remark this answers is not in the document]
Blisto91: If you want to include both then it should be `R"(\\(UbisoftConnect|upc)\.exe$)"`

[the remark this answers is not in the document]
Blisto91: It is because the syntax `\\UbisoftConnect\\upc\.exe$` means that it will only apply the config to `upc.exe` if it resides in a folder called `UbisoftConnect` so i got confused and assumed you had a game install where that was the case.

src/util/config/config.cpp
Blisto91: Yes

doitsujin:
This band-aid is getting annoying and we **really** need to find a way to make this the default behaviour for all apps, but well, guess the band-aid fixes will do for now.

PR #4051, opened 2024-06-03
[util] Disable countLosableResources for Myst V

Blisto91:
Works around a crash regression in dxvk 2.3 upon resolution change.


Blisto91:
Devs wanted to look deeper at the problem so created issue https://github.com/doitsujin/dxvk/issues/4052

K0bin (approved):

PR #4056, opened 2024-06-06
Limit frame rate to display refresh

doitsujin:
We cannot perform mode switches in non-Windows environments, but we still report display modes with lower refresh rates to the application. In some cases, this leads to bugs since games that have a built-in FPS limiter can reasonably assume that a 60 Hz display mode with Vsync enabled will also cap the frame rate at 60.

This implements a heuristic that measures the current frame rate, and if significantly higher than the target refresh rate, engages the FPS limiter after a few frames.

Existing configuration options as well as `DXVK_FRAME_RATE` will always take precedence, and setting the config to -1 unconditionally disables the limiter.

Might need some testing since some specific games might be a bit on the dumb side and default to something like 24 Hz for no reason at all with no way to change it in game (Crysis 1?).


Blisto91:
> Existing configuration options as well as DXVK_FRAME_RATE will always take precedence, and setting the config to -1 unconditionally disables the limiter.

If the env var works with -1 too now then remember to update the readme

PR #4061, opened 2024-06-06
Fixed GLFW exception at startup

Sanakan8472:
`GlfwWsiDriver::getInstanceExtensions` was creating an `std::vector` with a size argument in the ctor but then used `push_back` instead of filling the pre-allocated elements, leading to a bunch of nullptr entries at the start that caused an exception later on when accessed.


flibitijibibo (approved):
Pretty sure this was my fault when I moved the instance extensions to the wsi backend, lgtm

PR #4065, opened 2024-06-10
[util] Disable Crysis 1 refresh rate fps limiter

Blisto91:
Depends on https://github.com/doitsujin/dxvk/pull/4056

Game bug in its d3d10 mode where it selects the lowest supported refresh rate.


doitsujin:
Will merge once the other thing lands.

PR #4068, opened 2024-06-11
[util] Add auto installer

nafigator:
* Bash script for latest DXVK libs installation into wine prefixes. Release files copied only when they are newer than files in prefix or when prefix files is missing. As dependencies it uses grep, jq, cut and bash-helpers lib. See https://github.com/nafigator/bash-helpers. If bash-helpers not exists in bash includes than script sources it from github.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,27 @@ The most recent development builds can be found [here](https://github.com/doitsu
 
 Release builds can be found [here](https://github.com/doitsujin/dxvk/releases).
 
-## How to use
+## Installation
+#### Latest automatically
+Examples for systems with bash, curl, grep, cut and jq utils available:
+```
+export WINEPREFIX=/path/to/wineprefix
+curl -s https://raw.githubusercontent.com/doitsujin/dxvk/master/install.sh | bash
+```
+
+```
+cd /path/to/wineprefix
+curl -s https://raw.githubusercontent.com/doitsujin/dxvk/master/install.sh | bash
+```
+
+Or in one line:
+```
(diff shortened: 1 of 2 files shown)


nafigator:
For testing purposes you can use install.sh from my branch:
```curl -s https://raw.githubusercontent.com/nafigator/dxvk/auto-installer/install.sh | WINEPREFIX=~/.wine bash```

Also, script accepts params:
```
-v --version
-h --help
-d --debug
```

Combine them as follows:
`curl -s https://raw.githubusercontent.com/nafigator/dxvk/auto-installer/install.sh | WINEPREFIX=~/.wine bash -s - <params>`

Example:

```
curl -s https://raw.githubusercontent.com/nafigator/dxvk/auto-installer/install.sh | WINEPREFIX=~/.wine bash -s - -dv
```

WinterSnowfall:
To offer a bit of context: dxvk used to have an installation script, but everyone and their dog seemed to take issues with it for some reason or another, which created unnecessary focus on what is (ultimately) a less important aspect of the project, so the decision was taken to remove it from the repo and let people create and maintain their own, should they feel the need. I doubt this point will be walked back on.

nafigator:
> To offer a bit of context: dxvk used to have an installation script, but everyone and their dog seemed to take issues with it for some reason or another, which created unnecessary focus on what is (ultimately) a less important aspect of the project, so the decision was taken to remove it from the repo and let people create and maintain their own, should they feel the need. I doubt this point will be walked back on.

Fair enough. It's ok to close this without merge. In this case I'll think about  own gist or repo for this script. Hope it'll be useful for someone.

PR #4071, opened 2024-06-12
[util] Dead Space 2 - Limit to 60 FPS and enable vsync

Trevonn:
Similar to the first game it has a poor vsync implementation (limits to 30 fps) and physics issues when the frame rate is unlocked.

Limiting to 60 FPS and enabling vsync externally provides a better experience after the ingame vsync is disabled.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -511,6 +511,14 @@ namespace dxvk {
       { "dxvk.maxFrameRate",                     "60" },
       { "d3d9.presentInterval",                  "1" },
     }} },
+    /* Dead Space 2
+       Physics issues above 60 FPS
+       Built-in Vsync Locks the game to 30 FPS
+    */
+    { R"(\\deadspace2\.exe$)", {{
+      { "dxvk.maxFrameRate",                     "60" },
+      { "d3d9.presentInterval",                  "1" },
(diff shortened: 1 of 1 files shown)


brad86m:
Sadly, it apears this 60fps lock has been implemented where we no longer have the choice of disabling it. I have tried changing sync settings in the DXVK config file, but it is stuck tight at 60.

Personally, I think the issues between 60-120hz are pretty minimal, and only really become a large negative when going beyond 120. Same situation with the first game.

The option to leave it uncapped would be great.

Trevonn:
> Sadly, it apears this 60fps lock has been implemented where we no longer have the choice of disabling it. I have tried changing sync settings in the DXVK config file, but it is stuck tight at 60.
> 
> Personally, I think the issues between 60-120hz are pretty minimal, and only really become a large negative when going beyond 120. Same situation with the first game.
> 
> The option to leave it uncapped would be great.

Make sure you're using `d3d9.maxFrameRate` not `dxvk.maxFrameRate` this PR does not reflect the current built-in settings.

I just tested the below command on Dead Space 2 and it successfully raised the limit to 120 FPS

`DXVK_CONFIG="d3d9.maxFrameRate = 120" %command%`

PR #4086, opened 2024-06-23
[d3d9] Various vendor formats adjustments

### LINKED ISSUE #4098: [d3d9] The First Templar minimap depth issue
Noticed that the dots on the minimap in The First Templar can get hidden behind objects, e.g a bush, instead of always staying on top. I've only reproduced this on Nvidia as replaying the trace with amdvlk or radv it looked fine. Wined3d crashes on replay.

This seems to be a old regression in 1.5.1 as it looks fine before that

<details>
  <summary>Screenshot</summary>
  
image
</details>

### Software information
The First Templar (Steam version)
"Scene Resolution" set to High

### System information
- GPU: RTX 4080
- Driver: 555.58.0
- Wine version: Proton Bleeding Edge
- DXVK version: 1.5.1 - master. Bisect pending

### Apitrace file(s)
https://drive.proton.me/urls/PEY7HBZMNW#Vkl9L4sd0cjN
Replay with a Nvidia GPU

WinterSnowfall:
Fixes #4098.

Tests of both modern and d3d8/d3d9 era drivers show that D16_LOCKABLE is and was only ever supported on AMD.

P.S.: [Reference outputs here](https://drive.proton.me/urls/1WHHKT8P0C#cbJ6R2bTCCF8).

I've also taken the liberty to clean up the D32 configs and what was left of the validation logic, since we've removed support for it globally quite some time ago.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -569,15 +569,6 @@
 
 # d3d9.supportX4R4G4B4 = True
 
-# Support D32
-#
-# Support the D32 format.
-#
-# Supported values:
-# - True/False
-
-# d3d9.supportD32 = True
-
 # Disable A8 as a Render Target
 #
 # Disable support for A8 format render targets
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -569,6 +569,17 @@
 
 # d3d9.supportX4R4G4B4 = True
 
+# Support D16_LOCKABLE
+#
+# Support the D16_LOCKABLE format.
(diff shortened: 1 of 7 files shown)


WinterSnowfall:
So, it turns out the D3D9Options used in format selection came from the interface... where the vendorID is always returned as 0. Thanks to @Blisto91 for spotting the problem. I've moved back the formats selection logic directly "on site", which should be more robust.

Disabling DF formats on Nvidia properly fixes #4098.

PR #4089, opened 2024-06-25
[util] Add folder to Beyond Good and Evil config

Blisto91:
So the frame limit doesn't apply to the new remaster.

I checked that both the Steam and Gog version of the original is `Beyond Good and Evil` by default. For the remaster the Steam folder is `Beyond Good and Evil - 20th Anniversary Edition`


doitsujin:
Reverted the commit that consolidated options instead.

Issue #4066, opened 2024-06-10
The Sims Castaway Stories - Choppy FPS


Blayer98:
### Software information
The Sims Castaway Stories
Application Type: 32-Bit
Resolution: 1280x800
Fullscreen: Yes

Unfortunately, I cannot test this using WineD3D as I get a "D3DERR_INVALIDCALL" error, and I have tried using the DLL files while using apitrace, I got a game crash before I could enter the neighbourhood view or while entering a lot from the tutorial menu.  With this method, the game had massive frame drops upon launching the game, and this continued during the testing. Upon launching the game normally I have 60 frames per second until I enter the neighbourhood view or enter a lot from the tutorial menu, and from there, I get around 20 - 35 frames per second for the rest of my play session. I just want to point out that I don't get any frame rate issues with The Sims 2 at all, not on any of it's packs, nor the base game or it's compilations.

### System information
- System: Steam Deck (LED)
- GPU: AMD AMD Custom GPU 0405 (vangogh, LLVM 15.0.7, DRM 3.54, 6.1.52-valve9-1-neptune-61)
- Driver: 4.6 (Compatibility Profile) Mesa 23.1.3 (git-58f7632462)
- Wine version: 9.0.1
- DXVK version: v2.3-47-ge2a46a34

### Apitrace file(s)
- https://drive.google.com/file/d/16tBd51m738G8-mgs481tJiU7HkrRgJew/view?usp=sharing

### Log files
Proton Log without WineD3D and using apitrace with the DLLs: (To be posted shortly...)
Proton Log when playing normally: (To be posted shortly...)

Blayer98:
I can confirm that this issue occurs with The Sims Life Stories and The Sims Pet Stories as well.

Blayer98:
I just had a look at config.cpp, I was just thinking if maybe one of the settings is causing the issues? I’m not entirely sure and I can’t read apitraces (I don’t know how to tbh) since The Sims 2 runs fine so maybe there’s something else happening?

K0bin:
Your Apitrace ends just after loading into the game before showing any actual gameplay. So unfortunately that's not particularly useful.

Blayer98:
Hey, Yeah, that’s because the game lags really badly when I was trying to apitrace, and I literally couldn’t get into the neighbourhood view whatsoever.

K0bin:
Poor performance when using apitrace is normal, it's recording every single D3D9 call and the data that's part of that.

[...]

PR #4094, opened 2024-06-28
[util] Set cachedDynamicBuffers for The Sims 2 and friends

Blisto91:
Helps very low CPU bound performance especially in Castaway Stories.

https://github.com/doitsujin/dxvk/issues/4066 is kept open since even though this helps a bunch the performance is still oddly low. 
I haven't been able to compare to Windows since it crashes for me before i get ingame without dxvk and so far the game doesn't agree with apitrace.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -502,6 +502,7 @@ namespace dxvk {
       { "d3d9.supportX4R4G4B4",             "False" },
       { "d3d9.maxAvailableMemory",          "2048" },
       { "d3d9.memoryTrackTest",             "True" },
+      { "d3d9.cachedDynamicBuffers",        "True" },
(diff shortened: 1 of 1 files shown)


K0bin:
Maybe we should start considering just always using cached memory for `D3DUSAGE_DYNAMIC`. @Joshua-Ashton This issue seems to be pretty common for D3D9 games.

WinterSnowfall:
It's also common in several d3d8 games and benchmarks such as 3DMark2001 SE.

WinterSnowfall:
Gave using cached memory for D3DUSAGE_DYNAMIC a go and it didn't turn out that great (see #4103), so I think this can be merged now.

K0bin (approved):

PR #4099, opened 2024-07-04
WRC4 Added

ToRRent1812:
There is a game engine bug where playing above 60fps, causes audio to brake when switching maps. It would be nice to lock the framerate to avoid that issue.


Blisto91:
Isn't this a d3d9 game?

ToRRent1812:
Oops, You are right

src/util/config/config.cpp
```diff
@@ -959,6 +959,10 @@ namespace dxvk {
     { R"(\\(The Battle for Middle-earth (\(tm\))? II( Demo)?|The Lord of the Rings, The Rise of the Witch-king)\\game\.dat$)", {{
       { "d3d9.cachedDynamicBuffers",        "True" },
     }} },
+    /* WRC4 - Audio brakes above 60fps */
```
Blisto91, line 962: Minor spelling nit.

brakes -> breaks

Blisto91 (approved):

PR #4103, opened 2024-07-05
[util] Enable cachedDynamicBuffers by default

WinterSnowfall:
I haven't seen any negative impact so far, probably also because it's kind of difficult to be GPU limited in d3d9 these days.

On the other hand, the positive impact is massive in games that read back from DYNAMIC buffers. This practice appears to have been quite common around the end of the AGP GPU era, with some games even performing read backs from DYNAMIC WRITEONLY buffers. Yeah...

At some point I think we should remove the config option entirely and just go with it, but probably a good idea to keep it around for a while. Draft for now, until it gets more testing.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -639,13 +639,13 @@
 #
 # Allocates dynamic resources in D3DPOOL_DEFAULT in
 # cached system memory rather than uncached memory or host-visible
-# VRAM, in order to allow fast readback from the CPU. This is only
-# useful for buggy applications, and may reduce GPU-bound performance.
+# VRAM, in order to allow fast readback from the CPU.
+# Note that this may reduce GPU-bound performance in certain cases.
 #
 # Supported values:
 # - True/False
 
-# d3d9.cachedDynamicBuffers = False
+# d3d9.cachedDynamicBuffers = True
 
 # Seamless Cubes
 #
--- a/src/d3d9/d3d9_common_buffer.cpp
+++ b/src/d3d9/d3d9_common_buffer.cpp
@@ -134,7 +134,9 @@ namespace dxvk {
       memoryFlags |= VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
     }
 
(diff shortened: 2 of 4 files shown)


Blisto91:
We'll see what Tonga has to say about this. Just you wait

WinterSnowfall:
> I know this is about D3D9, but would enabling `d3d11.cachedDynamicResources` by default be good as well?

No. We're only considering doing this for d3d9 because there are a lot of games and benchmarks which require this sort of read-back optimized access pattern for dynamic buffers. That's not the general case with d3d11.

WinterSnowfall:
Looks like there can be a noticeable performance impact in some later d3d9 games, of roughly 20% when GPU limited. Note that we're still talking about 3 digit fps numbers observed on a R9 380, but that being said it is not something exactly negligible.

While we can add config options to disable cachedDynamicBuffers for said titles to undo the performance hit, we'll have to see just how extensive the affected games list turns out to be. Right now it's not that scary, but I'll wait for @Blisto91 to test some more.

K0bin:
Blisto has tested a lot of games and it would imact a bunch of them negatively. So we're not doing this.

WinterSnowfall:
> Blisto has tested a lot of games and it would imact a bunch of them negatively. So we're not doing this.

Agreed the initial approach is no longer feasible, but we're now looking at maybe using cached only for DYNAMIC and excluding DYNAMIC + WRITE ONLY. That fixes the performance regressions, question now is how many of the game that previously needed cachedDynamicBuffers work well now. I'd appreciate if we could keep this open for a little longer, just to experiment.

WinterSnowfall:
Tested the new approach but it turned out to be a dead end. The problem is indeed related to read-backs from DYNAMIC + WRITEONLY buffers, a thing that is technically discouraged in the documentation (and should in hindsight have been outright disallowed IMHO). Anyway, it seems like we'll have to keep adding workarounds whenever needed.

Blisto91:
Tonga has once again shown it does not play around.
Except when i played those games of course.

PR #4111, opened 2024-07-07
[d3d8] Implement monotonic state block tokens

WinterSnowfall:
Would need feedback from @AlpyneDreams .

The idea is to get rid of reinterpret casts and stick with what native does, namely a monotonically increasing counter that is passed back as the token whenever new state blocks are created. Tests have shown there's no token reuse, not even between device resets.

I also don't think we should worry about the address space being exhausted, because I did some quick math and if let's say a game creates 8 state blocks per frame @ 60 fps, it will take about 100+ days to overflow a DWORD. That's "let's not worry about it" territory in my book... but I can in theory implement a check to see if a token is already taken before allocating it.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -63,11 +63,6 @@ namespace dxvk {
   D3D8Device::~D3D8Device() {
     if (m_batcher)
       delete m_batcher;
-    
-    // Delete any remaining state blocks.
-    for (D3D8StateBlock* block : m_stateBlocks) {
-      delete block;
-    }
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::GetInfo(DWORD DevInfoID, void* pDevInfoStruct, DWORD DevInfoStructSize) {
@@ -993,21 +988,37 @@ namespace dxvk {
     Com<d3d9::IDirect3DStateBlock9> pStateBlock9;
     HRESULT res = GetD3D9()->CreateStateBlock(d3d9::D3DSTATEBLOCKTYPE(Type), &pStateBlock9);
 
-    D3D8StateBlock* pStateBlock = new D3D8StateBlock(this, Type, pStateBlock9.ref());
-    m_stateBlocks.insert(pStateBlock);
-
-    *pToken = DWORD(reinterpret_cast<uintptr_t>(pStateBlock));
+    m_stateBlocks.emplace(std::piecewise_construct,
+                          std::forward_as_tuple(m_token),
+                          std::forward_as_tuple(this, Type, pStateBlock9.ref()));
+    *pToken = m_token;
+    m_token++;
 
     return res;
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::CaptureStateBlock(DWORD Token) {
-    return reinterpret_cast<D3D8StateBlock*>(Token)->Capture();
+    auto stateBlockIter = m_stateBlocks.find(Token);
+
+    if (unlikely(stateBlockIter == m_stateBlocks.end())) {
+      Logger::err("Invalid token passed to CaptureStateBlock");
+      return D3DERR_INVALIDCALL;
+    }
+
+    return stateBlockIter->second.Capture();
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::ApplyStateBlock(DWORD Token) {
     StateChange();
-    return reinterpret_cast<D3D8StateBlock*>(Token)->Apply();
+
+    auto stateBlockIter = m_stateBlocks.find(Token);
+
+    if (unlikely(stateBlockIter == m_stateBlocks.end())) {
+      Logger::err("Invalid token passed to ApplyStateBlock");
+      return D3DERR_INVALIDCALL;
+    }
+
+    return stateBlockIter->second.Apply();
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::DeleteStateBlock(DWORD Token) {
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
I've tested ~140 d3d8 games with this PR and everything checks out. I've only seen one obscure game (FireStarter), at times, passing a state block token to apply a previously deleted state block, which is why I added the validations. Native drivers on Windows XP don't do any validations and would outright crash in this situation.

src/d3d8/d3d8_device.cpp
```diff
@@ -994,30 +994,46 @@ namespace dxvk {
...
+    m_token++;
 
     return res;
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::CaptureStateBlock(DWORD Token) {
-    return reinterpret_cast<D3D8StateBlock*>(Token)->Capture();
+    try {
+      return m_stateBlocks.at(Token)->Capture();
+    } catch (const std::out_of_range& e) {
```
Joshua-Ashton, line 1008: Please no try/catch. Please use `.find` and check if != end. Maybe add a helper to do that and return nullptr.

WinterSnowfall: Done. I'd like to stress that these are highly unlikely to ever be hit, I just thought it would be a good idea for us to be a bit more robust and not like... crash.

src/d3d8/d3d8_device.cpp
```diff
@@ -994,30 +994,46 @@ namespace dxvk {
     HRESULT res = GetD3D9()->CreateStateBlock(d3d9::D3DSTATEBLOCKTYPE(Type), &pStateBlock9);
 
     D3D8StateBlock* pStateBlock = new D3D8StateBlock(this, Type, pStateBlock9.ref());
-    m_stateBlocks.insert(pStateBlock);
 
-    *pToken = DWORD(reinterpret_cast<uintptr_t>(pStateBlock));
+    m_stateBlocks.insert({m_token, pStateBlock});
```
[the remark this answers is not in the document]
WinterSnowfall: I don't see why we would. It worked fine without one so far.

src/d3d8/d3d8_device.h
```diff
@@ -416,9 +416,11 @@ namespace dxvk {
 
     D3DPRESENT_PARAMETERS m_presentParams;
 
-    D3D8StateBlock*                      m_recorder = nullptr;
-    std::unordered_set<D3D8StateBlock*>  m_stateBlocks;
-    D3D8Batcher*                         m_batcher  = nullptr;
+    D3D8StateBlock*                             m_recorder = nullptr;
+    DWORD                                       m_recorderToken = 0;
+    DWORD                                       m_token    = 0;
+    std::unordered_map<DWORD, D3D8StateBlock*>  m_stateBlocks;
```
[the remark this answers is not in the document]
WinterSnowfall: I'm not sure if we can avoid all the ugliness since we own the state blocks (unlike in d3d9), but I'll give it a shot.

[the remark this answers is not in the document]
doitsujin: Should be possible if you aggressively use `emplace` rather than `insert` for creating new blocks.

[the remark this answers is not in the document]
WinterSnowfall: m_recorder would still need to be available outside of the map though. Looking it up whenever is needed is more of a pain IMHO.

src/d3d8/d3d8_device.cpp
```diff
@@ -994,29 +994,46 @@ namespace dxvk {
...
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::DeleteStateBlock(DWORD Token) {
     // "Applications cannot delete a device-state block while another is being recorded"
     if (unlikely(ShouldRecord()))
       return D3DERR_INVALIDCALL;
 
-    D3D8StateBlock* block = reinterpret_cast<D3D8StateBlock*>(Token);
-    m_stateBlocks.erase(block);
+    if (unlikely(m_stateBlocks.find(Token) == m_stateBlocks.end())) {
```
Joshua-Ashton, line 1030: I said before in my review to save off the iterator from find. Right now you are looking it up twice for no reason.

src/d3d8/d3d8_device.cpp
```diff
@@ -994,29 +994,46 @@ namespace dxvk {
...
+      Logger::err("Invalid token passed to DeleteStateBlock");
+      return D3DERR_INVALIDCALL;
+    }
+
+    D3D8StateBlock* block = m_stateBlocks.at(Token);
+    m_stateBlocks.erase(Token);
```
Joshua-Ashton, line 1036: Just delete the iter from find.

src/d3d8/d3d8_device.h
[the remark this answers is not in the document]
WinterSnowfall: Ok, this bit _should_ be fine now, hopefully.

src/d3d8/d3d8_device.cpp
WinterSnowfall: Doing that won't allow us to return `D3DERR_INVALIDCALL`, which is the whole point of this exercise? Unless I'm missing something.

src/d3d8/d3d8_device.cpp
WinterSnowfall: Hopefully fine now :sweat_smile:.

WinterSnowfall:
Finalized now, hopefully. Thanks for all the nudging in the right directions.

AlpyneDreams:
Looks good to me. I don't have permission to merge so one of the maintainers will have to do that.

PR #4123, opened 2024-07-08
Add more d3d8 to README

artemyto:


artemyto:
```
samplers: Shows the current number of sampler pairs used [D3D9 Only]
```
Does samplers DXVK_HUD option also support D3D8?

artemyto:
```
DXVK Native comes with a slim set of Windows header definitions required for D3D9/11 and the MinGW headers for D3D9/11.
```
Does DXVK Native also support D3D8?

README.md
```diff
@@ -9,7 +9,7 @@ The most recent development builds can be found [here](https://github.com/doitsu
 Release builds can be found [here](https://github.com/doitsujin/dxvk/releases).
 
 ## How to use
-In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, copy or symlink the DLLs into the following directories as follows, then open `winecfg` and manually add DLL overrides for `d3d11`, `d3d10core`, `dxgi`, `d3d9` and `d3d8`.
+In order to install a DXVK package obtained from the [release](https://github.com/doitsujin/dxvk/releases) page into a given wine prefix, copy or symlink the DLLs into the following directories as follows, then open `winecfg` and manually add DLL overrides for `d3d11`, `d3d10core`, `dxgi`, `d3d9` and `d3d8`. To use d3d8 with dxvk you have to use both the d3d8 and d3d9 DLLs as d3d8.dll depends on d3d9.dll.
```
Blisto91, line 12: I'm not sure there is much of a need to explicitly point this out as on Linux I think it is generally expected that you'd either use all the dxvk DLLs in your Wine prefix or none of them. Or at least I expect not doing so is rather niche.

We don't currently point out any of the other interdependencied like d3d10core needing d3d11 and dxgi or d3d11 needing dxgi.

Blisto91: Or at least if this was added I think it might be better with a separate note paragraph thatbaldo mentions the other ones

artemyto: I think this is a really important information, which should be written somewhere. In my opinion software should be documented to minimize users' questions and to help future development (just in case new developer jouned the team, new contirutor wanted to help, the project has been forked etc). Even if for now some information seems to be obvious and you think most of the users won't try to use the software wrong way.

Blisto91: Yes I agree it would make sense to mention it somewhere. But it should be done for all the DLLs. Not sure if it is the best location though.

I'll let the devs comment on it

artemyto:
```
But it should be done for all the DLLs.
```
I agree
```
I'll let the devs comment on it
```
Thank you

PR #4128, opened 2024-07-09
[d3d9] Add a forceModeHeights config option

WinterSnowfall:
The emperor's new clothes, based off of #3615, with the following distinctions:
- naming
- it adds only the config option, no defaults for any game in particular
- forced mode heights parsing happens during adapter creation

This is needed with certain hardware combinations of monitors/GPUs that report an excessive amount of modes, as some games have troubles starting in such situations. In particular the [Aquanox](https://github.com/doitsujin/dxvk/issues/3255) series of games are a pain in this regard.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -614,6 +614,19 @@
 
 # d3d9.forceAspectRatio = ""
 
+# Force Mode Heights
+#
+# Only exposes modes with certain heights, if they are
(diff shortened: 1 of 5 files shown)


WinterSnowfall:
I no longer have the hardware that ran into these type of issues with the AquaNox games, and since there are other options out there for resolution/mode limiting (on Linux at least), might as well not keep this around forever.

PR #4132, opened 2024-07-10
[util] Disable direct buffer mapping for Max Payne 1

K0bin:
Improves performance by avoiding stalls.

The game locks an index buffer multiple times per frame.

warn:  Waiting for resource: Lock Buffer
- Pool: D3DPOOL_DEFAULT
- Usage: D3DUSAGE_WRITEONLY
- OriginalFlags: 0
- Flags: 0
- OffsetToLock: 0
- SizeToLock: 240
- Type: INDEXBUFFER
- Total buffer Size: 240
- Map Mode: DIRECT

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1129,6 +1129,10 @@ namespace dxvk {
     { R"(\\Art of Murder - FBI Confidential\\game\.exe$)", {{
       { "d3d9.cachedDynamicBuffers",       "True" },
     }} },
+    /* Max Payne 1 - Stalls waiting for an index buffer */
+    { R"(\\MaxPayne\.exe$)", {{
+      { "d3d9.allowDirectBufferMapping",    "False" },
+    }} },
   }};
 
 


WinterSnowfall:
@K0bin We might actually want hold on a bit here. I'm currently down a very deep rabbit hole of how D3DLOCK_DISCARD behaves differently between d3d8 and d3d9, namely:

- d3d9:

> The application discards all memory within the locked region. For vertex and index buffers, the entire buffer will be discarded. This option is only valid when the resource is created with dynamic usage (see [D3DUSAGE](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dusage)).

- d3d8:

> The application discards the entire vertex buffer. This enables Direct3D to return a pointer to a new memory area so that the dynamic memory access (DMA) and rendering from the old area do not stall. See Remarks.

Removing the D3DLOCK_DISCARD flag for D3DUSAGE_WRITEONLY buffers seem to get rid of some of the queue syncs and also fixes texture rendering in Rayman 3.

K0bin:
Max Payne doesn't use DISCARD for the relevant locks here.

WinterSnowfall:
> Max Payne doesn't use DISCARD for the relevant locks here.

Hmmm, alright then, but there's 2 less queue syncs now than before.

src/util/config/config.cpp
```diff
@@ -1129,6 +1129,10 @@ namespace dxvk {
     { R"(\\Art of Murder - FBI Confidential\\game\.exe$)", {{
       { "d3d9.cachedDynamicBuffers",       "True" },
     }} },
+    /* Max Payne 1 - Stalls waiting for an index buffer */
+    { R"(\\MaxPayne\.exe$)", {{
+      { "d3d9.allowDirectBufferMapping",    "False" },
+    }} }
```
Blisto91, line 1135: Missing comma?

Blisto91 (approved):

PR #4134, opened 2024-07-11
[util] Reorder config options by API

WinterSnowfall:
Suggested by @Blisto91. Makes more sense to order them this way, now that d3d8 is part of the picture. I'll let @Blisto91 follow up if he also wants to order them alphabetically :wink: .

I've also checked if we missed to transition any d3d8 games from the d3d9 section, but it all looks good as is.


Blisto91:
Verified the Middle Earth configs works as indented

PR #4135, opened 2024-07-11
[meta] Rearrange readme

Blisto91:
Supersedes https://github.com/doitsujin/dxvk/pull/4123

- Rearrange the readme so the entries are consistently listed in a given section where it makes sense. They are ordered from project info -> usage -> building -> native.
- Moved up `Online multi-player games` so it sits near the start of the How to use section
- Also moved up `HUD` above Logs since it is mentioned under How to use so you don't have to scroll far.
- Change `Troubleshooting` to a `Build troubleshooting` subsection under Build instructions.
- Added a `DLL dependencies` subsection under How to use.
- Remove `Setup has to be done manually in this case.` under Compiling manually as i believe it is a reference to the old build script.
- Add a missing `d3d8` in the same subsection and rearrange the API mentions so they consistently go from lowest to highest.

Rendered version https://github.com/Blisto91/dxvk/blob/readme-boomboomboom/README.md


artemyto:
I think the interoperability between dxvk's dxgi.dll and vkd3d-proton also should be mentioned.

Also maybe the Vulkan 1.3 requirement.

Calinou:
By the way, the GitHub repository's description should be changed as it currently says:

> Vulkan-based implementation of D3D9, D3D10 and D3D11 for Linux / Wine

It might be worth mentioning Windows support while we're at it :slightly_smiling_face:

PR #4143, opened 2024-07-13
[d3d8] Add an option to respect DISCARD only for dynamic write-only buffers

WinterSnowfall:
Fixes missing textures in Rayman 3, a problem that based on online testimonies started to occur with the transition to Windows 7.

Test on Windows XP have shown that the game renders properly, so native behavior on age-accurate systems seem to differ from d3d9, as in D3DLOCK_DISCARD appears to be respected only for D3DUSAGE_DYNAMIC + D3DUSAGE_WRITEONLY.

Making it a draft for now, as it would need more testing to ensure there are no regressions in other titles.

### DIFF
--- a/src/d3d8/d3d8_buffer.h
+++ b/src/d3d8/d3d8_buffer.h
@@ -18,6 +18,7 @@ namespace dxvk {
       : D3D8Resource<D3D9, D3D8> (pDevice, std::move(pBuffer))
       , m_pool                   (Pool)
       , m_usage                  (Usage) {
+      m_options = this->GetParent()->GetOptions();
     }
 
     HRESULT STDMETHODCALLTYPE Lock(
@@ -25,6 +26,13 @@ namespace dxvk {
             UINT   SizeToLock,
             BYTE** ppbData,
             DWORD  Flags) {
+
+      if (m_options->forceLegacyDiscard &&
+          (Flags & D3DLOCK_DISCARD) &&
+         !((m_usage & D3DUSAGE_DYNAMIC) &&
+           (m_usage & D3DUSAGE_WRITEONLY)))
+          Flags &= ~D3DLOCK_DISCARD;
+
       return this->GetD3D9()->Lock(
         OffsetToLock,
         SizeToLock,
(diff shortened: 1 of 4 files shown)


WinterSnowfall:
Turns out this isn't exactly free (performance-wise), so I've hidden it behind a config option and enabled it only for Rayman 3. I have yet to see another game relying on this type of behavior, anyway.

mrdeathjr28:
i can confirm game render are fixed

rmn3fx

K0bin:
The docs only mention DYNAMIC. Does WRITEONLY make a difference for DISCARD?

WinterSnowfall:
> The docs only mention DYNAMIC. Does WRITEONLY make a difference for DISCARD?

For this game in particular, yes. For what it's worth d3d8to9 [is already doing this globally](https://github.com/crosire/d3d8to9/blob/2b25f270c4119437e80e52f41d0dc30236c62865/source/d3d8to9_vertex_buffer.cpp#L88), however handling it this way appears to add a few queue sync and is generally slower at times, hence the config option.

As mentioned in the commit, this behavior seems to have been default on Windows XP, but modern Windows will behave inline with what D3D9 does, namely allow DISCARD for both DYNAMIC and DYNAMIC + WRITEONLY, which is why the game is also broken on native currently. 

The workarounds [you'll find online](https://www.gog.com/forum/rayman_series/rayman_3_missing_textures) involve disabling HWVP entirely, using a game ini config option, specifically in order to disallow DISCARD on the affected buffers.

RibShark:
It makes a difference because Rayman 3 is doing undefined behaviour, namely writing to buffers after unlocking them (off by one error in reference counting I believe). Handling that is probably the more correct fix for this issue.

WinterSnowfall:
> It makes a difference because Rayman 3 is doing undefined behaviour, namely writing to buffers after unlocking them (off by one error in reference counting I believe). Handling that is probably the more correct fix for this issue.

If that were indeed the only problem, it would have been fixed just by disabling direct buffer mapping. Which it was not.

doitsujin:
> It makes a difference because Rayman 3 is doing undefined behaviour, namely writing to buffers after unlocking them (off by one error in reference counting I believe). Handling that is probably the more correct fix for this issue.

Pretty sure we already handle this on the D3D9 side, which this calls into. D3D9 games do this all the time too.

PR #4144, opened 2024-07-13
Small fix for the config template and the profiles

tobiasjakobi:
Like stated in the commit message, the feature was present in one of the d8vk branches at some point. But it never landed in DXVK.

For reference, I restored the feature here, rebase on top of current master: https://github.com/tobiasjakobi/dxvk/tree/d3d8-shadowbuf

However we might need some documentation of `d3d8.scaleDref` in the config template.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -709,16 +709,3 @@
 # - True/False
 
 # d3d9.countLosableResources = True
-
-# Use NVIDIA Shadow Buffers
-# 
-# Vendor behavior for GeForce3 and GeForce4 cards that allows
-# sampling depth textures with non-normalized Z coordinates
-# and applies hardware shadow filtering.
-#
-# Supported values:
-# - True/False
-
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
Hmmm, I guess this was meant to be PR-ed separately, but got lost in translation during upstreaming. @AlpyneDreams might want to rectify it at some point, I suggest we leave it as is for now.

dxvk.conf
```diff
@@ -707,18 +707,3 @@
 #
 # Supported values:
 # - True/False
-
-# d3d9.countLosableResources = True
```
artemyto, line 711: This line shouldn't be deleted

tobiasjakobi: Good point, guess I was a bit overzealous there :smiling_face_with_tear: 

Fixed.

K0bin (approved):

PR #4152, opened 2024-07-15
[utils] Hide Intel Gpu for Batman Arkham Knight

FelixK15:
Without this fix, Batman Arkham Knight won't boot on Intel GPUs since the code explicitly checks for vendor id `0x10DE` & `0x1002` and if the vendor id is different, the game won't boot.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -55,6 +55,11 @@ namespace dxvk {
     /* D3D11 GAMES                                */
     /**********************************************/
 
+    /* Batman Arkham Knight - doesn't like intel vendor id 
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
Why not `dxgi.hideIntelGpu = True` then?

FelixK15:
If I read the code in https://github.com/doitsujin/dxvk/blob/master/src/dxgi/dxgi_adapter.cpp#L289 correctly, wouldn't that result in a vendorId of `0xDEAD`? This would also fail for Batman Arkham Knight because it only whitelists vendoir Id `0x10DE` & `0x1002`.

FelixK15:
Changed to `dxgi.hiteIntelGpu = True` as suggested

PR #4155, opened 2024-07-16
[util] Add configs for Splinter Cell Conviction

Blisto91:
The game will sometimes black screen upon alt-tab without deviceLossOnFocusLoss. 
Spoofing the GPU through dxgi to a known one from its list prevents the System Detection tool from complaining that it is unsupported on some setups.

The reason for the spoof being dxgi and not d3d9 is that it is what is used during system validation. Deleting systemdetection.dll will make the game not load dxgi.dll at all (besides via the ubisoft overlay).
The game itself uses d3d9 to determine vendor for e.g. if Nvapi should be loaded or not.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -996,9 +996,17 @@ namespace dxvk {
       R"(|The Lord of the Rings, The Rise of the Witch-king)\\game\.dat$)", {{
       { "d3d9.cachedDynamicBuffers",        "True" },
     }} },
-    /* WRC4 - Audio breaks above 60fps */
+    /* WRC4 - Audio breaks above 60fps           */
     { R"(\\WRC4\.exe$)", {{
       { "d3d9.maxFrameRate",                "60" },
+    }} },
+    /* Splinter Cell Conviction - Alt-tab black  *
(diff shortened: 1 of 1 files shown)


mrdeathjr28:
> Deleting systemdetection.dll will make the game not load dxgi.dll at all (besides via the ubisoft overlay). The game itself uses d3d9 to determine vendor for e.g. if Nvapi should be loaded or not.

this a very good date and in my case works

Blisto91:
It shouldn't be necessary to delete the DLL anymore on Proton bleeding edge for this game. The issue that prevented you from continuing from the system validation pop up have been fixed.
Game should have a CPU Affinity with 4 threads like on Windows instead of just 1 then.

mrdeathjr28:
> It shouldn't be necessary to delete the DLL anymore on Proton bleeding edge for this game. The issue that prevented you from continuing from the system validation pop up have been fixed. Game should have a CPU Affinity with 4 threads like on Windows instead of just 1 then.

in my case appear another solution is change this value* (ConvictionUserPC):

[SystemValidation]
RunValidation=True

to

[SystemValidation]
RunValidation=False

*also works on im alive too (UserSettings)

PR #4161, opened 2024-07-19
Fix issues with CheckDeviceFormat & CheckDeviceFormatConversion

### LINKED ISSUE #4158: Serious Sam 2 Videos are broken
image

Affects Serious Sam 2.
Reported by OnMars and bisected down to af0009c5debe33cffd5e412d0ff3d8adcd064c55.

K0bin:
Fixes #4158

Serious Sam 2 has 3 ways of uploading video frames:

- Lock XRGB texture and write XRGB data to that texture.
That happens when: YUY2 **textures** are supported in `CheckDeviceFormat`. DXVK master hits this code path on Nvidia GPUs because we check for sampling support on `VK_FORMAT_G8B8G8R8_422_UNORM` and Nvidia supports that.
- Lock YUY2 offscreen surface and write YUY2 data to that surface. Copy it to a XRGB texture afterwards using `StretchRect`.
That happens when: YUY2 **textures** are **un**supported but YUY2 **surfaces** are supported in `CheckDeviceFormat`. Additionally YUY2 is supported as a source format in `CheckDeviceFormatConversion`.
- Lock XRGB texture and write YUY2 data to that texture.
That happens when:
    - either YUY2 **textures** are **un**supported in `CheckDeviceFormat` but YUY2 **surfaces** are supported in `CheckDeviceFormat`. At the same time YUY2 is **not** supported as a source format in `CheckDeviceFormatConversion`.
    - neither YUY2 **textures** nor **surfaces** are supported in `CheckDeviceFormat`. This is what happens with DXVK master on RADV because it does **not** support `VK_FORMAT_G8B8G8R8_422_UNORM` images.

Given that it also manages to get writing to the XRGB texture right in one of the cases, I'm pretty sure this is a game bug because no actual hardware hits that.
I do not know why support for YUY2 textures makes it use the code path where it writes XRGB data.

The case that YUY2 surfaces are supported in `CheckDeviceFormat` but not as a source format in `CheckDeviceFormatConversion` is almost certainly wrong anyway.

The first commit is pretty straightforward and should be clear.

The second one is because `CheckDeviceFormatConversion` also decides whether or not a format is supported in `StretchRect`. We do support YUY2 there.

The docs say (emphasis mine):

> CheckDeviceFormatConversion can also be used to determine which combinations of source surface formats and destination surface formats are permissible in calls to [StretchRect](https://learn.microsoft.com/en-us/windows/desktop/api/d3d9helper/nf-d3d9helper-idirect3ddevice9-stretchrect).
[...]
The source format must be **a FOURCC format** or a valid back buffer format.

### DIFF
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -251,7 +251,8 @@ namespace dxvk {
           D3D9Format SourceFormat,
           D3D9Format TargetFormat) {
     bool sourceSupported = SourceFormat != D3D9Format::Unknown
-                        && IsSupportedBackBufferFormat(SourceFormat);
+                        && (IsSupportedBackBufferFormat(SourceFormat)
+                        || (IsFourCCFormat(SourceFormat) && !IsVendorFormat(SourceFormat)));
     bool targetSupported = TargetFormat == D3D9Format::X1R5G5B5
                         || TargetFormat == D3D9Format::A1R5G5B5
                         || TargetFormat == D3D9Format::R5G6B5
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -169,7 +169,11 @@ namespace dxvk {
       return D3D_OK;
 
     // Let's actually ask Vulkan now that we got some quirks out the way!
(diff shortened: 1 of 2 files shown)


mrdeathjr28:
yeah in my case fix cinematics

ss2fx

PR #4166, opened 2024-07-24
Update shouldSubmit to correctly handle descriptorPoolOverallocation

esullivan-nvidia:
Currently shouldSubmit will force the dxvk context to be flushed when too many descriptor pools have been allocated. This heuristic does not work when VK_NV_descriptor_pool_overallocation is in use because there will only ever be a single pool.

This change updates the heuristic to use the number of allocated sets when VK_NV_descriptor_pool_overallocation is in use.

Resolves the issue described in https://github.com/ValveSoftware/Proton/issues/7862

### DIFF
--- a/src/dxvk/dxvk_descriptor.cpp
+++ b/src/dxvk/dxvk_descriptor.cpp
@@ -72,7 +72,9 @@ namespace dxvk {
     // memory bloat. This may be necessary for off-screen
     // rendering applications, or in situations where games
     // pre-render a lot of images without presenting in between.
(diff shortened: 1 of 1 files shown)


esullivan-nvidia:
For the bug I included in the description the application ends up repeatedly calling vkAllocateDescriptorSets so the driver will eventually hit an OOM if the pool isn't reset. The game doesn't use d3d for presentation so in this case the descriptor pool ends up growing indefinitely.

doitsujin:
alright, I was kind of expecting pool memory to still be limited in some way, good to know.

PR #4168, opened 2024-07-26
[Build/CI] Build PDBs for MSVC builds + upload MSVC build artifacts.

num0005:
While DXVK seems mostly meant for usage with Wine, a lot of people end up using it anyways as a D3D9 replacement on Windows due to native driver support getting worse or being dropped altogether and replaced with D3D9on12.

Generating PDBs for MSVC builds makes debugging DXVK issues a lot easier on Windows, especially when porting legacy D3D9/D3D8 code to use it. I believe PDB symbols would have been generated anyways for debug builds, but testing with release builds is sometimes required. 

Not sure if uploading the artifacts for the MSVC builds is a good idea, I presume there might have been a reason those weren't uploaded before. If those builds are not intended for anything other than testing, I could look into whatever PDB generation is possible under `mingw` without breaking anything but I don't have (much) experience with cross compiling for Windows, and it seems there are some special considerations for symbol generation when Wine is involved.


doitsujin:
Sounds good, our release builds are MinGW so we wouldn't be shipping these, but having publicly available artifacts with PDBs seems useful. Probably don't need to hide pdb generation behind a config option either.

@Joshua-Ashton is more involved with the CI stuff, gonna wait for their feedback.

PR #4176, opened 2024-07-30
[d3d9] Handle remaining edge cases of Discard & Lockable

K0bin:
A bunch of edge cases we weren't handling properly.
This is gonna need some testing.

In theory we could also use VK_STORE_OP_DONT_CARE for discardable depth stencil surfaces but I don't think that's worth the additional complexity. Tracking the load op is already a bit of a mess. Games don't really use discardable depth stencil surfaces all that much and it would only help tilers anyway as far as I know.

### DIFF
--- a/src/d3d9/d3d9_common_texture.cpp
+++ b/src/d3d9/d3d9_common_texture.cpp
@@ -184,6 +184,17 @@ namespace dxvk {
     if (pDesc->MipLevels == 0 || pDesc->MipLevels > maxMipLevelCount)
       pDesc->MipLevels = maxMipLevelCount;
 
+    if (unlikely(pDesc->Discard)) {
+      if (!IsDepthStencilFormat(pDesc->Format))
+        return D3DERR_INVALIDCALL;
+
+      if (pDesc->Format == D3D9Format::D32_LOCKABLE
+        || pDesc->Format == D3D9Format::D32F_LOCKABLE
+        || pDesc->Format == D3D9Format::D16_LOCKABLE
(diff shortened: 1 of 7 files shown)


Blisto91:
Tested with 9 random d3d9 games of varied engines so far, not seen any issues. Tested on Nvidia.

src/d3d9/d3d9_common_texture.cpp
```diff
@@ -184,6 +184,14 @@ namespace dxvk {
     if (pDesc->MipLevels == 0 || pDesc->MipLevels > maxMipLevelCount)
       pDesc->MipLevels = maxMipLevelCount;
 
+    if (unlikely(pDesc->Discard)) {
+      if (!IsDepthStencilFormat(pDesc->Format))
+        return D3DERR_INVALIDCALL;
+
+      if (pDesc->Format == D3D9Format::D32_LOCKABLE || pDesc->Format == D3D9Format::D32F_LOCKABLE || pDesc->Format == D3D9Format::D16_LOCKABLE)
```
PatrickvL, line 191: Is the S8_LOCKABLE case left out intentionally here, or would it be possible to call IsLockableDepthStencilFormat here too?

PatrickvL, line 191: Is the S8_LOCKABLE case left out intentionally here, or would it be possible to call IsLockableDepthStencilFormat here too?

PatrickvL:
Is the S8_LOCKABLE case left out intentionally in common texture, or would it be possible to call IsLockableDepthStencilFormat here too?

Joshua-Ashton:
Looks fine

Issue #3725, opened 2023-11-20
[dxgi] Metro Exodus: Enhanced edition multi GPU crash regression


Blisto91:
Since commit https://github.com/doitsujin/dxvk/commit/83a294285e90c65554bc851c4d4a126aa15916a4 [Metro Exodus: EE](https://store.steampowered.com/app/412020/Metro_Exodus/) seems to crash right on startup on multi GPU systems.
The reason i believe it might be multi GPU only is because the issue goes away when i either force select my RX 6800 through `MESA_VK_DEVICE_SELECT=1002:73bf MESA_VK_DEVICE_SELECT_FORCE_DEFAULT_DEVICE=1` or `DXVK_FILTER_DEVICE_NAME=6800`

A extra note that i don't know is relevant to the issue or not is that both my GPUs support ray tracing which that game requires.

### Software information
[Metro Exodus: Enhanced edition](https://store.steampowered.com/app/412020/Metro_Exodus/)

If not using Bleeding Edge or master vkd3d-proton `VKD3D_CONFIG=dxr` is needed to launch the game as it requires ray tracing. Using radv below version 23.2.1 will also need `RADV_PERFTEST=rt`

### System information
- GPUs: RX 6800 and Zen4 rdna2 iGPU
- Monitors: Single monitor through displayport on RX 6800
- DE: KDE Wayland 5.27.9
- Driver: Days old mesa main
- Wine version: Proton Experimental
- DXVK version: https://github.com/doitsujin/dxvk/commit/83a294285e90c65554bc851c4d4a126aa15916a4 and up

### Log files
Proton and game log
[steam-1449560.log](https://github.com/doitsujin/dxvk/files/13420124/steam-1449560.log)
[metroexodus.log](https://github.com/doitsujin/dxvk/files/13420126/metroexodus.log)

Possible lines of interest from Proton log
```
18522.785:0124:0128:info:vkd3d-proton:dxgi_vk_swap_chain_init: Creating swapchain (1920 x 1080), BufferCount = 2.
18522.785:0124:0128:warn:vkd3d-proton:d3d12_resource_validate_desc: Invalid format 0.
err:   DXGI: CreateSwapChainForHwnd: Failed to create swap chain, hr -2147024809
```

Blisto91:
Log with extra logging. Made on X11
[steam-1449560-extra-logging.log](https://github.com/doitsujin/dxvk/files/13430257/steam-1449560-extra-logging.log)

Extra test scenarios:
[...]

PR #4192, opened 2024-08-07
[util] Set hideIntegratedGraphics = True for The Talos Principle 2

artemyto:
On my system with AMD Radeon RX 6800S and AMD Radeon 680M the game picks integrated card by default.

It is literally unplayable on 680M (like much less than 1 fps).

The hideIntegratedGraphics = True option fixes the issue.

Other users also reported the problem, for example you can read about it on Proton issue tracker: https://github.com/ValveSoftware/Proton/issues/7237#issuecomment-1798124091

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -45,6 +45,12 @@ namespace dxvk {
     { R"(\\Metro Exodus Enhanced Edition\\MetroExodus\.exe$)", {{
       { "dxvk.hideIntegratedGraphics",      "True" },
     }} },
+    /* The Talos Principle 2 picks GPU adapters
+     * by available VRAM, which causes issues on some
+     * systems with integrated graphics. */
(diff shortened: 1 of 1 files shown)


artemyto:
I don't know the root case of the problem, but maybe it is similar to https://github.com/doitsujin/dxvk/issues/3725?

artemyto:
Also you can see many reports from people about the game selecting integrated GPU on protondb: https://www.protondb.com/app/835960.

And the game don't work well for some users with  both integrated and discrete cards there, maybe those users didn't try to disable integrated graphics.

doitsujin:
I'm generally very hesitant to do workarounds like this since it doesn't typically affect all systems, e.g. on my own laptop the game works fine as-is. Not sure why the game would default to integrated either after all the various things we've done to ensure our adapter / monitor enumeration etc matches Windows as closely as possible, so this needs some investigation.

PR #4202, opened 2024-08-12
Fix various shader-related issues

doitsujin:
DCE Fixes invalid SPIR-V in some edge cases, e.g. unreachable continue blocks in a loop. Doing this at the front-end level would theoretically be possible, but much more difficult to reason about.

As for the rest, all of these changes mess with fossilize DBs so do it all in one go.

### DIFF
--- a/src/dxbc/dxbc_compiler.cpp
+++ b/src/dxbc/dxbc_compiler.cpp
@@ -1623,7 +1623,9 @@ namespace dxvk {
       
       case DxbcOpcode::Mad:
       case DxbcOpcode::DFma:
-        if (likely(!m_moduleInfo.options.longMad)) {
+        if (ins.controls.precise()) {
+          // FXC only emits precise mad if the shader explicitly uses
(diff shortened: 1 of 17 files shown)


src/dxbc/dxbc_compiler.cpp
```diff
@@ -2046,37 +2048,25 @@ namespace dxvk {
...
-            m_module.opCompositeExtract(componentType, src.at(0).id, 1, &idx),
-            m_module.opCompositeExtract(componentType, src.at(1).id, 1, &idx));
-        }
-
-        // Unconditionally mark as precise since the exact order of operation
-        // matters for some games, even if the instruction itself is not marked
-        // as precise.
-        m_module.decorate(dst.id, spv::DecorationNoContraction);
+    for (uint32_t i = 0; i < componentCount; i++) {
+      if (dst.id) {
```
PatrickvL, line 2055: This could alternatively check `i > 0` (or `i == 0` when the if/else blocks were swapped)

PR #4209, opened 2024-08-17
[native] Don't leave empty log files

Blisto91:


Blisto91:
Hm i may have misunderstood the original situation that made me make this PR.

I was noticing that the dxvk version shipping with the Linux native version of Men of War 2 was always leaving empty dxvk logs in its game folder. That i find now is even when i set dxvk log level to info myself.
If i then shove dxvk master in to be used the game will create and fill a d3d11 log like expected but then also create a empty dxgi log. The dxgi log will also get filled if i myself set log level to info.
So the empty files kinda confused me as to what was going on.

PR #4210, opened 2024-08-18
[util] Solve Swapchain Error for Time Leap Paradise

qinlili23333:
Solve this error `err:   D3D9SwapChainEx: Failed to recreate swap chain: VK_ERROR_NATIVE_WINDOW_IN_USE_KHR`
image

After fix it works

PS: if run under 4K, LAA patch is suggested to avoid crash

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -661,6 +661,10 @@ namespace dxvk {
     { R"(\\k2\.exe$)", {{
       { "d3d9.memoryTrackTest",             "True" },
     }} },
+    /* Time Leap Paradise SUPER LIVE             */
+    { R"(\\tlpsl\.exe$)", {{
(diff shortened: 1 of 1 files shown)


Blisto91:
@K0bin Wanted to explore maybe doing this be default when `VK_ERROR_NATIVE_WINDOW_IN_USE_KHR` is encountered

qinlili23333:
[tlpsl.zip](https://github.com/user-attachments/files/16822932/tlpsl.zip)
Added trace.
image
And now I'm 100% sure this is due to the game first create device without a swapchain just for detecting MSAA and resolution the release it, after configuration dialog it creats a new device again and then swapchain.

qinlili23333:
Try to write a universal fix https://github.com/qinlili23333/dxvk/commit/18ffee14ec6da80ca7e10839007ed3934df4db06
Not a good solution, but works.

PR #4217, opened 2024-08-24
[d3d8] Improve handling of failed d3d9 calls

WinterSnowfall:
Mostly nits uncovered while I was (yet again) reviewing refcounts...

The only real concern is that in some cases we are creating & reffing objects even after the proxied calls to d3d9 are failing, which of course doesn't make any sense. Probably less of a problem with games that choke on such errors anyway, but as we know there are some that *expect* calls to error out in order to function properly.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -96,7 +96,7 @@ namespace dxvk {
           DWORD         MagicNumber;
         };
 
-        if(FAILED(res)) {
+        if (FAILED(res)) {
           if (DevInfoStructSize != sizeof(D3DDEVINFO_VCACHE))
             return D3DERR_INVALIDCALL;
 
@@ -170,7 +170,10 @@ namespace dxvk {
   HRESULT STDMETHODCALLTYPE D3D8Device::GetDeviceCaps(D3DCAPS8* pCaps) {
     d3d9::D3DCAPS9 caps9;
     HRESULT res = GetD3D9()->GetDeviceCaps(&caps9);
-    dxvk::ConvertCaps8(caps9, pCaps);
+
+    if (likely(SUCCEEDED(res)))
+      dxvk::ConvertCaps8(caps9, pCaps);
+
     return res;
   }
 
@@ -207,6 +210,10 @@ namespace dxvk {
   HRESULT STDMETHODCALLTYPE D3D8Device::CreateAdditionalSwapChain(
       D3DPRESENT_PARAMETERS* pPresentationParameters,
       IDirect3DSwapChain8** ppSwapChain) {
+    InitReturnPtr(ppSwapChain);
+
+    if (unlikely(pPresentationParameters == nullptr || ppSwapChain == nullptr))
+      return D3DERR_INVALIDCALL;
     
     Com<d3d9::IDirect3DSwapChain9> pSwapChain9;
     d3d9::D3DPRESENT_PARAMETERS params = ConvertPresentParameters9(pPresentationParameters);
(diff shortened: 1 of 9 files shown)


src/d3d8/d3d8_device.cpp
```diff
@@ -207,15 +207,17 @@ namespace dxvk {
...
     Com<d3d9::IDirect3DSwapChain9> pSwapChain9;
     d3d9::D3DPRESENT_PARAMETERS params = ConvertPresentParameters9(pPresentationParameters);
     HRESULT res = GetD3D9()->CreateAdditionalSwapChain(
       &params,
       &pSwapChain9
     );
-    
-    *ppSwapChain = ref(new D3D8SwapChain(this, std::move(pSwapChain9)));
+
+    if (likely(!FAILED(res)))
```
doitsujin, line 219: `SUCCEEDED`?

WinterSnowfall: That's a macro antonym I tend to always spell wrong :sweat_smile: , but done.

WinterSnowfall:
I've added some general robustness to d3d8 device calls - although highly unlikely, there was some potential null pointer dereference going on. Also added additional validations where applicable, especially around CopyRects (we still need to make it more restrictive, but it's a start).

WinterSnowfall:
Looks good after a quick round of random game tests - on the added validation part it now (mostly) mirrors d9vk anyway, so no surprise there.

WinterSnowfall:
The ghost of Managed Buffer Placement still haunts us to this day. I've also now removed some remaining sleight of hand (on buffer GetDesc) we were doing to facilitate it.

WinterSnowfall:
I've tested this PR with the entire battery of ~140 d3d8 games I have and they're all fine. Should be good to mainline.

Blisto91:
Tested it quickly with 12 random games to verify amd radv is fine too and haven't hit any issues

src/d3d8/d3d8_device.cpp
```diff
@@ -1542,26 +1636,36 @@ namespace dxvk {
...
-      // Cache current shader
-      m_currentVertexShader = Handle;
-      
       GetD3D9()->SetVertexDeclaration(info->pVertexDecl);
-      return GetD3D9()->SetVertexShader(info->pVertexShader);
+      res = GetD3D9()->SetVertexShader(info->pVertexShader);
+
+      if (likely(SUCCEEDED(res)))
+        // Cache current shader
+        m_currentVertexShader = Handle;
```
doitsujin, line 1642: looks weird to have multiple lines here (even if one is a comment) but no { }

src/d3d8/d3d8_device.cpp
```diff
@@ -1542,26 +1636,36 @@ namespace dxvk {
...
-      m_currentVertexShader = Handle;
-
       //GetD3D9()->SetVertexDeclaration(nullptr);
       GetD3D9()->SetVertexShader(nullptr);
-      return GetD3D9()->SetFVF( Handle );
+      res = GetD3D9()->SetFVF(Handle);
+
+      if (likely(SUCCEEDED(res)))
+        // Cache current FVF
+        m_currentVertexShader = Handle;
```
doitsujin, line 1655: same here

src/d3d8/d3d8_device.cpp
```diff
@@ -1715,14 +1823,19 @@ namespace dxvk {
     }
 
     StateChange();
+    HRESULT res = GetD3D9()->SetPixelShader(pPixelShader);
 
-    // Cache current pixel shader
-    m_currentPixelShader = Handle;
+    if (likely(SUCCEEDED(res)))
+      // Cache current pixel shader
+      m_currentPixelShader = Handle;
```
doitsujin, line 1828: and here

doitsujin (commented):
Haven't investigated *every* single line of code in detail but overall this makes sense.

I generally prefer the `if (FAILED(hr)) return hr;` kind of pattern over what's done here since it makes the intended code path easier to follow in some cases, but I don't think should be changed now for the sake of it since it'll just break stuff. Might be something to keep in mind if these methods are getting touched again in the future though.

LGTM overall, just formatting nits that caught me off-guard.

WinterSnowfall:
> I generally prefer the `if (FAILED(hr)) return hr;` kind of pattern over what's done here since it makes the intended code path easier to follow in some cases

Hmmm, fair enough. I sort of went for the "let's not have 5 returns per call" approach where it could be helped, but I'll keep it in mind in the future.

A dressed the rest.

PR #4223, opened 2024-08-29
[meta] Add more information about anti-cheats

Tiagoquix:
Add more information about anti-cheats with examples to provide more security when using DXVK.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,14 +53,14 @@ Manipulation of Direct3D libraries in online multi-player games may be considere
 #### Linux (Proton/Wine)
 Most of the time, if a game runs without any issues via Proton, it probably also means it's safe to use DXVK with it.
 
-If a game is Steam Deck "Verified", you can probably use DXVK without worrying.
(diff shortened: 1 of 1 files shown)


mirh:
I'm not sure that mouthful is really doing more good than harm. 
Except destiny 2 (and even then, that was like 5 years ago before they added checks to gracefully acknowledge at least the linux case) I'm not aware of any other actively bad scenario.

doitsujin:
I don't like this. Even if a game is Steam Deck Verified, that rating solely pertains to the Steam Deck, not desktop Proton, and while it is often true that a Verified game will do just fine on desktop including any potential anti-cheat, we can't really pull statements like this out of our arse.

The original statement isn't there to scare people, it's there because I don't want to be responsible if someone *does* get in trouble for or while using DXVK with any given game.

mirh:
If you do actually find instances of games that in the year of the lord 2024 still outright bans you just for meekly swapping a dll, you could note that in the dxvk PCGW page.

PR #2466, opened 2022-01-28
Disable direct buffer mapping mode for Resident Evil games

K0bin:
Fixes #2453 
Fixes #2084
Fixes #2266

RE games lock the same D3DPOOL_DEFAULT + D3DUSAGE_DYNAMIC buffer every frame. DXVK chooses MAP_MODE_DIRECT for it, so all our measures to prevent stalling in `Lock` don't apply here and the game syncs with the GPU every frame.

I'd personally just get rid of MAP_MODE_DIRECT but @Joshua-Ashton told me on Discord to make this an option instead.

This will have to be rebased against #2157 if that PR is merged first.

PR #4224, opened 2024-08-29
[util] Document "d3d9.allowDirectBufferMapping" in dxvk.conf

Tiagoquix:
I'm not sure if the description of the option is correct; I described it as best I could interpret it from past commits and current usage.


mirh:
#2466 and #4132 make it seem kinda more complex and nuanced

PR #4225, opened 2024-08-30
[util] Make frame rate limiter enablement heuristic more robust

doitsujin:
Might need a bit of testing to ensure it still applies where necessary.

Should fix some cases where some mixture of swapchain buffering and timing errors cause us to enable the limiter even if the game has its own or if we're running at display refresh in a non-`present_wait` environment.


WinterSnowfall:
Tested this for a bit with 20-ish games, where I was previously seeing it trigger without cause. Appears to be fine now. Hopefully some VRR heroes can test the "positive" scenarios.

PR #4228, opened 2024-08-30
[d3d9] Add R16 and AL16 as known unsupported formats + some cleanup

WinterSnowfall:
Some games will use these two formats (excessively even) with CreateOffscreenPlainSurface (d3d9) / CreateImageSurface (d3d8), which won't fail even for known unsupported formats. We've had the same problem with P8 on d3d8 side in the past.

WineD3D supports both R16 and AL16 along these lines, but not with CreateTexture (nor does it advertise them as supported formats). This PR brings us on par and is also meant to silence a lot of `ConvertFormat: Unknown format encountered` warnings we've seen in logs for these two formats in particular.

I've also cleaned up some supported formats from GetUnsupportedFormatInfo, for good measure (I'm assuming nobody bothered to remove them from there once support was added). 

Draft for now as I still have to:
- test the behavior on native WinXP
- see if any other dubious formats need inclusion based on above

### DIFF
--- a/src/d3d9/d3d9_format.cpp
+++ b/src/d3d9/d3d9_format.cpp
@@ -429,6 +429,10 @@ namespace dxvk {
 
       case D3D9Format::RAWZ: return {}; // Unsupported
 
+      case D3D9Format::R16:  return {}; // Unsupported
+
+      case D3D9Format::AL16: return {}; // Unsupported
+
       default:
         Logger::warn(str::format("ConvertFormat: Unknown format encountered: ", Format));
         return {}; // Unsupported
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
After checks with both d3d8 and d3d9 on native Nvidia and AMD (thanks to @Blisto91), it looks like none of these formats are or were ever supported (thanks, WinXP), not even with CreateOffscreenPlainSurface and D3DPOOL_SCRATCH.

So, I've only left the necessary entries for them not to generate log spam, since they're now proven to be red herrings that might mislead people into thinking we're not supporting something we should be (I too was once one of them, I should know :).

Odd that WineD3D would support them, but following native behavior is of course the best option. Even stranger that a fair number of games from the era seem to attempt to employ them.

K0bin (approved):

PR #4230, opened 2024-08-31
[d3d8] Clamp BaseVertexIndex to INT_MAX during use

WinterSnowfall:
There's a minor type difference between BaseVertexIndex in d3d8 (UINT set by SetIndices) and in d3d9 (INT as input parameter for DrawIndexedPrimitive). The values are, I expect, too high to ever be hit, however if that happens we shouldn't overflow into a negative value with d3d9.

Clamping is perhaps not ideal, but in order to simulate native d3d8 behavior we technically have to accept the full range of UINT values of BaseVertexIndex, which we should also return during calls to GetIndices. I have verified this works fine with native.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -1168,7 +1168,7 @@ namespace dxvk {
           UINT             PrimitiveCount) {
     return GetD3D9()->DrawIndexedPrimitive(
       d3d9::D3DPRIMITIVETYPE(PrimitiveType),
-      m_baseVertexIndex, // set by SetIndices
+      static_cast<INT>(std::min(m_baseVertexIndex, static_cast<UINT>(INT_MAX))), // set by SetIndices
       MinVertexIndex,
       NumVertices,
       StartIndex,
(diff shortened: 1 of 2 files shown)


K0bin (approved):

src/d3d8/d3d8_device.cpp
```diff
@@ -1168,7 +1168,7 @@ namespace dxvk {
           UINT             PrimitiveCount) {
     return GetD3D9()->DrawIndexedPrimitive(
       d3d9::D3DPRIMITIVETYPE(PrimitiveType),
-      m_baseVertexIndex, // set by SetIndices
+      static_cast<INT>(std::min(m_baseVertexIndex, static_cast<UINT>(INT_MAX))), // set by SetIndices
```
K0bin, line 1171: I'd prefer `std::numeric_limits<uint32_t>::max()` but the climits one is fine too.

WinterSnowfall: I mean, that will ultimately return the climits macro anyway, so meh: https://en.cppreference.com/w/cpp/types/numeric_limits/max

K0bin: Yeah but it's more consistent with the rest of DXVK.
But like I said, I think the other version is fine too.

PR #4242, opened 2024-09-06
[util] Use cdecl calling convention for __wine_dbg_output

bylaws:
This matches wine's definition, fixes crashes when wine is built with clang.

### DIFF
--- a/src/util/log/log.cpp
+++ b/src/util/log/log.cpp
@@ -77,10 +77,14 @@ namespace dxvk {
         std::string adjusted = outstream.str();
 
         if (!adjusted.empty()) {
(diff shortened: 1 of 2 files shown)


doitsujin:
Needs some sort of fix to not break native builds, lgtm otherwise. Not sure how we ended up declaring this the wrong way in the first place.

Joshua-Ashton:
Why would this break native builds? They don't use this

bylaws:
> Why would this break native builds? They don't use this

__cdecl is a windows-ism

doitsujin (approved):
not a fan of the ifdef spam but it'll do for now i guess.

PR #4244, opened 2024-09-07
[util] Enable cachedDynamicBuffers for Steel Soldiers and FIFA 2003

WinterSnowfall:
Steel Soldiers does some "classic" CPU read-backs from DYNAMIC buffers. So does FIFA Football 2003. Plus some minor config file alignment, as it was a bit all over the place for the last two entries.


Blisto91:
Have the demo been checked for if the exe name is the same?

WinterSnowfall:
> Have the demo been checked for if the exe name is the same?

I've included it in the regex for FIFA 2003, but the demo for Steel Soldiers uses "demo.exe", so I'll let it rest.

Blisto91 (approved):

PR #4245, opened 2024-09-07
[d3d8] Validate viewport dimensions on SetViewport

WinterSnowfall:
D3D8 apparently validates this, while D3D9 does not. I've checked that WineD3D also performs this validation, ~~I am yet to check olden (WinXP) native behavior~~. Native d3d8 drivers, of WinXP fame, also perform this validation.

P.S.: @Blisto91 Also tested on modern Windows with AMD and it also performs the validation.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -926,6 +926,22 @@ namespace dxvk {
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::SetViewport(const D3DVIEWPORT8* pViewport) {
+    if (likely(pViewport != nullptr)) {
+      // we need a valid render target to validate the viewport
+      if (unlikely(m_renderTarget == nullptr))
+          return D3DERR_INVALIDCALL;
+
+      D3DSURFACE_DESC rtDesc;
+      HRESULT res = m_renderTarget->GetDesc(&rtDesc);
+
+      // D3D8 will fail when setting a viewport that's outside of the
+      // current render target, although this apparently works in D3D9
+      if (likely(SUCCEEDED(res)) &&
+          unlikely(pViewport->X + pViewport->Width  > rtDesc.Width ||
+                   pViewport->Y + pViewport->Height > rtDesc.Height))
+        return D3DERR_INVALIDCALL;
+    }
+
     StateChange();
     return GetD3D9()->SetViewport(reinterpret_cast<const d3d9::D3DVIEWPORT9*>(pViewport));
   }


doitsujin:
Does anything happen to existing viewports when the app binds a smaller render target?

WinterSnowfall:
> Does anything happen to existing viewports when the app binds a smaller render target?

Guess it's worth a quick test, but I doubt that affects anything unless the app also sets a new viewport.

WinterSnowfall:
@doitsujin After setting a smaller sized render target, apparently we already automatically adjust the viewport to fit (at least with a following call to GetViewport its new dimensions are equal to the smaller render target). WineD3D behaves the same. So no need to worry about this bit, just games being silly with SetViewport as per above.

WinterSnowfall:
Hmmm... I hadn't considered the X and Y (upper left) coordinates in the validation actually, and some games might do that. I'll adjust and retest this.

WinterSnowfall:
Alright, this should be sufficient. Verified with WineD3D and native WinXP drivers that the validation also takes the X and Y offsets into account (which makes sense).

src/d3d8/d3d8_device.cpp
```diff
@@ -926,6 +926,31 @@ namespace dxvk {
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::SetViewport(const D3DVIEWPORT8* pViewport) {
+    if (likely(pViewport != nullptr)) {
+      HRESULT res;
+
+      // get the D3D9 render target if we haven't cached it yet
+      if (unlikely(m_renderTarget == nullptr)) {
+        Com<d3d9::IDirect3DSurface9> pRT9 = nullptr;
```
K0bin, line 934: It would be nicer if you moved that into a little helper function. As far as I can tell, we have that same code in a bunch of places.

src/d3d8/d3d8_device.cpp
```diff
@@ -926,6 +926,31 @@ namespace dxvk {
...
+        // current render target, although this apparently works in D3D9
+        if (likely(SUCCEEDED(res)) &&
+            unlikely(pViewport->X + pViewport->Width  > rtDesc.Width ||
+                     pViewport->Y + pViewport->Height > rtDesc.Height))
+          return D3DERR_INVALIDCALL;
+      }
+    }
+
     StateChange();
     return GetD3D9()->SetViewport(reinterpret_cast<const d3d9::D3DVIEWPORT9*>(pViewport));
```
K0bin, line 955: Did you verify that it always accepts the viewport if no RT is bound? Seems a bit odd to me.

WinterSnowfall: Well, I guess we should fail if no RT is bound, that's true. I'll correct that.

K0bin (approved):
The `m_renderTarget` null check is technically unnecessary as far as I can tell but it doesn't hurt either so whatever.

WinterSnowfall:
Should fix the aspect ratio problem in #4246, so it fixes it for good in conjunction with #4258.

PR #4252, opened 2024-09-11
d3d{9,11}_swapchain: recreate swapchain on VK_SUBOPTIMAL_KHR

llyyr:
The vulkan wayland wsi returns suboptimal to when the window is fullscreened and not directly scanned out, and there are modifiers available that would allow the window to be directly scanned out. Recreate the swapchain if we receive suboptimal result.

This allows us recreate the swapchain to use a modifier that allows direct scan-out under winewayland on wayland compositors.


llyyr:
> We already recreate the swap chain if present returns `SUBOPTIMAL_KHR`.

Yeah I see it now. I'm not too familiar with dxvk but I'll look into why that isn't sufficient. Marking this as draft for now.

> if WSI can't give us what we want.

I might be misunderstanding [this comment](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/vulkan/wsi/wsi_common_wayland.c#L1536-1550), but from my understanding we're still picking the format but the WSI is responsible picking what modifier to use for the client. It would be a WSI bug if it picks a modifier but it's still suboptimal no?

doitsujin:
> Yeah I see it now. I'm not too familiar with dxvk but I'll look into why that isn't sufficient.

Looking at the mesa code it seems like present always returns `VK_SUCCESS`? Not sure if this is intended by the vk spec that return values ping-pong between `SUBOPTIMAL` and `SUCCESS`, but this would at least explain the issue.

We can definitely land this change in that case. Does this currently affect XWayland and/or Gamescope environments too?

> It would be a WSI bug if it picks a modifier but it's still suboptimal no?

Arguably yes, but Wayland and X11 aren't the only WSI implementations we end up running on even if those are obviously the priority. I'd prefer to be robust here.

PR #4255, opened 2024-09-11
[util] Fix string conversion 

kasper93:


doitsujin:
Would have been nice to at least have a PR description on **why** this fixes anything, but looking at the details of `GetEnvironmentVariableW` this looks correct at least.

PR #4259, opened 2024-09-13
Global buffer

doitsujin:
Basically do the vkd3d-proton thing of only creating one single VkBuffer per memory allocation whenever possible to open up some larger refactor. Works best if both maintenance4 and maintenance5 are supported, we still support creating a regular `VkBuffer` if necessary.

Needs testing with various games and various drivers.


artemyto:
Seems like these commits were merged into master. Shouldn't this PR be closed?

PR #4264, opened 2024-09-15
[d3d8] Relax viewport validation in windowed mode

WinterSnowfall:
Technically shouldn't be needed, but it addresses some problems @Blisto91 has been having with Rainbow Six 3 (in windowed mode), which sets the viewport a whole pixel wider than the render target, after some window resize shenanigans cause the swapchain to be reset like 3 times and end up with a width of 1 pixel less than the initial size that the game sets it to.

I have no idea why this happens, but this workaround should be pretty harmless.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -927,7 +927,7 @@ namespace dxvk {
 
   HRESULT STDMETHODCALLTYPE D3D8Device::SetViewport(const D3DVIEWPORT8* pViewport) {
     if (likely(pViewport != nullptr)) {
-      // we need a valid render target to validate the viewport
+      // We need a valid render target to validate the viewport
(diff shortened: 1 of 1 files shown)


Blisto91:
Verified

Blisto91 (approved):

WinterSnowfall:
If it works around the issue, I guess it's ready to ship. Also verified that GameMaker games are still fine, but those had a fullscreen viewport problem in the first place anyway.

K0bin:
This is pretty ugly but I don't really know how to debug why this happens. I guess it would be fine as a totally not permanent temporary workaround.

WinterSnowfall:
> I guess it would be fine as a totally not permanent temporary workaround.

My thoughts exactly. If it gets some problematic game behavior out of the way, I don't mind stepping an inch out of spec (at least for now).

PR #4270, opened 2024-09-16
[d3d8] Respect D3DCREATE_MULTITHREADED and make d3d8 thread safe

WinterSnowfall:
By popular demand, thread-safe d3d8, ported over from d3d9 and sprinkled wherever it affects state. Unsure if anything is actually affected by it (based on testing so far doesn't seem like it), but the overhead is probably rather negligible anyway in d3d8 space.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -44,7 +44,8 @@ namespace dxvk {
     , m_presentParams(*pParams)
     , m_deviceType(DeviceType)
     , m_window(hFocusWindow)
-    , m_behaviorFlags(BehaviorFlags) {
+    , m_behaviorFlags(BehaviorFlags)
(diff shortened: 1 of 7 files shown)


Blisto91:
> By popular demand

Didn't know I was popular

WinterSnowfall:
@Blisto91 Just so you are aware, this is now rebased on top of master, so we'll see for sure if it makes a difference in any of the problematic games. They're awaiting your fine and popular testing.

WinterSnowfall:
In short, this doesn't appear to fix anything... nor does it break anything. I guess it's a nice to have anyway. I'll test it a bit more with the full set of games to make sure it doesn't bite us back later.

WinterSnowfall:
Tested this with (almost) all the ~140 d3d8 games at my disposal and seems fine. No wins or losses as far as I can see. Thread safety might ultimately be a big meh for d3d8, but at least I also got in a small fix for the SetRenderState batcher logic.

Issue #4263, opened 2024-09-15
[d3d8] Old EverQuest client (takproject.net) still needs dgVoodoo


kai4785:
My working system looks like this:
Ubuntu 24.04
nvidia GeForce RTX 2060 driver 535.183.01 (from Ubuntu)
libvulkan1-1.3.275.0-1build1 (from Ubuntu)
[lutris-GE-Proton8-26-x86_64](https://github.com/GloriousEggroll/wine-ge-custom/releases/tag/GE-Proton8-26)
winecfg set `native,builtin` for all dxvk libraries: `d3d10core.dll`, `d3d11.dll`, `d3d8.dll`, `d3d9.dll`, `dxgi.dll`
`d3d8.dll` from [dgVoodoo2_8](https://dege.freeweb.hu/dgVoodoo2/dgVoodoo2/) ([direct link](http://dege.freeweb.hu/dgVoodoo2/bin/dgVoodoo2_8.zip))
`d3d11.dll` and `dxgi.dll` from dxvk-2.4

This works for the [old EverQuest client for The Al'Kabor Project server](https://wiki.takp.info/index.php/Getting_Started_on_Linux_(v2)).

If I copy the remaining dxvk libraries into place (`d3d10core.dll`, `d3d8.dll`, `d3d9.dll`), the game crashes after character select (which coincides with calling into d3d8).

Ideally I want to switch away from dgVoodoo and go with just dxvk now that there's d3d8 support.

I have tried looking at the diff of the log between the working environment and the broken one with `export WINEDEBUG=+loaddll,+d3d`, but nothing jumps out to me. I do notice that dxvk's `d3d8.dll` causes `d3d9.dll` to load when it does not when using dgVoodoo's version. But anything that looks like an error when it fails seems to also be in the logs when it is playable. Those logs are ~22MiB. Should I upload those?

Here's logs with just `export WINEDEBUG=+loaddll`. In the zip, `good.log` is with `dgVoodoo`, `bad.log` is with all dxvk libraries.
[logs.zip](https://github.com/user-attachments/files/17004265/logs.zip)

WinterSnowfall:
You'll need both the d3d8 and d3d8 dlls from dxvk to use with d3d8 games, [as descried in the readme](https://github.com/doitsujin/dxvk?tab=readme-ov-file#dll-dependencies). If a d3d8 game doesn't also load dxvk's d3d9.dll, it will crash.

kai4785:
[...]

PR #4274, opened 2024-09-18
[d3d9] Optimize SWVP devices

K0bin:
Needs lots of testing.

This makes D3D9 devices, that are configured to always use software vertex processing (so not `MIXED`), always use the late per draw buffer upload path. We copy the vertex data that each specific draw accesses to a temporary buffer and render from that, similar to Up-draws.
This makes sense because games that use pure SWVP expect vertex processing to be synchronous which has lead to both bugs and performance problems. For example we used to run into issues when respecting NOOVERWRITE or have dozens or even hundreds of queue syncs per frame. Considering that SWVP is supposed to run on the CPU, the amount of vertices is hopefully small.

I hope this won't impact more modern or demanding games.

The game that inspires this was Phantasmat from this comment:
https://github.com/doitsujin/dxvk/issues/4263#issuecomment-2358659238

It uses a single 96,000 byte vertex buffer (POOL = DEFAULT, USAGE = WRITEONLY, FVF != 0) and writes data to it before every single draw. Ofc it also doesn't specify a lock range, so we end up uploading the entire 96 KB buffer over and over again, run out of staging memory and then stall. It is a 2D game, so with this PR we upload 4 vertices for every draw.

### DIFF
--- a/src/d3d9/d3d9_common_buffer.cpp
+++ b/src/d3d9/d3d9_common_buffer.cpp
@@ -74,14 +74,6 @@ namespace dxvk {
     if (!(m_desc.Usage & (D3DUSAGE_DYNAMIC | D3DUSAGE_WRITEONLY)))
       return D3D9_COMMON_BUFFER_MAP_MODE_BUFFER;
 
-    // Tests show that DISCARD does not work for pure SWVP devices.
-    // So force staging buffer path to avoid stalls.
-    // Dark Romance: Vampire in Love also expects draws to be synchronous
-    // and breaks if we respect NOOVERWRITE.
-    // D&D Temple of Elemental Evil breaks if we respect DISCARD. 
-    if (m_parent->CanOnlySWVP())
-      return D3D9_COMMON_BUFFER_MAP_MODE_BUFFER;
-
     if (!options->allowDirectBufferMapping)
       return D3D9_COMMON_BUFFER_MAP_MODE_BUFFER;
 
@@ -134,7 +126,8 @@ namespace dxvk {
       memoryFlags |= VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
     }
 
-    if ((memoryFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT) && m_parent->GetOptions()->cachedDynamicBuffers) {
+    if ((memoryFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT) && (m_parent->GetOptions()->cachedDynamicBuffers || m_parent->CanOnlySWVP())) {
+      // Never use uncached memory on devices that support SWVP because we might end up reading from it.
       memoryFlags &= ~VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
       memoryFlags |= VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
                   |  VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
--- a/src/d3d9/d3d9_common_buffer.cpp
+++ b/src/d3d9/d3d9_common_buffer.cpp
@@ -111,12 +111,20 @@ namespace dxvk {
       info.stages |= VK_PIPELINE_STAGE_HOST_BIT;
       info.access |= VK_ACCESS_HOST_WRITE_BIT;
 
-      if (!(m_desc.Usage & D3DUSAGE_WRITEONLY))
-        info.access |= VK_ACCESS_HOST_READ_BIT;
-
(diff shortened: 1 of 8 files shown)


K0bin:
cc @WinterSnowfall

WinterSnowfall:
I can throw this at a bunch of games of course, but I think it would be useful and very helpful (since enhancing the HUD is a trend now) to also add the type of VP (based on device type and on m_isSWVP in case of Mixed) as an element to the D3D9 HUD.

The use of **D3DCREATE_SOFTWARE_VERTEXPROCESSING** devices is AFAIK very limited even in d3d8 and generally only used as a fallback in case HW or Mixed modes fail. A very limited set of games let you pick which to use.

WinterSnowfall:
This PR also properly fixes https://github.com/AlpyneDreams/d8vk/issues/179 , on which we had more or less given up in d8vk. The Supreme Commander d3d8 games can now be played with correct text rendering even without the "Nvidia workaround" configuration (which affected performance very negatively).

dungeon007:
> This PR also properly fixes [AlpyneDreams#179](https://github.com/AlpyneDreams/d8vk/issues/179) , on which we had more or less given up in d8vk. The Supreme Ruler d3d8 games can now be played with correct text rendering even without the "Nvidia driver workaround" configuration option (which affected performance very negatively).

Back then i give up too, but it worked fine as is on Gallium NINE. Now we know what is a problem, as NINE have SVPs optimized.
Now you got it too.

src/d3d9/d3d9_hud.cpp
```diff
@@ -125,4 +125,40 @@ namespace dxvk::hud {
     return position;
   }
 
+  HudSWVPState::HudSWVPState(D3D9DeviceEx* device)
+          : m_device          (device)
+          , m_isSWVPText ("") {}
+
+
+  void HudSWVPState::update(dxvk::high_resolution_clock::time_point time) {
```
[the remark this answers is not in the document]
K0bin: Done

dungeon007:
> > Now you know what is a problem, as NINE have SVPs optimized.
> 
> It has nothing to do with that.

Of course, could be something else as. Phantasmat regressed from yesterday.

src/d3d9/d3d9_common_buffer.cpp
```diff
@@ -74,14 +74,6 @@ namespace dxvk {
     if (!(m_desc.Usage & (D3DUSAGE_DYNAMIC | D3DUSAGE_WRITEONLY)))
       return D3D9_COMMON_BUFFER_MAP_MODE_BUFFER;
 
-    // Tests show that DISCARD does not work for pure SWVP devices.
-    // So force staging buffer path to avoid stalls.
-    // Dark Romance: Vampire in Love also expects draws to be synchronous
-    // and breaks if we respect NOOVERWRITE.
-    // D&D Temple of Elemental Evil breaks if we respect DISCARD. 
-    if (m_parent->CanOnlySWVP())
```
misyltoad, line 82: Have we tested this on `D&D Temple of Elemental Evil` that it is actually this?

WinterSnowfall: Current version of the PR is fine with ToEE (at some point it broke it, but was corrected).

src/d3d9/d3d9_common_buffer.cpp
```diff
@@ -111,12 +111,20 @@ namespace dxvk {
...
-        info.access |= VK_ACCESS_HOST_READ_BIT;
-
       memoryFlags |= VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT
-                  |  VK_MEMORY_PROPERTY_HOST_COHERENT_BIT
-                  |  VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
+                  |  VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+
+      if ((m_desc.Usage & (D3DUSAGE_WRITEONLY | D3DUSAGE_DYNAMIC)) == 0
+        || DoPerDrawUpload()
+        || m_parent->CanSWVP()
```
WinterSnowfall, line 119: This may introduce a performance regression for MIXED mode HWVP games (which aren't that uncommon). I'd recommend bringing back CanOnlySWVP() to check for D3DCREATE_SOFTWARE_VERTEXPROCESSING exclusively and using it here. 

We shouldn't hit MIXED mode SWVP paths with dxvk anyway, because if a game sees it can HWVP in MIXED mode, it most likely will.

WinterSnowfall (changes requested):

WinterSnowfall (approved):

PR #4275, opened 2024-09-18
[d3d9] Optimize late buffer uploads with dirty bitmasks

K0bin:
Based on #4274

Optimizes late buffer uploads (managed or per-draw) similar to how we're doing it for textures.
This avoids having to iterate over all 16 vertex buffers multiple times for every draw.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4812,8 +4812,6 @@ namespace dxvk {
 
         if (texInfo == pResource) {
           m_activeTexturesToUpload |= 1 << i;
-          // We can early out here, no need to add another index for this.
-          break;
         }
       }
     }
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5300,16 +5300,10 @@ namespace dxvk {
       }
 
       if (unlikely(vbo->NeedsReadback())) {
-        // There's two ways the GPU can write to buffers in D3D9:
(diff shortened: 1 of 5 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -5237,11 +5259,9 @@ namespace dxvk {
           bool*                   pDynamicVBOs,
           bool*                   pDynamicIBO
   ) {
-    bool dynamicSysmemVBOs = true;
-    for (uint32_t i = 0; i < caps::MaxStreams && dynamicSysmemVBOs; i++) {
-      auto* vbo = GetCommonBuffer(m_state.vertexBuffers[i].vertexBuffer);
-      dynamicSysmemVBOs &= vbo == nullptr || (vbo->DoPerDrawUpload() || CanOnlySWVP());
-    }
+    const uint32_t usedBuffersMask = (m_state.vertexDecl != nullptr ? m_state.vertexDecl->GetStreamMask() : ~0u) & m_activeVertexBuffers;
+    bool dynamicSysmemVBOs = usedBuffersMask == m_activeVertexBuffersToUploadPerDraw;
```
doitsujin, line 5261: What's the intention here with the `==`?

doitsujin (commented):
Question, not sure how any of this is really supposed to work though.

src/d3d9/d3d9_device.cpp
K0bin: The problem is that we change the BaseVertexIndex so it can only work if we copy the data for all bound vertex buffers.

=> all bound buffers that get actually used by the draw need to be part of the bitmask.

PR #4277, opened 2024-09-19
[util] Fix small_vector move constructor

K0bin:
Sry about the bug.

### DIFF
--- a/src/util/util_small_vector.h
+++ b/src/util/util_small_vector.h
@@ -33,12 +33,36 @@ namespace dxvk {
     };
 
     small_vector(small_vector&& other) {
-      if (other.m_capacity == N) {
+      if (other.m_size < N) {
         for (size_t i = 0; i < other.m_size; i++) {
-          u.m_data[i] = std::move(other.u.m_data[i]);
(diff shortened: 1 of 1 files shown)


src/util/util_small_vector.h
```diff
@@ -33,13 +33,37 @@ namespace dxvk {
...
         for (size_t i = 0; i < other.m_size; i++) {
-          u.m_data[i] = std::move(other.u.m_data[i]);
+          new (&u.m_data[i]) T(std::move(*other.ptr(i)));
         }
       } else {
         u.m_ptr = other.u.m_ptr;
-        other.m_capacity = N;
+        m_capacity = other.m_capacity;
+        other.u.m_ptr = nullptr;
+        other.m_capacity = 1;
```
doitsujin, line 44: why 1? We initialize this to N for obvious reasons, it shouldn't ever be smaller.

doitsujin (changes requested):

src/util/util_small_vector.h
K0bin: I just wanted anything that's smaller than N of the other vendor to avoid calling free on the pointer after it has been moved to the new vector.

PR #4285, opened 2024-09-22
[util] Cap Prince of Persia (2008) max fps

Blisto91:
Game can get stuck when loading at framerates above 1000

Edit: Changed to 240fps


K0bin (approved):

K0bin:
Does 1000 really work? Shouldn't we go for something lower to make sure?

Blisto91:
From my testing to far yes. But i don't mind going lower if you think. The clothes physics behave a bit twitchy at those high framerates anyway

K0bin:
360 fps is the state of the hard for competitive esports shooters, I think it's fine going with 240 or 360 fps for a casual single player action adventure.

K0bin (approved):

PR #4292, opened 2024-09-24
[d3d9] ResolveZ: Only copy aspects that both images support

### LINKED ISSUE #4291: [d3d9] The First Templar has lighting issues, not present with wined3d
### Software information
 The First Templar, GOG copy, all settings set to maximum possible

### System information
- GPU: 7900 XTX
- Driver: Mesa 24.2.3
- Wine version: Proton Experimental
- DXVK version: Whatever Experimental has as of now

### Apitrace file(s)

wined3d trace:

https://drive.google.com/file/d/1Et4a9dbP9nOk7I3Nfkv8muZcozK26lhh/view?usp=sharing

dxvk trace:

https://drive.google.com/file/d/1PSOT1wePe9PWaDDG6_iLywewEsvOpWmp/view?usp=sharing

### Log files

Proton log: 
[steam-heroic-The First Templar.log](https://github.com/user-attachments/files/17119608/steam-heroic-The.First.Templar.log)

Shows a bunch of:

```
err:   DxvkContext: copyImageFb: Unsupported operation:
err:     srcFormat = VK_FORMAT_D32_SFLOAT_S8_UINT (aspect 6)
err:     dstFormat = VK_FORMAT_D32_SFLOAT (aspect 2)
```

### Screenshots that shows the issue

Ekran Görüntüsü_20240924_213851
Ekran Görüntüsü_20240924_214014
Ekran Görüntüsü_20240924_214027

K0bin:
Fixes #4291

The game uses the RESZ hack to copy from a `D3DFMT_D24S8` texture to one with format `D3DFMT_DF24`.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -7640,16 +7640,13 @@ namespace dxvk {
     const D3D9_VK_FORMAT_MAPPING srcFormatInfo = LookupFormat(srcDesc->Format);
     const D3D9_VK_FORMAT_MAPPING dstFormatInfo = LookupFormat(dstDesc->Format);
 
-    auto srcVulkanFormatInfo = lookupFormatInfo(srcFormatInfo.FormatColor);
-    auto dstVulkanFormatInfo = lookupFormatInfo(dstFormatInfo.FormatColor);
-
     const VkImageSubresource dstSubresource =
       dstTextureInfo->GetSubresourceFromIndex(
-        dstVulkanFormatInfo->aspectMask, 0);
+        dstFormatInfo.Aspect, 0);
 
     const VkImageSubresource srcSubresource =
       srcTextureInfo->GetSubresourceFromIndex(
-        srcVulkanFormatInfo->aspectMask, src->GetSubresource());
+        srcFormatInfo.Aspect, src->GetSubresource());
 
     const VkImageSubresourceLayers dstSubresourceLayers = {
       dstSubresource.aspectMask,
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -7640,14 +7640,20 @@ namespace dxvk {
...
+    VkImageSubresource dstSubresource =
       dstTextureInfo->GetSubresourceFromIndex(
         dstFormatInfo.Aspect, 0);
 
-    const VkImageSubresource srcSubresource =
+    VkImageSubresource srcSubresource =
       srcTextureInfo->GetSubresourceFromIndex(
         srcFormatInfo.Aspect, src->GetSubresource());
 
+    if ((dstSubresource.aspectMask & srcSubresource.aspectMask) != 0) {
```
doitsujin, line 7651: Can this ever be 0, i.e. are stencil-only formats a thing in D3D9?

doitsujin (commented):
question, lgtm otherwise

src/d3d9/d3d9_device.cpp
K0bin: I was worried about Depth -> Color like D32F -> R32F. I don't know whether that's allowed in RESZ, the docs are pretty much nonexistent. I *think* it should probably work as long as we don't pass 0 as the aspect masks to the backend.

K0bin: Oh and yes, stencil only formats are a thing: https://github.com/doitsujin/dxvk/blob/ca3aa2014b2f7252d84043e20d6e349ff849a786/src/d3d9/d3d9_format.cpp#L275

Should we explicitly error out int hat case?

doitsujin: Fair point about color dst, I *think* the backend supports that.

I guess we should error out if the ANDed mask is 0 and dst is not color.

PR #4296, opened 2024-09-26
[d3d8] GetInfo rework and various assorted fixes

WinterSnowfall:
I've reworked our implementation of GetInfo() to more closely follow what d9vk is doing, namely:
- `pQuery->Issue()` will never fail in d9vk (and even if can fail according to spec, we shouldn't care much, as `pQuery->GetData()` would also then fail).
- `D3DDEVINFOID_RESOURCEMANAGER` and `D3DDEVINFOID_VERTEXSTATS`, though technically supported by d3d8 are not implemented in d9vk. D3D9's `CreateQuery()` will always return `D3DERR_NOTAVAILABLE` for both as it stands, thus we shouldn't really speculate on the outcome (removed the "unlikely" hint).
- According to specs, D3D9 `GetData()` should always be called with `D3DGETDATA_FLUSH` for the queries that D3D8 supports.
- D3D9 `GetData()` returns are inline with what `GetInfo()` should return if it reaches that point.

The above allows us to get rid of the goto and restructure the flow a bit.

Also as part of this PR I've changed some `CopyRects` behavior:
- removed format difference check from `copyTextureBuffers`, since this is now done globally on `CopyRects` calls, as per spec.
- `CopyRects` depth stencil checks are done after the format difference checks, so there's no need to check both src and dest.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -528,10 +528,6 @@ namespace dxvk {
     HRESULT res = D3D_OK;
     D3DLOCKED_RECT srcLocked, dstLocked;
 
-    // CopyRects cannot perform format conversions.
-    if (srcDesc.Format != dstDesc.Format)
-      return D3DERR_INVALIDCALL;
-
     bool compressed = isDXT(srcDesc.Format);
 
     res = src->LockRect(&srcLocked, &srcRect, D3DLOCK_READONLY);
@@ -647,10 +643,8 @@ namespace dxvk {
 
     // This method cannot be applied to surfaces whose formats
     // are classified as depth stencil formats.
-    if (unlikely(isDepthStencilFormat(D3DFORMAT(srcDesc.Format)) ||
-                 isDepthStencilFormat(D3DFORMAT(dstDesc.Format)))) {
+    if (unlikely(isDepthStencilFormat(D3DFORMAT(srcDesc.Format))))
       return D3DERR_INVALIDCALL;
-    }
 
     StateChange();
 
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -18,7 +18,7 @@ namespace dxvk {
 
   static constexpr DWORD getShaderIndex(DWORD Handle) {
     if ((Handle & D3DFVF_RESERVED0) != 0) {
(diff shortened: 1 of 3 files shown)


WinterSnowfall:
Fixed a couple of other things I stumbled into:

- We were creating an invalid auto depth stencil if the D3D9 device wasn't created with the `EnableAutoDepthStencil` present param flag enabled. The stencil surface would crash in these cases on any direct calls, such as `GetDesc()`. Keeping it as nullptr is pretty much the correct behavior here.
- I honestly don't see how it's possible to create a depth stencil with D3DFMT_UNKNOWN as *everything* validates that away, however using D32 was incorrect, because it's technically unsupported. Replaced it with D24X8, although I don't think it's actually going to end up used anyway.

WinterSnowfall:
- While looking at #4308, I've noticed native d3d8 starts generating shader handles at 3 (then 5, 7, 9 etc. since the FVF codes are powers of two... what a bloody mess this API is, seriously). We were starting at 1, which was practically fine in our detection logic, but native avoids it because it technically **is** a FVF code, namely `D3DFVF_RESERVED0` (0x0001). I've bumped the mapping to skip 1 and also start at 3, so we're in line with native d3d8, although I don't expect it to matter much.

src/d3d8/d3d8_device.cpp
```diff
@@ -119,28 +120,22 @@ namespace dxvk {
...
         return S_FALSE;
     }
 
+    if (pQuery != nullptr) {
+      // Immediately issue the query. D3D9 will begin it automatically before ending.
+      pQuery->Issue(D3DISSUE_END);
+      // TODO: Will immediately issuing the query actually yield meaingful results?
+      // Only relevant once RESOURCEMANAGER or VERTEXSTATS are implemented by D9VK,
+      // since VCACHE queries will immediately return data during this call.
+      res = pQuery->GetData(pDevInfoStruct, DevInfoStructSize, D3DGETDATA_FLUSH);
```
doitsujin, line 136: tldr yes it will because there's no GPU involved here.

WinterSnowfall: I'd leave that note until (if ever) there's an implementation. Native drivers actually mull over if anything is returned on an initial call with D3DGETDATA_FLUSH, recommending one checks in a loop until results are available. That sounds... like something I'd rather not have done anyway.

PR #4300, opened 2024-09-27
Sampler pool

doitsujin:
This definitely needs at least one more pair of eyes since there's a lot of subtle bug / typo potential here, and some D3D9 testing.

Basically implements a sampler pool in the backend, to replace the D3D9 one, and to deduplicate sampler objects in general. This builds on the recent ref counting changes and works as follows:
- `DxvkSampler` is now allocated from an `unordered_map`. It's still a heap allocation, but at least only *one*.
- If a sampler's ref count reaches 0, we'll generally keep it alive and add it to a least-recently-used list. This is an extremely simple doubly-linked list of sampler objects and is based around the idea that resource lifetime tracking gives us the correct order for free.
- If we look up an existing sampler that has a ref count of 0, we simply take it out of the list.
- If there are too many samplers around, we'll just destroy some of the least recently used ones.
  The clean-up thread has a stricter threshold here so if we do end up destroying samplers, we keep that work off the CS thread.

Additionally, this uses a very compressed representation of sampler state, only using 24 bytes (16 of which are the border color, which isn't always used). LOD ranges and bias are stored in a fixed-point representation that meets the minimum requirements of the D3D11.3 spec, but does **not** implement round to nearest even.

This *should* hopefully be fast, but need to benchmark the overhead of calling `vkCreateSampler` all the time as we do in some D3D9 games.

**Edit:** Updated with a safe-guard against creating too many samplers. We have a hard limit especially if border colors are in play, but the only way we'll ever hit that is when a game decides to queue up thousands of draws with unique samplers every time.


src/d3d9/d3d9_device.cpp
```diff
@@ -6972,6 +6981,67 @@ namespace dxvk {
...
+    uint64_t submissionId = m_submissionFence->value();
+
+    while (++submissionId <= m_submissionId) {
+      m_submissionFence->wait(submissionId);
+
+      if (m_dxvkDevice->getSamplerStats().liveCount <= MaxSamplerCount)
+        return;
+    }
+
+    // If we end up *here*, good luck.
```
misyltoad, line 7041: Should maybe log something, but very unlikely to hit it.

doitsujin: Done. If we ever hit that, that's going to be our bug anyway.

src/d3d9/d3d9_device.cpp
```diff
@@ -6972,6 +6981,68 @@ namespace dxvk {
...
+    // many samplers that we need to wait for the GPU to free some.
+    // We should absolutely never hit this path in the real world.
+    Logger::warn("Sampler pool exhausted, synchronizing with GPU.");
+
+    Flush();
+    SynchronizeCsThread(DxvkCsThread::SynchronizeAll);
+
+    // Need to reset sampler stats here since we'd
+    // otherwise hit this path again the next time
+    m_lastSamplerStats = 0u;
```
DadSchoorse, line 7030: Shouldn't this be `m_lastSamplerStats = m_samplerBindCount << SamplerCountBits;`? Otherwise, the next time `m_samplerBindCount` is still some high value, and m_lastSamplerBindCount won't offset it because it's 0.

DadSchoorse: Or maybe even move it just before the return and store `(m_samplerBindCount << SamplerCountBits) | m_dxvkDevice->getSamplerStats().liveCount`. Or maybe I'm too stupid to understand threaded code, that's also always a possibility. :frog:

doitsujin: Yeah, I think we do actually need to update this with the live count while CS is idle.

The idea here was that CS would give us new stats before the next call, but that's not guaranteed to happen.

doitsujin: Fixed.

K0bin (approved):
LGTM

PR #4305, opened 2024-09-29
Implement resource relocation and rework meta passes

doitsujin:
Does three things:
- Uses new view/sampler for copies, blits, resolves etc instead of creating separate Vulkan objects. This part is largely a cleanup.
- Introduces functionality to relocate resources and recreate images with new usage flags on demand.
  This enables us to drop the whole meta copy usage thing in the front-ends and just enable those bits as necessary.
- Fixes a long-standing problem with Cuda/NVAPI interop where querying GPU addresses races with context methods.
  This might in fact need even more work since there are other methods relying on stable view handles, but this has technically always been broken at least for buffers.

Needs a bunch of testing with both D3D11 and D3D9.


K0bin (approved):
Nice work!

PR #4310, opened 2024-10-01
[dxvk] Fix use-after-move in DxvkBufferSlice ctor

bylaws:


doitsujin:
I had already done this yesterday actually.

PR #4320, opened 2024-10-04
[d3d8] Various Wine device tests fixes (Part 1)

WinterSnowfall:
Fixes a rather large portion (in terms of failures, at least) of Wine's D3D8 device tests.

The changes are:
- fixed a type derp on texture subresource lookup
- don't allow mix-matched adapter and buffer formats in windowed mode (apparently that's not a thing in D3D8)
- make some D3D8 present parameter validations visible to the calling application
- cache (and hold references to) all the back buffers tied to an additional swapchain
- various additional CheckDeviceMultiSampleType validations that native also does (~TODO: check these on D3D9 side directly also~ They indeed check out on D3D9 as well)
- failure of device resets due to invalid losable count should return D3DERR_DEVICELOST
- can't create a 0 x 0 backbuffer in fullscreen (but windowed mode works, absolutely lovely)
- D3D8 SetStreamSource won't update the stride if the buffer data is NULL (so the old one is retained)
- calls to SetRenderTarget should always be forwarded to D3D9, since they can affect the viewport
- stub ValidatePixelShader and ValidateVertexShader (needed mostly for Wine tests)
- textures/surfaces using DXT formats need to be checked for block alignment during creation

### DIFF
--- a/src/d3d9/d3d9_common_texture.cpp
+++ b/src/d3d9/d3d9_common_texture.cpp
@@ -156,6 +156,16 @@ namespace dxvk {
 
     if (pDesc->Width == 0 || pDesc->Height == 0 || pDesc->Depth == 0)
       return D3DERR_INVALIDCALL;
+
+    // Native drivers won't allow the creation of DXT format
+    // textures that aren't aligned to block dimensions.
+    if (IsDXTFormat(pDesc->Format)) {
+      D3D9_FORMAT_BLOCK_SIZE blockSize = GetFormatBlockSize(pDesc->Format);
+
+      if ((blockSize.Width  && (pDesc->Width  & (blockSize.Width  - 1)))
+       || (blockSize.Height && (pDesc->Height & (blockSize.Height - 1))))
+        return D3DERR_INVALIDCALL;
+    }
     
     if (FAILED(DecodeMultiSampleType(pDevice->GetDXVKDevice(), pDesc->MultiSample, pDesc->MultisampleQuality, nullptr)))
       return D3DERR_INVALIDCALL;
--- a/src/d3d8/d3d8.def
+++ b/src/d3d8/d3d8.def
@@ -1,3 +1,6 @@
 LIBRARY D3D8.DLL
 EXPORTS
+  ValidatePixelShader @ 2
+  ValidateVertexShader @ 3
+
   Direct3DCreate8 @ 5
--- a/src/d3d8/d3d8.sym
+++ b/src/d3d8/d3d8.sym
@@ -1,5 +1,7 @@
 {
   global:
+    ValidatePixelShader;
+    ValidateVertexShader;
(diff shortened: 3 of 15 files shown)


WinterSnowfall:
I'm happy to report that with the latest additions (Stubs of ValidatePixelShader and ValidateVertexShader) the D3D8 Wine device tests complete, rather than crash mid-way:

`0020:device: 38225 tests executed (34 marked as todo, 0 as flaky, 14247 failures), 1 skipped.`

A lot of the above failures are related to DXT texture support in various situations, so tentatively I'll fix those and leave this PR at that.

WinterSnowfall:
Checking block alignment during surface/texture creation for DXT formats has **drastically** reduced the number of failures in both d3d8 and d3d9 Wine device tests. Here are some numbers:

- Current master:
-- D3D8: _Doesn't complete due to missing ValidatePixelShader and ValidateVertexShader, but has roughly 14000+ failures when those are stubbed_
-- D3D9: `0020:device: 141598 tests executed (78 marked as todo, 0 as flaky, 15756 failures), 6 skipped.`

- This PR:
-- D3D8: `0020:device: 51365 tests executed (34 marked as todo, 0 as flaky, 1107 failures), 1 skipped.`
-- D3D9: `0020:device: 155338 tests executed (78 marked as todo, 0 as flaky, 2019 failures), 6 skipped.`

Just as a note I've checked both against Windows XP and there are only a bunch of failures (e.g. 7 for the D3D8 devices tests), so these mostly check out vs native behavior.

I still have a few checks to make, but in terms of PR scope I will leave it at that and split any future work in another PR.

WinterSnowfall:
Thanks to @Blisto91 for testing and finding a very odd regression with Comanche 4 (now fixed). I've also thrown this at quite a few d3d8 and d3d9 games and it all seems fine. Would potentially appreciate a bit more d3d9 testing before merging.

src/d3d9/d3d9_common_texture.cpp
```diff
@@ -156,6 +156,16 @@ namespace dxvk {
 
     if (pDesc->Width == 0 || pDesc->Height == 0 || pDesc->Depth == 0)
       return D3DERR_INVALIDCALL;
+
+    // Native drivers won't allow the creation of DXT format
+    // textures that aren't aligned to block dimensions.
```
doitsujin, line 161: does this requirement also pertain to smaller mips or only the top level, i.e. is something like 12x8 with two, three, four mips allowed respectively?

doitsujin (commented):
question

src/d3d9/d3d9_common_texture.cpp
WinterSnowfall: To be fair, I haven't checked, but the Wine tests at least only validate the top level.

Issue #3691, opened 2023-10-14
Civilization: Beyond Earth options menu crash


Blisto91:
Civ Beyond Earth crashes if once you get to the main menu you then press the options button. I have bisected this to not always be the case and that it seems to start with commit https://github.com/doitsujin/dxvk/commit/b746e1352b4c93b7d5f51d013f6aaef249c8aecd

While the logs makes it appear as a 32bit memory crash it acts a bit special in that the sequence above is the only way I've gotten it to crash in the little testing I've done so far. If i instead go in and play a game (starts fine without a crash) i can then access the options menu fine from there. Same if i from in game exit back to the main menu it will also work fine then.

### Software information
Sid Meier's Civilization: Beyond Earth

### System information
- GPU: 7900 xtx
- Driver: Few days old mesa-git
- Wine version: Proton Experimental
- DXVK version: https://github.com/doitsujin/dxvk/commit/b746e1352b4c93b7d5f51d013f6aaef249c8aecd and above

### Apitrace file(s)
Pending. Isn't cooperating on Linux at least so far.

Edit: Windows native trace https://drive.proton.me/urls/ZW6SBVVEE8#3YcjC5XZx662

### Log files
[CivilizationBE_DX11_dxgi.log](https://github.com/doitsujin/dxvk/files/12907847/CivilizationBE_DX11_dxgi.log)
[CivilizationBE_DX11_d3d11.log](https://github.com/doitsujin/dxvk/files/12907848/CivilizationBE_DX11_d3d11.log)
[CivilizationBE_DX11_d3d9.log](https://github.com/doitsujin/dxvk/files/12907849/CivilizationBE_DX11_d3d9.log)

c3Ls1US:
@Blisto91 I have encountered a similar menu-crashing issue in another game using Proton 8.04.

This hasn't happened before and I would like to compare your logs in order to precisely discern whether the bug is a regression from DXVK or wine, so would you mind providing another debug log, perhaps by setting the `PROTON_LOG` environment variable?

Blisto91:
[steam-65980.log](https://github.com/doitsujin/dxvk/files/13261158/steam-65980.log)

Note that the commit bisected above was in use before Proton 8.
The easiest way to find if it's a dxvk regression is just to shove a old dxvk version in to the new Proton. E.g the one from 7.0-6.

Blisto91:
Added a Windows native trace

PR #4325, opened 2024-10-05
Resource upload rework

doitsujin:
Mostly affects D3D11. The main goal here is to shift responsibility for allocating staging memory away from the context to the initializer so that we can *eventually* move those calls to the CS thread.

This also changes some D3D11 resource update logic in general. Previously, most `UpdateSubresource` calls went through a 4MB per-context staging buffer with a linear allocator because allocation overhead was actually a thing, but when the game is done doing resource updates we'd have an essentially unused 4MB block of memory just sitting there potentially keeping a much larger allocation alive.

To solve this, we reduce the staging buffer size to 256kB and free it after every queue submission. We still want to benefit from the linear allocator for tiny resource updates, but also want to make sure the memory gets freed as soon as possible.

Things still to be done:
- [x] Rework `copyBufferToImage` to do DS format conversion as necessary
- [x] Rework `copyImageToBuffer` to do DS format conversion as necessary
- [x] Use that to get rid of the turbo-ugly `copyPackedBufferToDepthStencilImage` method (and its friend)


Blisto91:
Fixes https://github.com/doitsujin/dxvk/issues/3691

PR #4326, opened 2024-10-05
native: Add SDL3 WSI

### LINKED ISSUE #4323: native: Add SDL3 WSI backend
SDL 3.1.3 was released today, and the ABI has been frozen! Making an SDL3 backend should now be safe - we can copypaste most of SDL2 and just fix the small bits that dxvk depends on (I think only monitor enumeration changed significantly, everything else is pretty much the same).

flibitijibibo:
Took care of all the busy work, all that's left is to actually port from 2 to 3 in the new files. Probably won't be able to get to this for a while, so think of this as a small exercise for anyone that wants to see what changed in SDL3 :)

Fixes #4323

### DIFF
--- a/src/wsi/sdl3/wsi_platform_sdl3_funcs.h
+++ b/src/wsi/sdl3/wsi_platform_sdl3_funcs.h
@@ -1,0 +1,2 @@
+// TODO SDL3
+#undef SDL_PROC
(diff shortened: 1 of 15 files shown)


src/wsi/sdl3/wsi_platform_sdl3_funcs.h
```diff
@@ -0,0 +1,2 @@
+// TODO SDL3
```
flibitijibibo, line 1: Once the build works again, fill this in to remove the linker requirement. See wsi_platform_sdl2_funcs for an example.

PR #4328, opened 2024-10-06
[d3d9] Silence unhandled render state D3DRS_ADAPTIVETESS_Y

Blisto91:
This doesn't fix anything besides getting rid of a warning but @WinterSnowfall said it would be good to add it so it isn't potentially missed whenever TruForm support is worked on.


WinterSnowfall (approved):

WinterSnowfall:
I don't know how you could get it wrong on the first attempt. Someone who had no idea what they were doing must have suggested it :frog: .

WinterSnowfall (approved):

PR #4336, opened 2024-10-08
D3D9: Disable direct buffer mapping by default

K0bin:
In my experience this causes nothing but problems. Lets disable it and see whether people complain about worse perf in some games.


WinterSnowfall:
After our little managedBufferPlacement experiment in d8vk (which was essentially this, but a bit worse), I would really advise against it. We'd end up needing even more workaround for the games that regress because of the new default.

PR #4342, opened 2024-10-10
[d3d8] Stub DebugSetMute to fix C&C:Generals performance

### LINKED ISSUE #4112: [d3d8] C&C: Generals/Zero Hour - menu slowdown
The game drops to ~15 fps when menus that are "heavy" on the UI like options or skirmish are accessed. This doesn't happen with WineD3D. The game is perfectly playable otherwise. Note that the game engine is capped at 31.2 fps.

The game and its addon are also available on Steam now, and are rather popular even these days, so we should look into this.

P.S.: The above problem is unaffected by cachedDynamicBuffers.

### Software information
C&C: Generals (& Zero Hour), retail.

### System information
- GPU: RTX 4070
- Driver: 535.183.01
- Wine version: 9.12 Staging
- DXVK version: fe9c875

### Apitrace file(s)
[generals.trace.tar.xz](https://drive.proton.me/urls/8XC0DHRN6M#YBYdw3mz1spl)

### Log files
- d3d8.log: [generals_d3d8.log](https://github.com/AlpyneDreams/d8vk/files/11440472/generals_d3d8.log)

WinterSnowfall:
This has to be like the dumbest thing ever, but C&C: Generals is spamming calls to DebugSetMute() *every frame* whenever the slowdowns occur. Stubbing it makes it not choke. Fixes #4112.

### DIFF
--- a/src/d3d8/d3d8_main.cpp
+++ b/src/d3d8/d3d8_main.cpp
@@ -51,6 +51,10 @@ extern "C" {
     return S_OK;
   }
 
+  DLLEXPORT void __stdcall DebugSetMute() {
+    dxvk::Logger::debug("D3D8: DebugSetMute: Stub");
+  }
+
   DLLEXPORT IDirect3D8* __stdcall Direct3DCreate8(UINT nSDKVersion) {
     IDirect3D8* pDirect3D = nullptr;
     dxvk::CreateD3D8(&pDirect3D);
--- a/src/d3d8/d3d8.def
+++ b/src/d3d8/d3d8.def
@@ -2,5 +2,5 @@ LIBRARY D3D8.DLL
 EXPORTS
   ValidatePixelShader @ 2
   ValidateVertexShader @ 3
-
+  DebugSetMute @ 4
   Direct3DCreate8 @ 5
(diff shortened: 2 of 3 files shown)


src/d3d8/d3d8_main.cpp
```diff
@@ -51,6 +51,10 @@ extern "C" {
     return S_OK;
   }
 
+  DLLEXPORT void __stdcall DebugSetMute() {
+    dxvk::Logger::debug("D3D8: DebugSetMute: Stub");
```
doitsujin, line 55: Can we make sure this is only logged once, especially if the game is spamming it?

doitsujin (commented):
> This function is internal and not intended to be used by software developers.

Ah yeah, as if that has ever stopped anyone.

src/d3d8/d3d8_main.cpp
WinterSnowfall: Can do, but I forgot how that was being done :sweat_smile: .

netborg-afps:
Well to my understanding it does, as running the game through a simple wrapper on the pre-patch dxvk was solving this issue too, where DebugSetMute() was not defined.

misyltoad:
because that wrapper stubs the function.

netborg-afps:
No, it was my own experimental wrapper only defining Direct3DCreate8.

PR #4347, opened 2024-10-11
[d3d9] Make proper use of X/YHotSpot for software cursors

WinterSnowfall:
Apparently only now it dawned on me that XHotSpot/YHotSpot are not in fact initial cursor positions, rather where the cursor pointer tip should be set within the bitmap coordinates :man_facepalming:. That means all the bitmap clipping & processing I was doing is rather superfluous, and in retrospect the new approach makes a lot more sense.

Anyway, this fixes some snapping on cursor bitmap transitions that I noticed and shouldn't be there. Now everything is properly aligned vs hotspot coordinates, as the application expects it to be.

This also means we frequently end up rendering the bitmap rectangle outside of the screen when the cursor is near screen edges, so that backend viewport-based positioning wasn't a bad idea after all :wink:.

### DIFF
--- a/src/d3d9/d3d9_cursor.cpp
+++ b/src/d3d9/d3d9_cursor.cpp
@@ -41,8 +41,8 @@ namespace dxvk {
     POINT currentPos = { };
     ::GetCursorPos(&currentPos);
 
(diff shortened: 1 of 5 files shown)


Blisto91:
This cursored API

Issue #4343, opened 2024-10-10
[d3d9][regression] El Matador broken shadows


Blisto91:
Shadows in the game can show up as either static, or moving with the camera, lines since commit https://github.com/doitsujin/dxvk/commit/3004026db848c5478a1905c7e76d80952e62da40

Video showing the issue with dxvk https://youtu.be/X0Qv3HWGT9U
<details>
  <summary>Wined3d screenshot</summary>

![image](https://github.com/user-attachments/assets/80331266-2408-4110-bc79-762a59356423)
</details>

### Software information
El Matador

### System information
- GPU: RTX 4080
- Driver: 550.40.71
- Wine version: Proton Bleeding Edge
- DXVK version: https://github.com/doitsujin/dxvk/commit/3004026db848c5478a1905c7e76d80952e62da40 and above

### Apitrace file(s)
https://drive.proton.me/urls/4FN6054GHW#vkiOuPyX0TSk

K0bin:
Looks the same on Windows.

![image](https://github.com/user-attachments/assets/fd8141cc-a1a6-4ad5-8523-10f29c5a5244)

It's Wine that renders it incorrectly. And yes, it does look weird.

Blisto91:
rip in peace

PR #4349, opened 2024-10-11
[d3d9] Only dirty frame buffer on render state changes if render area is impacted

K0bin:
In the past we had bugs because render targets that were disabled using the color write mask limited the render area.
We also had the same issue with a depth stencil view which was unused (no depth test, no stencil test, no depth write).
#2573
Our solution was to only bind RTs & the DSV if it will actually get used in any way.

However this can lead to a bunch of unnecessary render pass spills if the RT/DSV we unbind has the same size as the rest of the others anyway. Those unnecessary render pass spills are what this PR optimizes away.

Investigation of #4343 is what gave me the idea. The PR ends up reducing the number of render passes from 15 to 13 in this particular game (which actually needs us to unbind the disabled RTs). It might of course impact other games even more.

Saving render passes would be especially nice for Turnip or Asahi.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -477,6 +477,9 @@ namespace dxvk {
       m_autoDepthStencil = nullptr;
     } else {
       // Extended devices only reset the bound render targets
+      m_rtLimitsRenderArea = (1 << 4) - 1;
+      m_dsvLimitsRenderArea = true;
+      m_renderArea = { ~0u, ~0u };
       for (uint32_t i = 0; i < caps::MaxSimultaneousRenderTargets; i++) {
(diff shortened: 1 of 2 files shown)


doitsujin:
This seems to introduce quite a lot of extra state tracking that seems brittle and hard to reason about, is there a universe where we can evaluate this purely inside `BindFramebuffer` (and skip the call if we don't need to change anything) without too much overhead?

K0bin:
The backend checks if the new frame buffer would be identical to the old one, so we could get rid of most of the tracking and rely on that.
We'd only pay the cost of `D3D9DeviceEx::BindFramebuffer` on the next draw after changing the color write mask from/to 0.

K0bin:
Alternative proposal up there. ^

PR #4352, opened 2024-10-12
[d3d9] Add ID3D9VkInteropDevice::CreateImage to create custom textures/surfaces

AlpyneDreams:
This extends the D3D9 device interop interface to allow creating surfaces and textures with custom image usage flags. This is useful, for example, if you want to add usage flags like `VK_IMAGE_USAGE_STORAGE_BIT` for writing to D3D9 textures in compute shaders.

### DIFF
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -48,6 +48,9 @@ namespace dxvk {
     bool                IsBackBuffer;
     bool                IsAttachmentOnly;
     bool                IsLockable;
+
+    // Additional paramters for ID3D9VkInteropDevice
+    VkImageUsageFlags   ImageUsage = 0;
   };
 
   struct D3D9ColorView {
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -49,7 +49,7 @@ namespace dxvk {
     bool                IsAttachmentOnly;
     bool                IsLockable;
 
-    // Additional paramters for ID3D9VkInteropDevice
+    // Additional parameters for ID3D9VkInteropDevice
     VkImageUsageFlags   ImageUsage = 0;
   };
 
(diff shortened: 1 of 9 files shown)


src/d3d9/d3d9_surface.h
```diff
@@ -20,8 +20,8 @@ namespace dxvk {
     D3D9Surface(
             D3D9DeviceEx*             pDevice,
       const D3D9_COMMON_TEXTURE_DESC* pDesc,
-            IUnknown*                 pContainer,
-            HANDLE*                   pSharedHandle);
+            IUnknown*                 pContainer    = nullptr,
+            HANDLE*                   pSharedHandle = nullptr);
```
doitsujin, line 24: Not a huge fan of default parameters where  we can avoid them, can we just make a separate constructor instead that just forwards to this one?

src/d3d9/d3d9_texture.h
```diff
@@ -132,7 +132,7 @@ namespace dxvk {
     D3D9Texture2D(
             D3D9DeviceEx*             pDevice,
       const D3D9_COMMON_TEXTURE_DESC* pDesc,
-            HANDLE*                   pSharedHandle);
+            HANDLE*                   pSharedHandle = nullptr);
```
doitsujin, line 135: same here

doitsujin (changes requested):
nit, also one question: Is there a universe where you might need view formats that don't exactly match the D3D9 image format?

src/d3d9/d3d9_common_texture.h
```diff
@@ -48,6 +48,9 @@ namespace dxvk {
     bool                IsBackBuffer;
     bool                IsAttachmentOnly;
     bool                IsLockable;
+
+    // Additional paramters for ID3D9VkInteropDevice
```
PatrickvL, line 52: nit: spelling of parameters

PatrickvL (commented):
Small spelling error in a comment

AlpyneDreams:
> nit, also one question: Is there a universe where you might need view formats that don't exactly match the D3D9 image format?

I can't think of a situation where that'd be needed but I imagine one might exist. Could always add another field to `D3D9VkExtImageDesc` for setting a separate view format later on.

PR #4353, opened 2024-10-12
Refactor barrier tracking

doitsujin:
Cleans up a lot of code and paves the way for future work where we will need to track which subresources of an image have been transitioned from `UNDEFINED` or `PREINITIALIZED`.

This also replaces the old tracking code with a more efficient approach involving a fixed-size hash table backed by red-black trees rather than a linear list, and more importantly, uses resource allocations rather than Vulkan resource handles as a base in order to fix the hash table conflicts introduced with the global buffer rework.

Lots and lots of really tiny commits in order to help locate regressions if there are any. These changes are scary and will require **very** broad testing for rendering issues, as well as validation errors related to barriers or image layouts.


src/d3d11/d3d11_texture.cpp
```diff
@@ -730,33 +730,34 @@ namespace dxvk {
   
   VkImageLayout D3D11CommonTexture::OptimizeLayout(VkImageUsageFlags Usage) {
```
K0bin, line 731: Have you considered moving this function to the backend? It's used in both the d3d9 and d3d11 frontend and only uses the Vulkan usage as parameter.

doitsujin: I have, but for now I think it's more useful to keep it here since we don't *always* want to change the default layout when adding usage flags, and the backend doesn't really know how the front-end intends to actually use the image.

PR #4358, opened 2024-10-13
[d3d9] SIMPLIFIED: Only dirty frame buffer on render state changes if render area is impacted

K0bin:
Simplified version of #4349 that removes some tracking.
It removes some more tracking that has kind of bothered me.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -181,8 +181,6 @@ namespace dxvk {
     m_flags.set(D3D9DeviceFlag::DirtySpecializationEntries);
 
     // Bitfields can't be initialized in header.
-    m_boundRTs = 0;
-    m_anyColorWrites = 0;
     m_activeRTsWhichAreTextures = 0;
     m_alphaSwizzleRTs = 0;
     m_lastHazardsRT = 0;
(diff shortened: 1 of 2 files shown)


doitsujin:
I like this approach more, however I think we need to do two passes over the render target array:
- one to determine the actual render area from only actively written RTV/DSV
- one to filter out non-written RTV/DSV that are *actually* smaller than the render area

Otherwise we'll filter out false positives, especially if an active DSV is what actually limits the render area.

src/d3d9/d3d9_device.cpp
```diff
@@ -6325,40 +6317,96 @@ namespace dxvk {
...
-          m_state.depthStencil->GetDepthStencilView(),
-          m_state.depthStencil->GetDepthStencilLayout(depthWrite, m_activeHazardsDS != 0, m_hazardLayout) };
+      dsvBound = sampleCountMatches && (anyDSStateEnabled || !dsvLimitsRenderArea);
+      if (sampleCountMatches && anyDSStateEnabled && dsvExtent.width < renderArea.width && dsvExtent.height < renderArea.height) {
+        // It's smaller on both axis, so the previous RTs no longer limit the size
+        limitsRenderAreaMask = 0u;
       }
     }
 
+    uint32_t invertedLimitsAreaMask = ~limitsRenderAreaMask & ((1 << m_state.renderTargets.size()) - 1);
```
doitsujin, line 6391: the & seems redundant here since boundMask can only ever have the lowest four bits set anyway

K0bin: You're right.

PR #4359, opened 2024-10-13
[d3d8] Wine device tests fixes (Part 3)

WinterSnowfall:
Based on #4332, trying to fix some more low-hanging Wine tests, after which I'll call it a day probably.

A list of changes:
- fixed error return codes for GetPrivateData and added size validation for SetPrivateData
- In D3D8, a call to BeginStateBlock() must be ended with EndStateBlock(); all other state block related calls will fail while recording via BeginStateBlock()
- added various present parameter validations on device creation and reset
- added D3DERR_INVALIDCALL return code for various calls with D3DFMT_UNKNOWN
- stripped some stray d3d9 caps from the d3d8 caps
- GetInfo/Query error code adjustments
- added various usage and pool validations for index and vertex buffers on creation
- prevent calls to MultiplyTransform() to be captured in state blocks when in D3D8 compatibility mode (yeah, don't ask, but I've checked and indeed seems to behave this way).
- a few other minor fixes and adjustments


WinterSnowfall:
As an indication of progress:

Current master:
-- D3D8: `0020:device: 51187 tests executed (34 marked as todo, 0 as flaky, 835 failures), 1 skipped.`
-- D3D9: `0020:device: 155155 tests executed (73 marked as todo, 0 as flaky, 1816 failures), 6 skipped.`

This PR:
-- D3D8: `0020:device: 51187 tests executed (34 marked as todo, 0 as flaky, 829 failures), 1 skipped.`
-- D3D9: `0020:device: 155156 tests executed (78 marked as todo, 0 as flaky, 1801 failures), 6 skipped.`

WinterSnowfall:
Alright, also done on this one. Seems fine now on initial tests. I'll wait on #4332 until I give it the full game prefix test treatment.

WinterSnowfall:
Tested with 250+ d3d8 and d3d9 titles. Apart from a minor regression hiccup with RC Cars, which uses D3DSWAPEFFECT_COPY_VSYNC with 3 back buffers, all was good.

I've decided to let the game do its thing and not enforce the maximum one buffer limit with D3DSWAPEFFECT_COPY_VSYNC, even though Windows XP and WineD3D both seem to validate it, and yes, the game indeed doesn't work with WineD3D because of it.

src/d3d8/d3d8_device.cpp
```diff
@@ -496,8 +515,13 @@ namespace dxvk {
           UINT                Height,
           D3DFORMAT           Format,
           IDirect3DSurface8** ppSurface) {
+    // Only CreateImageSurface clears the content of ppSurface
+    // before checking if Format is equal to D3DFMT_UNKNOWN.
     InitReturnPtr(ppSurface);
 
+    if (unlikely(Format == D3DFMT_UNKNOWN))
```
K0bin, line 522: What happens on the D3D9 side if the format is UNKNOWN?
Shouldn't we handle this there?

src/d3d9/d3d9_device.h
```diff
@@ -983,13 +983,6 @@ namespace dxvk {
       return m_isD3D8Compatible;
     }
 
-    void SetD3D8CompatibilityMode(bool compatMode) {
```
K0bin, line 986: I like the change to just set that once at the start.

K0bin (commented):
Looks fine to me except for that one question.

src/d3d8/d3d8_device.cpp
WinterSnowfall: D3D9 will always clear the return pointers on all these calls regardless of D3DFMT_UNKNOWN, so it's more of a D3D8 quick.

K0bin: Same thing for all the other instances of this ofc.

src/d3d9/d3d9_device.h
WinterSnowfall: Yeah, not to mention the old way of doing things allowed toggling compatibility state during a device's lifecycle, which makes no sense.

src/d3d8/d3d8_device.cpp
K0bin: Oh right. I'd appreciate a bunch of short comments about this.

WinterSnowfall: Sure, can do. As you noticed during the previous LockRect/LockBox PRs, D3D8 has an interesting way of doing clears that wasn't preserved in D3D9. Not that I noticed anything caring, but Wine does test for them.

K0bin (approved):

PR #4361, opened 2024-10-14
[util] Add Secret World Legends config and move Far Cry 2

Blisto91:
Otherwise d3d9 mode thinks we only have 512MB vram which will lock the higher graphics presets.

~~I opted to put this in the d3d11 section as going by the Far Cry 2 config I assume we sort the games by their highest supported dxvk renderer (d3d12 not included).~~ See comment below

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -982,6 +982,11 @@ namespace dxvk {
     { R"(\\FEAR(MP|XP|XP2)?\.exe$)", {{
       { "d3d9.maxFrameRate",                 "360" },
     }} },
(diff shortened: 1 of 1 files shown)


Blisto91:
> ~~I opted to put this in the d3d11 section as going by the Far Cry 2 config I assume we sort the games by their highest supported dxvk renderer (d3d12 not included).~~

After many decades of R&D, committee meetings and talking to costumers at the world famous international DXVK convention i have decided it makes more sense to keep it in the d3d9 section. Makes sense to only sort by highest API if we set configs for both.
I have also taken the liberty of moving Far Cry 2 to the d3d9 section.

PR #4373, opened 2024-10-16
Remove D3D9 staging buffer marker

doitsujin:
And replace with a simple fence. Functionally the same, but with significantly less code.


K0bin (approved):
Nice cleanup.

PR #4375, opened 2024-10-17
Update dxvk.conf

Oreojandro:


WinterSnowfall:
Feel free to use these values as local defaults for all applications or only specific applications, [as described here](https://github.com/doitsujin/dxvk/wiki/Configuration). Your PR attempts to change global defaults however, which isn't a good idea.

K0bin:
Besides that, when you're making a PR to an open source repo, you need to actually describe what your changes do and why. Same goes for the commit message.

Issue #4118, opened 2024-07-07
[d3d11] Zenless Zone Zero Zooting Ze VRAM


cybik:
Zenless Zone Zero (not a Steam-published game but hey) seems to be suffering from a VRAM leak. Whether it's from upstream or a DXVK quirk, I can't be sure, but Windows players seem to be unaffected, or at least affected far more slowly. Comparatively, there have been numerous reports of both AMD/ATI *and* nVidia users getting game crashes 30 minutes to an hour into a gameplay session. I myself have been suffering these slowdowns, so much so that at the mere HINT of frames being dropped, I close the game and restart it to be able to play another relatively stable bout of 30 minutes.

Before anyone wonders about the issue title, I was trying to go for maximum alliteration. Basically, ZZZ is "oversubscribing" the VRAM and it seems as if it's assuming more memory is available than there is. Could be due to some DX11 APIs related to resource loading that have implementation quirks, i dunno. 

### Software information
*Game:* Zenless Zone Zero
*Type:* ARPG (Gacha, F2P - can be installed without paying)
*Core Engine:* Unity 2019.4.40 (ultra-customized, most likely)
*Settings:* High on everything. (No ultra available)
*Resolution:* 3840x1600 (ultrawide 21:9)

### System information
- GPU: RTX 3080 10gb
- Driver: 555.58.02
- Wine version: GE-Proton 9-10
- DXVK version: `2.3.1-120-gfe9c---`
- Desktop: KDE5 on Wayland

### Apitrace file(s)
- None for now

### Log files
- Proton logfile of a run that had a slowdown at the end: [steam-18233669865720250368.log](https://gist.github.com/cybik/939c506fe1937e8023a3bc422f700123)

doitsujin:
Unity Engine games are kinda known to load all resources into VRAM regardless of the GPU in use, this is generally not a leak. DXVK's memory management also cannot be improved due to some fundamental design flaws.

cybik:
Fair point, but usually DXVK itself doesn't come close to the VRAM limit. For that matter, the game looks like it's averaging at 5.5~gb (for now), with occasional peaks that then go down after a bit (when they're not crashing the system). I'm guessing that everything else running in the background (composition, accelerated desktop, etc) takes precious resources away from DXVK, but still.

Is there an obvious way to limit the total allowed vidmem heap that I'm missing, or is this another one of these "can't be solved unless/until the dev stops being rude to memory" issues (which, per your statement, it certainly might be)?

[...]

PR #4378, opened 2024-10-18
Memory defragmentation

doitsujin:
Builds on all the reworks from the past couple of weeks to implement memory defragmentation.

## Things still to do
- [x] Get rid of the format conversion context in the D3D9 front-end. **Some D3D9 games may break before this is solved.** (#4388)
- [x] Put a limit on the number of resources to relocate at once, processing thousands in one go isn't free and might lead to frame time spikes.
- [x] Only consider chunk allocations rather than the entire heap when deciding whether to defrag at all.
- [x] Fix some useless error messages when we can't relocate a resource
- [x] Work around some Nvidia driver bug which causes all apps to hang as soon as defragmentation happens (#4380)
- [x] Test this a whole bunch

## What this does
As described in #4280, our memory allocator works on chunks of 256MB that are allocated from the system. The goal here is to make more efficient use of these chunks and *actually* return memory back to the system if we have a lot of unused memory sitting around. This is especially important under memory pressure (even more so on Nvidia due to the need of dedicated allocations for e.g. render targets), or if the app in question isn't actually a game but rather a launcher that temporarily eats over 1GB of VRAM but frees most of it when getting minimized.

As an example, Metaphor ReFantazio (the demo version in this case) allocates over 4GB of memory right at the start, just to free most of it right away. Because some small allocations are scattered all across the memory chunks, we still need to keep the full 4.6GB of memory around:
Bildschirmfoto-696

With defragmentation, we get a lot closer to what the game is actively using:
Bildschirmfoto-697

This **only** affects VRAM allocations that are not mapped into CPU address space. For CPU-accessible memory we require pointer stability, so moving those around dynamically is not feasible - that said, most mapped allocations are short-lived anyway so the problem usually solves itself.

The algorithm used here is very simple, we periodically look at the chunk that has the lowest amount of memory used and try to move those resources to existing chunks, while preventing the allocator from reusing that chunk until the memory is *actually needed*. This way, we essentially produce empty chunks which can subsequently be freed. While not optimal in any way, this generally seems to work well in practice.

## What this does not (yet) do
We don't migrate any resources between VRAM and system memory yet. This is planned as a future PR and will likely be necessary in order to make Unity Engine games work better on cards with less than 12GB of VRAM (e.g. #4118).

This also means that if we do currently allocate a resource in system memory, we **will not** move it back to VRAM even if enough space would be available, so performance in those games is still going to be an issue.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -388,6 +388,20 @@
 # dxvk.trackPipelineLifetime = Auto
 
 
+# Controls memory defragmentation
+#
+# By default, DXVK will try to defragment video memory if there is a
+# significant amount of memory wasted, or if the allocation budget of
+# the application is exceeded. This option is provided solely for
+# debug purposes.
+#
+# Supported values:
+# - True: Enable defragmentation
+# - False: Disable defragmentation
+
+# dxvk.enableMemoryDefrag = True
+
+
 # Sets enabled HUD elements
 # 
 # Behaves like the DXVK_HUD environment variable if the
--- a/src/dxvk/dxvk_allocator.cpp
+++ b/src/dxvk/dxvk_allocator.cpp
@@ -32,6 +32,13 @@ namespace dxvk {
     while (index--) {
       PageRange entry = m_freeList[index];
 
+      // The chunk index is the same regardless of alignment.
+      // Skip chunk if it does not accept new allocations.
+      uint32_t chunkIndex = entry.index >> ChunkPageBits;
+
+      if (unlikely(m_chunks[chunkIndex].disabled))
+        continue;
+
       if (likely(!(entry.index & (alignment - 1u)))) {
         // If the current free range is sufficiently aligned, we can use
         // it as-is and simply modify the remaining free list entry.
@@ -42,7 +49,6 @@ namespace dxvk {
 
         insertFreeRange(entry, index);
 
-        uint32_t chunkIndex = pageIndex >> ChunkPageBits;
         m_chunks[chunkIndex].pagesUsed += count;
         return pageIndex;
       } else {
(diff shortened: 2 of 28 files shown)


Mupli:
Is it the same thing that affects Final Fantasy XVI on steam deck (it eats 11 GB RAM and 7 GB VRAM) ? 

Will this solve the problem on AMD cards with less VRAM memory.

Mupli:
@doitsujin Hi, was wonder if you could answer my previous questions. Thx.

Blisto91:
Note that it is weekend and he is currently away.

This feature mostly "only" help in situations where a game might temporarily spike in VRAM usage or if it holds on to some allocations that were previously in use but no longer is. E.g going between zones or menus in a game. Though that might also happen during the initial loading screen.

Current master, which this builds on, already should be better in regards to VRAM used amount versus allocated amount. The benefit differs per game obviously.

mbriar:
@Mupli it will change nothing for FFXVI because that uses D3D12.

Mupli:
@Blisto91  - There is no a weekend for passionate devs :D
@mbriar - got it. Thx. On AMD game tries to fill 12GB VRAM (PS 5 port) into 4-8 GB cards. It felt like this could help. But yeah different tech.

jams3223:
> @Blisto91 - There is no weekend for a passionate devs :D @mbriar - got it. Thx. On AMD game tries to fill 12GB VRAM (PS 5 port) into 4-8 GB cards. I felt like this could help. But yeah different tech.

Have you given SteamOS 3.6 and Proton Experimental a shot? I'm interested to know!

Blisto91:
Please keep the discussions unrelated to this PR to a minimum.

ViNi-Arco:
> our memory allocator works on chunks of 256MB that are allocated from the system.

Mr. Philip, would changing to 192mb bring an improvement, or is it small enough not to be worth it?

doitsujin:
I don't see why reducing the chunk size would help with anything, we still want to avoid having to put large resources into their own vulkan allocations and small chunks increase fragmentation.

256 works well in practice, and for apps that don't use a lot of VRAM, we even start with much smaller chunks now to reduce memory waste, this was a change in 2.4.1 primarily intended to improve the whole game launcher situation.

PR #4380, opened 2024-10-18
Rework GPU synchronization

doitsujin:
Part of #4378, but I'd like to get this extensively tested (and merged) first.

TL;DR Nvidia hangs when submitting a binary semaphore wait without any command buffers attached, even though this *should* be legal and we haven't seen any validation errors. Switching to timeline semaphores works around the issue and makes the whole synchronization thing easier to reason about in general.

We've tried this some two years ago already, but it didn't end up working because timeline semaphores were broken on 32-bit Proton. Things have moved on a little though and I hope that it works out this time, because otherwise we have a big problem.

Another thing that this does is introduce the use of present fences if `VK_EXT_swapchain_maintenance1` is supported. This shouldn't functionally change anything, but silences some sporaic validation errors and is kind of neeed for correctness anyway, it's just scary because anything synchronization-related can cause bugs, hangs, crashes and whatnot.

### DIFF
--- a/src/dxvk/dxvk_cmdlist.cpp
+++ b/src/dxvk/dxvk_cmdlist.cpp
@@ -174,18 +174,6 @@ namespace dxvk {
     const auto& graphicsQueue = m_device->queues().graphics;
     const auto& transferQueue = m_device->queues().transfer;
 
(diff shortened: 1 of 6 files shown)


leegao (commented):
Design doc: https://docs.google.com/document/d/1ggdFKIRHjO2DBonrvjQQMNFCNIcBHwh_NAfJ2i_m5fI/edit?usp=sharing

PR #4384, opened 2024-10-20
[util] Set forceSamplerTypeSpecConstants for Alpha Protocol

Blisto91:
Some objects like metal containers will have unwanted reflection when they shouldn't otherwise.

<details>
  
Without the config
image

With the config. Matches wined3d and native (latter confirmed via YouTube)
image
</details>

There was some talk of maybe just doing this by default so this PR also acts as a reminder issue.


Blisto91:
Randomly discovered to actually maybe be a regression so needs to be looked more at https://github.com/doitsujin/dxvk/issues/4387

doitsujin:
Merging for now since the whole thing needs a bit of a re-think and the old behaviour was probably broken too.

PR #4389, opened 2024-10-22
[d3d9] Enable strict float emulation by default on new Nvidia drivers

K0bin:
Not yet tested.

### DIFF
--- a/src/d3d9/d3d9_options.cpp
+++ b/src/d3d9/d3d9_options.cpp
@@ -95,7 +95,8 @@ namespace dxvk {
       bool hasMulz = adapter != nullptr
                   && (adapter->matchesDriver(VK_DRIVER_ID_MESA_RADV)
                    || adapter->matchesDriver(VK_DRIVER_ID_MESA_NVK)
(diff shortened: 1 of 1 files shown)


doitsujin (approved):

Blisto91:
Since my GPU is in the higher end it is a bit hard to get bound in any of the games we already default to `Strict` where it doesn't also fluctuate, but using the classic test Witcher 2 it is looking fine.

<details>

  `d3d9.floatEmulation = True` with 560.35.03
witcher2true560

  `d3d9.floatEmulation = Strict` with 560.35.03
witcher2strict560

  `d3d9.floatEmulation = Strict` with 565.57.01
witcher2strict565
</details>

Blisto91 (approved):

doitsujin:
Why do we not already have an app profile for TR Legend if it's literally broken right now?

Blisto91:
It was just recently discovered that the issue shows when floatEmulation is set to `Strict` and it haven't been looked at why yet I believe.
I would be careful with just making a config setting it to `True` without some testing as the game seems sensitive to it in general. The reporter says that disabling invariantPosition also fixes it, but that doesn't seem to be the case on Nvidia from my own testing. Game also had a similar issue in another in the past before longMad was enabled.

PR #4398, opened 2024-10-24
[d3d9] Implement several IDirect3DShaderValidator9 validations

### LINKED ISSUE #2045: [D3D9] The Void needs bits of IDirect3DShaderValidator9 implementation.
I've debugged 'The Void' Steam game some time ago (which was crashing at certain locations both with dxvk and Wine's d3d9) and figured that is the first known game which requires some bits of real IDirect3DShaderValidator9 implementation to work (I've put some details and PoC Wine implementation which adds the necessary bits to https://bugs.winehq.org/show_bug.cgi?id=47229).

The game disassembles and assembles back some shaders (I have no idea why). Some shaders fail to assemble back due to: 'dcl_texcoord10_pp v10' (only 10 inputs are supported in ps_3_0 and this is the eleventh) which result in the crash as the game doesn't consider D3DXAssembleShader() failures. With (partially) working shader validator the problematic shaders get rejected at the earlier stage (IIRC shader compilation from HLSL) and the game handles that fine.

The PoC patch for Wine's d3d9 attached to the Wine bug is enough to make game work.

For testing with the game, here is a save file and reproduction instructions: https://github.com/ValveSoftware/Proton/issues/4527. Also, here is the cheat sheet which explains how to open dev console and get to any location at once: https://forum.ice-pick.com/viewtopic.php?f=28&t=9044. Without that I would probably never get to a problematic spot.

Also, WRT game testing, it creates the compiled shader cache and once you have passed the problematic place successfully
the next run won't trigger problematic behaviour (e. g., if you first run on Windows without dxvk and then with dxvk you won't reproduce it until the cache is cleared). So it is better to delete users/Public/Documents/My\ Games/Void/Cache/* before each run.

WinterSnowfall:
Fixes #2045.

(Working) rebase of @misyltoad's former d3d9-shader-val2 branch, which doesn't crash on the main menu with The Void. On top of it I've added a crude adaptation of @gofman's PS input counts validation. It seems to work fine in terms of actually doing what's expected of it ~, but it doesn't currently fix the crashing issues in the game~.

A save file that can reproduce the crashing can be found [here](https://github.com/ValveSoftware/Proton/issues/4527#issuecomment-756287834).

### DIFF
--- a/src/d3d9/d3d9_shader_validator.h
+++ b/src/d3d9/d3d9_shader_validator.h
@@ -2,22 +2,56 @@
 
 #include "d3d9_include.h"
 
+#include "../dxso/dxso_header.h"
+#include "../dxso/dxso_decoder.h"
+
 namespace dxvk {
+
+  enum class D3D9ShaderValidatorMessage : uint32_t {
+    BeginOutOfOrder = 0xeb,
+    InstructionOutOfOrder = 0xec,
+    InstructionEndOfShader = 0xed,
+    InstructionNullArgs = 0xee,
+    BadVersionTokenLength = 0xef,
+    BadVersionTokenType = 0xf0,
+    BadEndToken = 0xf1,
+    EndOutOfOrder = 0xf2,
+    MissingEndToken = 0xf3,
+    BadInputRegisterDeclaration = 0x12c,
+    BadInputRegister = 0x167,
+    BadInstructionLength = 0x21e,
+  };
+
+  enum class D3D9ShaderValidatorState {
+    Begin,
+    ValidatingHeader,
+    ValidatingInstructions,
+    EndOfShader,
+    Error,
+  };
+
+  using D3D9ShaderValidatorCallback = HRESULT(STDMETHODCALLTYPE *)(
+    const char*                      pFile,
+          UINT                       Line,
+          DWORD                      Unknown,
+          D3D9ShaderValidatorMessage MessageID,
+    const char*                      pMessage,
+          void*                      pUserData);
 
   class IDirect3DShaderValidator9 : public IUnknown {
 
   public:
 
     virtual HRESULT STDMETHODCALLTYPE Begin(
-            void* pCallback,
-            void* pUserParam,
-            DWORD Unknown) = 0;
+            D3D9ShaderValidatorCallback pCallback,
+            void*                       pUserParam,
+            DWORD                       Unknown) = 0;
 
     virtual HRESULT STDMETHODCALLTYPE Instruction(
-      const char*  pUnknown1,
-            UINT   Unknown2,
-      const DWORD* pInstruction,
-            DWORD  InstructionLength) = 0;
+      const char*  pFile,
+            UINT   Line,
+      const DWORD* pdwInst,
+            DWORD  cdw) = 0;
 
     virtual HRESULT STDMETHODCALLTYPE End() = 0;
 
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
I've had to comment out:

```
      // + 1 to account for the opcode...
      uint32_t dwordLength = m_ctx->getInstructionContext().instruction.tokenLength + 1;
      if (cdw != dwordLength) {
        return ErrorCallback(pFile, Line, 0x2,
          D3D9ShaderValidatorMessage::BadInstructionLength,
          str::format("Instruction length specified for instruction (", cdw, ") does not match the token count encountered (", dwordLength, "). Aborting validation."));
      }
```

because on many occasions it seems like:
- the dwordLength for `mov, mul, mad` and `dp4` is -1 vs cdw
- the dwordLength for `tex` is -2 vs cdw

That causes the game to crash before getting to the main menu. @misyltoad I hope you have a better memory than I do and maybe have a clue what's going on here?

Using:

`if (cdw < dwordLength)`

also works, but I'm not sure if that's the intended behavior here.

WinterSnowfall:
@Blisto91 *Temba, his arms wide.*

Seems to not crash for me now in the problematic parts. Please also test when you can.

WinterSnowfall:
Left the BadInstructionLength validations commented out and with a TODO, since I'm not sure what's going on there. Perhaps it's something we can improve upon in the future, if any game actually cares.

I'd welcome more testing, but at this stage it seems to be enough for The Void. The risk of breaking other stuff I think is rather minimal, since all other validations deal with shader type/structure validations and IDirect3DShaderValidator9 call order.

WinterSnowfall:
Act of War - Direct Action tries to use register input number 11 in a PS and expects that to work, otherwise it hangs before starting a game. Back to the drawing board, I guess.

WinterSnowfall:
Should be good to go post a review. It's been tested with quite a few games, and at this point I'm fairly confident it shouldn't regress anything. That being said, it's d3d9, so we won't know until it's merged.

src/d3d9/d3d9_shader_validator.h
```diff
@@ -37,32 +71,203 @@ namespace dxvk {
...
+        }
+      }
 
       return D3D_OK;
     }
 
 
     HRESULT STDMETHODCALLTYPE End() {
-      Logger::debug("D3D9ShaderValidator::End: Stub");
+      if (unlikely(m_state == D3D9ShaderValidatorState::Error)) {
```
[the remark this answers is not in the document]
WinterSnowfall: A good question for @misyltoad . I'm afraid I have no idea how to go about testing this. Games do seem to recreate the validator instance for each shader from what I've noticed, but I have no idea if this is intended by design.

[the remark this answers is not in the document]
WinterSnowfall: P.S.: According to [these tests](https://bugs.winehq.org/attachment.cgi?id=69141) it does indeed seem that once a validation error is hit, the validator instances must be released and recreated. Great API, I know.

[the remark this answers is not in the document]
K0bin: So it can be reused if the shader is fine but not if there's issues with it?

src/d3d9/d3d9_shader_validator.h
```diff
@@ -37,32 +71,203 @@ namespace dxvk {
...
+        switch (instContext.instruction.opcode) {
+          case DxsoOpcode::Comment:
+          case DxsoOpcode::Def:
+          case DxsoOpcode::DefB:
+          case DxsoOpcode::DefI:
+            break;
+
+          default:
+            for (uint32_t instNum = 1; pdwInst[instNum] & 0x80000000; instNum++) {
+              DWORD regType  = ((pdwInst[instNum] & D3DSP_REGTYPE_MASK)  >> D3DSP_REGTYPE_SHIFT)
```
K0bin, line 150: Is it possible to also get those from `DxsoInstructionContext`?

src/d3d9/d3d9_shader_validator.h
```diff
@@ -37,32 +71,203 @@ namespace dxvk {
...
+      if (m_isPixelShader && m_majorVersion == 3) {
+        switch (instContext.instruction.opcode) {
+          case DxsoOpcode::Comment:
+          case DxsoOpcode::Def:
+          case DxsoOpcode::DefB:
+          case DxsoOpcode::DefI:
+            break;
+
+          default:
+            for (uint32_t instNum = 1; pdwInst[instNum] & 0x80000000; instNum++) {
```
K0bin, line 149: Please add a comment explaining this. What does the `0x80000000` mask give you here?

src/d3d9/d3d9_shader_validator.h
```diff
@@ -37,32 +71,203 @@ namespace dxvk {
...
+    HRESULT ValidateHeader(const char* pFile, UINT Line, const DWORD* pdwInst, DWORD cdw) {
+      if (unlikely(cdw != 1)) {
+        return ErrorCallback(pFile, Line, 0x6,
+          D3D9ShaderValidatorMessage::BadVersionTokenLength,
+          "IDirect3DShaderValidator9::Instruction: Bad version token. DWORD count > 1 given. Expected DWORD count to be 1 for version token.");
+      }
+
+      DxsoReader reader = { reinterpret_cast<const char*>(pdwInst) };
+      uint32_t headerToken = reader.readu32();
+      uint32_t shaderType  = headerToken & 0xffff0000;
```
K0bin, line 204: I'd prefer it if we could reuse `DxsoHeader` & `DxsoProgramInfo` here unless there's a good reason for why that doesn't work.

You need `DxsoProgramInfo` later anyway so you might as well let the constructor of `DxsoHeader` read out the program type.

src/d3d9/d3d9_shader_validator.h
```diff
@@ -37,32 +71,203 @@ namespace dxvk {
...
+          case DxsoOpcode::DefI:
+            break;
+
+          default:
+            for (uint32_t instNum = 1; pdwInst[instNum] & 0x80000000; instNum++) {
+              DWORD regType  = ((pdwInst[instNum] & D3DSP_REGTYPE_MASK)  >> D3DSP_REGTYPE_SHIFT)
+                             | ((pdwInst[instNum] & D3DSP_REGTYPE_MASK2) >> D3DSP_REGTYPE_SHIFT2);
+              DWORD regIndex = pdwInst[instNum] & D3DSP_REGNUM_MASK;
+
+              if (unlikely(regType == D3DSPR_INPUT && regIndex >= 10)) {
```
K0bin, line 154: `DxsoRegisterType::Input`

K0bin (commented):
Looks mostly alright to me but I don't really like how inconsistent it is about re-using our Dxso compiler types in some places but not in others. I think being a bit more consistent about that and using them every where as much as possible would also make parts of it a bit more readable than using the raw bit masks and bit shifts.

K0bin:
We also need to test it with Sims 2.

src/d3d9/d3d9_shader_validator.h
WinterSnowfall: No other reason than I didn't quite know they existed (tried to reuse our own Dxso stuff as much as possible, but it's a brave new world for me). Will give it a go.

src/d3d9/d3d9_shader_validator.h
WinterSnowfall: Will check.

src/d3d9/d3d9_shader_validator.h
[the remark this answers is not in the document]
WinterSnowfall: > So it can be reused if the shader is fine but not if there's issues with it?

Appears that way, yes. I think I'll end up writing a very targeted test on this behavior by "borrowing" some valid/invalid definitions from the Wine tests.

src/d3d9/d3d9_shader_validator.h
K0bin: You're right:
https://learn.microsoft.com/en-us/windows-hardware/drivers/display/source-parameter-token
https://learn.microsoft.com/en-us/windows-hardware/drivers/display/destination-parameter-token
https://learn.microsoft.com/en-us/windows-hardware/drivers/display/label-token

I still think this warrants a short comment.
Something like: 
> Iterate over register tokens. Bit 31 of register tokens is always 1.

src/d3d9/d3d9_shader_validator.h
K0bin: Might be problematic because `DxsoInstructionContext` gets the register token count from `decodeInstructionLength` and that might be different than iterating over all tokens that have bit 31 set.
And broken shaders are why we're implementing this after all.

WinterSnowfall: Ugh, I see... decodeInstructionLength does not appear to work correctly in some cases in its current implementation, as per the commented section in the above validation that was added by @misyltoad . That's a bit beyond my scope to fix at this point.

K0bin: We *could* replace the for loop by iterating over source registers decoded by `decodeInstruction` and `decodeInstructionLength` but I guess just leave it as it is for now before we make this more complicated than it has to be.

src/d3d9/d3d9_shader_validator.h
WinterSnowfall: Done.

src/d3d9/d3d9_shader_validator.h
WinterSnowfall: Added that very comment, as I think it sums it up nicely.

src/d3d9/d3d9_shader_validator.h
K0bin: As discussed on Discord: `DxsoHeader` throws an exception when it encounters invalid values which ofc isn't great for the validator. So we'll keep this as it is atm.

WinterSnowfall: `DxsoHeader` will thrown an exception for invalid versions, so I'd rather not use it here. As far as `DxsoProgramInfo` goes, we are creating one futher down (and saving it under m_ctx).

K0bin (approved):
Looks fine now to me.

WinterSnowfall:
Will hold on this until:
- @Blisto91 tests Sims 2, which allegedly pass a cdw and could explode with this PR
- I test the native reuse behavior of IDirect3DShaderValidator9 once an exception is hit (though I don't expect any surprises)

Blisto91:
It has passed the Sims 2 and Sims Castaway Stories test

PR #4402, opened 2024-10-25
Fixes for building with clang-cl and wine headers.

rbernon:
A couple of compilation failures found while experimenting with clang msvc mode.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -973,21 +973,21 @@ namespace dxvk {
           float    Z,
           DWORD    Stencil) {
     StateChange();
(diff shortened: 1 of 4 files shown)


rbernon:
Eh I was indeed curious why these casts were not missing in the first place, looks like it'll need to be fixed differently...

WinterSnowfall:
I don't think we include `d3d9types.h` in the d3d9 namespace in d8vk, so there's that. I hate compilers. The definitions for D3DRECT/D3DMATRIX are identical, no casting should be necessary.

rbernon:
For some reason they end up in the namespace when I build with clang-cl, and it definitely fails to build. But I'll look into it.

PR #4411, opened 2024-10-29
[d3d9] Add ID3D9InteropBuffer for vertex buffers, index buffers, surfaces, and volumes

AlpyneDreams:
Exposes everything that has a DxvkBuffer to have its underlying VkBuffer(s) accessed the same way that VkImages can already be accessed for textures. Also extends ID3D9InteropTexture to be returned by volumes in addition to textures and surfaces. Interfacing with vertex buffers, index buffers, and volumes are essential features for interop.

### DIFF
--- a/src/d3d9/d3d9_interop.cpp
+++ b/src/d3d9/d3d9_interop.cpp
@@ -1,6 +1,7 @@
 #include "d3d9_interop.h"
 #include "d3d9_interface.h"
 #include "d3d9_common_texture.h"
+#include "d3d9_common_buffer.h"
 #include "d3d9_device.h"
 #include "d3d9_texture.h"
 #include "d3d9_buffer.h"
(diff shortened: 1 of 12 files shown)


src/dxvk/dxvk_buffer.h
```diff
@@ -182,6 +182,14 @@ namespace dxvk {
...
       return m_info;
     }
+
+    /**
+     * \brief Sharing mode info
+     * \returns Sharing mode info
+     */
+    const DxvkSharingModeInfo& sharingModeInfo() const {
+      return m_sharingMode;
+    }
```
doitsujin, line 192: This should be retrieved from the `DxvkDevice` instead, the only reason the buffer even stores it locally is to avoid pointer chasing when discarding buffers (which is an extremely hot code path in D3D11).

src/d3d9/d3d9_surface.cpp
```diff
@@ -88,6 +88,11 @@ namespace dxvk {
       return S_OK;
     }
 
+    if (riid == __uuidof(ID3D9VkInteropBuffer)) {
+      *ppvObject = ref(m_texture->GetVkBufferInterop());
+      return S_OK;
+    }
```
doitsujin, line 94: Is it necessary to expose implementation details like this? Not that it's likely that we ever move away from textures having buffers, but I don't really want this to end up making any changes at all more difficult.

src/d3d9/d3d9_interop.cpp
```diff
@@ -117,6 +118,86 @@ namespace dxvk {
...
+          break;
+        default:
+          return D3DERR_INVALIDCALL;
+      }
+    } else if (m_texture) {
+      buffer = m_texture->GetBuffer();
+    }
+
+    if (pHandle) {
+      *pHandle = buffer->storage()->getBufferInfo().buffer;
```
doitsujin, line 185: This is not safe, backing storage can change at any time either due to defrag or copy optimizations. This either needs a full sync with the CS thread with a guarantee that the interop client finishes using the buffer before submitting any more DXVK work, or lock the buffer in place by calling `DxvkContext::ensureBufferAddress` *from the CS thread* in order to prevent the backend from relocaing it (see `D3D9DeviceEx::InjectCs`).

If this is ever used to expose mapped buffers that can get discarded by the front-end for any reason, things get even more problematic.

Moreover, this is missing the buffer offset. DXVK only creates one `VkBuffer` per memory allocation these days unless there's a good reason not to (e.g. index buffers on a device that doesn't support `KHR_maintenance5`).

doitsujin (changes requested):
This is extremely sketchy and honestly looks like a bad idea. Why is this needed anyway?

If some sort of data exchange with client apps is actually required, I'd much rather do that as a self-contained API that lets you copy data between a D3D9 buffer and app-managed Vulkan buffers, but as it stands, this is not safe and effectively locks in our locking and memory management implementations, which is unacceptable.

FWIW I think images are technically broken as well right now due to defrag, but that's far less problematic to fix.

src/dxvk/dxvk_buffer.h
AlpyneDreams: Gotcha

src/d3d9/d3d9_surface.cpp
AlpyneDreams: Yeah, I suppose I can't think of a use for this specific hook at the moment.

src/d3d9/d3d9_interop.cpp
AlpyneDreams: > This is not safe, backing storage can change at any time either due to defrag or copy optimizations. This either needs a full sync with the CS thread with a guarantee that the interop client finishes using the buffer before submitting any more DXVK work, or lock the buffer in place by calling `DxvkContext::ensureBufferAddress` _from the CS thread_ in order to prevent the backend from relocaing it (see `D3D9DeviceEx::InjectCs`).
> 
> If this is ever used to expose mapped buffers that can get discarded by the front-end for any reason, things get even more problematic.
> 
> Moreover, this is missing the buffer offset. DXVK only creates one `VkBuffer` per memory allocation these days unless there's a good reason not to (e.g. index buffers on a device that doesn't support `KHR_maintenance5`).

The context I'm using interop in is completely manually managed in terms of thread/sharing safety. DXVK and the graphics queue are flushed and locked while submitting work, and I'm looking into the possibility of submitting work directly to the DXVK command stream as a separate (much smaller) PR. This is a very good point though and I can envision this being a footgun for parallel stuff or for clients that aren't managing their locks. I'll see if I can come up with a better solution.

AlpyneDreams: > Moreover, this is missing the buffer offset. DXVK only creates one `VkBuffer` per memory allocation these days unless there's a good reason not to (e.g. index buffers on a device that doesn't support `KHR_maintenance5`).

Gotcha

AlpyneDreams:
> Is there anyone asking for this functionality?

> This is extremely sketchy and honestly looks like a bad idea. Why is this needed anyway?

The project I am working would benefit immensely from being able to bind (or access bindlessly) vertex and index buffers for compute shaders or to use them to build raytracing acceleration structures etc. Most of this is work that would be done while DXVK is locked and no commands are issued.

> FWIW I think image interop is technically broken as well right now due to defrag, but that's far less problematic to fix.

I have been using it pretty extensively just fine for a couple months now, however I think there are sometimes issues after the device resets or changes mode so you're probably right there.

> If some sort of data exchange with client apps is actually required, I'd much rather do that as a self-contained API that lets you copy data between a D3D9 buffer and app-managed Vulkan buffers, but as it stands, this is not safe and effectively locks in our locking and memory management implementations, which is not really acceptable.

That's a very fair point. Since there's so many issues with this PR, I'll go back to the drawing board and see if there's a more narrow  set of functionality we can expose. Ideally I think it would be preferrable to avoid duplicating all buffers just for the interop client, so I'll see if I can think of  an API that's adaptable to any kind of implementation details while still exposing access to the device-side VkBuffers for just geometry under narrow circumstances.

PR #4413, opened 2024-10-30
D3D11: Improve direct image mapping

doitsujin:
Changes a bunch of things for mapped images:
- `MAP_WRITE_DISCARD` on mapped images actualy discards the image now. This is rare, but avoids GPU syncs when it does happen and makes it more viable to map images directly in the first place, rather than going through a buffer.
- For images with a height and depth of 1, we can actually do exactly that by default now. 1D images are typically small and aren't expected to benefit from `OPTIMAL` tiling, so we can just map those directly without any real downside. Various :jp: games and BG3 actually hit this, with the consequence that we can save some system memory because we don't have to keep tiny staging buffers around
- Fixes some bad decision making w.r.t. memory types of mapped images.
- Fixes various potential issues in code paths that are mostly untested and not really used by games.
- Fixes some wine tests.

I also want to explore whether there's a way to fix the Warhammer 3 without an app profile, and without compromising performance in other games. The issue with that game was that it maps a **massive** dynamic texture multiple times per frame while only writing a few pixels each time, so the buffer path is extremely slow and ends up saturating PCIe bandwidth. As a band-aid, we force `LINEAR` images that we can map directly for that game.

This will need a fair bit of (targeted) testing since messing around with any sort of resource mapping is scary.


doitsujin:
Actually I think the better way to deal with this is to nuke the direct mapping path entirely. It creates a massive maintenance / testing burden, is only used in one game that happens to have a native Linux port anyway, and we can probably solve the perf issues there by moving the mapped buffer to HVV so that we only do a VRAM -> VRAM copy. Sucks on Deck either way, but

PR #4415, opened 2024-10-30
D3D11: Remove direct image mapping

doitsujin:
Supercedes #4413.

There are numerous problems with trying to use`VK_IMAGE_TILING_LINEAR` and exposing that directly to the app:
- It's an extremely poorly tested code path. We only ever enabled it for Warhammer 3 and some feature that nobody uses.
- The performance hit from using the suboptimal tiling mode is very difficult to predict.
- Driver support for linear tiling is extremely inconsistent, which makes testing it a nightmare.
- Games are broken and ignore `D3D11_MAPPED_SUBRESOURCE::RowPitch`, expecting it to be tightly packed anyway.
- There were numerous bugs in our implementation.

This **will** regress Warhammer 3 performance. A possible workaround might be to put the staging buffer in host-visible VRAM, at least on ReBAR systems. Ultimately, the game really just shouldn't be doing what it's doing.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -276,14 +276,6 @@
 # dxvk.zeroMappedMemory = False
 
 
-# Resource size limit for buffer-mapped dynamic images, in kilobytes.
-# A higher threshold may reduce memory usage and PCI-E bandwidth in
-# some games, but may also increase GPU synchronizations. Setting it
-# to -1 disables the feature.
-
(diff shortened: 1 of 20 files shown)


doitsujin:
Turns out this sucks even more. Merged the good commits but it looks like we will need to properly fix up the whole direct image mapping thing after all and test it extensively in a variety of games.

PR #4414, opened 2024-10-30
[d3d9] Order modes descending by refresh rate

WinterSnowfall:
Fixes #4288. Not the shadows, those are fine. The game does however pick the first (lowest ordinal) mode for a particular resolution. 

Since we were ordering them ascending by refresh rates, this has the potential to end up on 24 Hz. The game offers no means of manually selecting refresh rates. Which is why I suspect the reporter of the issue gets stuck on 50 Hz, being on AMD (see below output sample).

Behavior on native drivers is as follows:
- Nvidia:

>     D3DFMT_X8R8G8B8 1920 x 1080 @ 60 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 59 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 50 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 120 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 100 Hz

- AMD:

>     D3DFMT_X8R8G8B8 1920 x 1080 @ 50 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 59 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 60 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 100 Hz
>     D3DFMT_X8R8G8B8 1920 x 1080 @ 120 Hz

- Intel:

>     D3DFMT_R5G6B5 1920 x 1080 @ 60 Hz
>     D3DFMT_R5G6B5 1920 x 1080 @ 48 Hz

- WineD3D:

>     D3DFMT_R5G6B5 1920 x 1080 @ 75 Hz
>     D3DFMT_R5G6B5 1920 x 1080 @ 60 Hz
>     D3DFMT_R5G6B5 1920 x 1080 @ 60 Hz
>     D3DFMT_R5G6B5 1920 x 1080 @ 50 Hz

These were collected on different systems, but you get the idea. Fun fact: Intel doesn't even order by height & width ascending, rather puts the current resolution first, and then "sticks to the plan".

I guess this would need some regression testing, but at some point we'll have to decide if we just want to YOLO it and see if any of the 99.999 d3d9 games out there are affected.

PR #4416, opened 2024-10-30
[d3d9] Ignore modes with refresh rates under 60 Hz

### LINKED ISSUE #4288: Price of Persia 2008
### Software information
Price of Persia 2008 

### System information
- GPU: 3060 12gb
- Driver: latest nvidia
- Wine version: windows 11
- DXVK version: latest 2.4

### Apitrace text:
Original game give me are 1gb video memory and ~60 fps. When I install are dxvk = 1700 mb + 50 stable fps with both of bugs with shaders and smoothness. Please fix it

WinterSnowfall:
Also fixes #4288. Still not the shadows, they're fine.

An alternative to #4414, and IMHO a better way to address the problem the game has, while keeping with ascending ordering, which seems to have been respected on Windows XP at least, is to simply ignore modes with refresh rates under 60 Hz.

I've taken the liberty of adding a config option to revert to the old behavior, for purists, but the filtering gets enabled by default.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -333,6 +333,19 @@
 # d3d9.reproducibleCommandStream = False
 
 
+# Low Refresh Rate Support
+#
+# List modes with refresh rates lower than 60 Hz. Note that this
+# was typically uncommon in the D3D8/D3D9 era, and some games will
+# automatically select the first listed mode for a particular
+# resolution, expecting it to be equal to 60 Hz.
+#
+# Supported values:
+# - True/False
+
+# d3d9.lowRefreshRateSupport = False
+
+
 # Sets number of pipeline compiler threads.
 # 
 # If the graphics pipeline library feature is enabled, the given
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -796,6 +796,7 @@ namespace dxvk {
     uint32_t modeIndex = 0;
 
     const auto forcedRatio = Ratio<DWORD>(options.forceAspectRatio);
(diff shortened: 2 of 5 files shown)


PatrickvL:
50 Hz is the default refresh rate in most of Europe and many other countries. Are you sure disabling that by default is a wise choice?

WinterSnowfall:
Here are some sample outputs on Windows XP, using olden GPUs:

[d3d8-triangle_WinXP_ATI_HD4650_8.65.txt](https://github.com/user-attachments/files/17585106/d3d8-triangle_WinXP_ATI_HD4650_8.65.txt)

[d3d8-triangle_WinXP_Nvidia_GF4_45.23.txt](https://github.com/user-attachments/files/17585107/d3d8-triangle_WinXP_Nvidia_GF4_45.23.txt)

[d3d8-triangle_WinXP_Nvidia_FX5500_93.71.txt](https://github.com/user-attachments/files/17585111/d3d8-triangle_WinXP_Nvidia_FX5500_93.71.txt)

Are you connecting your modern GPU to an analogue CRT for that authentic retro experience? That's nice, but also not typical for the average dxvk use-case. Still there's a config option you can use in case you want to have access to absolutely all the modes exposed by your monitor.

doitsujin:
Not sure if this is such a great idea to begin with, but if we do go for it, this absolutely *needs* a safeguard to report the mode with the highest refresh rate for each resolution, even if said mode is < 60.

WinterSnowfall:
After some debate, it was agreed #4414 is less invasive, so we should go with that and hope nothing regresses. Though if WineD3D is sorting modes descending by refresh rate anyway, I guess it should be fine.

PR #4423, opened 2024-11-02
[util] Re-enable direct buffer mapping for Rayman 3

### LINKED ISSUE #4422: Rayman 3: visual issue (D3D8)
Hello DXVK team,

if I use DXVK for Rayman 3 game, I have a visual issue:
https://youtu.be/l4Ickp4R9wQ
As you can see in this video, this visual issue appearing when I fight an elf or recovering yellow prism.

image

This visual issue could be related to "hardware vertex processing".
Is there a way to ask DXVK to disable "hardware vertex processing"?

Thank you for your help.

My config:
Manjaro 24.1.1 Xahea
Kernel:  x86_64 Linux 6.6.54-2-MANJARO
Proc: AMD Ryzen 9 5900HX
GPU: AMD Radeon RX 6600M
Drivers: Mesa 24.2.4
Resolution: 1920x1080
32 Go RAM
DE: Xfce4

wine log:
[rayman3_wine_dxvk.log](https://github.com/user-attachments/files/17606262/rayman3_wine_dxvk.log)

WinterSnowfall:
Fixes #4422. While disabling direct buffer mapping got rid of some queue syncs caused by buffer locks, which improved the game's quirky frame-rate heuristics, it seems to break some in-game effects. Guess it is what it is.

The games's SWVP mode (hackable through the game's ini) seems to rely on disabling direct buffer mapping to render properly, but since it's not recommended or exposed by the game and HWVP now works properly, that's not that big of a problem.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1150,7 +1150,7 @@ namespace dxvk {
      * Missing geometry and textures without      *
      * legacy DISCARD behavior                    */
     { R"(\\Rayman3\.exe$)", {{
-      { "d3d9.allowDirectBufferMapping",   "False" },
+      { "d3d9.maxFrameRate",                  "60" },
       { "d3d8.forceLegacyDiscard",          "True" },
     }} },
   }};


WinterSnowfall:
Also added a 60 fps lock, since the game appears to break above that, but its internal heuristics seem to rely on the set refresh rate, ergo it will probably still be less than idea on modes above 60 Hz.

K0bin:
IMO we should look into why it causes visual issues rather than just reenabling direct mapping.

WinterSnowfall:
Well, the game writes to buffers outside of Unlock(), so I'm assuming that's messing with a number of things. I would be nice to solve this otherwise, because re-enabling direct mapping has broken SWVP mode (note forcing that is more of a hack, but still).

RibShark:
Yeah, this issue is caused by the game writing to certain locked buffers due to reference miscounting. I can fix it on the game's side, but not sure what the "correct" fix on the wrapper side would look like.

WinterSnowfall:
> Yeah, this issue is caused by the game writing to certain unlocked buffers due to reference miscounting. I can fix it on the game's side, but not sure what the "correct" fix on the wrapper side would look like.

Well that is indeed the problem, since we have to cater for the entire breadth and depth of the d3d8/9 spectrum, and making some games happy without breaking others is sometime nearly impossible. The good news is that the game should be fine now in terms of rendering with its default settings, except for game engine specific jankiness, which is out of our scope to handle, so we have to rely on patches like Better Rayman 3 to address.

> Also worth noting that this happens on Windows without any graphics wrappers (at least on NVIDIA cards) so isn't strictly a dxvk bug.

Yeah, I am aware the game is also broken on some modern Windows systems, however it does miraculously work fine on my Windows XP/GeForce 4 retro PC. So at some point, when all the planets were in alignment, it was indeed fully functional.

PR #4424, opened 2024-11-02
[d3d9] Early return D3D_OK on present with a NULL m_window

### LINKED ISSUE #3894: Support creating D3D9 device without a window
In spite of what the documentation says, it's possible to create a D3D9 device with NULL passed for both `PresentationParams::hDeviceWindow` and `FocusWindow`. This will still create an implicit swapchain and you can query the backbuffers and render to those. You can even call `Present` and that returns `S_OK`.

According to Wine commit logs, Planetary Annihilation: Titans depends on this. This also makes the "visuals" Wine test crash with DXVK.

WinterSnowfall:
Tenatively fixes #3894.

Creating a device with NULL hFocusWindow and hDeviceWindow already works, so does getting the backbuffers. However, Present crashes because  `m_wctx` is uninitialized when m_window is nullptr.

Moving the return a bit earlier makes it work just fine however. Haven't checked if this messes with anything else, but I doubt any of the `m_wctx` stuff is relevant for this use case.

I have yet to check if rendering to the backbuffer works in this situation, but I think it will. Drafting until I actually check on that bit.

### DIFF
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -153,6 +153,9 @@ namespace dxvk {
     if (hDestWindowOverride != nullptr)
       m_window = hDestWindowOverride;
 
+    if (m_window == nullptr)
+      return D3D_OK;
+
     UpdateWindowCtx();
 
     bool recreate = false;
@@ -170,9 +173,6 @@ namespace dxvk {
     m_dirty    |= recreate;
 
     m_lastDialog = m_dialog;
-
-    if (m_window == nullptr)
-      return D3D_OK;
 
 #ifdef _WIN32
     const bool useGDIFallback = m_partialCopy && !HasFrontBuffer();


WinterSnowfall:
Wrote a test that fully works with this PR: https://github.com/WinterSnowfall/d8vk-tests/blob/da4c2a1b9cdecb6b915f60e4aad1587ad06fa8cd/d3d9/d3d9_triangle.cpp#L671 and checks:
- device creation with null hFocusWindow/hDeviceWindow
- getting the back buffer
- drawing to the back buffer
- presenting

As a side-note the back buffer dimensions are the ones set in the present params, even if in this case, so rendering and everything else should work just as it normally would.

WinterSnowfall:
P.S.: Unfortunately can't confirm that this fixes the crash in the d3d9 Wine visual tests, because those have regressed on master with: b4ed1081058be5361784f686aa07b194057dd8a6. Namely it will now hang during `fog_test` following:
`visual.c:2412: Test failed: Untransformed vertex with vertex fog and z = 1.0 has color 00ff0000`

WinterSnowfall:
While the above test passes on WineD3D and dxvk with this PR, WinXP actually fails straight up on device creation. I've asked @Blisto91 to check what happens on modern Windows.

WinterSnowfall:
I'll undraft, because it technically does solve a crash on Present in this scenario. Remains to be seen at a later date if we want to support it at all, but things should be fine in null window territory with this minor fix.

K0bin (approved):

WinterSnowfall:
> I've asked @Blisto91 to check what happens on modern Windows.

And ofc it passes on modern Windows. Guess the Wine tests were right after all.

PR #4439, opened 2024-11-10
[util] Fix Tomb Raider: Legend flicking result from #4389

### LINKED ISSUE #4319: [d3d9] Tomb Raider: Legend has some form of flickering with DXVK, does not occur with wined3d
### Software information

Tomb Raider: Legend, GOG copy, all settings maxed, everything enabled

Ekran Görüntüsü_20241004_000622

### System information
- GPU: 7900 XTX
- Driver: Mesa 24.2.3
- Wine version: Proton Experimental
- DXVK version: Whatever Experimental has and also tested with latest release, occurs with master too

### Apitrace file(s)

- wined3d trace:

https://drive.google.com/file/d/1l1bGAeI62rfjLC91BePooemQTipn6i0H/view?usp=sharing

- dxvk trace:

https://drive.google.com/file/d/1_k7pPDNML1ki_AWsHQKdd7iXhv0TfFJh/view?usp=sharing

### Log files

[steam-heroic-Tomb Raider Legend.log](https://github.com/user-attachments/files/17250195/steam-heroic-Tomb.Raider.Legend.log)

Has this in it but not sure if that is related:

```
warn:  Failed to create global buffer:
warn:    size:  157810696
warn:    usage: 1ff
warn:    type:  0
warn:  D3D9DeviceEx::SetRenderState: Unhandled render state 181
```

### Video that shows the issue

https://youtu.be/PGoJLkdbvvA?si=c6WlcYPici43iT79

qinlili23333:
This should fix #4319 .
But I'm not sure whether it will bring negative effect to TRA and TRU, although it should also fix them due to similar engine implementation.
I don't own TRA or TRU, so need someone own them to test them.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -580,6 +580,7 @@ namespace dxvk {
     { R"(\\(trl|tra|tru)\.exe$)", {{
       { "d3d9.cachedDynamicBuffers",        "True" },
       { "d3d9.maxFrameRate",                "60" },
+      { "d3d9.floatEmulation",              "False" },
     }} },
     /* Everquest                                 */
     { R"(\\eqgame\.exe$)", {{
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -572,12 +572,17 @@ namespace dxvk {
     { R"(\\SpellForce2.*\.exe$)", {{
       { "d3d9.forceSamplerTypeSpecConstants", "True" },
     }} },
-    /* Tomb Raider: Legend, Anniversary, Underworld  *
+    /* Tomb Raider: Anniversary, Underworld          *
      * Read from a buffer created with:              *
      * D3DPOOL_DEFAULT,                              *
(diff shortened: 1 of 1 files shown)


Blisto91:
If we want to do this then Legends should be split out in to its own entry yes. The issue only happens with the "Next Generation Content" option which the other two games don't have.
Also we'll have to test to make sure it doesn't seemingly regress anything else on the big vendors. E.g. AMD and Nvidia

qinlili23333:
I have split Legend to dedicated one.

I have tested on Nvidia RTX 4080 Laptop (565.90, Windows 23H2) and Intel Xe 80eu (101.5768, Windows 24H2), works perfect.

Need someone to test it on AMD (and Intel on Linux) before merge, that's why I put it as a draft.

WinterSnowfall:
Since the former default value was "True", why not use that here?

qinlili23333:
Superseded by #4442

PR #4448, opened 2024-11-12
[d3d9] Do not assume 16-byte alignment in replaceNaN

doitsujin:
Stack alignment is 4 bytes on 32-bit, and we cannot align variables on the stack, so this is technically broken. Use unaligned stores instead even if those are a bit slower.

### DIFF
--- a/src/util/util_vector.h
+++ b/src/util/util_vector.h
@@ -152,11 +152,11 @@ namespace dxvk {
 
   inline Vector4 replaceNaN(Vector4 a) {
     #ifdef DXVK_ARCH_X86
-    alignas(16) Vector4 result;
+    Vector4 result;
     __m128 value = _mm_loadu_ps(a.data);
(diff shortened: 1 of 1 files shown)


AlpyneDreams:
For what it's worth, I don't remember why I added the alignas in #3547 instead of just making it unaligned like in #3515, but I don't think there was a good reason. It already caused unfathomable anguish during the development of d8vk.

misyltoad:
why can variables not be aligned? surely it can just add some and & off the stack ptr?

doitsujin:
> why can variables not be aligned? surely it can just add some and & off the stack ptr?

Ishi explained the issue, something with SEH requiring stack alignment to be done in a specific way but GCC doing it in a different way, so with a Windows target platform, GCC will just not align stack at all and silently generate code that can explode at any time.

hammadnadeemx (approved):

qinlili23333 (approved):

PR #4450, opened 2024-11-12
[util] Hide amd for GTA The Definitive Edition trilogy

Blisto91:
Crashes because of static ags on amd systems which report HDR support.


Blisto91:
Btw an alternative to this would be to just extend the UE4 exe check to include these.
https://github.com/doitsujin/dxvk/blob/8dc1fe1262f1108841c72834d83714bf87dac163/src/dxgi/dxgi_options.cpp#L73

PR #4451, opened 2024-11-13
Update README.md

m1m1k4tz:
Programmatically set registry keys

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ wine reg delete 'HKEY_CURRENT_USER\Software\Wine\DllOverrides' /v "d3d11" /d nat
 wine reg delete 'HKEY_CURRENT_USER\Software\Wine\DllOverrides' /v "dxgi" /d native,builtin /f >/dev/null 2>&1
 ```
 
-#### DLL dependencies 
+### DLL dependencies 
 Listed below are the DLL requirements for using DXVK with any single API.
 
 - d3d8: `d3d8.dll` and `d3d9.dll`
--- a/README.md
+++ b/README.md
@@ -27,7 +27,11 @@ For a pure 32-bit Wine prefix (non default) the 32-bit DLLs instead go to the `s
 ```
 export WINEPREFIX=/path/to/wineprefix
 cp x32/*.dll $WINEPREFIX/drive_c/windows/system32
-winecfg
+wine reg add 'HKEY_CURRENT_USER\Software\Wine\DllOverrides' /v "d3d8" /d native,builtin /f >/dev/null 2>&1
+wine reg add 'HKEY_CURRENT_USER\Software\Wine\DllOverrides' /v "d3d9" /d native,builtin /f >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


Blisto91:
I think this might make things too verbose. At least the uninstall section is quite unpleasant with all the extra lines imo.
You'd also need to update the case of a 32bit prefix.

m1m1k4tz:
Okay, I've added the 32 bit section and also some other fixes.

README.md
```diff
@@ -9,30 +9,76 @@ The most recent development builds can be found [here](https://github.com/doitsu
...
+In order to remove DXVK from a prefix, remove the DLLs and DLL overrides, and run wineboot -u to restore the original DLL files.
+
+In a default Wine prefix that would be as follows:
+```
+export WINEPREFIX=/path/to/wineprefix
+rm $WINEPREFIX/drive_c/windows/system32/d3d8.dll
+rm $WINEPREFIX/drive_c/windows/system32/d3d9.dll
+rm $WINEPREFIX/drive_c/windows/system32/d3d10core.dll
+rm $WINEPREFIX/drive_c/windows/system32/d3d11.dll
+rm $WINEPREFIX/drive_c/windows/system32/dxgi.dll
```
LucasSnatiago, line 51: What do you think about changing this line to something like:
rm -f $WINEPREFIX/drive_c/windows/system32/d3d{8,9,10core,11}.dll
rm -f $WINEPREFIX/drive_c/windows/system32/dxgi.dll

or

rm -f $WINEPREFIX/drive_c/windows/system32/d{3d8,3d9,3d10core,3d11,xgi}.dll

So it will take only 1 or 2 lines intead of 5.

README.md
```diff
@@ -9,30 +9,76 @@ The most recent development builds can be found [here](https://github.com/doitsu
...
+rm $WINEPREFIX/drive_c/windows/syswow64/d3d8.dll
+rm $WINEPREFIX/drive_c/windows/syswow64/d3d9.dll
+rm $WINEPREFIX/drive_c/windows/syswow64/d3d10core.dll
+rm $WINEPREFIX/drive_c/windows/syswow64/d3d11.dll
+rm $WINEPREFIX/drive_c/windows/syswow64/dxgi.dll
```
LucasSnatiago, line 56: If want to make it even more compact, it is possible to do:

rm -f $WINEPREFIX/drive_c/windows/**{system32,syswow64}**/d{3d8,3d9,3d10core,3d11,xgi}.dll

So it will be just one command to remove all files at once.

m1m1k4tz: Okay, thanks I've added those changes in

LucasSnatiago:
@Blisto91 is it ok to use curly braces in scripts like this? I'm new to programming and shell script. I don't know if it works across all the shells available for linux.

m1m1k4tz:
> @Blisto91 is it ok to use curly braces in scripts like this? I'm new to programming and shell script. I don't know if it works across all the shells available for linux.

https://www.gnu.org/software/bash/manual/html_node/Brace-Expansion.html

I think most shells would probably implement this and it should work unless theres a specific bug in the shell.

Issue #4447, opened 2024-11-12
Compiling latest dxvk with -march=native (znver2) causes games to crash


ghost:
compile commands:
```
meson setup --cross-file build-win64.txt --buildtype release \
--prefix /mnt/gitdisk/root/dxvk --libdir lib64 --bindir bin64 build64 \
-Dc_args="-march=native -pipe" -Dc_link_args="-march=native -pipe" \
-Dcpp_args="-march=native -pipe" -Dcpp_link_args="-march=native -pipe" \
-Dstrip=true \
&& ninja -C build64/ install

meson setup --cross-file build-win32.txt --buildtype release \
--prefix /mnt/gitdisk/root/dxvk --libdir lib32 --bindir bin32 build32 \
-Dc_args="-march=native -pipe" -Dc_link_args="-march=native -pipe" \
-Dcpp_args="-march=native -pipe" -Dcpp_link_args="-march=native -pipe" \
-Dstrip=true \
&& ninja -C build32/ install
```

test environment:
I recreated Proton Experimental bleeding edge's folder structure in /Steam/compatibilitytools.d/ and symlinked all files sans dxvk dlls.
Games tested: Arma 3 for 64bit and Company of Heroes 1 for 32bit. Arma 3 makes it to the splash screen then crashes. CoH pops up it's crash report mailer.

-march=x86-64-v3 also fails. -march=x86-64-v2 works with reduced performance. compiling without custom flags works fine. My cpu is AMD zen2 3500x
I skimmed the discussion about how adding newer instruction sets does nothing basicly. So no biggie I suppose.
Hopefully I'm not being an idiot again.

WinterSnowfall:
Sounds like a compiler bug? That being said, why not simply add `-march=native` in dxvk's Meson script instead of `-msse3` and friends? You'd still want the sse math among other things that are already in there.

Also note `-march=native` can be rather problematic in this day and age when (as an example) disabling your E cores may expose instructions specific to your P cores. And indeed, I have tested this myself a while back (worked for me at least), but it brings little if any benefit.

Blisto91:
What error does the games crash with in log?

doitsujin:
Can repro, got a crash with `0x006ffffd515001 d3d11+0xc5001: vmovdqa %ymm0, 0x40(%rsp)`

[...]

PR #4454, opened 2024-11-13
 [build] prefer avx128

ghost:
partial fix for x86_64-v3+ gcc builds
closes https://github.com/doitsujin/dxvk/issues/4447

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -14,12 +14,12 @@ compiler_args = [
   '-msse2',
   '-msse3',
   '-mfpmath=sse',
+  '-mprefer-avx128',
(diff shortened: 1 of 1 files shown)


doitsujin (changes requested):
I don't really want to pass flags by default when they interact with a feature that is disabled by default, and when it's only there to work around a compiler bug anyway (as far as I understand, clang does not need this workaround).

Honestly I'm not sure if this is something we'd want to clobber the build files with **at all** given that users need to manually opt into AVX in order to trigger the issue in the first place.

PR #4458, opened 2024-11-15
Add --64-only and --32-only build options to package-release.sh

60IQ:
Copy paste from package-native.sh. These options are very convenient for tinkering. Thank you.


WinterSnowfall:
This would be somewhat useful in my humble opinion, as I sometimes find myself commenting out the x64 bits in the script anyway to speed up compilation during d3d8/d3d9 development. Having a command line flag for it would be nice.

PR #4460, opened 2024-11-15
[d3d11] Avoid UB declaring the UnmappedSubresource constant

bylaws:
See https://github.com/llvm/llvm-project/issues/59036


doitsujin:
This doesn't really change the underlying issue, we need store the map type as a uint32 instead and do a cast every single time the functionality is used.

PR #4464, opened 2024-11-16
[d3d9] Disable exclusive fullscreen for KiriKiri Z games

qinlili23333:
KiriKiri Z games rely on GDI for top menu bar. If exclusive FS enabled, display or hide menu will result black flicker. So we disable exclusive FS for KiriKiri Z games. Tested on 9-nine- series and Corona Blossom series on Windows 11 and Kali latest with Nvidia GPU.

### DIFF
--- a/src/d3d9/d3d9_options.cpp
+++ b/src/d3d9/d3d9_options.cpp
@@ -28,6 +28,19 @@ namespace dxvk {
     return id;
   }
 
+  static bool isExclusiveFullscreenAllowed() {
+    // KiriKiri Z games rely on GDI for top menu bar
+    // If exclusive FS enabled, display or hide menu will result black flicker
+    // So we disable exclusive FS for KiriKiri Z games
+    //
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
Just how many KiriKiri Z games are there and why can't we simply use executable-name based config overrides, which is like what we almost universally do for everything else?

qinlili23333:
> Just how many KiriKiri Z games are there and why can't we simply use executable-name based config overrides, which is like what we almost universally do for everything else?

Approx 100-200 titles on Steam, while a large amount of titles are not released on Steam. It will be a difficult work to collect file names for all these titles. 
Also, galgames usually have a lot of fan made translated version with different file name, which makes it more difficult to write rules for all of them.

That't why I prefer to use universal fix for krkrz games just like what we do for UE4 games.


While for CatSystem2 games (only approx less than 30 titles and have no explicit dll handle to detect), I approve that rules can be better.

PR #4465, opened 2024-11-16
[util] add d3d9.enableDialogMode for CatSystem2 games already on Steam

qinlili23333:
Like krkrz, CatSystem2 games also need GDI for top menu bar which cannot work well under exclusive FS.
May lack games not on Steam.
CatSystem2 games are not too much, so adding them to config is easier than detect it.


qinlili23333:
> I don't really want to do this. Together with the other thing we're talking like several hundred games that no one has really heard about here, I'd rather just disable FSE by default and have these work without littering the code base with this. Doing that will need some discussion + testing though.

I approve disabling FSE by default should solve the most compatibility issues. Although there may be some performance drop without FSE, d3d9 games usually won't reach GPU bottleneck for most modern GPUs.

PR #4479, opened 2024-11-19
[d3d8] Wine device tests fixes (Part 4)

WinterSnowfall:
I'm aiming to fix a few more d3d8/d3d9 Wine devices tests. 

So far:
- readdressed a relaxation of validations done to prevent crashing in Black Desert in a way that now passes the affected Wine device tests
- fixed invalid storage of VS/PS functions and declarations as well as other d3d8 VS/PS behavior
- added DS dimensions validation on SetRenderTarget (d3d8 specific)
- added PS/VS version validation on shader creation (d3d8)
- partially implemented ValidateVertexShader/ValidatePixelShader, with version validations (d3d8)
- fixed d3d9 multiple image unlock behavior
- validated RT parent device during SetRenderTarget
- added various NormalizeTextureProperties validations

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -1696,6 +1696,17 @@ namespace dxvk {
     if (unlikely(pDeclaration == nullptr || pHandle == nullptr))
       return D3DERR_INVALIDCALL;
 
+    // Validate VS version for non-FF shaders
+    if (pFunction != nullptr) {
+      uint32_t majorVersion = (pFunction[0] >> 8) & 0xff;
+      uint32_t minorVersion = pFunction[0] & 0xff;
+
+      if (unlikely(majorVersion != 1 || minorVersion > 1)) {
+        Logger::err(str::format("D3D8Device::CreateVertexShader: Unsupported VS version ", majorVersion, ".", minorVersion));
+        return D3DERR_INVALIDCALL;
+      }
+    }
+
     D3D8VertexShaderInfo& info = m_vertexShaders.emplace_back();
 
     // Store D3D8 bytecodes in the shader info
@@ -1925,6 +1936,14 @@ namespace dxvk {
 
     if (unlikely(pFunction == nullptr || pHandle == nullptr))
       return D3DERR_INVALIDCALL;
+
+    uint32_t majorVersion = (pFunction[0] >> 8) & 0xff;
+    uint32_t minorVersion = pFunction[0] & 0xff;
+
+    if (unlikely(majorVersion != 1 || minorVersion > 4)) {
+      Logger::err(str::format("D3D8Device::CreatePixelShader: Unsupported PS version ", majorVersion, ".", minorVersion));
+      return D3DERR_INVALIDCALL;
+    }
 
     d3d9::IDirect3DPixelShader9* pPixelShader;
     
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -1683,13 +1683,15 @@ namespace dxvk {
     D3D8VertexShaderInfo& info = m_vertexShaders.emplace_back();
 
     // Store D3D8 bytecodes in the shader info
-    if (pDeclaration != nullptr)
-      for (UINT i = 0; pDeclaration[i+1] != D3DVSD_END(); i++)
-        info.declaration.push_back(pDeclaration[i]);
+    for (UINT i = 0; pDeclaration[i] != D3DVSD_END(); i++)
+      info.declaration.push_back(pDeclaration[i]);
+    info.declaration.push_back(D3DVSD_END());
 
-    if (pFunction != nullptr)
-      for (UINT i = 0; pFunction[i+1] != D3DVS_END(); i++)
+    if (pFunction != nullptr) {
+      for (UINT i = 0; pFunction[i] != D3DVS_END(); i++)
         info.function.push_back(pFunction[i]);
+      info.function.push_back(D3DVS_END());
+    }
     
     D3D9VertexShaderCode result = TranslateVertexShader8(pDeclaration, pFunction, m_d3d8Options);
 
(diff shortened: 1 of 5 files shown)


src/d3d8/d3d8_main.cpp
```diff
@@ -19,14 +19,38 @@ extern "C" {
...
+    const char* errorMessage = "";
+
+    if (unlikely(pPixelShader == nullptr)) {
+      errorMessage = "D3D8: ValidatePixelShader: Null pPixelShader";
+    } else {
+      uint32_t majorVersion = (pPixelShader[0] >> 8) & 0xff;
+      uint32_t minorVersion = pPixelShader[0] & 0xff;
+
+      if (unlikely(majorVersion != 1 || minorVersion > 4)) {
+        errorMessage = dxvk::str::format("D3D8: ValidatePixelShader: Unsupported PS version ",
```
turol, line 31: This looks like UAF. The returned string is freed immediately but you're still its contents.

WinterSnowfall: Nice catch, thanks. Fixed.

WinterSnowfall:
Obligatory indication of progress:

Current master:
-- D3D8: `0020:device: 50761 tests executed (34 marked as todo, 0 as flaky, 696 failures), 1 skipped.`
-- D3D9: `0020:device: 154696 tests executed (78 marked as todo, 0 as flaky, 1695 failures), 6 skipped.`

This PR:
-- D3D8: `0020:device: 50755 tests executed (34 marked as todo, 0 as flaky, 622 failures), 1 skipped.`
-- D3D9: `0020:device: 154681 tests executed (76 marked as todo, 0 as flaky, 1631 failures), 6 skipped.`

src/d3d8/d3d8_device.cpp
```diff
@@ -935,15 +935,31 @@ namespace dxvk {
     // SetDepthStencilSurface is a separate call
     D3D8Surface* zStencil = static_cast<D3D8Surface*>(pNewZStencil);
 
-    if (likely(m_depthStencil != zStencil)) {
-      StateChange();
-      res = GetD3D9()->SetDepthStencilSurface(D3D8Surface::GetD3D9Nullable(zStencil));
+    // Depth stencil dimensions can not be lower than
+    // those of the currently set render target.
+    if (m_renderTarget != nullptr && zStencil != nullptr) {
```
K0bin, line 940: You're no longer checking whether `zStencil` is different from the one thats bound rn.

src/d3d8/d3d8_device.cpp
```diff
@@ -1680,16 +1696,29 @@ namespace dxvk {
...
+
     D3D8VertexShaderInfo& info = m_vertexShaders.emplace_back();
 
     // Store D3D8 bytecodes in the shader info
-    if (pDeclaration != nullptr)
-      for (UINT i = 0; pDeclaration[i+1] != D3DVSD_END(); i++)
-        info.declaration.push_back(pDeclaration[i]);
+    for (UINT i = 0; pDeclaration[i] != D3DVSD_END(); i++)
+      info.declaration.push_back(pDeclaration[i]);
+    info.declaration.push_back(D3DVSD_END());
```
K0bin, line 1715: The old code was straight up broken, right?

`      for (UINT i = 0; pDeclaration[i+1] != D3DVSD_END(); i++)`

This misses the last token **before** the end and the end token. The new code makes sense to me, I'd just like to make sure there isn't some clever intention behind the old code that you accidentally removed here.

src/d3d9/d3d9_device.cpp
```diff
@@ -4060,7 +4070,7 @@ namespace dxvk {
     desc.IsAttachmentOnly   = TRUE;
     desc.IsLockable         = Lockable;
 
-    if (FAILED(D3D9CommonTexture::NormalizeTextureProperties(this, D3DRTYPE_TEXTURE, &desc)))
+    if (FAILED(D3D9CommonTexture::NormalizeTextureProperties(this, D3DRTYPE_SURFACE, &desc)))
```
K0bin, line 4073: Nice find actually. We don't have a lot of resource type based checks in NormalizeTextureProperties, so I don't worry too much that this will cause regressions.

src/d3d9/d3d9_device.cpp
```diff
@@ -4975,9 +4985,13 @@ namespace dxvk {
 
     UINT Subresource = pResource->CalcSubresource(Face, MipLevel);
 
-    // We weren't locked anyway!
-    if (unlikely(!pResource->GetLocked(Subresource)))
-      return D3D_OK;
+    // Don't allow multiple unlockings, except for D3DRTYPE_TEXTURE
+    if (unlikely(!pResource->GetLocked(Subresource))) {
+      if (pResource->GetType() == D3DRTYPE_TEXTURE)
+        return D3D_OK;
```
K0bin, line 4991: D3D9 being weird as usual.

K0bin (approved):
LGTM, just a little question.

src/d3d8/d3d8_device.cpp
WinterSnowfall: Yeah, that's because stuff like changing the render target and setting the very same depth stencil can error out now depending on the result of the dimensions check. That check is done on d3d9 side anyway, so it will return early anyway in most cases.

src/d3d8/d3d8_device.cpp
WinterSnowfall: Yep, was pretty much broken since the entire definition/function needed to be stored, including the end token. Nothing seems to care about this, but luckily there were a few wine tests that helped me spot it.

src/d3d9/d3d9_device.cpp
WinterSnowfall: It didn't affect anything really, since there were no D3DRTYPE_SURFACE type validations in NormalizeTextureProperties before, but there is one now for plain surfaces, so that's how I spotted this bit.

src/d3d8/d3d8_device.cpp
K0bin: Interesting, I guess if any game actually used the declaration from GetVertexVertexShader it would've read way out of bounds.

src/d3d9/d3d9_device.cpp
WinterSnowfall: Pretty much :frog: .

PR #4482, opened 2024-11-23
[util] Set Borderlands 2 floatEmulation to Strict

Blisto91:
Indirect bug report on the Borderlands 2 PCGamingWiki page


K0bin (approved):
Told you earlier that I would've merged it even with just the PCGamingWiki text as a source but now you even went ahead and confirmed it.

Issue #4147, opened 2024-07-15
[d3d9] Scarface: The World Is Yours doesn't even boot with DVXK 2.4


Sagatt:
### Software information
- **Name of the game:** Scarface: The World Is Yours
- **Settings used:** Default

### System information
- **GPU:** NVIDIA GeForce RTX 2060
- **Driver:** Latest NVIDIA driver (please specify version if known)
- **Operating System:** Windows 11
- **Wine version:** (Not applicable for Windows users)
- **DXVK version:** (Not applicable for Windows users)

### Apitrace file(s)
- None

### Log files

**Describe the issue:** Game doesn't even boot.

Blisto91:
Please post dxvk logs and make a apitrace if able.

mirh:
Does the version in the title imply that it was working for you with 2.3.1?

koffiato:
Game gets stuck on black screen on my setup. Works fine with native DirectX 9, apitrace and logs are attached. Will try making the game run windowed as my monitor blacks out for a bit, the game may be changing exclusive fullscreen resolution in a funky way.

### System information
- GPU: AMD Radeon RX 6700 XT
- Driver: 24.6.1
- OS version: Windows 11 23H2
- DXVK version: 1.10.3, 2.3.1 and 2.4.0

### Apitrace file(s)
- [scarface.trace](https://drive.google.com/file/d/1ds-B596fqMRY9GRCS95S3fHbCz1wc7b_/view?usp=sharing)

### Log files
- [scarface_d3d9.log](https://github.com/user-attachments/files/16234887/scarface_d3d9.log)

doitsujin:
If the log is anything to go by it locks up before even creating a swap chain, which is weird. Has this game *ever* worked with any version of DXVK?

Trace replays fine here, so there might be window message shenanigans going on perhaps, but not sure.

Sagatt:
Never did work with any version at all, simply gives me a black screen.

koffiato:
I tried going back to ancient versions with RADV driver, no dice; it never worked with DXVK. WINED3D works fine however.

Blisto91:
Can you also post a apitrace where it doesn't work with dxvk. Might get something from comparing them

K0bin:
The apitrace replays fine on my system as well. The fact that you cannot buy this game anymore, makes looking into this very problematic.

Blisto's idea was pretty good. Please try to make an apitrace with DXVK where the game crashes during recording of the trace.

[...]

Issue #3963, opened 2024-04-21
Any chances add "d3d9.enableDialogMode = True" for dx11?


7AllenWalker7:
Windows 10 x64
Nvidia 552.22
add borderless mode for vulkan on dx11 games.

7AllenWalker7:
In some games, you can't set the borderless mode for vulcan, the settings in game not work.

"Valheim" work on exclusive fullscreen - (X)
"Tribes of Midgard" work on borderless for game settings- (V)

doitsujin:
Not entirely sure what you mean? The problem that the dialogue mode option solves literally does not exist in D3D11.

7AllenWalker7:
This command allows you to run the application in bordeless fullscreen instead of an exclusive full screen on dx9.
In dx11 applications, vulkan is launched in exclusive full screen mode in most cases. If it is possible to launch a vгlлanoin borderless mode, I would like to use it.

vlad54rus:
> Not entirely sure what you mean? The problem that the dialogue mode option solves does not really exist in D3D11.

How about making "VK_EXT_full_screen_exclusive" controllable from a config file? Name it dxvk.AllowExclusiveFullscreen, (Auto/True/False).

Blisto91:
The extension is a Windows only thing fwiw which is not officially a target platform.

7AllenWalker7:
**lextra2**
No. 
Vulkan ignores the game settings and always uses the exclusive full screen in some games (black scren when Alt+Tab), in others it uses the game settings and runs in exclusive and Borderless mode (good games)

Issue #3819, opened 2024-01-27
World of Warcraft: WotLK - Two monitor long blackscreen flash


Vulcum:
I use two monitors and when i click on second monitor the screen with game just act like ALT+TAB. Blackscreen pop up really slow and after 2-3 seconds I see game again. After it when I click to primary screen with game the same scene happen. Blackscreen and after shows game. But is it possible to remove this? Just like normal fullscreen(windowed) settings. I tried older GPU driver and the blackscreen flash was max 1 second in worst. But when I use the latest GPU drivers I have these problems. I "solved" it by self with game settings to only windowed mode and by my support program Dual-Monitor Tool (using for locking my cursor in one screen) i made shortcut to fill screen with open window. So I see Windows bar sadly.

How I said. This blackscreen flash happend all the time but it was under 1 second so I could still manage control my character in-game. 

**So Is it possible to remove the blackscreen flash?**

### Software information
World of Warcraft: Wrath of the Lich King (3.3.5a) - Warmane private server.

### System information
- GPU: MSI GeForce RTX 4080 SUPRIM X 16G
- CPU: AMD Ryzen 9 5900X
- MB: ASUS ROG STRIX X570-E GAMING - AMD X570
- GPU Driver: 551.23 - 01/24/2024
- Primary Monitor: Dell S2721DGFA - LED monitor 27" (DisplayPort) - 165 Hz
- Secondary Monitor: BenQ GW2320 23" - HDMI - 60 Hz
- Windows 11 Pro - version 23H2
- DXVK version: v2.0-async (using config by Milor123 ---> https://gist.github.com/Milor123/351517ec9741786b910f874826943975 )

### Apitrace file(s)
https://gist.github.com/Milor123/351517ec9741786b910f874826943975
Files: https://github.com/Sporif/dxvk-async/releases
Tutorial: https://forum.warmane.com/showthread.php?t=440657

Vulcum:
> This just sounds like Windows is changing resolution or refresh rate or other monitor mode since the exclusive fullscreen app can make Windows do that when it is running. That isn't the fastest on Windows. I assume you have tested latest regular dxvk too as old versions aren't supported here (nor async).
> 
> > So I see Windows bar sadly.
> 
> Can't you solve this by using the `maximized` option below `Windowed Mode`? This will turn it in to a borderless window.

I tried set Windowed Maximized (its default) but still happening. 
Its interesting when i just remove dxvk config all of this stop.
[...]

Issue #2640, opened 2022-05-13
[Windows] Final Fantasy XIV seems to run in exclusive fullscreen even when set to borderless


JaneSmith:
I'm playing Final Fantasy XIV. It uses DirectX 11 and is set to borderless fullscreen mode. There are various third-party programs that provide helpful information for this game in the form of on-screen overlays.

I recently started using DXVK, and was very pleased to see that it has helped resolve some issues I was having with DirectX 11. However, I've also noticed that none of my overlays are working anymore. It is as if the game is running in exclusive fullscreen mode, instead of borderless mode.

I did a search around, and I found the old issue #1397. As far as I can understand from this, setting the game to borderless should work, but it isn't working.

### Software information

Final Fantasy XIV, set to use DirectX 11 and borderless fullscreen mode.

### System information

- GPU:  AMD Radeon RX 5700 XT
- Driver: 21.50.21.07 and/or 30.0.15021.7000
- Wine version: N/A
- DXVK version: 1.10.1

### Apitrace file(s)

- Not sure this is applicable as I'm trying to use this on Windows rather than Linux.

### Log files

- Not sure this is applicable as I'm trying to use this on Windows rather than Linux.

doitsujin:
This is not really a bug, Vulkan doesn't really have the concept of exclusive fullscreen mode the same way that DXGI does, so drivers on Windows tend to "optimize" Vulkan applications into exclusive fullscreen mode because it gives them a slightly faster presentation path.

TL;DR won't/can't fix, especially since this is a complete non-issue on Linux.

vlad54rus:
In the dxvk.conf file try setting dxgi.tearFree = True and dxgi.syncInterval = 0

AFNGP:
Facing the same issue on another MMO called Soulworker. Borderless Windowed runs Exclusive Fullscreen. I have apps showing on top of the game generally usually (i.e. a DPS parser) and Exclusive Fullscreen shows on top of such apps no matter what.

Would be interesting if someone found a workaround to the problem. Will try the one proposed by @vlad54rus once I get home and report back.

JaneSmith:
Thank you for the suggestion. Unfortunately, I tried it and it didn't make any difference. I agree, it would be great if there were a workaround to this. But if it isn't possible, then I understand.

For the record, I am a Linux user, but I'm still stuck on a Windows virtual machine for my gaming as one of my games still uses anti-cheat software that refuses to work on Linux. Thanks for all your hard work on this.

[...]

PR #4485, opened 2024-11-23
Always disable exclusive full-screen on Windows

doitsujin:
So far we're always letting the driver decide whatever it wants to, but since Windows doesn't seem to give drivers any sort of low-level access to anything more modern than D3D9-level presentation APIs we just end up in a situation that sucks for everyone anyway, so migth as well disable it and at least have a slim chance of things like alt+tab actually working.

Should also fix #4465 and related issues, I can't be arsed to add hundreds of app profiles for this garbage.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -500,16 +500,6 @@
 # d3d9.floatEmulation = Auto
 
 
-# Enable dialog box mode
-#
-# Changes the default state of dialog box mode.
-# *Disables* exclusive fullscreen when enabled.
-#
-# Supported values:
-# - True, False: Always enable / disable
-
-# d3d9.enableDialogMode = False
-
 # Overrides the application's MSAA level on the swapchain
 # 
 # Supported values: -1 (application) and 0 to 16 (user override)
--- a/src/d3d11/d3d11_swapchain.cpp
+++ b/src/d3d11/d3d11_swapchain.cpp
@@ -502,7 +502,6 @@ namespace dxvk {
     presenterDesc.imageExtent     = { m_desc.Width, m_desc.Height };
     presenterDesc.imageCount      = PickImageCount(m_desc.BufferCount + 1);
     presenterDesc.numFormats      = PickFormats(m_desc.Format, presenterDesc.formats);
-    presenterDesc.fullScreenExclusive = PickFullscreenMode();
 
     VkResult vr = m_presenter->recreateSwapChain(presenterDesc);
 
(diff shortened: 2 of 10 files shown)


Blisto91:
Touches issue
https://github.com/doitsujin/dxvk/issues/2640
https://github.com/doitsujin/dxvk/issues/3963
And maybe or maybe not
https://github.com/doitsujin/dxvk/issues/3819
https://github.com/doitsujin/dxvk/issues/1694
https://github.com/doitsujin/dxvk/issues/4147

qinlili23333:
Approve for short term until we can find some better way to handle FSE. 

Actually these days I'm also considering about detecting WS_SYSMENU flag and only enable DSE for those without this flag. I tested variety of games including normal AAA titles and AVG with different engines. I found that those games cannot work well under FSE all have WS_SYSMENU flag even under fullscreen while most AAA titles like UE3 titles usually remove this flag when entering fullscreen. I'm not sure whether we can detect this flag and do sth better, so I approve we should disable FSE globally until we have developed sth better.

doitsujin:
I want to avoid fucking around with window flags / messages at all costs. I also just don't want to deal with annoying Windows quirks that simply do not exist on other platforms.

mirh:
> Windows doesn't seem to give drivers any sort of low-level access to anything more modern than D3D9-level presentation APIs

I'm not sure I understand the referent. 
Are we talking about modern presentation as in having vulkan use DXGI rather than GDI, or low-level access as in using something like D3DKMTWaitForVerticalBlankEvent rather than old ass (and/or [finnicky](https://github.com/SpecialKO/SpecialK/commit/a1f63d3aa6dfcdb0b1fdf97bc2a7421057a16350)) WaitForVBlank & friends?

> Sadly it seems WS_SYSMENU is still not a good idea.

How about that *and* if it has any children? 
(notably enough WS_CHILD and WS_POPUP are incompatible, the later being the old way I knew you could set non-d3d FSE)

qinlili23333:
> Are we talking about modern presentation as in having vulkan use DXGI rather than GDI
Sadly, many applications rely on GDI compatibility, and some of them need DXGI in the same time. For a wrapper which prefer compatibility instead of performance, it's unacceptable to use these solutions which may break compatibility. SpecialK is more performance preferred and focus on game only, that's why it can use many aggressive optimization which may break compatibility. (e.g. You can even use DXVK for libcef based applications, but you cannot use SpeciakK)

PR #4491, opened 2024-11-25
[util] Separate Borderlands 2 fixes from Borderlands: The Pre-Sequel

Tiagoquix:
The `d3d9.floatEmulation = Strict` was tested by @Blisto91 and it fixes a problem, but it is only relevant for Borderlands 2, as the test was not performed in The Pre-Sequel nor are there any bugs reported for the Pre-Sequel game regarding the need of Strict floats.

This PR reorganizes Borderlands games fixes and only keeps Strict float emulation for BO2.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -469,17 +469,22 @@ namespace dxvk {
     { R"(\\anarchyonline\.exe$)", {{
       { "d3d9.memoryTrackTest",             "True" },
     }} },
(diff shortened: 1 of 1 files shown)


qinlili23333:
Have you tested it for presequel?
Since currently we have enabled strict for most modern AMD and NVIDIA drivers, you need to test with Intel or manually override config.

I own this game but recently I can only access to a Windows NVIDIA laptop, so my test is not enough.

Blisto91 (approved):

PR #4492, opened 2024-11-25
[d3d9] Several Wine device test fixes

WinterSnowfall:
~WIP to fix several low hanging d3d9-specific Wine device tests.~ Done.

So far:
- fixed SRCCOLOR2 & INVSCRCCOLOR2 caps being exposed with non-9Ex devices/interfaces (allegedly only WARP does this)
- fixed the `test_refcount` Wine test by preventing device child refcount underruns (we were already doing this in d3d8, thanks to Motor City Online)
- apply MultiplyTransform state block behavior universally (it is not d3d8 specific)
- operations on clip planes with an out of bounds index will reflect on the highest valid clip plane
- add surface format validations to `D3D9Surface::GetDC()` (note that valid formats still don't seem to work properly here, which might be something we'd want to revisit at some point)
- fixed a d3d8/d3d9 difference of expected error codes on failed device reset (due to losable resurces)
- fixed d3d9 state block clearing and recording behavior (the validations work similarly to the system we have in d3d8, but have to be implemented differently due to d3d9 state block implementation differences).

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -2025,10 +2025,6 @@ namespace dxvk {
 
     const uint32_t idx = GetTransformIndex(TransformState);
 
-    // D3D8 state blocks ignore capturing calls to MultiplyTransform().
-    if (unlikely(!m_isD3D8Compatible && ShouldRecord()))
-      return m_recorder->MultiplyStateTransform(idx, pMatrix);
-
     m_state.transforms[idx] = m_state.transforms[idx] * ConvertMatrix(pMatrix);
 
     m_flags.set(D3D9DeviceFlag::DirtyFFVertexData);
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -514,7 +514,8 @@ namespace dxvk {
     if (unlikely(m_losableResourceCounter.load() != 0 && !IsExtended() && m_d3d9Options.countLosableResources)) {
       Logger::warn(str::format("Device reset failed because device still has alive losable resources: Device not reset. Remaining resources: ", m_losableResourceCounter.load()));
       m_deviceLostState = D3D9DeviceLostState::NotReset;
-      return D3DERR_DEVICELOST;
+      // D3D8 returns D3DERR_DEVICELOST here, whereas D3D9 returns D3DERR_INVALIDCALL.
+      return m_isD3D8Compatible ? D3DERR_DEVICELOST : D3DERR_INVALIDCALL;
     }
 
     hr = ResetSwapChain(pPresentationParameters, nullptr);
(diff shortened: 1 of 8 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -309,6 +309,14 @@ namespace dxvk {
 
     // When in SWVP mode, 256 matrices can be used for indexed vertex blending
     pCaps->MaxVertexBlendMatrixIndex = m_isSWVP ? 255 : 8;
+    // Only 9Ex devices advertise D3DPBLENDCAPS_SRCCOLOR2 and D3DPBLENDCAPS_INVSRCCOLOR2
+    if (!IsExtended()) {
```
K0bin, line 313: It would be cleaner to just conditionally add those two flags if the adapter is extended in Adapter::GetDeviceCaps rather than conditionally removing them in **both** the device and the interface.

src/d3d9/d3d9_device.cpp
```diff
@@ -522,7 +522,8 @@ namespace dxvk {
     if (unlikely(m_losableResourceCounter.load() != 0 && !IsExtended() && m_d3d9Options.countLosableResources)) {
       Logger::warn(str::format("Device reset failed because device still has alive losable resources: Device not reset. Remaining resources: ", m_losableResourceCounter.load()));
       m_deviceLostState = D3D9DeviceLostState::NotReset;
-      return D3DERR_DEVICELOST;
+      // D3D8 expects a D3DERR_DEVICELOST error code
+      return m_isD3D8Compatible ? D3DERR_DEVICELOST : D3DERR_INVALIDCALL;
```
K0bin, line 526: ... and D3D9 expects INVALIDCALL?

K0bin (changes requested):
Not a huge fan of how you did the D3D9Ex specific caps, aside from that it's fine.

src/d3d9/d3d9_device.cpp
WinterSnowfall: Done. Agreed it's cleaner this way. I was operating under the false impression it would be a pain to check if a device is extended in the adapter (I stopped asking my brain why it keeps doing these things a long time ago).

src/d3d9/d3d9_device.cpp
WinterSnowfall: Reworded it so it's a bit more clear what happens here.

K0bin:
Thanks for changing it. I'm still a bit worried about the ref counting commit

WinterSnowfall:
> Thanks for changing it. I'm still a bit worried about the ref counting commit

Well, as things stand now, if an application releases an object with a 0 ref count we will underrun the refcount on it. That's what the failing Wine test was checking.

FWIW we've had the same logic in d3d8 for a while now, because Motor City Online does just that and will call Release() on some resources several times in succesion until it gets 0 on each call.

PR #4493, opened 2024-11-26
[native] Include all supported wsi backend headers

Blisto91:
So applications actually can support multiple backends with the same .so

### DIFF
--- a/include/native/wsi/native_wsi.h
+++ b/include/native/wsi/native_wsi.h
@@ -2,12 +2,20 @@
 
 #ifdef DXVK_WSI_WIN32
 #error You shouldnt be using this code path.
-#elif DXVK_WSI_SDL3
+#endif
+
+#ifdef DXVK_WSI_SDL3
 #include "wsi/native_sdl3.h"
-#elif DXVK_WSI_SDL2
+#endif
+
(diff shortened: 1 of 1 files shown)


pchome:
1. Are you sure  `#if not defined(DXVK_WSI_SDL3) || DXVK_WSI_SDL2 || DXVK_WSI_GLFW` will work?
  should be `#if !defined(DXVK_WSI_SDL3) && !defined(DXVK_WSI_SDL2) && !defined(DXVK_WSI_GLFW)`
2. Are you sure those headers are not overlap?
  I know almost nothing about SDL, so maybe yes or maybe not (overlap).

Blisto91:
No it was indeed wrong. I drafted it because i had to go away so couldn't fix it immediately

Blisto91:
The second question I'll CC @flibitijibibo for

flibitijibibo:
Can confirm that this would cause conflicts. For native applications that actually plan to support multiple backends, odds are the backends in the application itself would also be split up to avoid the same kind of issues, so each backend can separately include the correct wsi.h without having to worry about collisions.

Blisto91:
Thank you for the input both of you. I'll go ahead and close this.

PR #4495, opened 2024-11-27
[util] Enable deferSurfaceCreation for the Codename Panzer series

WinterSnowfall:
Intros are played back using ddraw, and main menu isn't rendered otherwise.


Blisto91:
The old enableDialogMode option apparently only did anything for the game in Proton. I can confirm that with this it works both with regular Wine and Proton. Nice find

Blisto91 (approved):

PR #4497, opened 2024-11-28
[d3d9] Device import API

AlpyneDreams:
**EDIT**: See below. Instead of creating an async compute queue we instead allow the interop client to provide their own device

<s>Currently, the DXVK interop app I'm working on needs to use the async compute queue. However, DXVK creates the device without any queues from the compute family. This PR simply adds an extra compute queue to the `VkDeviceCreateInfo`.

If there's some reason that we'd want to avoid doing this unless the interop client asks for it, for example if it might cause unexpected performance or behavior changes, then I can add another commit to this PR to make it conditional, but I wouldn't expect this to have any side effects at all since DXVK never uses async compute.</s>

### DIFF
--- a/src/d3d9/d3d9_interfaces.h
+++ b/src/d3d9/d3d9_interfaces.h
@@ -3,6 +3,27 @@
 #include "d3d9_include.h"
 #include "../vulkan/vulkan_loader.h"
 
+using D3D9VkQueueLockCallback = void(bool);
+
+/**
+ * \brief Device import info
+ */
+struct D3D9VkDeviceImportInfo {
+  VkDevice                          device;
+  D3DDEVTYPE                        deviceType;
+  VkQueue                           graphicsQueue;
+  uint32_t                          graphicsQueueFamily;
+  VkQueue                           transferQueue;
+  uint32_t                          transferQueueFamily;
+  VkQueue                           sparseQueue;
+  uint32_t                          sparseQueueFamily;
+  uint32_t                          extensionCount;
+  const char**                      extensionNames;
+  const VkPhysicalDeviceFeatures2*  features;
+  D3D9VkQueueLockCallback*          queueLockCallback;
+};
+
+
 /**
  * \brief D3D9 interface for Vulkan interop
  *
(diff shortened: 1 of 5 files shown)


AlpyneDreams:
> Not gonna lie, I don't like these interop PRs very much. This might not functionally change much, but it also makes zero sense for us to have this upstream since DXVK does not use or require a compute queue, all this does is create a maintenance burden for something that isn't even shipping.
> 
> We don't currently have a good API to import an existing Vulkan device (_some_ functionality is there for importing a vkd3d-proton device, but there's no way to query required extensions / features etc right now), but I think that would be the way forward here.

That's alright, I'm probably going to do most other changes I had planned on a fork from now on since they're too specialized for upstream. I can still try to add a device import API. Something like this?

```cpp
/**
 * \brief Device import info
 */
struct D3D9VkExtDeviceImportInfo {
  VkDevice                          device;
  VkQueue                           graphicsQueue;
  uint32_t                          graphicsQueueFamily;
  VkQueue                           transferQueue;
  uint32_t                          transferQueueFamily;
  VkQueue                           sparseQueue;
  uint32_t                          sparseQueueFamily;
  uint32_t                          extensionCount;
  const char**                      extensionNames;
  const VkPhysicalDeviceFeatures2*  features;
  D3D9VkExtQueueLockCallback*       queueLockCallback;
};

/**
 * \brief D3D9 interface for Vulkan interop
 *
 * Provides access to the instance and physical device
 * handles for the given D3D9 interface and adapter ordinals.
 */
MIDL_INTERFACE("3461a81b-ce41-485b-b6b5-fcf08ba6a6bd")
ID3D9VkInteropInterface : public IUnknown {

          ...

  /**
   * \brief Gets the VkDeviceCreateInfo for a D3D9 adapter
   *
   * \param [in] Adapter Adapter ordinal
   * \param [in] ... Arguments to IDirect3D9Ex::CreateDeviceEx
   * \param [out] ppCreateInfo The Vulkan device create info
   */
  virtual HRESULT GetDeviceCreateInfo(
          UINT                   Adapter,
          D3DDEVTYPE             DeviceType,
          HWND                   hFocusWindow,
          DWORD                  BehaviorFlags,
          D3DPRESENT_PARAMETERS* pPresentationParameters,
          D3DDISPLAYMODEEX*      pFullscreenDisplayMode,
          VkDeviceCreateInfo**   ppCreateInfo) = 0;

  /**
   * \brief Create a D3D9 device for an existing Vulkan device
   * 
   * It is suggested to create the device with the
   * VkDeviceCreateInfo returned by GetDeviceCreateInfo,
   * which will specify everything the device needs for
   * DXVK to function.
   * 
   * \param [in] Adapter Adapter ordinal
   * \param [in] pInfo Info about the created device
   * \param [out] ppReturnedDevice The D3D9 device
   */
  virtual HRESULT ImportDevice(
          UINT                        Adapter,
          D3D9VkExtDeviceImportInfo*  pInfo,
          IDirect3DDevice9Ex**        ppReturnedDevice) = 0;
};
```

AlpyneDreams:
The latest push has the implementation for a device import API. Here is an [example](https://gist.github.com/AlpyneDreams/210dc9156f1a63c278c1fc6a92bebe10) of how it would be used. I have tested this with my project and I can confirm that it works.

Any thoughts or recommended changes on the implementation here? Nits welcome too @misyltoad @doitsujin

AlpyneDreams:
Going to move this to a new PR when I'm done with some improvements to the API.

PR #4505, opened 2024-12-04
[d3d9] Allow interop to query enabled instance extensions

AlpyneDreams:
Just adds a simple function to get a list of enabled instance extensions. This allows the interop user to, for example, check if debug utils are enabled.

### DIFF
--- a/src/d3d9/d3d9_interfaces.h
+++ b/src/d3d9/d3d9_interfaces.h
@@ -28,6 +28,17 @@ ID3D9VkInteropInterface : public IUnknown {
   virtual void STDMETHODCALLTYPE GetPhysicalDeviceHandle(
           UINT                  Adapter,
           VkPhysicalDevice*     pPhysicalDevice) = 0;
+
+  /**
+   * \brief Gets a list of enabled instance extensions
+   * 
+   * \param [out] pExtensionCount Number of extensions
+   * \param [out] ppExtensions List of extension names
+   * \returns D3DERR_MOREDATA if the list was truncated
+   */
+  virtual HRESULT STDMETHODCALLTYPE GetInstanceExtensions(
+          UINT*                       pExtensionCount,
+    const char**                      ppExtensions) = 0;
 };
 
 /**
--- a/src/dxvk/dxvk_extensions.h
+++ b/src/dxvk/dxvk_extensions.h
@@ -343,6 +343,14 @@ namespace dxvk {
     DxvkExt extSurfaceMaintenance1          = { VK_EXT_SURFACE_MAINTENANCE_1_EXTENSION_NAME,            DxvkExtMode::Optional };
     DxvkExt khrGetSurfaceCapabilities2      = { VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME,       DxvkExtMode::Optional };
     DxvkExt khrSurface                      = { VK_KHR_SURFACE_EXTENSION_NAME,                          DxvkExtMode::Required };
+
+    /**
+     * \brief Get list of supported extensions
+     * \returns List of supported extensions
+     */
+    std::initializer_list<const DxvkExt*> getExtensionList() const {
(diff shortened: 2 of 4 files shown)


src/dxvk/dxvk_extensions.h
```diff
@@ -343,6 +343,14 @@ namespace dxvk {
     DxvkExt extSurfaceMaintenance1          = { VK_EXT_SURFACE_MAINTENANCE_1_EXTENSION_NAME,            DxvkExtMode::Optional };
     DxvkExt khrGetSurfaceCapabilities2      = { VK_KHR_GET_SURFACE_CAPABILITIES_2_EXTENSION_NAME,       DxvkExtMode::Optional };
     DxvkExt khrSurface                      = { VK_KHR_SURFACE_EXTENSION_NAME,                          DxvkExtMode::Required };
+
+    /**
+     * \brief Get list of supported extensions
+     * \returns List of supported extensions
+     */
+    std::initializer_list<const DxvkExt*> getExtensionList() const {
```
doitsujin, line 351: This is UB, need to return a proper container here.

Would also prefer this method to be moved to `DxvkInstance` so that it actually and unambiguously returns the enabled extensions.

src/d3d9/d3d9_interfaces.h
```diff
@@ -28,6 +28,17 @@ ID3D9VkInteropInterface : public IUnknown {
...
           VkPhysicalDevice*     pPhysicalDevice) = 0;
+
+  /**
+   * \brief Gets a list of enabled instance extensions
+   * 
+   * \param [out] pExtensionCount Number of extensions
+   * \param [out] ppExtensions List of extension names
+   * \returns D3DERR_MOREDATA if the list was truncated
+   */
+  virtual HRESULT STDMETHODCALLTYPE GetInstanceExtensions(
```
doitsujin, line 39: This needs to be moved to `ID3D9VkInteropInterface1` with its own GUID, otherwise anything using this will explode violently.

doitsujin (changes requested):

src/d3d9/d3d9_interfaces.h
AlpyneDreams: Done. When this merges I'll move the stuff in #4506 to that interface too.

src/dxvk/dxvk_extensions.h
AlpyneDreams: Oops, that was supposed to be a vector. Fixed it. It doesn't return the enabled extensions, just a list of all `DxvtExt` objects in `DxvkInstanceExtensions` that can be queried for their enabled status. The point of putting it in `DxvkInstanceExtensions` was that if someone adds a new extension then they'll know to add it to the list since it's right there in the struct.

doitsujin: We already have similar code for device extensions in `DxvkAdapter`. I don't want this done in two different ways for no reason.

AlpyneDreams: Okay, I'll rewrite it.

AlpyneDreams: With b436477, we now return the extension name list that the instance was created with. This fixes extensions from extension providers not being included and entirely avoids the problem of having to change the code if new extensions are supported.

AlpyneDreams:
@doitsujin Rebased and squashed. Everything should be good to go.

src/d3d9/d3d9_interface.cpp
```diff
@@ -98,6 +98,11 @@ namespace dxvk {
       return S_OK;
     }
 
+    if (riid == __uuidof(ID3D9VkInteropInterface1)) {
```
doitsujin, line 101: nit: Since this returns the same object as `ID3D9VkInteropInterface`, this should be folded into the if statement above (see the `IUnknown`/`IDirect3D9` thing further up, or any other instance of versioned COM interfaces).

AlpyneDreams: Done.

PR #4506, opened 2024-12-04
[d3d9] Add device import interop API

AlpyneDreams:
As requested by @doitsujin in #4497, I have added an API to import existing Vulkan devices. Here is an [example](https://gist.github.com/AlpyneDreams/210dc9156f1a63c278c1fc6a92bebe10) of how it could be used to add an async compute queue when creating the device. First, all device creation info is queried with `GetDeviceCreateInfo`, then the device is created in userspace and passed to `ImportDevice`. I have tested this extensively in a project under active development and can confirm that it works. Feedback or nits are welcomed!

I have personally tested it on AMD and NVIDIA on Linux and NVIDIA on Windows (both MSVC and Mingw).

EDIT: The new API in of July 2025 has three separate calls to get device info: `QueryDeviceExtensions`, `QueryDeviceQueues`, and `QueryDeviceFeatures`. These correspond to calls in `DxvkDeviceCapabilities`. The only changes to the backend are to support importing separate transfer and sparse queues. A separate transfer queue may be preferable for concurrency. A separate sparse queue may be necessary for compatibility with devices that do not support sparse binding on the main graphics queue.

### DIFF
--- a/src/d3d9/d3d9_interfaces.h
+++ b/src/d3d9/d3d9_interfaces.h
@@ -4,6 +4,30 @@
 #include "../vulkan/vulkan_loader.h"
 
 /**
+ * \brief Queue family and handle
+ */
+struct D3D9VkDeviceQueue {
+  uint32_t family = VK_QUEUE_FAMILY_IGNORED;
+  VkQueue  queue  = VK_NULL_HANDLE;
+};
+
+/**
+ * \brief D3D9 VkDevice import info
+ */
+struct D3D9VkDeviceImportInfo {
+    VkDevice                          device          = VK_NULL_HANDLE;
+    D3D9VkDeviceQueue                 graphicsQueue   = {};
+    uint32_t                          extensionCount  = 0u;
+    const char**                      extensionNames  = nullptr;
+    const VkPhysicalDeviceFeatures2*  features        = nullptr;
+    D3DDEVTYPE                        deviceType      = D3DDEVTYPE_HAL;
+    HWND                              focusWindow     = NULL;
+    DWORD                             behaviorFlags   = 0u;
+    D3DPRESENT_PARAMETERS*            presentParams   = nullptr;
+    D3DDISPLAYMODEEX*                 fullscreenMode  = nullptr;
+};
+
+/**
  * \brief D3D9 interface for Vulkan interop
  *
  * Provides access to the instance and physical device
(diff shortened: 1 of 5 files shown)


WinterSnowfall:
@AlpyneDreams Not sure if you saw it, but there's a warning that's probably worth fixing:

```
../../../src/d3d9/d3d9_interop.cpp: In member function ‘virtual HRESULT dxvk::D3D9VkInteropInterface::GetDeviceCreateInfo(UINT, D3D9VkDeviceCreateInfo**)’:
Warning: ../../../src/d3d9/d3d9_interop.cpp:92:23: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const size_t’ {aka ‘const unsigned int’} [-Wsign-compare]
   92 |     for (int i = 0; i < queueCount; i++) {
      |                     ~~^~~~~~~~~~~~
```

AlpyneDreams:
@WinterSnowfall Thanks, fixed!

AlpyneDreams:
Rebased.

AlpyneDreams:
Rebased to 2.5.2

doitsujin:
Apologies for the silence, will look at this again after the next release, we had some regressions to take care of.

AlpyneDreams:
Fixed an issue where `VkDeviceMemoryOverallocationCreateInfoAMD` was stored locally instead of with everything else in the `pNext` chain that was causing problems on AMD.

AlpyneDreams:
Fixed an issue caused by yesterday's fix :)

AlpyneDreams:
Rebased to 2.5.3 and squashed some fixes.

doitsujin:
Been a while, but I am going to start working on improving device / instance creation now, at least to the point where required features / extensions etc can be queried from DXVK and then be fed back into it.

AlpyneDreams:
@doitsujin 
> Been a while, but I am going to start working on improving device / instance creation now, at least to the point where required features / extensions etc can be queried from DXVK and then be fed back into it.

Okay, I've updated the PR to use those new query functions, and I also removed a lot of stupid cruft I left in the initial version of the PR. I'm going to switch this to draft until I have time to do more testing, though.

It seems like `importDevice` still doesn't support specifying a transfer or sparse queue. Should I extend it to support those as optional parameters like I did in the initial version of the PR?

AlpyneDreams:
I have updated the [example code](https://gist.github.com/AlpyneDreams/210dc9156f1a63c278c1fc6a92bebe10) to show how the current device import API is used to add extra queues in a production environment

AlpyneDreams:
Rebased to 2.7

AlpyneDreams:
Added commits to support separate transfer and sparse queues. A separate transfer queue may be preferable for concurrency. A separate sparse queue may be necessary for compatibility with devices that do not support sparse binding on the main graphics queue. This makes minor tweaks to the backend, should not affect existing functionality.

AlpyneDreams:
I have tested the latest additions in production and can confirm it works, it creates a separate transfer queue and uses the main queue for sparse. I've updated the example with new changes required to do this.

AlpyneDreams:
Closed as I don't currently have time to work on interop. This PR is complete and there's no reason it couldn't be revived in future, but anyone intending to use interop should probably figure out a bespoke solution until best practices can be established.

PR #4508, opened 2024-12-04
[d3d9] Spec-constant out writes to clip distances when disabled

misyltoad:
Add a new spec constant with a mask of the enabled clip planes such that they can be optimized out to improve performance.

For GPL shaders, override what we return here so it's always true and don't bother putting the mask in the UBO.

### DIFF
--- a/src/dxso/dxso_compiler.cpp
+++ b/src/dxso/dxso_compiler.cpp
@@ -3482,6 +3482,7 @@ void DxsoCompiler::emitControlFlowGenericLoop(
     
     uint32_t floatType = m_module.defFloatType(32);
     uint32_t vec4Type  = m_module.defVectorType(floatType, 4);
+    uint32_t boolType  = m_module.defBoolType();
     
     // Declare uniform buffer containing clip planes
     uint32_t clipPlaneArray  = m_module.defArrayTypeUnique(vec4Type, clipPlaneCountId);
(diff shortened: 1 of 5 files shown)


src/dxso/dxso_compiler.cpp
```diff
@@ -3552,9 +3552,9 @@ void DxsoCompiler::emitControlFlowGenericLoop(
 
       DxsoRegisterValue dist = emitDot(position, plane);
 
-      // Always consider clip planes enabled when doing GPL by forcing a mask of 0xffffffff for the quick value.
-      uint32_t clipPlaneEnabledBit = m_spec.get(m_module, m_specUbo, SpecClipPlaneMask, i, 1, m_module.constu32(0xffffffff));
-      uint32_t clipPlaneEnabled = m_module.opINotEqual(boolType, clipPlaneEnabledBit, m_module.constu32(0));
+      // Always consider clip planes enabled when doing GPL by forcing 6 the quick value.
+      uint32_t clipPlaneCount = m_spec.get(m_module, m_specUbo, SpecClipPlaneCount, 0, 32, m_module.constu32(6));
```
DadSchoorse, line 3556: Probably good to move this outside the loop for shorter spirv.

misyltoad: Done.

src/dxso/dxso_compiler.cpp
```diff
@@ -3536,6 +3537,9 @@ void DxsoCompiler::emitControlFlowGenericLoop(
     DxsoRegisterValue position;
     position.type = { DxsoScalarType::Float32, 4 };
     position.id = m_module.opLoad(vec4Type, positionPtr);
+
+    // Always consider clip planes enabled when doing GPL by forcing 6 the quick value.
```
DadSchoorse, line 3541:
```suggestion
    // Always consider clip planes enabled when doing GPL by forcing 6 for the quick value.
```

src/dxso/dxso_compiler.cpp
```diff
@@ -3536,6 +3537,9 @@ void DxsoCompiler::emitControlFlowGenericLoop(
     DxsoRegisterValue position;
     position.type = { DxsoScalarType::Float32, 4 };
     position.id = m_module.opLoad(vec4Type, positionPtr);
+
+    // Always consider clip planes enabled when doing GPL by forcing 6 for the quick value.
+    uint32_t clipPlaneCount = m_spec.get(m_module, m_specUbo, SpecClipPlaneCount, 0, 32, m_module.constu32(6));
```
DadSchoorse, line 3542:
```suggestion
    uint32_t clipPlaneCount = m_spec.get(m_module, m_specUbo, SpecClipPlaneCount, 0, 32, m_module.constu32(caps::MaxClipPlanes));
```

K0bin (approved):

PR #4513, opened 2024-12-08
D3D9: Handle binding mismatching texture types

### LINKED ISSUE #4387: [d3d9][regression?] Alpha Protocol weird reflections
Since commit https://github.com/doitsujin/dxvk/commit/25076d9220d37fbac589a83193a571d965d7f81a the two metal containers in the corner of the first room will have a weird reflection that moves with the camera, it looks like the outside of a building maybe. Prior to this they didn't have any reflections which match wined3d and native.

`d3d9.forceSamplerTypeSpecConstants = True` seemingly works around the issue. 
Also if i change window mode via either the game options or alt + enter the issue also disappears until i load the map again.

<details>

native Nvidia
image

dxvk
image
</details>

### Software information
[Alpha Protocol](https://store.steampowered.com/agecheck/app/34010/)
Using the default settings.

### System information
- GPU: RTX 4080
- Driver: 560.35.03
- Wine version: Proton Bleeding Edge / Windows 11
- DXVK version: https://github.com/doitsujin/dxvk/commit/25076d9220d37fbac589a83193a571d965d7f81a and above

### Apitrace file(s)
TBD

Edit: added
https://drive.proton.me/urls/XB50DYT3JR#MiZEjE0tSOf1

K0bin:
Fixes #4387

Alpha Protocol binds a cube texture to a slot which is declared as a 2d texture in the shader. Binding a mismatching texture type leads yields (0,0,0,0) (yes, alpha 0, unlike binding NULL) on native D3D9. See the table inside the issue.

I also checked that this is checked at draw time, not at the time of the SetTexture call. So binding a shader with a correct texture type afterwards results in correct rendering and the other way around as well. 

Because if there's one thing the D3D9 frontend needed, it is more bit field tracking... :)

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3372,9 +3372,16 @@ namespace dxvk {
 
       BindShader<DxsoProgramTypes::VertexShader>(GetCommonShader(shader));
       m_vsShaderMasks = newShader->GetShaderMask();
+
+      UpdateTextureTypeMismatchesForShader(newShader, m_vsShaderMasks.samplerMask, caps::MaxTexturesPS + 1);
     }
-    else
+    else {
       m_vsShaderMasks = D3D9ShaderMasks();
+
+      // Fixed function vertex shaders don't support sampling textures.
+      m_dirtyTextures = m_vsShaderMasks.samplerMask & m_mismatchingTextureTypes;
+      m_mismatchingTextureTypes &= ~m_vsShaderMasks.samplerMask;
+    }
 
     m_flags.set(D3D9DeviceFlag::DirtyInputLayout);
 
@@ -3730,6 +3737,8 @@ namespace dxvk {
 
       BindShader<DxsoProgramTypes::PixelShader>(newShader);
       newShaderMasks = newShader->GetShaderMask();
+
+      UpdateTextureTypeMismatchesForShader(newShader, newShaderMasks.samplerMask, 0);
     }
     else {
       // TODO: What fixed function textures are in use?
@@ -3737,6 +3746,10 @@ namespace dxvk {
 
       // The RT output is always 0 for fixed function.
       newShaderMasks = FixedFunctionMask;
+
+      // Fixed function always uses spec constants to decide the texture type.
+      m_dirtyTextures = newShaderMasks.samplerMask & m_mismatchingTextureTypes;
+      m_mismatchingTextureTypes &= ~newShaderMasks.samplerMask;
     }
 
     // If we have any RTs we would have bound to the the FB
@@ -4341,12 +4354,12 @@ namespace dxvk {
       const uint32_t textureType = newTexture != nullptr
         ? uint32_t(newTexture->GetType() - D3DRTYPE_TEXTURE)
         : 0;
-      // There are 4 texture types, so we need 2 bits.
+      // There are 3 texture types, so we need 2 bits.
       const uint32_t offset = StateSampler * 2;
       const uint32_t textureBitMask = 0b11u       << offset;
       const uint32_t textureBits    = textureType << offset;
 
-      // In fixed function shaders and SM < 3 we put the type mask
+      // In fixed function shaders and SM < 2 we put the type mask
       // into a spec constant to select the used sampler type.
       m_textureTypes &= ~textureBitMask;
       m_textureTypes |=  textureBits;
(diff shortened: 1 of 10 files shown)


K0bin:
Fun game bug in Alpha Protocol:

> warn:  Mismatching view type. Slot: 2. Expected: VK_IMAGE_VIEW_TYPE_2D. Got: VK_IMAGE_VIEW_TYPE_CUBE
warn:  Mismatching view type. Slot: 2. Expected: VK_IMAGE_VIEW_TYPE_CUBE. Got: VK_IMAGE_VIEW_TYPE_2D

K0bin:
I also tried it with Star Wars Force Unleashed 2 with forceSamplerTypeSpecConst disabled and didn't see any issues. But at the same time I don't remember how difficult it was to hit the bug that made me enable that option for the game in the first place so idk, I think I'll just keep the option for it for now.

src/d3d9/d3d9_device.cpp
```diff
@@ -6297,6 +6313,64 @@ namespace dxvk {
   }
 
 
+  void D3D9DeviceEx::UpdateTextureTypeMismatchesForShader(const D3D9CommonShader* shader, uint32_t shaderSamplerMask, uint32_t shaderSamplerOffset) {
+    if (unlikely(shader->GetInfo().majorVersion() < 2 || m_d3d9Options.forceSamplerTypeSpecConstants)) {
```
WinterSnowfall, line 6317: Why are you unlikely-ing the entire d3d8 frontend? :stuck_out_tongue: I'd just not hint this bit at all.

K0bin: Because old games are less unlikely to be demanding, so I'd rather optimize for the newer ones, especially when we're talking about tiny things like the `likely/unlikely` hints.

WinterSnowfall: There are plenty of SM1 d3d9 games out there too, but fine, have it your way.

Blisto91:
Tested a few games without issue

src/d3d9/d3d9_device.cpp
```diff
@@ -6322,6 +6338,64 @@ namespace dxvk {
...
+    }
+
+    if (unlikely(shader == nullptr || shader->GetInfo().majorVersion() < 2)) {
+      return;
+    }
+
+    const D3D9CommonTexture* tex = GetCommonTexture(m_state.textures[stateSampler]);
+    VkImageViewType boundViewType  = D3D9CommonTexture::GetImageViewTypeFromResourceType(tex->GetType(), D3D9CommonTexture::AllLayers);
+    VkImageViewType shaderViewType = shader->GetImageViewType(shaderTextureIndex);
+    bool shaderUsesTexture = shaderViewType != VK_IMAGE_VIEW_TYPE_1D; // D3D9 does not have 1D textures. The value of VIEW_TYPE_1D is 0 which is the default when there is no declaration for the type.
```
doitsujin, line 6391: can we write this as `VkImageViewType(0)` then?

src/dxso/dxso_compiler.cpp
```diff
@@ -762,6 +763,10 @@ namespace dxvk {
         // We could also be depth compared!
         DclSampler(idx, binding, samplerType, true, implicit);
       }
+
+      const uint32_t offset = idx * 2;
+      uint32_t textureBits = uint32_t(viewType - VK_IMAGE_VIEW_TYPE_1D);
```
doitsujin, line 768: subtracting `VK_IMAGE_VIEW_TYPE_1D` (i.e. 0) here seems pointless? We're just encoding the enum as-is anyway.

src/d3d9/d3d9_shader.h
```diff
@@ -54,11 +54,18 @@ namespace dxvk {
 
     uint32_t GetMaxDefinedConstant() const { return m_maxDefinedConst; }
 
+    VkImageViewType GetImageViewType(uint32_t samplerSlot) const {
+      const uint32_t offset = samplerSlot * 2;
+      const uint32_t mask = 0b11;
+      return static_cast<VkImageViewType>(((m_textureTypes >> offset) & mask) + VK_IMAGE_VIEW_TYPE_1D);
```
doitsujin, line 60: kinda same here

doitsujin (commented):
nits

K0bin:
Done.

PR #4517, opened 2024-12-09
[d3d9] Validate depth and stencil clears

WinterSnowfall:
Yet another validation we were missing. Tested both d3d8 and d3d9, they both behave the same here.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1840,6 +1840,12 @@ namespace dxvk {
 
     D3D9DeviceLock lock = LockDevice();
 
+    // D3DCLEAR_ZBUFFER and D3DCLEAR_STENCIL are invalid flags
+    // if there is no currently bound DS (which can be the autoDS)
+    if (unlikely(m_state.depthStencil == nullptr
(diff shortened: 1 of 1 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -1840,6 +1840,12 @@ namespace dxvk {
 
     D3D9DeviceLock lock = LockDevice();
 
+    // D3DCLEAR_ZBUFFER and D3DCLEAR_STENCIL are invalid flags
+    // if there is no currently bound DS (which can be the autoDS)
+    if (unlikely(m_state.depthStencil == nullptr
+            && ((Flags & D3DCLEAR_ZBUFFER) || (Flags & D3DCLEAR_STENCIL))))
```
doitsujin, line 1846: Looks weird when this can be written more concisely as `Flags & (D3DCLEAR_ZBUFFER | D3DCLEAR_STENCIL)`?

doitsujin (commented):
nit

src/d3d9/d3d9_device.cpp
WinterSnowfall: Makes sense, of course. I have to stop writing code when I'm half asleep.

PR #4538, opened 2024-12-21
[d3d9] Refactor D3D9ShaderValidator

### LINKED ISSUE #4537: DXVK 2.5.2 causes FPS drop due to shader issues
FPS go to 5 due to shader issue
I tested builds from actions and I found that this become happening after merging #4398

### System information
- GPU: NVIDIA GeForce GTX 1650
- Driver: 32.0.15.6636
- DXVK version: 2.5.2

### Apitrace file(s)
[logs + videos](https://drive.google.com/drive/folders/1hrbNvCZVigGu9whdRWkPLE4WTW5pmcvl?usp=sharing)

WinterSnowfall:
I had hoped the input count validation would work globally, but apparently it regresses that particular game.

There's no nicer way to handle this unfortunately, since we're dealing with entirely undocumented behavior, so I've hidden the offending validation behind a config option and enabled it only for The Void, the only known game that actually needs it.

Also part of this PR, I've refactored the code to sit in its own .cpp file (as I should have done from the start really).

### DIFF
--- a/src/d3d9/d3d9_shader_validator.cpp
+++ b/src/d3d9/d3d9_shader_validator.cpp
@@ -1,0 +1,205 @@
+#include "d3d9_shader_validator.h"
+
+namespace dxvk {
+
+  HRESULT STDMETHODCALLTYPE D3D9ShaderValidator::QueryInterface(REFIID riid, void** ppvObject) {
+    if (ppvObject == nullptr)
+      return E_POINTER;
+
+    *ppvObject = ref(this);
+    return S_OK;
+  }
+
+  HRESULT STDMETHODCALLTYPE D3D9ShaderValidator::Begin(
+      D3D9ShaderValidatorCallback pCallback,
+      void*                       pUserData,
+      DWORD                       Unknown) {
+    if (unlikely(m_state != D3D9ShaderValidatorState::Begin)) {
+      return ErrorCallback(nullptr, -1, 0,
+        D3D9ShaderValidatorMessage::BeginOutOfOrder,
+        "IDirect3DShaderValidator9::Begin called out of order. ::End must be called first.");
+    }
+
+    m_callback = pCallback;
+    m_userData = pUserData;
+    m_state    = D3D9ShaderValidatorState::ValidatingHeader;
+
+    return D3D_OK;
+  }
+
+
+  HRESULT STDMETHODCALLTYPE D3D9ShaderValidator::Instruction(
+      const char*  pFile,
+            UINT   Line,
+      const DWORD* pdwInst,
+            DWORD  cdw) {
+    if (unlikely(pdwInst == nullptr || !cdw)) {
+      return ErrorCallback(pFile, Line, 0,
+        D3D9ShaderValidatorMessage::InstructionNullArgs,
+        "IDirect3DShaderValidator9::Instruction called with NULL == pdwInst or 0 == cdw.");
+    }
+
+    if (unlikely(m_state == D3D9ShaderValidatorState::Begin)) {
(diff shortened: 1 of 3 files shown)


doitsujin:
What exactly does the validation in question do, what do the shaders in The Void look like that cause it to fail, and what is the problem with the other game exactly?

WinterSnowfall:
> What exactly does the validation in question do, what do the shaders in The Void look like that cause it to fail, and why does it break the other game?

The validation in question fails if a PS 3.0 shader uses more than 10 (higher than index 9) input registers. This is (apparently) what The Void relies on to work correctly.

As to why the game on #4537 takes issue with it, no idea, as it's no longer available for purchase, but since the validation is pretty specific, I'm sure it uses a higher than 9 index in a PS 3.0 shader for some reason. As of now it's the only other game (with the exception of The Void of course) that we know of to do that, but in its case it *doesn't* expect the validation to fail.

TLDR, this entire thing is a mess even by d3d9 standards, but since in practice only The Void needs it, I think the PR isn't too far of an ask. I lost count how many config options we have specifically for Sims 2, so we might as well have a very dubious one for The Void, at least until (if ever) we get an exact idea what the indented behavior is here, if it differs based on OS versions etc.

doitsujin:
My concern here is that we're essentially making irreversible changes since it's going to be impossible to even *try* to understand what's going on. I'd prefer to ask the user to dump shaders (with a build that has the workaround), so we can at least try and see where we fail.

WinterSnowfall:
> My concern here is that we're essentially making irreversible changes since it's going to be impossible to even _try_ to understand what's going on. 

Not sure how anything is irreversible at this point? The offending validation is already in effect on master, this PR only *restricts* it to The Void.

> I'd prefer to ask the user to dump shaders (with a build that has the workaround), so we can at least try and see where we fail.

I can do that, but as I said the validation is very specific and only fails if an input register index above 9 is used with a PS 3.0. I won't fail on anything else.

I can try to write a test for it, but not sure what good that will do except maybe to confirm how native behaves. Based on the fact that The Void and the affected game in the issue expect different things, I doubt it's going to be consistent anyway.

doitsujin:
> I can do that, but as I said the validation is very specific and only fails if an input register index above 9 is used with a PS 3.0. It won't fail on anything else.

I would like to at least see what the offending shader *is* and make sure we're not missing something silly (built-in input or something like that?). If it happens to be a scenario that should actually fail but doesn't for reasons we don't know then we can still do an app profile, but I'd just like to at least investigate it first.

WinterSnowfall:
@Blisto91 has asked the reporter to dump the shaders, but the annoying thing is the offending game is a) obscure, b) of the MMO variety, c) no longer available for purchase.

The alternative is to simply no care about it, but I am concerned it just *might* affect other things as well long term. And I'd rather not do that just to fix The Void, which was really the intention here.

WinterSnowfall:
Dumped shaders are here: https://github.com/doitsujin/dxvk/issues/4537#issuecomment-2571759696

There seems to be quite a lot of PS 3.0 in there, and some are using up to register index 14 as input, if I'm reading it correctly... which should really fail validation based on our assumptions, but for some reason that game expects them not to?

Since we've not seen/hit this in many other games we've tested that actually validate PS 3.0, I'm not sure what's going on to be honest. I've probably hit a "goes above my head" wall at this point, so would appreciate your input.

It might be the validation calls have a slient failure mode of sorts, where they return the right error strings but pass validation? Kind of hard to tell when there's no actual documentation on any of this, and I went with what misyltoad was able to divine as ground truth.

PR #4543, opened 2024-12-22
[d3d8/9] Clear pLockedRect/Box contents universally on lock

WinterSnowfall:
Based on a test that I wrote and ran on Windows XP it looks like:
- d3d9 behaves similarly to d3d8 during LockRect/Box
- ~LockRect doesn't discriminate based on RTYPE and textures also get their pLockedRect contents cleared~

So the Wine tests were partially incorrect for d3d8 and incomplete for d3d9 (the Wine test in question was d3d8 specific, in all fairness).

I'm waiting for @Blisto91 to confirm modern native behavior, and in the meantime I will do some general testing. Not that I expect anything to blow up due to this minor nit, but you never know.

Edit: All the above is confirmed on modern AMD and Intel Windows drivers.

### DIFF
--- a/src/d3d9/d3d9_surface.cpp
+++ b/src/d3d9/d3d9_surface.cpp
@@ -124,17 +124,19 @@ namespace dxvk {
       return D3DERR_INVALIDCALL;
 
     D3DBOX box;
+    auto& desc = *(m_texture->Desc());
     D3DRESOURCETYPE type = m_texture->GetType();
 
-    if (m_texture->Device()->IsD3D8Compatible() && type != D3DRTYPE_TEXTURE) {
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
As tradition dictates, it appears the Wine d3d8 wine test was ultimately correct, and D3DRTYPE_TEXTURE surfaces don't get cleared... **outside of D3DPOOL_DEFAULT**. This API...

The behavior is however consistent between d3d8 and d3d9, that much still holds.

WinterSnowfall:
*Sigh*... in the true spirit of d3d9, turns out there is a difference on LockRect. Anyway, it's all good now.

PR #4566, opened 2024-12-30
Limit frame rate for AH3LM to 60 fps

spiffeeroo:
https://store.steampowered.com/app/370460

Arcana Heart 3 Love Max!!!!! runs too fast when frame rate is above 60 fps (like on high refresh rate monitor). Limit frame rate to 60 fps so game runs at correct speed.


waspennator:
Could this fix also be applied for the XTEND version too,  i tested it out again on my OLED Deck to check and it ran at 90 FPS out of the box, fresh launch with no save file unless I add "DXVK_FRAME_RATE=60 %command%" to cap it.

download

spiffeeroo:
Updated to include Xtend version too.

Blisto91:
I haven't been able to reproduce any issues with the builtin frame limiter in the Xtend version, but if we are going to apply this to them all it's better to merge them. No reason to have the exe's split in to separate configs here.

PR #4576, opened 2025-01-03
[spirv] Emit the grad and const offset image ops in the correct order

esullivan-nvidia:
Currently the grad and const offset image operand ids are emitted in the incorrect order. This causes incorrect code gen if both the grad and const offset image operands are used.

https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#Image_Operands

This fixes the compilation error found when running TopSpin 2k25 through DXVK using the NVIDIA proprietary Vulkan driver.

https://github.com/ValveSoftware/Proton/issues/7684

### DIFF
--- a/src/spirv/spirv_module.cpp
+++ b/src/spirv/spirv_module.cpp
@@ -3892,15 +3892,15 @@ namespace dxvk {
       
       if (op.flags & spv::ImageOperandsLodMask)
         m_code.putWord(op.sLod);
-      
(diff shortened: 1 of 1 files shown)


adamdmoss (approved):
LGTM according to my reading of the spec (and code)...

lahm911:
Does this game Topspin work on Winlator

K0bin:
We don't officially support Android and the DXVK repo is not the right place to ask whether a game works.

PR #4577, opened 2025-01-03
[d3d9] Skip texture type check when forceSpecConst option is active

K0bin:
Fixes https://github.com/ValveSoftware/Proton/issues/1277


K0bin:
That must have gotten lost in one of the rebases of the original PR, I'm actually sure I did the same change before.

PR #4587, opened 2025-01-06
Add more logging to D3D9 shader validation

doitsujin:
If something fails, print the entire instruction as dwords.

### DIFF
--- a/src/d3d9/d3d9_shader_validator.cpp
+++ b/src/d3d9/d3d9_shader_validator.cpp
@@ -93,12 +93,11 @@ namespace dxvk {
             DWORD regIndex = pdwInst[instNum] & D3DSP_REGNUM_MASK;
 
             if (unlikely(regType == static_cast<DWORD>(DxsoRegisterType::Input) && regIndex >= 10)) {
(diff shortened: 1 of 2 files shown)


src/d3d9/d3d9_shader_validator.cpp
```diff
@@ -188,10 +187,20 @@ namespace dxvk {
...
+            DWORD                      InstrLength,
             D3D9ShaderValidatorMessage MessageID,
-      const std::string&               Message) {
+            std::string                Message) {
+    if (pInstr && InstrLength) {
+      for (uint32_t i = 0; i < InstrLength; i++) {
+        Message += str::format(i ? "," : " [",
+          std::hex, std::setfill('0'), std::setw(8), pInstr[i],
+          i + 1 == InstrLength ? "]" : "");
+      }
```
WinterSnowfall, line 199: Rather than appending the message, it's better to just use a separate string for a debug logger here, since the callback error message will be outputted as warning.

WinterSnowfall (changes requested):

PR #4588, opened 2025-01-06
Added ability to override VK_LOADER_LAYERS_DISABLE env var.

aleiby:
When testing against an old Unity app on Windows, I discovered someone was setting VK_LOADER_LAYERS_DISABLE=~implicit~ which was preventing me from using RenderDoc to compare DX11 vs Vulkan output.  Since dxvk is right in the middle of all of this, it provides an opportune place to fix that.

### DIFF
--- a/src/util/util_env.cpp
+++ b/src/util/util_env.cpp
@@ -35,6 +35,13 @@ namespace dxvk::env {
 #endif
   }
 
+  void setEnvVar(const char* name, const char* value) {
+#ifdef _WIN32
+      ::SetEnvironmentVariableA(name, value);
+#else
+      std::setenv(name, value);
+#endif
+  }
 
   size_t matchFileExtension(const std::string& name, const char* ext) {
     auto pos = name.find_last_of('.');
--- a/src/dxvk/dxvk_instance.cpp
+++ b/src/dxvk/dxvk_instance.cpp
@@ -31,6 +31,19 @@ namespace dxvk {
 
     m_options = DxvkOptions(m_config);
 
+    std::string vkLoaderLayersDisableEnv = env::getEnvVar("VK_LOADER_LAYERS_DISABLE");
+    if (!vkLoaderLayersDisableEnv.empty())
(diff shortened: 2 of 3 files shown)


src/util/util_env.cpp
```diff
@@ -35,6 +35,13 @@ namespace dxvk::env {
 #endif
   }
 
+  void setEnvVar(const char* name, const char* value) {
+#ifdef _WIN32
+      ::SetEnvironmentVariableA(name, value);
+#else
+      std::setenv(name, value);
```
doitsujin, line 42: This doesn't exist in the std namespace and has an extra parameter.

I'd be fine with printing a warning and not supporting this in native builds at all for the time being; apps using dxvk-native shouldn't be messing around with the Vulkan environment in the first place and we can still add it later if it does turn out to be needed.

Otherwise, we'd need to explicitly include posix headers here and also make sure it behaves the same as the Windows path (mostly relevant for handling the `value == nullptr` case, but also unicode considerations).

doitsujin (changes requested):
I'm fine with this change overall since it's a debug band-aid, just needs a fix for native builds.

pchome:
> `VK_LOADER_LAYERS_DISABLE`
> This is also checked before other layer environment variables (such as `VK_LOADER_LAYERS_ENABLE`) so that a user may easily disable all layers and then selectively re-enable individual layers using the enable environment variable.

> `export VK_LOADER_LAYERS_DISABLE=*MESA*,~implicit~`
> `set VK_LOADER_LAYERS_DISABLE=*MESA*,~implicit~`
> The above would disable any Mesa layer and all other implicit layers that would normally be enabled on the system. 

> `export VK_LOADER_LAYERS_ENABLE=*validation,*recon*`
> `set VK_LOADER_LAYERS_ENABLE=*validation,*recon*`
> The above would enable the Khronos validation layer and the GfxReconstruct layer, if both were present on the system and already visible to the loader. 

So, I guess if `VK_LOADER_LAYERS_ENABLE` wasn't reset by app then naming implicit layer there should work.
Or through deprecated `VK_INSTANCE_LAYERS`. No need to hack this in dxvk.

PR #4589, opened 2025-01-06
Set private data debug object name

aleiby:
Set debug names for buffers, fences, textures, and views to match the app's DX11 set values.

### DIFF
--- a/src/d3d11/d3d11_buffer.h
+++ b/src/d3d11/d3d11_buffer.h
@@ -181,7 +181,12 @@ namespace dxvk {
             D3D11_BUFFER_DESC*      pBufferDesc);
 
   private:
-    
+
+    virtual void SetD3DDebugObjectName(const char* name) {
+      if (m_buffer)
+        m_buffer->setDebugName(name);
+    }
+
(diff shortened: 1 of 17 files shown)


doitsujin:
This approach doesn't really work as there is no 1:1 relation between D3D11 and Vulkan objects. Quite the opposite actually, the backing storage for *any* resource can change at any point in time for various reasons (memory defrag, `D3D11_MAP_WRITE_DISCARD`, swapchain present, ...), so accessing Vulkan resources like this is not only unsafe but also doesn't work as soon as a resource gets relocated.

Unfortunately this is going to take quite a bit of work to get right. I have a [branch](https://github.com/doitsujin/dxvk/tree/debug-utils-stuff) lying around with some basic debug improvements, but put it on hold because it didn't seem overly useful at the time. I can revisit this in the near future though and add support for D3D debug names where it makes sense.

doitsujin:
Merged the debug rework, as mentioned it does require setting `DXVK_DEBUG=markers` to work since this is neither transparent to the app, nor does DXVK behave exactly the way it does without it enabled as mentioned in the other PR.

PR #4590, opened 2025-01-07
[d3d9] Use cdw to iterate over register tokens in D3D9ShaderValidator

### LINKED ISSUE #4537: DXVK 2.5.2 causes FPS drop due to shader issues

WinterSnowfall:
Tentatively fixes #4537, by making sure we don't read tokens out of bounds. The Void still works fine with the change (ergo the validation still fails, as it's supposed to).

### DIFF
--- a/src/d3d9/d3d9_shader_validator.cpp
+++ b/src/d3d9/d3d9_shader_validator.cpp
@@ -87,7 +87,7 @@ namespace dxvk {
 
         default:
           // Iterate over register tokens. Bit 31 of register tokens is always 1.
-          for (uint32_t instNum = 1; pdwInst[instNum] & 0x80000000; instNum++) {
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
[Confirmed to fix the above issue](https://github.com/doitsujin/dxvk/issues/4537#issuecomment-2575765783). Since it's a more restrictive change the chances of it blowing up anything (at least vs current master) are minimal, so should be good to go.

PR #4596, opened 2025-01-09
[d3d8] Various quality-of-life nitpicks

WinterSnowfall:
Long overdue quality-of-life d3d8-only nitpicks (or a bit of housekeeping, if you will), which should be pretty harmless overall:
- 48ab9c6 changes all d3d8 source files to use LF (Unix) endline delimiters, as it currently is a spotty mix of LF and CRLF
- 4ce90de ensures we don't leave any state block data uninitialized, so that we don't risk applying garbage with empty state blocks (and also refactors things a bit, such as using the dxvk namespace properly)
- 0d9c13c splits the D3D8SwapChain header into its own cpp file (no functional changes otherwise)
- b6aeaeb splits the D3D8Texture2D/3D/Cube header into its own cpp file (no functional changes otherwise)
- 62ea36f splits the D3D8Volume header into its own cpp file (no functional changes otherwise)
- af47004 moves some definitions from the D3D8Surface header to the cpp file (no functional changes otherwise)
- ed7b018 splits the D3D8Vertex/IndexBuffer header into its own cpp file (no functional changes otherwise)
- 69fb979 wraps up with fixing some minor formatting and alignment nits


WinterSnowfall:
Done a round of sanity checking with ~50 d3d8 games and it's fine, as one would expect.

Note that I don't think it's worth too close of an inspection, as it's mostly 2000 lines of nothing. That being said the d3d8 codebase should now be somewhat more maintainable/presentable.

Should anything blow up (unlikely as it is), I'll address it - everything here is entirely contained to d3d8.

Issue #4583, opened 2025-01-06
Very choppy framerate


quake1982:
DXVK v2.5.2 causes very unstable and choppy framerate.
Games: C&C Red Alert 2
             StarCraft 2
 Also, from version 2.5, DXVK no longer works on C&C Tiberian Sun.

CPU: AMD Ryzen 9 3950X 16-Core Processor 3.49 GHz
GPU: AMD Radeon VII

mbriar:
> DXVK no longer works on C&C Tiberian Sun

That game uses ddraw6 by default, and would have never worked with DXVK, unless you use some kind of mod?

quake1982:
> > DXVK no longer works on C&C Tiberian Sun
> 
> That game uses ddraw6 by default, and would have never worked with DXVK, unless you use some kind of mod?

I'm using CNC TS Client 7.01, and I switch the render to the CNC-DDRAW.

WinterSnowfall:
> Games: C&C Red Alert 2

C&C Red Alert 2 is also a ddraw game. I'm assuming you're using cnc-ddraw's d3d9 renderer for that as well? Note that cnc-ddraw has an opengl backend and you *should* be using that directly for better performance anyway.

quake1982:
Choppy FPS fixed, by switching on the Radeon Enhanced Sync of AMD video cards in the driver.

Blisto91:
@quake1982 Does this also work fine with 2.5.3 if you make a file called `dxvk.conf` next to the exe and insert `dxvk.allowFse = True` ?

quake1982:
> [@quake1982](https://github.com/quake1982) Does this also work fine with 2.5.3 if you make a file called `dxvk.conf` next to the exe and insert `dxvk.allowFse = True` ?

Thank you for the reply, the latest version works fine for the game that I mentioned. But, I'll try the way you provided. Version 2.5.2 with choppy fps was quite strange on my PC, the major issue was because of the driver of my AMD card. The Radeon 7 is a pretty old video card comparing the RX 6000 series.

Blisto91:
@quake1982 Friendly ping

Issue #4572, opened 2025-01-01
Need for Speed Hot Pursuit 2 (2002) broken rendering and worse fps since 2.5.2


De-M-oN:
I use d3d8to9 https://github.com/crosire/d3d8to9/releases
otherwise there is no fog in the game.
And the dxvk d3d9.dll

2.5.1 is a bless. The game runs 3-times better fps than dgvoodoo2 DX11 (DX12 a bit faster but crashes at 4k)
Thats some insane magic by dxvk holy moly!
Thanks to this I can finally enjoy custom high poly car models with actual playable fps. Thank you so much!

I'm on Windows 10.

But now its broken with 2.5.2 as you can see at the attached screenshots.
Also the fps got a big impact now.
main menu was before above 1000 fps, now just 450 fps
ingame was before 675 fps, dropped down to 520-ish.

Not sure if the fps drop is due to the broken rendering too, or because of the forced borderless mode now.
It was exclusive before and alt+tab worked perfectly fine. Not sure if it is, but whether exclusive or borderless should be a config option.
The changelog did sound like it is now borderless with no possibility to get exclusive again?

I'm not alone with this broken rendering.
Also people in the Retro Racing Point Discord report the same issue for this game.

### Software information
Need for Speed Hot Pursuit 2 (2002)

### System information
- GPU: RTX 2080 Ti
- Driver: 561.09 Studio
- Wine version: 
- DXVK version: 2.5.2 D3D9
<img width="1920" alt="Need for Speed Hot Pursuit 2(Nfshp2 exe) Screenshot 2025 01 01 - 19 15 54 94" src="https://github.com/user-attachments/assets/044fe1d0-f45c-41d1-96db-162113d7070c" />
<img width="1920" alt="Need for Speed Hot Pursuit 2(Nfshp2 exe) Screenshot 2025 01 01 - 19 16 05 49" src="https://github.com/user-attachments/assets/871be333-6382-4a6f-9ada-128418bb9082" />

<img width="1920" src="https://cdn.discordapp.com/attachments/737226135259316224/1323696472319066236/hp2dxvk252.png?ex=67761ca8&is=6774cb28&hm=2722a6ff65ff0559b1472fb288d3994cda6461d1d00369d55c747156c4b42106&" />

Blisto91:
Try using this build https://github.com/doitsujin/dxvk/actions/runs/12435734651

De-M-oN:
fixed the rendering. FPS still that much worse.

[...]

PR #4598, opened 2025-01-09
Reintroduce config option to enable FSE on Windows

### LINKED ISSUE #4595: Request: Bring Fullscreen Exclusive back to Windows
Or at least allow us to re-enable it via dxvk.conf. Exclusive Fullscreen is required for Gsync to probably work, otherwise it's all stuttery and I'm stuck with v2.5.1. Thank you.

doitsujin:
It sucks that Windows users have to choose between functioning VRR, HDR and present_wait (aka any sort of frame latency guarantees) on the one hand, or being able to alt+tab out of their games on the other, but there's not much we can really do as long as win32 vulkan WSI is stuck in d3d9 kinds of stone age.

Currently untested. Closes #4595.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -18,6 +18,18 @@
 # dxgi.enableHDR = True
 
 
+# Allows the Vulkan driver to opt-in to exclusive full-screen mode on
+# Windows. Certain features, such as variable refresh rate or HDR, will
+# not work without this setting, however enabling it will break certain
+# games that use additional GDI windows, and it will also break alt+tab.
+#
+# This setting has no effect on non-Windows platforms.
+#
+# Supported values: True, False
+
+# dxvk.allowFse = False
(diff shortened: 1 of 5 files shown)


K0bin:
Also seems to be the cause of #4583 and probably the performance regression mentioned in #4572

qinlili23333:
> Although we should perhaps mention it in the README, otherwise I assume those issues will keep coming.

I think we should update the wiki page for Windows. We should add more about FSE, VRR, HDR and other various common issues.

Maybe I can write a draft, but seems wiki git cannot accept pull requests. I'll try to write a comprehensive guide on my own site and it can be copied into wiki later if acceptable.

K0bin:
Unfortunately it's not that common that people check the wiki but we could at least link it in response to issues then.

PR #4602, opened 2025-01-10
[d3d8] d3d9 bridge & various other enhancements

WinterSnowfall:
A few d3d8 enhancements, tentatively targeted to land after 2.5.3. Currently WIP.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -1588,12 +1588,11 @@ namespace dxvk {
       default:
         break;
 
-      // TODO: D3DRS_LINEPATTERN - vkCmdSetLineRasterizationModeEXT
-      case D3DRS_LINEPATTERN: {
-        [[maybe_unused]]
-        D3DLINEPATTERN pattern = bit::cast<D3DLINEPATTERN>(Value);
-        stateChange = false;
-      } break;
+      // TODO: Implement D3DRS_LINEPATTERN - vkCmdSetLineRasterizationModeEXT
+      case D3DRS_LINEPATTERN:
+        Logger::warn("D3D8Device::SetRenderState: Unimplemented render state D3DRS_LINEPATTERN");
+        m_linePattern = bit::cast<D3DLINEPATTERN>(Value);
+        return D3D_OK;
 
       // Not supported by D3D8.
       case D3DRS_ZVISIBLE:
@@ -1653,9 +1652,9 @@ namespace dxvk {
       default:
         break;
 
-      // TODO: D3DRS_LINEPATTERN
       case D3DRS_LINEPATTERN:
-        break;
+        *pValue = bit::cast<DWORD>(m_linePattern);
+        return D3D_OK;
 
       // Not supported by D3D8.
       case D3DRS_ZVISIBLE:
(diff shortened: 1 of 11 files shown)


WinterSnowfall:
This is actually good to go even now. I've given it the full d3d8 test treatment in ~200 games or so (in light of the housekeeping I've done a bit earlier too) , and haven't run into any issues.

The lone extra d3d9-side validation this includes has been checked against native behavior, so the risk is on that is also minimal.

WinterSnowfall:
Note that this PR has grown a bit in scope, as I've taken upon me to address the remaining D3DPOOL_SCRATCH paths in CopyRects.

Not that anything about that is too controversial, since d9vk pretty much treats D3DPOOL_SCRATCH as a more lax type of D3DPOOL_SYSTEMMEM anyway, but it is going to need a bit of blanket testing to ensure there are no regressions.

Only a bunch of d3d8 games I've looked at so far actually use D3DPOOL_SCRATCH at all, and none are known to employ it with CopyRects, so this is more in the interest of "future proofing".

hammadnadeemx (approved):
Looks good.

src/d3d8/d3d8_device.cpp
```diff
@@ -640,14 +646,14 @@ namespace dxvk {
    * The following table shows the possible combinations of source
    * and destination surface pools, and how we handle each of them.
    *
-   *     ┌────────────┬───────────────────────────┬───────────────────────┬───────────────────────┬──────────┐
-   *     │ Src/Dst    │ DEFAULT                   │ MANAGED               │ SYSTEMMEM             │ SCRATCH  │
-   *     ├────────────┼───────────────────────────┼───────────────────────┼───────────────────────┼──────────┤
-   *     │ DEFAULT    │  StretchRect              │  GetRenderTargetData  │  GetRenderTargetData  │ -        │
-   *     │ MANAGED    │  UpdateTextureFromBuffer  │  memcpy               │  memcpy               │ -        │
-   *     │ SYSTEMMEM  │  UpdateSurface            │  memcpy               │  memcpy               │ -        │
-   *     │ SCRATCH    │  -                        │  -                    │  -                    │ -        │
```
K0bin, line 649: UpdateSurface errors out if Src isn't SystemMem and Dst isn't Default.
https://github.com/doitsujin/dxvk/blob/736ee20eb56a293dd066dcda49db4e48a4ff6057/src/d3d9/d3d9_device.cpp#L942

K0bin (changes requested):
Tiny nit, looks fine to me otherwise.

src/d3d8/d3d8_device.cpp
WinterSnowfall: Ugh... fine, guess I'll have to create a temporary buffer in SYSTEMMEM and UpdateSurface from that.

WinterSnowfall: Done. Although I've never actually seen anything in the wild doing SCRATCH -> DEFAULT with CopyRects as it doesn't make much sense.

Nice catch, btw. I totally missed it.

K0bin: I would've just kept that as a TODO and reverted that one comment change.

But this way works too. :)

WinterSnowfall: Well, the point of the PR was to get rid of most of the TODOs, so :P... There's still some I'd like to look at in d3d8_shader, but I'll address them separately some day, as that entire deal is definitely outside my comfort zone.

K0bin (approved):

PR #4605, opened 2025-01-11
[util] Enable hideIntegratedGraphics for Bright memory

Blisto91:
The game will prefer other vendors over Intel even if the latter is the dGPU in the system. It likely assumes that Intel is always a iGPU. This also happens with the native game on Windows and so isn't a Linux exclusive. Both in d3d11 and d3d12 mode.

Bright Memory: Infinite which is a newer game does not seem to suffer from this.


Blisto91:
As a general note for this.

I did a bit of further digging and this is in general likely a pre Unreal Engine 4.26 quirk.
The changelog for 4.26 have this line
> New: Added support for Intel hardware with non-local memory. 

This also seems to match up with the UE4 games I've tried so far

Issue #4222, opened 2024-08-28
[d3d8] Pure Pinball 2.0 Redux locks up/crashes during startup


WinterSnowfall:
I guess I'll probably have to record a video of what's going on, since apparently the trace captured invalid vertex positions for all the elements of the pinball table :frog:.

Here's a recording of the issue: https://drive.proton.me/urls/G24CPQ94BR#A3v4Ckv8g7fW

Anyway, there's geometry/lighting flickering in the upper side of all tables, which is apparent whenever the camera changes position. The game doesn't work with WineD3D, but I do remember it working well with d8vk 1.0, so this is most likely a post-merge regression.

### Software information
Pure Pinball 2.0 Redux

### System information
- GPU: RTX 4070
- Driver: 535.183.01
- Wine version: 9.15 Staging
- DXVK version: e4fd9ff

### Apitrace file(s)
[Pure Pinball 2.0 REDUX.trace.tar.xz](https://drive.proton.me/urls/J35H3P6AFW#A2m3iMx078hh)

### Log files
[Pure Pinball 2.0 REDUX_d3d8.log](https://github.com/user-attachments/files/16781673/Pure.Pinball.2.0.REDUX_d3d8.log)

WinterSnowfall:
The game's broken on master currently, and behaves similarly to WineD3D. After the loading screen and a short promo picture, the game ends up on a perpetual black screen instead of the main menu.

I've bisected it down to:
```
207e15eb24b8b4d562a8e461a668eefe1fcf3c0c is the first bad commit
commit 207e15eb24b8b4d562a8e461a668eefe1fcf3c0c
Author: Philip Rebohle <philip.rebohle@tu-dortmund.de>
Date:   Mon Sep 30 13:53:31 2024 +0200

    [dxvk] Refactor swap chain blitter to use plain Vulkan
    
    Temporarily disable the HUD until that is refactored too.
```

WinterSnowfall:
Flickering should be fixed by b99012d, I'll keep this to track the crashing (maybe someday we'll figure out what this game wants).

WinterSnowfall:
Seems that the game does start just fine at times, when the planets are in alignment, so I doubt any of the jank is actually dxvk specific, especially since it behaves the same with WineD3D. Verified that rendering is now correct when it does decide to work.

PR #4609, opened 2025-01-13
Rework swapchain management

doitsujin:
Big refactor that moves a lot of responsibility to the backend, where it should have been in the first place.

Key changes:
- Vulkan surface and swapchain creation now happens *entirely* in the backend. Instead of this clusterfuck where the front-end needs to manually synchronize everything through four different threads and then pray that it actually works, we now just set swapchain properties via a bunch of functions that can then get applied during the next `acquireNextImage` call.
- Enabled automatic conversion between HDR10 and scRGB color spaces if the system only supports one of those and the game uses the other. This is definitely not well tested with actual HDR content though.
- Fixed broken blending with the HUD and software cursor in non-linear color spaces (sRGB if the image format is non-sRGB; HDR10)

Needs **lots** of testing, especially in D3D9 apps since there are so many weird edge cases wrt swapchain usage.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -140,15 +140,6 @@
 # dxgi.maxSharedMemory = 0
 
 
-# Override back buffer count for the Vulkan swap chain.
-# Setting this to 0 or less will have no effect.
-#
-# Supported values: Any number greater than or equal to 2.
-
-# dxgi.numBackBuffers = 0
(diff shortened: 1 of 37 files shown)


WinterSnowfall:
It appears this PR fixes #4222 , for not entirely known reasons.

PR #4623, opened 2025-01-17
Latency tracker

doitsujin:
This isn't Reflex integration yet, but sort of serves as groundwork for it.

`NV_low_latency2` requires us to pass frame IDs to queue submissions, and some markers will need to be submitted at different times due to internal threading, so we kind of need to pass a bunch of state and internal markers around anyway, which is already 90% of the work needed to support latency sleep in many (older) games.

This PR adds this on top of `NV_low_latency2` as well as a homebrew latency sleep solution that's based on our internal timings. This can be enabled by setting `dxvk.latencySleep = True` in the config file.

The next step would be to add another implementation that consumes markers from the application (and handles all sorts of potential bugs) and implement the `ID3DLowLatencyDevice` interface for dxvk-nvapi.

Obviously not tested very well, especially D3D9 is a mess.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -106,6 +106,15 @@
 # dxvk.latencyTolerance = 1000
 
 
+# Disables the use of VK_NV_low_latency2. This will make Reflex unavailable
+# in games, and if dxvk.latencySleep is set to True, a custom algorithm will
+# be used for latency control.
+#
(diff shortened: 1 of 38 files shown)


doitsujin:
Closing, will make follow-up PR with full Reflex support since I mostly have that working locally now.

PR #4629, opened 2025-01-20
[d3d8/9] Proper (and age accurate) handling of d3d9.shaderModel = 0

### LINKED ISSUE #4625: Skybox position, scale, and orientation incorrect and variable when using d3d9.shaderModel = 0
### Description of the issue
When using d3d9.shaderModel = 0;
The skybox (hemisphere) is rendered too small, and changes position / orientation / scale seemingly randomly as the camera position and orientation changes.
This does not occur in the stock base game's fixed function pipeline running without DXVK.

### Replication
Run Painkiller Black Edition with d3d9.shaderModel = 0 set in dxvk.conf.
Load a level which uses a skybox (level 1 for example).
Observe the skybox in the incorrect scale and location.
Move around and change viewing angle to see the skybox occasionally teleport to a different location/orientation/scale

### Software information
Game: Painkiller Black Edition
Settings: Doesn't matter, as long as the skybox is not disabled in settings

### System information
- GPU: RTX 3080 10GB
- Driver: Latest 566.36
- Wine version: Using windows
- DXVK version: 2.5.3

### API Trace
I only seem to be able to run apitrace without dxvk, replacing the dxvk d3d9.dll with the apitrace d3d9.dll.
I tried the method with the command line as such:
- apitrace trace --api d3d9 F:\SteamLibrary\steamapps\common\Painkiller Black Edition\Bin\Painkiller.exe
- apitrace trace --api dxgi F:\SteamLibrary\steamapps\common\Painkiller Black Edition\Bin\Painkiller.exe

However with --api d3d9 I received the message:
"\lib\wrappers\d3d9.dll was never used: application probably does not use this API"
With --api dxgi I received no message and no trace was generated.

Here are two screenshots, the first shows the issue clearly, and the second shows the skybox smaller and off in the distance when altering the viewing angle.

Image

Image

As I was unable to find another wrapper which could force the Pain engine into fixed function mode, I used a patched .dll to force the engine to use the fixed function path without using any wrappers - the issue did not occur and the sky rendered correctly in the "stock" fixed function path without DXVK.

### Log

[Painkiller_d3d9.log](https://github.com/user-attachments/files/18476531/Painkiller_d3d9.log)

WinterSnowfall:
Fixes #4625 and follows up on an idea I've had in the d8vk days, see https://github.com/AlpyneDreams/d8vk/pull/208.

Extends D3D7-compatible FF only caps (aka disabling programmable shader support) to D3D8 when using `d3d9.shaderModel = 0`, and also validates shader versions on creation in D3D9.

Useful for debugging and for nostalgia reasons (with games that do handle such situations properly).

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3314,6 +3314,19 @@ namespace dxvk {
     if (unlikely(ppShader == nullptr))
       return D3DERR_INVALIDCALL;
 
+    const uint32_t majorVersion = (pFunction[0] >> 8) & 0xff;
+    const uint32_t minorVersion = pFunction[0] & 0xff;
+
+    // Late fixed-function capable hardware exposed support for VS 1.1
+    const uint32_t shaderModelVS = m_d3d9Options.shaderModel == 0 ? 1 : m_d3d9Options.shaderModel;
+
+    if (unlikely(majorVersion > shaderModelVS
+             || (majorVersion == 1 && minorVersion > 1)
+             || (majorVersion > 1  && minorVersion != 0))) {
+      Logger::err(str::format("D3D9DeviceEx::CreateVertexShader: Unsupported VS version ", majorVersion, ".", minorVersion));
+      return D3DERR_INVALIDCALL;
+    }
+
     DxsoModuleInfo moduleInfo;
     moduleInfo.options = m_dxsoOptions;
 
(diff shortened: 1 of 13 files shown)


WinterSnowfall:
I wasn't expecting any revelations from testing the caps on my GeForce 4 MX 440 card, but apparently I was wrong.

Late generations of fixed-function only hardware actually advertised support for VS 1.1 (with no PS support), which makes a lot of sense actually, since games would try to use VS 1.1 anyway, even if we tried to force them to not create programmable shaders of any kind.

This PR replicates the above behavior when `d3d9.shaderModel = 0` is used and also fixes some other minor SM1 derps.

src/d3d9/d3d9_device.cpp
```diff
@@ -3314,6 +3314,17 @@ namespace dxvk {
     if (unlikely(ppShader == nullptr))
       return D3DERR_INVALIDCALL;
 
+    uint32_t majorVersion = (pFunction[0] >> 8) & 0xff;
+    uint32_t minorVersion = pFunction[0] & 0xff;
+
+    // Late fixed-function capable hardware exposed support for VS 1.1
+    uint32_t shaderModelVS = m_d3d9Options.shaderModel == 0 ? 1 : m_d3d9Options.shaderModel;
+
+    if (unlikely(majorVersion > shaderModelVS)) {
```
K0bin, line 3323: This seems redundant. We already check the shader model in `D3D9Shader::GetShaderModule`. 
It would be better to take a look at the check in there and make sure that's correct.

Or just remove the one in `D3D9Shader::GetShaderModule` because the shader model numbers seem different anyway.

WinterSnowfall: Done, and I dumped the shader model number from the dxso options completely since it was only used during that validation.

K0bin (approved):
Good job.

PR #4632, opened 2025-01-21
Nvidia Reflex support

doitsujin:
This implements Reflex support through dxvk-nvapi's `ID3DLowLatencyDevice` on drivers that support `VK_NV_low_latency2`. Fairly invasive change since we need to pass frame IDs all over the place and mess around with certain markers to account for internal threading.

Tested in a few games:
- **Ghostrunner**: Broken, game never even calls the function to enable Reflex at all, let alone the sleep function.
- **Ghostrunner 2**: Works as expected.
- **God of War**: Works as expected.
- **Mordhau**: Game sets present markers but thinks that Reflex is unsupported, so you can't enable it in game. (???)
- **Overwatch 2**: Works, but only if Vsync is disabled. I don't know if this is expected behaviour.
- **Quake Champions**: Doesn't use markers, but appears to work.

Not sure if there are any other Reflex-enabled D3D11 games that aren't simultaneously blocked by anti-cheat measures, ~~it's a bit unfortunate that the only UE4 test case seems to be bugged at a game level~~.

This also adds a HUD item (`latency`) that will show up when a game uses latency markers, and will display the average time it takes from the game setting the `SIMULATION_START` (or `INPUT_SAMPLING`, if it uses that) marker to presentation of the given frame completing on the GPU, as well as the average latency sleep duration.

Bildschirmfoto-829

Things that are not really supported very well include non-monitonic frame IDs (dxvk-nvapi sorts these out for us so that should never happen anyway), and the whole thing probably isn't very well behaved when a game randomly skips markers for a frame since it breaks our way of correlating app-provided frame IDs with the ones we use for the Vulkan swapchain.

Finally, this adds an option to enable latency sleep for older games; the obvious downside here is that we can only really sleep after Present, so this isn't going to do all that much in games that have an asynchronous render thread, besides reducing some buffering inside DXVK. This can be enabled by setting `dxvk.latencySleep = True` in the configuration file.

Supersedes #4623, #3690.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -106,6 +106,16 @@
 # dxvk.latencyTolerance = 1000
 
 
+# Disables the use of VK_NV_low_latency2. This will make Reflex unavailable
+# in games, and if dxvk.latencySleep is set to True, a custom algorithm will
+# be used for latency control. By default, the extension will not be used in
+# 32-bit applications due to driver issues.
+#
+# Supported values: Auto, True, False
+
+# dxvk.disableNvLowLatency2 = Auto
+
+
 # Override PCI vendor and device IDs reported to the application. Can
 # cause the app to adjust behaviour depending on the selected values.
 #
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -76,6 +76,36 @@
 # d3d9.maxFrameRate = 0
 
 
+# Controls latency sleep and Nvidia Reflex support.
+#
+# Supported values:
+# - Auto: By default, DXVK only supports latency sleep in D3D11 games that
+#         use Reflex if the graphics driver supports VK_NV_low_latency2,
+#         and if dxvk-nvapi is enabled in Proton.
+# - True: Enables built-in latency reduction based on internal timings.
+#         This assumes that input sampling for any given frame happens after
+#         the D3D9 or DXGI Present call returns; games that render and present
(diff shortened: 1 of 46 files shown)


WinterSnowfall:
I've tested out `dxvk.latencySleep = True` in quite a few d3d8 and d3d9 games and it does seem to work as intended, pretty much negating the latency introduced by VSync. How observable the effect is ultimately depends on other aspects of a game and of course the eye of the beholder.

I haven't run into any ill effects or instability, beside a minor quirk in Aquanox 2 - it, for some reason, seems to half its framerate at times with the setting enabled, which doesn't happen with it disabled.

qinlili23333:
Suggest game for test: My Time at Sandrock (D3D11, Unity 2021, Single player game without anticheat)
Party Animals may also be useful for test.

The full list of Reflex enabled games: https://www.nvidia.com/en-au/geforce/technologies/reflex/supported-products/

doitsujin:
I'm aware of the full list, but 99% of the games on that are D3D12 only so it's kinda hard to really pick something out.

Anyway, will give that Unity game a go.

doitsujin:
Nevermind my previous comment, it does actually do something, just wasn't testing the correct dxvk branch :frog:

Too bad the game is CPU-bound like hell so there's no actual latency sleep happening.

PR #4645, opened 2025-01-26
[d3d9] Relax SM2 minor version checks on shader creation

WinterSnowfall:
Thanks to @Blisto91 for noticing it.

D3D9 archeology strikes again, as some PS version 2.1 have been spotted in the wild. I can only assume they correspond to the more lenient 2_x SM2 mode, although that doesn't come with an equivalent cap at device level.

I have found obscure references that ATI provided an override to 2.1 at some point at least: https://www.bodnara.co.kr/bbs/article.html?num=49848 .

The conventions around 2_x are very unclear, so this PR relaxes the minor version checks with SM2, to ensure we don't cause any issues with late D3D9 games.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3322,7 +3322,8 @@ namespace dxvk {
 
     if (unlikely(majorVersion > shaderModelVS
              || (majorVersion == 1 && minorVersion > 1)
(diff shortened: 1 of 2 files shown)


K0bin (approved):

Blisto91:
As a public note one game that would crash without this change is Dragon Age Origins when Graphics Detail is set to Low

PR #4646, opened 2025-01-26
[d3d9] Route operations on unmappable memory through allocator

### LINKED ISSUE #4613: D3D9: Multithreaded Texture creation may crash
Hi,

I was curious if i could apply dxvk on my d3d9 project and came across one issue:

I create my d3d9ex device with multithreaded flag and use it to load my assets in a separate thread.

And this is where DXVK seems to have its issue:

```
AV: Exception thrown at 0x67FC3BDC (d3d9.dll) in project.exe: 0xC0000005: Access violation writing location 0x00000000.


[Inline Frame] d3d9.dll!dxvk::D3D9MemoryChunk::Unmap(dxvk::D3D9Memory *) Line 157	C++
d3d9.dll!dxvk::D3D9Memory::Unmap() Line 285	C++
[Inline Frame] d3d9.dll!dxvk::D3D9CommonTexture::UnmapData() Line 238	C++
d3d9.dll!dxvk::D3D9Initializer::InitTexture(dxvk::D3D9CommonTexture * pTexture, void * pInitialData) Line 63	C++
d3d9.dll!dxvk::D3D9DeviceEx::CreateTexture(unsigned int Width, unsigned int Height, unsigned int Levels, unsigned long Usage, _D3DFORMAT Format, _D3DPOOL Pool, IDirect3DTexture9 * * ppTexture, void * * pSharedHandle) Line 663	C++
d3dx9_43.dll!_D3DXLoadVolumeFromResourceW@36()	Unknown
d3dx9_43.dll!_D3DXCreateTextureFromFileInMemoryEx@60()	Unknown
```

Looking at the assembly at the eip of the exception, it seems to out-of-bounds the m_mappingRanges vector.
In my case it seems to access index 0, but the vector is empty. 

Likely a race condition as it may work at some times and other times not.

Throwing all work onto a single thread solves this and dxvk works just fine then.

### Software information
AMD64 Windows 24H2 26100.2605, 32bit application

### System information
- GPU: RTX 4090
- Driver: 566.45
- Wine version: n/a
- DXVK version: 32bit 2.5.3*(*)

*windows-msvc CI: dxvk-merge-69a3d78b3435c39caf561ea71778ae6be0e9f44b-msvc-output
**This bug also appears on older versions

### Apitrace file(s)
/

### Log files
/

K0bin:
Hopefully fixes #4613

There was a race that could lead to freeing a chunk and allocating memory from it at the same time. This was extremely rare because D3D9 games usually don't make use of multi threading for rendering and just have a single rendering.
On top of that nearly all calls into any methods related to this are protected by the device lock anyway.

To fix this we just always route all operations through the general allocator which acquires its own lock before
calling the equivalent function inside the chunk.

### DIFF
--- a/src/d3d9/d3d9_mem.cpp
+++ b/src/d3d9/d3d9_mem.cpp
@@ -19,6 +19,7 @@ namespace dxvk {
     SYSTEM_INFO sysInfo;
     GetSystemInfo(&sysInfo);
     m_allocationGranularity = sysInfo.dwAllocationGranularity;
+    m_mappingGranularity = m_allocationGranularity * 16;
   }
 
   D3D9Memory D3D9MemoryAllocator::Alloc(uint32_t Size) {
@@ -26,7 +27,7 @@ namespace dxvk {
 
     uint32_t alignedSize = align(Size, CACHE_LINE_SIZE);
     for (auto& chunk : m_chunks) {
-      D3D9Memory memory = chunk->Alloc(alignedSize);
+      D3D9Memory memory = chunk->AllocLocked(alignedSize);
       if (memory) {
         m_usedMemory += memory.GetSize();
         return memory;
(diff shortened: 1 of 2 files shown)


enginelesscc:
Tested it for my scenario in https://github.com/doitsujin/dxvk/issues/4613
Seems fixed, haven't encountered any problems so far :)

K0bin:
Great, thanks for testing.

PR #4647, opened 2025-01-26
[d3d9] Adjust SM2 VS/PS caps

WinterSnowfall:
These were ultimately (largely) correct, but they do have D3D constants defined.

~I still have to check how all of these show up on a SM1 capable card, so WIP for now.~ Update: I've checked, and they're all 0 for SM1 (and fixed-function ofc).

Doubt it matters much beyond esthetics, but let's unleash it in experimental for a while.


doitsujin:
Has this been compared against native, or do the actuial values not change and we just report 0 if we disable sm2 support?

WinterSnowfall:
> Has this been compared against native, or do the actuial values not change and we just report 0 if we disable sm2 support?

For now I've only compared it with WineD3D, which seems to follow the constants, so I don't have much concerns there. And for SM1, I checked on Windows XP / a GeForce 4 Ti and they are all 0.

PR #4651, opened 2025-01-27
Assign viewInfo.minLayer for D3D11VideoProcessorInputView

VPInventions:
Issue: D3D11VideoProcessorInputView constructor ignores D3D11_VIDEO_PROCESSOR_INPUT_VIEW_DESC::Texture2D.ArraySlice parameter and always sets viewInfo.minLayer to 0.

The fix assigns the correct array layer value.

### DIFF
--- a/src/d3d11/d3d11_video.cpp
+++ b/src/d3d11/d3d11_video.cpp
@@ -200,7 +200,7 @@ namespace dxvk {
         viewInfo.viewType   = VK_IMAGE_VIEW_TYPE_2D;
         viewInfo.mipIndex   = m_desc.Texture2D.MipSlice;
         viewInfo.mipCount   = 1;
-        viewInfo.layerIndex = 0;
+        viewInfo.layerIndex = m_desc.Texture2D.ArraySlice;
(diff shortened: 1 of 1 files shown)


doitsujin (approved):
Will merge once CI passes, was any actual app running into this?

VPInventions:
I'm implementing video decode acceleration for VirtualBox (I'm one of developers), i.e. D3D11VideoDecoder and D3D11VideoDecoderOutputView and corresponding dxvk bits. VirtualBox uses dxvk for 3D acceleration on Linux hosts.

The guest creates D3D11VideoProcessorInputViews (and D3D11VideoDecoderOutputViews) from a 2D texture array with 10 layers. So the guest displays every 10th frame without the proposed fix.

BTW, would you be interested in including video acceleration in dxvk at some point in the future? Currently it is still a work in process. There are a number of hardcoded things and I have not figured out yet how to correctly synchronize video queue with graphics queue in dxvk. But the code already works in principle and decodes H.264 videos. Other formats will be added later.

PR #4653, opened 2025-01-28
[d3d8] Fix up swizzle for all opcodes requiring a replicate swizzle

### LINKED ISSUE #4113: [d3d8] Missing f(r)og in NFS Hot Pursuit 2
I don't know where the f(r)og is. Both WineD3D and d3d8to9 + dxvk render fog correctly, while d8vk does not.

I've captured a trace using WineD3D which reproduces the issue. Fog should be instantly visible (or not) on race start, although it's a somewhat subtle effect. Looking at the lighthouse in the distance near the end of the trace provides more clarity on the matter.

P.S.: The game does indeed use a lot of VS with `exp oFog.x, -r1`, but that doesn't currently do anything fog-wise.

### Software information
Need for Speed: Hot Pursuit 2

### System information
- GPU: RTX 4070
- Driver: 535.183.01
- Wine version: 9.12 Staging
- DXVK version: fe9c875

### Apitrace file(s)
[NFSHP2.trace.tar.xz](https://drive.proton.me/urls/GBG8K3XZ0M#qcfvVyFZGtx2)

### Log files
[NFSHP2.log](https://github.com/AlpyneDreams/d8vk/files/13268992/NFSHP2.log)

WinterSnowfall:
Fixes #4113.

https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/d3d9types/ne-d3d9types-_d3dshader_instruction_opcode_type

The information in the docs isn't clear or complete, but it is at least clear that RSQ isn't the only opcode which requires the explicit use of a replicate swizzle for its inputs.

If the swizzle on inputs is somehow `.xyzw` / D3DVS_NOSWIZZLE (although that is technically illegal in the context of the affected opcodes) it appears native d3d8 does indeed impose a replicate swizzle, as @AlpyneDreams has already determined when looking at **Indiana Jones And The Emperor's Tomb**, and also based on empirical evidence it appears `.w` is the one that makes things look as expected, and in this case also fixes the missing fog in **NFS Hot Pursuit 2**.

Needs testing to ensure nothing regresses otherwise, but once that is done, I'll undraft.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -1788,8 +1788,8 @@ namespace dxvk {
 
     // Validate VS version for non-FF shaders
     if (pFunction != nullptr) {
-      const uint32_t majorVersion = (pFunction[0] >> 8) & 0xff;
-      const uint32_t minorVersion = pFunction[0] & 0xff;
+      const uint32_t majorVersion = D3DSHADER_VERSION_MAJOR(pFunction[0]);
+      const uint32_t minorVersion = D3DSHADER_VERSION_MINOR(pFunction[0]);
 
       if (unlikely(majorVersion != 1 || minorVersion > 1)) {
         Logger::err(str::format("D3D8Device::CreateVertexShader: Unsupported VS version ", majorVersion, ".", minorVersion));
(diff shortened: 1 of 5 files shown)


K0bin:
The shader bytecode came with the game and wasn't compiled at runtime, right?

WinterSnowfall:
> The shader bytecode came with the game and wasn't compiled at runtime, right?

Pretty sure that's the case for almost all d3d8 games, if not, in fact, all.

WinterSnowfall:
Tested on the full complement of ~200 d3d8 games - found no regressions or other types of issues.

K0bin (approved):

Issue #4082, opened 2024-06-20
[d3d9, regressions] Alt-tab issues in fullscreen at non native resolutions


Blisto91:
I've noticed a alt-tab issue that seemingly affects d3d9 games in general with dxvk. This behavior is a regression compared to older versions of dxvk.
The issue happens you set the game to fullscreen mode and a non native resolution, that is different than 1080p for me. When you then do a alt-tab the game will be a mostly black screen except a small corner in the top left that will either show the game or a part of a background app.

The 3 games i have tested for this so far are Battlestations Midway, Borderlands 2 and Guild Wars 1. In all these cases setting `d3d9.deviceLossOnFocusLoss = True` works around the issue, but that might be a undesirable solution in general.

#### Midway and Borderlands 2
I have bisected the issue to start with commit https://github.com/doitsujin/dxvk/commit/83f6400a07035aba10538bd21336dcf89fc22393
@WinterSnowfall Gave me a little experiment and inserting the code below the linked line makes the issue go away. Not that it is a proper solution.
```
windowData.swapchain->RecreateSwapChain();
```
https://github.com/doitsujin/dxvk/blob/e9fc071d9548a6e865f09b9c03f9349ef0aa5c1e/src/d3d9/d3d9_window.cpp#L75 

#### Guild Wars 1
Bisected to instead start with https://github.com/doitsujin/dxvk/commit/b0ed9e30ceeb8ef16d242ea189f53e717ab37201. 
The code trick above for Midway and Borderlands 2 does not work here for this one.
I recently [added a app profile](https://github.com/doitsujin/dxvk/pull/4080) to work around this before i knew of this being a general issue, so maybe that isn't a desired fix.

Note that the regressing commit for Midway and Borderlands 2 also have an effect for GW1 in that the mouse cursor will be restricted to the small corner up in the left side while visuals stay the same. This goes away in later dxvk versions (not bisected) before it then regresses visually the same with the commit noted in the GW1 section.

### System information
- GPU: RTX 4080
- Driver: 550.40.63
- Wine version: Proton Bleeding Edge

Blisto91:
Extra note since d8vk was merged since I made this.

This problem affects some d3d8 games the same and is probably a bigger issue there in practice since a lot of games from that era won't support 1080p and above.
I personally consider this issue the biggest showstopper in regards to enabling d3d8 support by default in Proton as I would consider it a regression versus currently used wined3d.

WinterSnowfall:
[...]

PR #4655, opened 2025-01-29
[util] Cleanup some deviceLossOnFocusLoss configs

Blisto91:
A bunch of alt-tab related issues are gone on current master which allows a small builtin config cleanup.
This has been tested on a RTX 4080 in KDE Wayland, KDE X11 and Windows.

Delta Force, The Sims 3, Guild Wars 1 and Splinter Cell Conviction work out of the box since https://github.com/doitsujin/dxvk/pull/4609 so the configs aren't needed.
Rise of Nations is fine since the 2.4.1 release without the config.

Closes https://github.com/doitsujin/dxvk/issues/4082

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -936,12 +936,7 @@ namespace dxvk {
     /* Delta Force: Xtreme 1 & 2                 *
      * Black screen on Alt-Tab and performance   */
     { R"(\\(DFX|dfx2)\.exe$)", {{
-      { "d3d9.deviceLossOnFocusLoss",       "True" },
       { "d3d9.cachedDynamicBuffers",        "True" },
-    }} },
-    /* The Sims 3 - Black screen on alt-tab      */
-    { R"(\\TS3(W)?\.exe$)", {{ 
-      { "d3d9.deviceLossOnFocusLoss",       "True" },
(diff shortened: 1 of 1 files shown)


doitsujin:
Do we understand why these are no longer needed?

Blisto91:
For the 4 games mentioned first and the ones in https://github.com/doitsujin/dxvk/issues/4082 then, the ones that wanted to work with the commit, i reversed bisected them to start working on their own from https://github.com/doitsujin/dxvk/commit/43838d3df8c1e5e1d3ec1f69413598b7ff296c6e
The problem mentioned in the linked issue are old regressions from a previous time when the swapchain code was touched.
Though i can't give a technical reason as to why.

Rise of Nations i have not bisected which specific commit that fixed the issue in 2.4.1

doitsujin:
The strange part here is that the swapchain backend code really **should** be transparent to the app, so this is a little bit worrying.

Will merge for now, but probably good to keep an eye on things.

turbedi:
Aren't the comments for Splinter Cell Conviction and Delta Force outdated now? They still mention the black screen on Alt-Tab

Blisto91:
Yes. Derp

PR #4670, opened 2025-02-04
[d3d9] Update software cursor position using SetCursorPosition

### LINKED ISSUE #3020: Dungeon Siege 2 invisible mouse cursor

WinterSnowfall:
Fixes #3020 on Windows as well.

Games relying on software cursors apparently also call SetCursorPosition on each frame, and since we call ::SetCursorPos(X, Y) when that happens, on native Windows that tends to cause some strange movement feedback for some reason when we later retrieve the cursor position before present, using ::GetCursorPos(). And, you guessed it, this works just fine in Wine...

Regardless, what seems to work just as fine both on Windows and Wine is simply using the coordinates passed by the application with SetCursorPosition calls instead of us getting the position of the win32 cursor before present.

The downside is that if any game expects software cursors to work without it calling SetCursorPosition on each frame, it will be very very disappointed now.

Needs a bit more testing to confirm the above doesn't happen in practice. For now it has been confirmed to work on Windows and Linux with Dungeon Siege 2 at least.

### DIFF
--- a/src/d3d9/d3d9_cursor.cpp
+++ b/src/d3d9/d3d9_cursor.cpp
@@ -29,20 +29,20 @@ namespace dxvk {
 
 
   void D3D9Cursor::UpdateCursor(int X, int Y) {
+    // SetCursorPosition is used to directly update the position of software cursors,
+    // but keep track of the cursor position even when using hardware cursors, in order
+    // to ensure a smooth transition/overlap from one type to the other.
+    m_sCursor.X = X;
+    m_sCursor.Y = Y;
+
+    if (unlikely(m_sCursor.Width > 0 && m_sCursor.Height > 0))
+      return;
+
     POINT currentPos = { };
     if (::GetCursorPos(&currentPos) && currentPos == POINT{ X, Y })
         return;
 
     ::SetCursorPos(X, Y);
-  }
-
-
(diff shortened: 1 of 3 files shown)


WinterSnowfall:
> An operating system cursor is created and used under either of these conditions:
> 
>  a) The hardware has set D3DCURSORCAPS_COLOR (see [D3DCURSORCAPS](https://learn.microsoft.com/en-us/windows/desktop/direct3d9/d3dcursorcaps)), and the cursor size is 32x32 (which is the cursor size in the operating system).
>  b) The application is running in windowed mode.
> 
> Otherwise, DirectX uses an emulated cursor. An application uses [IDirect3DDevice9::SetCursorPosition](https://learn.microsoft.com/en-us/windows/desktop/api/d3d9/nf-d3d9-idirect3ddevice9-setcursorposition) to move an emulated cursor to follow mouse movement.

D3D9 docs appear to suggest this is indeed the expected approach for emulated/software cursors, so I think we should be good here. All known games using the path still work in Wine, @Blisto91 will graciously confirm it's all fine on Windows as well (hopefully).

PR #4673, opened 2025-02-06
[hud] Use compute shader for text, use Titillium Web font

synchromach:
dxvk font 1

Key changes:

- There's a new compute shader used in `drawTextIndirect`, glue code for which is mostly shamelessly copied from `HudFrameTimeItem`.
- Each line of text is now 2 triangle strips instead of 1, one for the color on the front and one for the outline on the back - so each character now takes 4 triangles instead of 2.
- It now uses a non-monospaced font, Titillium Web, at 64px on a 1020x334 texture (this might hurt for people with low memory)

I have not put any effort at all into optimizing this and most numbers in the HUD are now out of visual-alignment. Please suggest any and all changes.


synchromach:
> > It now uses a non-monospaced font, Titillium Web, at 64px on a 1020x334 texture (this might hurt for people with low memory)
> 
> IMO it looks a lot worse than the current font.

Do you have a different font in mind, or do you want the old font back? I'll generate it.

K0bin:
> Do you have a different font in mind, or do you want the old font back? I'll generate it.

I just don't really understand the point of this PR. The current HUD is a debugging tool. It works and looks fine.

synchromach:
> > Do you have a different font in mind, or do you want the old font back? I'll generate it.
> 
> I just don't really understand the point of this PR. The current HUD is a debugging tool. It works and looks fine.

Alright, closing.

PR #4682, opened 2025-02-11
Fix possible presenter deadlock (hopefully)

doitsujin:
TL;DR there are situations where swapchain destruction will wait for the frame worker to drain the queue, but the frame worker might lock up in the frame latency tracker because something is calling `setLatencyMarkerNV` which in turn stalls until swapchain recreation finishes and keeps the latency tracker locked.

It's a mess, needs testing.


K0bin:
It appears to fix it for Ready Or Not. I've had no issues reproducing the hang with master but can't manage to get it with the PR.

K0bin (approved):
LGTM

PR #4698, opened 2025-02-19
[d3d8/9] Prevent device child ref underruns on release

### LINKED ISSUE #4615: [d3d9] D3D9DeviceChild refcounts should be clamped on Release()
Calling Release() on D3D9DeviceChild currently causes refcount underruns if an application keeps calling it even after the refcount has reached 0. In some cases (especially due to race conditions) this can cause livelocking, although I expect this to be exceptionally rare if it's even a problem in practice.

In d3d8 land we have encountered a game that depends on this behaving correctly however (without underruning), namely Motor City Online, which is why we've added [this bit](https://github.com/doitsujin/dxvk/blob/c707d9026f33b6ab89639f154b6ac5f6326fa037/src/d3d8/d3d8_device_child.h#L34-L36). While it perhaps isn't fully thread safe, it does solve the underlying problem. I have PRed the same change on d3d9 side in the past but it was deemed to not be an ideal way of handling it.

There is also a WineD3D d3d9 device test centered around device refcounts which currently fails due to our mishandling of the situation.

We should, at some point, fully fix this properly in both d3d8 and d3d9.

WinterSnowfall:
Closes #4615.

If there's a more optimal way to do it or you have any suggestions to improve the code, they're more than welcome. It's the first time I'm implementing anything of the sort in C++, so there might be some rough edges.

The second commit is an homage to @K0bin's early 2025 quote :stuck_out_tongue: :

> The rest of DXVK uses numeric_limits<uint32_t>::max() instead.

Needs a bit of testing, will undraft once that is done.

### DIFF
--- a/src/d3d8/d3d8_device_child.h
+++ b/src/d3d8/d3d8_device_child.h
@@ -31,16 +31,28 @@ namespace dxvk {
     }
 
     ULONG STDMETHODCALLTYPE Release() {
-      // ignore Release calls on objects with 0 refCount
-      if(unlikely(!this->m_refCount))
-        return this->m_refCount;
+      uint32_t oldRefCount, refCount;
 
-      uint32_t refCount = --this->m_refCount;
+      do {
+        oldRefCount = this->m_refCount.load(std::memory_order_acquire);
+
+        // clamp value to 0 to prevent underruns
+        if (unlikely(!oldRefCount))
+          return 0;
+
+        refCount = oldRefCount - 1;
+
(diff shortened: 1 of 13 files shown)


K0bin:
🐸

WinterSnowfall:
Tested a bunch of d3d9 games and a few d3d8 games for good measure (though I don't expect any surprises there), and it seems all good. The Wine d3d9 refcount test is also happy now.

PR #4699, opened 2025-02-19
Lazy binding for D3D11

doitsujin:
~~Based on #4693 because the DXBC bits would otherwise turn into rebase hell, so that needs to land first.~~ Merged.

TL;DR is that we defer forwarding resource bindings (`SetShaderResourceViews` and friends) to the backend if:
- None of the active shaders use the binding, or
- The app sets a null buffer/view/sampler.

These will instead be applied on the next draw or dispatch that are actually known to use the given binding.

The rationale here is that forwarding unused bindings can confuse the backend a little and lead to descriptor set invalidations even if none of the *actually used* bindings have changed. As for null bindings, these are very rarely actually used in shaders, and instead we can assume that the app will either bind a non-null resource right after or binds a shader which doesn't use the given binding slot.

At the same time, immediately forwarding currently *active* binding avoids the extra overhead from tracking everything in well-behaved applications, even if this means that *some* bindings might be redundant (e.g. app binds a shader that doesn't use it right after, or binds two different resources to the same slot back-to-back).

In practice, this improves CPU-bound performance in God of War somewhat, and in Trine 5 this reduces the number of draw calls by ~50% in the menu because our indirect draw batching can now kick in properly. Almost all games run into this in *some* capacity, so it should be a nice overall improvement.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -3274,6 +3274,15 @@ namespace dxvk {
     auto dirtyMask = m_state.lazy.shadersDirty & m_state.lazy.shadersUsed;
     dirtyMask.clr(DxbcProgramType::ComputeShader);
 
+    if (unlikely(!(dirtyMask & m_state.lazy.graphicsUavShaders).isClear())) {
+      DxbcProgramType stage = DxbcProgramType::PixelShader;
+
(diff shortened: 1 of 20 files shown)


DanielPopRe:
Bro i have an idea can u intagrate even the dx 12 ?

Blisto91:
@DanielPopRe Don't comment on random PRs

PR #4717, opened 2025-02-26
Work around broken Watch Dogs 2 compute shaders

doitsujin:


src/d3d11/d3d11_options.h
```diff
@@ -26,6 +26,13 @@ namespace dxvk {
     /// without explicit synchronization.
     bool forceVolatileTgsmAccess = false;
 
+    /// Force UAV synchronization insided compute shaders
+    ///
+    /// Workaround for compute shaders that access overlapping
+    /// memory regions within a UAV without proper workgroup
+    /// synchroniation. Will have a negative performance impact.
```
Margen67, line 33:
```suggestion
    /// synchronization. Will have a negative performance impact.
```

PR #4721, opened 2025-03-01
[dxvk] Add CS thread load to the HUD

K0bin:
Same way we estimate the GPU load.

### DIFF
--- a/src/dxvk/dxvk_cs.cpp
+++ b/src/dxvk/dxvk_cs.cpp
@@ -202,11 +202,22 @@ namespace dxvk {
       while (!m_stopped.load()) {
         { std::unique_lock<dxvk::mutex> lock(m_mutex);
 
-          m_condOnAdd.wait(lock, [this] {
-            return (!m_queueOrdered.queue.empty())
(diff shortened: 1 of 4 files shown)


src/dxvk/dxvk_cs.cpp
```diff
@@ -202,11 +202,20 @@ namespace dxvk {
...
         { std::unique_lock<dxvk::mutex> lock(m_mutex);
 
-          m_condOnAdd.wait(lock, [this] {
-            return (!m_queueOrdered.queue.empty())
-                || (!m_queueHighPrio.queue.empty())
-                || (m_stopped.load());
-          });
+          if (unlikely(m_queueOrdered.queue.empty()
+            && m_queueHighPrio.queue.empty()
+            && (!m_stopped.load()))) {
```
doitsujin, line 207: could make this a bit more elegant and do something like
```c++
auto pred = [this] { return !queue.empty() && ... };

if (!pred()) {
  ...
  m_condOnAdd.wait(lock, pred)
}
```

since the condition isn't trivial her

doitsujin (commented):
small nit

Issue #1885, opened 2021-01-04
[d3d11] Total War Troy rendering issue


Leopard1907:
Game has a rendering issue. It has these weird looking cloud ( possibly?) shadow outlines on campaign map. Zooming in and out at campaign map affects visibility of it but that is hard to not notice it. Game has this issue with DXVK ( didn't try native d3d11 but i assume it doesn't have this) since it was released at August 2020. Since game updates over time didn't change anything on that regard, i assume that is a DXVK issue.

![Ekran Görüntüsü - 2021-01-05 01-16-36](https://user-images.githubusercontent.com/24878978/103587015-6139d200-4ef7-11eb-8dc2-7f7f6196eb02.png)
![Ekran Görüntüsü - 2021-01-05 01-17-26](https://user-images.githubusercontent.com/24878978/103587021-63039580-4ef7-11eb-93cf-34ce42239974.png)

### Software information
A Total War Saga:Troy
Setting used: High preset, 4X AA and 16x AF

### System information
- GPU: Nvidia GTX 1050
- Driver: 455.45.01
- Wine version: 6.0
- DXVK version: 1.7.3

### Apitrace file(s)
Apitrace didn't work with it. I have captured two rdc's.

https://drive.google.com/file/d/16UJ5OF2kD4PiydOg1CzE3nnYmQGQHwNX/view?usp=sharing

https://drive.google.com/file/d/1gNoB4-kjIvsMWNwkSRfoiYlpef6mbWYR/view?usp=sharing

### Log files
- d3d9.log: -
- d3d11.log: 
[Troy_d3d11.log](https://github.com/doitsujin/dxvk/files/5767103/Troy_d3d11.log)

- dxgi.log: 
[Troy_dxgi.log](https://github.com/doitsujin/dxvk/files/5767105/Troy_dxgi.log)

doitsujin:
Is this even a bug that doesn't happen on Windows?

CME42:
It's not reproducible on windows.
it seems like the game tries to sample from a multisampled depth buffer when things go wrong:
![Bildschirmfoto von 2021-01-12 11-48-03](https://user-images.githubusercontent.com/11275913/104304758-165a2480-54cc-11eb-8a51-bdcc8cae9aa5.png)

There are also some validation errors:
```
[...]

Issue #1216, opened 2019-10-05
World of Final Fantasy Offset issue


kust2708:
Rendering in game seems to have some issue, the screen is splitted by multiple horizontal bands with offset.

![20190818174433_1](https://user-images.githubusercontent.com/6338427/66259045-ed1cf580-e7ac-11e9-80fc-e172e67a84ec.jpg)
![20190818174534_1](https://user-images.githubusercontent.com/6338427/66259046-ed1cf580-e7ac-11e9-86ed-935a77708aba.jpg)
![20190928200535_1](https://user-images.githubusercontent.com/6338427/66259047-edb58c00-e7ac-11e9-8abc-2d1a2568e203.jpg)
![20190928200640_1](https://user-images.githubusercontent.com/6338427/66259048-edb58c00-e7ac-11e9-9842-dbeb69c52f02.jpg)

### Software information
Name : World of Final Fantasy
Setting doesn't have influence on the result, only full screen is required in order to launch the game.

### System information
- GPU: NVIDIA GeForce GTX 1080 Ti
- Driver: nvidia 435.21
- Wine version: 4.17 & Proton 4.11-6
- DXVK version: 1.4.2

### Apitrace file(s)
No result, the game crashes without DXVK enabled : 
wine apitrace.exe trace -a dxgi /run/media/jeux/WindowsSteamLibrary/steamapps/common/WOFF/WOFF.exe
apitrace: loaded into Z:\run\media\jeux\WindowsSteamLibrary\steamapps\common\WOFF\WOFF.exe
0031:fixme:kernelbase:AppPolicyGetProcessTerminationMethod FFFFFFFFFFFFFFFA, 000000000022FCD0
002f:fixme:kernelbase:AppPolicyGetProcessTerminationMethod FFFFFFFFFFFFFFFA, 000000000022FCD0

### Log files
- d3d11.log: 
[WOFF_d3d11.log](https://github.com/doitsujin/dxvk/files/3693592/WOFF_d3d11.log)
- dxgi.log: 
[WOFF_dxgi.log](https://github.com/doitsujin/dxvk/files/3693593/WOFF_dxgi.log)
- Proton log : 
[steam-552700.log](https://github.com/doitsujin/dxvk/files/3693597/steam-552700.log)

doitsujin:
Yes, this is a known game bug and cannot easily be fixed inside DXVK.

kust2708:
[...]

Issue #4480, opened 2024-11-20
[d3d11] Stalker Clear Sky dx10 mode msaa bugged


Blisto91:
In Stalker Clear Sky using the dx10 mode (which is actually d3d11 FL10_1) if you turn on Anti Aliasing (msaa) the main menu, and ingame i gather, can become fully garbled up. 

<details>
  <summary>Screenshot</summary>
  
![image](https://github.com/user-attachments/assets/2cfb47bd-90c2-4aae-ac64-4d26a213fe6a)
</details>

### Software information
[S.T.A.L.K.E.R.: Clear Sky](https://store.steampowered.com/app/20510/STALKER_Clear_Sky/)
dx10 mode with AA turned on

### System information
- GPU: RTX 4080
- Driver: 565.57.01
- Wine version: Proton Bleeding Edge
- DXVK version: Tested 1.0 to master

### Apitrace file(s)
https://drive.proton.me/urls/V0S952F3R8#JqRQnM7x2dpA
Reproduces the issue

### Log files
[xrEngine_d3d11.log](https://github.com/user-attachments/files/17835129/xrEngine_d3d11.log)
[xrEngine_dxgi.log](https://github.com/user-attachments/files/17835133/xrEngine_dxgi.log)

Blisto91:
I gather from @doitsujin that this is actually a game bug that doesn't have quite as drastic outcome on AMD. Disabling msaa entirely for the game is likely not wanted because it might sorta kinda work on AMD. Maybe kinda perhaps.

doitsujin:
```
VUID-RuntimeSpirv-samples-08725(ERROR / SPEC): msgNum: -1917995058 - Validation Error: [ VUID-RuntimeSpirv-samples-08725 ] Object 0: handle = 0xfe711100000001df, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; | MessageID = 0x8dadb7ce | vkCmdDrawIndexed():  the descriptor VkDescriptorSet 0xfe711100000001df[] [Set 0, Binding 2, Index 0, variable "t1"] has image created with VK_SAMPLE_COUNT_4_BIT. The Vulkan spec states: If an OpTypeImage has an MS operand 0, its bound image must have been created with VkImageCreateInfo::samples as VK_SAMPLE_COUNT_1_BIT (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-RuntimeSpirv-samples-08725)
```
[...]

PR #4731, opened 2025-03-04
Work around games sampling MSAA textures

doitsujin:
Huge pain in the arse, but this should fix striped patterns on Nvidia in a bunch of games that fuck up their MSAA usage.

TL;DR every Ubisoft game ever (and a bunch of others) try to bind a multisampled image as a plain `Texture2D` instead of either resolving or using `Texture2DMS` in the shader. This is light years deep into undefined behaviour territory, but happens to work on *some* native drivers. On AMD, this has trivially worked before with DXVK as well because the hardware samples from the first plane, but it's invalid Vulkan either way and made MSAA completely ineffective in a number of games that essentially draw the unresolved image to the swapchain (think World of Final Fantasy, #1216).

Instead, we can do the right thing and just do an implicit `SAMPLE_ZERO` resolve, which roughly matches what happens on AMD hardware anyway.


Blisto91:
Verified that it fixes https://github.com/doitsujin/dxvk/issues/1885 and https://github.com/doitsujin/dxvk/issues/4480

PR #4741, opened 2025-03-06
Don't override app-provided barrier control flags

doitsujin:
Needs a bit of Nvidia perf testing in the handful of games that we enable this for, but TL;DR if app uses UAV Overlap explicitly, don't try to outsmart it. Our implementation should match Windows behaviour now.


doitsujin:
Hrmpf, RE2 uses UAV overlap *extensively* on its compute passes, but not on graphics, so we'd end up with a lot of extra synchronization when it does occlusion culling. Might need to change how these config options work.

doitsujin:
God of War also has problems in some areas where it stores to a bunch of images from PS, but then doesn't use uav overlap even though it would actually be safe to do so. Starting to think that this is a bad idea.

PR #4743, opened 2025-03-07
[d3d8/9] Some static analysis-based rework and corrections

WinterSnowfall:
I've thrown Cppcheck at the d3d8 and d3d9 folders and basically this has come out of it. Some stuff is potentially scary and everything definitely needs a second pair of eyes. Happy to clarify what the problem was on each change.

Currently WIP.

### DIFF
--- a/src/dxso/dxso_compiler.cpp
+++ b/src/dxso/dxso_compiler.cpp
@@ -1163,8 +1163,7 @@ namespace dxvk {
             return m_vs.oPSize;
 
           default: {
-            DxsoRegisterPointer nullPointer;
-            nullPointer.id = 0;
+            DxsoRegisterPointer nullPointer = { };
             return nullPointer;
           }
         }
@@ -1274,8 +1273,7 @@ namespace dxvk {
       default: {
         //Logger::warn(str::format("emitGetOperandPtr: unhandled reg type: ", reg.id.type));
 
-        DxsoRegisterPointer nullPointer;
-        nullPointer.id = 0;
+        DxsoRegisterPointer nullPointer = { };
         return nullPointer;
       }
     }
@@ -2702,7 +2700,7 @@ void DxsoCompiler::emitControlFlowGenericLoop(
     const DxsoOpcode opcode = ctx.instruction.opcode;
 
     DxsoRegisterValue texcoordVar;
-    uint32_t samplerIdx;
+    uint32_t samplerIdx = 0u;
 
     DxsoRegMask vec3Mask(true, true, true,  false);
     DxsoRegMask srcMask (true, true, true,  true);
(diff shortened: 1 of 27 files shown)


WinterSnowfall:
I also threw in the dxso folder and didn't very much like what I found... not entirely sure I've used proper defaults in all cases, but some of the uninitialized variables had large potential to explode and were being flagged as outright errors by Cppcheck.

I'll leave it up to you to review and decide if we should merge this sooner rather and later. Ideally would also need a bit of testing, though none of this should change any behavior for the worse.

src/dxso/dxso_compiler.cpp
```diff
@@ -23,7 +23,11 @@ namespace dxvk {
     , m_programInfo( programInfo )
     , m_analysis   ( &analysis )
     , m_layout     ( &layout )
-    , m_module     ( spvVersion(1, 3) ) {
+    , m_module     ( spvVersion(1, 3) )
+    , m_rRegs      ( )
+    , m_pRegs      { }
+    , m_tRegs      { }
+    , m_samplers   { } {
```
doitsujin, line 30: nit, if we default-initialize things anyway then we can just do that in the actual declaration as well (`array<frog> frogs = { };` etc)

WinterSnowfall: Sure, will do. Regarding the defaults I've used (ouside of 0-ing memes), does everything look fair? :sweat_smile:

WinterSnowfall:
Gave this a go too on top of current master, with about 30-ish d3d8/9 games or so. Seems fine.

src/util/util_bit.h
```diff
@@ -690,9 +690,9 @@ namespace dxvk::bit {
     explicit uint48_t(uint64_t n)
     : a(uint16_t(n)), b(uint16_t(n >> 16)), c(uint16_t(n >> 32)) { }
 
-    uint16_t a;
-    uint16_t b;
-    uint16_t c;
+    uint16_t a = 0;
+    uint16_t b = 0;
+    uint16_t c = 0;
```
doitsujin, line 695: is there a reason for this? this struct is somewhat intended to be uninitialized since it's just a numeric type, and the barrier tracking code must manually initialize everything anyway. I'm worried that this pessimizes generated code.

src/dxso/dxso_compiler.cpp
doitsujin: should be.

src/util/util_bit.h
WinterSnowfall: Static analysis complains because there's a default constructor which will leave these uninitialized in theory. Haven't checked if it's actually used to be fair.

doitsujin: Might not be since the only place where this is actually used is `DxvkAddressRange`, but again, it's just a numeric type and those generally aren't default-initialized by default, so I don't really think this needs to be.

WinterSnowfall: Alright, I'll remove it then and leave these as they were.

PR #4744, opened 2025-03-07
[d3d9] Remove double presents with SyncInterval > 1

doitsujin:
Seems to work as-is, at least when forcing `d3d9.presentInterval=2` via config.


K0bin:
I guess `m_wctx->presenter->setSyncInterval(presentInterval);` this just handles it?

K0bin:
The PR is fine of course but doesn't doing it via the frame rate limiter introduce potential frame pacing issues compared to proper vsync?

doitsujin:
In theory there's a risk that the compositor latches at uneven intervals, yes, but spamming back-to-back presents also has real issues that we have far less control over (exhausting images unless we allocate extra, present stalling longer than it should, etc).

This isn't exactly a *common* scenario anyway, but I'd argue that our d3d11/12 experience with doing the exact same thing has been decent enough.

K0bin (approved):

PR #4745, opened 2025-03-08
[d3d9] Build shader constant UBOs on CS thread

K0bin:
As discussed on Discord, this moves the task of copying the constants into a bump allocated UBO for each draw onto the CS thread. Hopefully this will reduce the work on the game thread as the CS thread is usually the less busy one. Not a lot of games manage to saturate it.

So far I'm skeptical whether it provides any actual performance gains. Either way, this needs to be tested.


K0bin:
Without the PR:
image

With the PR:
image

When I disable flushing in SetShaderConstant:
image

PR #4757, opened 2025-03-10
Rework depth-stencil resolves

doitsujin:
Small cleanup, hence the negative delta. Also fixes an oversight where we accidentally broke partial resolves for certain D3D9 StretchRect cases.

Need to verify that this doesn't break anything else.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -909,7 +909,7 @@ namespace dxvk {
           cDstImage->mipLevelExtent(cDstLayers.mipLevel));
       });
     } else {
-      const VkFormat format = m_parent->LookupFormat(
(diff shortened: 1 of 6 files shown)


src/d3d11/d3d11_context.cpp
```diff
@@ -909,7 +909,7 @@ namespace dxvk {
           cDstImage->mipLevelExtent(cDstLayers.mipLevel));
       });
     } else {
-      const VkFormat format = m_parent->LookupFormat(
```
K0bin, line 912: This seems like a pointless change.

doitsujin: ye, fair, it only happened because I had put the code to use the src image format here first.

Not like it really changes anything.

PR #4763, opened 2025-03-13
[dxvk] Disable sparse zero buffer on pre-RDNA AMD hardware

doitsujin:
So yeah, apparently this hangs Renoir (Vega) and potentially some other (old) AMD GPUs that nobody really tests anymore as well.

Gonna file a mesa issue and see where that goes, will merge the workaround if no quick fix can be found.

### DIFF
--- a/src/dxvk/dxvk_context.cpp
+++ b/src/dxvk/dxvk_context.cpp
@@ -7671,10 +7671,16 @@ namespace dxvk {
     bufInfo.debugName = "Zero buffer";
 
     // If supported by the device, create a large sparse buffer and keep it unmapped
-    // in order to avoid having to allocate and clear any actual memory
+    // in order to avoid having to allocate and clear any actual memory. Some specific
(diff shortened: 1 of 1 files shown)


mrdeathjr28:
Many thanks lego videogame hang/freeze are fixed

dxvk26fx

dilworks:
Ha, this PR merge narrowly missed the 2.6 release, which was causing the exact same crash described  by @mrdeathjr28 but on very different games (a couple of old Japanese stuff like Recettear and Fatal Zero Action). In fact this is the very second Google hit for "The CS has been cancelled because the context is lost. This context is innocent." and "dxvk"!

Found the relevant artifact build, installed it, GPU crashes were gone. And I thought ATi driver hell was a thing of the past, but guess not :( As an owner of a Ryzen 7 5700U with Vega graphics (still for sale in 2025!), thanks for the workaround~!

Issue #4766, opened 2025-03-14
Manhunt 2 Doesn't Run


OGPoetoe:
Hello, I have trouble with Manhunt 2. While using dxvk 2,6, it doesn't run while previous Version 2.5.3, it works normally.

This is Manhunt 2 Log.
[Manhunt2_d3d9.log](https://github.com/user-attachments/files/19243654/Manhunt2_d3d9.log)

Blisto91:
Hi there. 
Are you able to get a apitrace of the game? Just follow the instructions here https://github.com/doitsujin/dxvk/wiki/Using-Apitrace and get past the part that usually crashes.

OGPoetoe:
> Hi there. Are you able to get a apitrace of the game? Just follow the instructions here https://github.com/doitsujin/dxvk/wiki/Using-Apitrace and get past the part that usually crashes with dxvk.

here you go : https://drive.google.com/file/d/12RWptIOIsO9XGQKPQDgjia2lzTLMb7nF/view?usp=sharing

WinterSnowfall:
> here you go : https://drive.google.com/file/d/12RWptIOIsO9XGQKPQDgjia2lzTLMb7nF/view?usp=sharing

Thank you for the trace, unfortunately it doesn't capture a whole lot, so won't help much in debugging the issue. I assume it was captured using dxvk? If so, can you please try to capture one without dxvk (so native Windows or WineD3D, depending on which platform you are on)?

OGPoetoe:
wait i try it.

> > i capture it without dxvk. wait i try again
> 
> In that case please:
> 
>     * capture one without dxvk, but until the game gets to the main menu at least
> 
>     * capture one with dxvk up until the point of the crash
> 
> 
> And share both here.


okay.

this is for non-dxvk : https://drive.google.com/file/d/1-lrxcxI9__bsp7vt1PQlc_sJ4sDMaWEV/view?usp=sharing

i try with dxvk but still doesn't launch because it conflicted between d3d9's dxvk and d3d9's apitrace.

WinterSnowfall:
> i try with dxvk but still doesn't launch because it conflicted between d3d9's dxvk and d3d9's apitrace.

You can leave only dxvk's d3d9.dll next to the game's executable and run: `C:\Path_To_Apitrace_Folder\bin\apitrace.exe trace -a d3d9 Game.exe` from the game folder.

OGPoetoe:
[...]

PR #4768, opened 2025-03-14
[d3d8/9] Validate D3DDEVTYPE on device creation and cap queries

WinterSnowfall:
Fixes some device type madness observed in #4766 , but probably not the main cause of concern in that issue.

Drafted until @Blisto91 helps validate the behavior/return codes against modern Windows, current implemented behavior is "the Windows XP way".

~P.S.: Also noticed Windows consistently reports 5 MB shy of 4096 MB when it comes to available texture memory by default, so let's do the same.~

### DIFF
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -292,8 +292,15 @@ namespace dxvk {
           D3DCAPS9*  pCaps) {
     using namespace dxvk::caps;
 
-    if (pCaps == nullptr)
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
Turns out modern native d3d9 only derps out for D3DDEVTYPE_SW. As for the reported memory size, that turned out to be a bit superfluous, as we're already accounting for 32-bit shenanigans and Windows drivers seem to report about the same too (very close to 4096).

PR #4779, opened 2025-03-17
Add workaround for ReShade misinterpreting swapchain contents

doitsujin:
Sigh, it would be real nice if we could actually use Vulkan features without everything blowing up.


doitsujin:
Decided that we shouldn't be working around this since being fixed upstream anyway.

PR #4789, opened 2025-03-22
Refactor and normalize render state

doitsujin:
Compresses some render state structures in a similar way to what the pipeline state thing already does, and changes all the D3D11 state binding methods to pass a copy of the full state rather than ref-counting the D3D11 state object itself. This is now much more feasible when things like depth-stencil state are now 10 bytes rather than 72.

Additionally, this normalizes depth-stencil and blend state to the extent possible in the frontends. The backend already does this to an extent at pipeline compile time, but sanitizing some state can be useful to not create extra pipeline states in the first place. 

This also benefits #4785 since we can now more easily prove that e.g. stencil isn't written, and also changes blend states that essentially disable writing to a given render target to use a write mask of 0.

Needs extensive regression testing for both D3D11 and D3D9.

### DIFF
--- a/src/d3d11/d3d11_blend.cpp
+++ b/src/d3d11/d3d11_blend.cpp
@@ -27,8 +27,8 @@ namespace dxvk {
     if (desc.IndependentBlendEnable && desc.RenderTarget[0].LogicOpEnable)
       Logger::warn("D3D11: Per-target logic ops not supported");
     
-    m_loState.enableLogicOp         = desc.RenderTarget[0].LogicOpEnable;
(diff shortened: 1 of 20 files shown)


misyltoad (approved):
DX9 changes look fine

K0bin (approved):

doitsujin:
> or are not accidentally overflowing things

I'm not too worried about *that* aspect since we're using the same bit sizes for everything in the pipeline state structs already. What I am scared of is dumb regressions though.

PR #4793, opened 2025-03-23
[dxbc] Always export point size on ANV

doitsujin:
Sigh.


llandwerlin-intel:
And yet it passes the CTS. So maybe we're missing coverage? Any way to reproduce?

llandwerlin-intel:
Also, the value is undefined if the value is written in a VS shader but then not written further down the pipeline (like geometry to tesselation for example)

PR #4794, opened 2025-03-23
[d3d9]  Modify StretchRect to not resolve if src and dst are at the same sample count.

TheIronWolfModding:
I was investigating the issue where one MSAA depth buffer would not get copied to another MSAA buffer.  It turned out that `StretchRect` does resolve even if src and dst sample counts are identical.   I then noticed that comment right above does not match what code is doing, so here's the fix.

Note, I have not tested this with 2.6 (I am still on 2.4.1) but I have not noticed any issues with this fix and no new validation errors surfaced in my tests.


K0bin:
We'll definitely squash those if we decide to merge it.  :laughing:

doitsujin:
Change looks simple enough but holy mother of god, what even is that commit history?

doitsujin:
Merged manually because commit history is still thoroughly fucked.

In general we avoid merge commits like the plague for this reason, it's easier to just maintain small changes and rebase on top of master if necessary.

PR #4795, opened 2025-03-24
Require maintenance4 and maintenance5

doitsujin:
Currently based on both #4785 and #4789 to avoid rebase hell later on.

This is more of a thing for 2.7 anyway since I don't want to break certain drivers in a point release, but now that RenderDoc supports `maintenance5`, we should just require it given that a bunch of our code already sort of relies on the feature being present anyway. This allows for some substantial code cleanups.


artemyto:
Currently maintenance5 is listed as optional in https://github.com/doitsujin/dxvk/blob/master/VP_DXVK_requirements.json

It would be great to change it to requirement in the profile in this PR to help driver developers and regular users who will refer to the profile in the future.

Digger1955:
Can you also check support for VK_EXT_load_store_op_none? 

This extension is a complete duplicate of VK_KHR_load_store_op_none according to 

https://registry.khronos.org/vulkan/specs/latest/man/html/VK_EXT_load_store_op_none.html 
and 
https://registry.khronos.org/vulkan/specs/latest/man/html/VK_KHR_load_store_op_none.html 

but this extension is supported by both newer and older drivers. 

For example, on Windows, AMD Vulkan Driver for Polaris and Vega supports ONLY VK_EXT_load_store_op_none (since these GPUs have been moved to the Legacy category), while Windows AMD Vulkan Driver for RDNA supports both VK_EXT_load_store_op_none and VK_KHR_load_store_op_none.

PR #4798, opened 2025-03-24
[util] Enable cachedDynamicBuffers for Lego Indiana Jones: The Original Adventures

SuperSamus:
I didn't make an API trace to investigate the game, I just randomly tried the option and saw the improvements. The Extras menu in particular has a 10000% performance improvement.
The game is downloaded from Amazon/Twitch Prime (Heroic).

## Before
Gameplay
Screenshot_20250324_180036
Pause menu > Extras
Screenshot_20250324_180050

## After
Gameplay
Screenshot_20250324_175803
Pause menu > Extras
Screenshot_20250324_175820

I kinda wonder if this is the only game that has this problem, since there are *many* games using the same engine: https://www.pcgamingwiki.com/wiki/Engine:Nu2 [^1]
I do have other LEGO games that I can test later, though.

[^1]: Ordering by release date, the first game supporting D3D11 is Lego Batman 3 (2014!), while the first to drop D3D9 support was Lego City Undercover (2017!?).

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1043,6 +1043,11 @@ namespace dxvk {
     { R"(\\sh2pc\.exe$)", {{
       { "d3d9.extraFrontbuffer",            "True" },
     }} },
+    /* Lego Indiana Jones: The Original Adventures *
+     * Fix UI performance                          */
+    { R"(\\LEGOIndy\.exe$)", {{
+      { "d3d9.cachedDynamicBuffers",        "True" },
+    }} },
 
     /**********************************************/
     /* D3D8 GAMES                                 */


K0bin:
> I kinda wonder if this is the only game that has this problem, since there are many games using the same engine

I've played a bunch of Lego Star Wars on the Steam Deck and that didn't have the problem at least.

qinlili23333:
I don't remember that I got such poor performance in menu when using dxvk with this game on Nvidia or Intel. 
Suggest to investigate a bit further before merging. I can do an apitrace later.
And maybe we also need to check other versions from GOG since pcgw says GOG version has some additional stability fix.

SuperSamus:
> I have made MD5 of my installation here

The .exe hash matches.

> Apitrace on Nvidia with Prime version

The Apitrace runs with no issues.
(BTW, an even faster way to reproduce it is in the control remapping menu: it runs at single digit frame rates!)

> testing over 60fps is not suggested

I was leaving uncapped for benchmarking purposes.
(Also, 60 FPS is way too conservative: I completed all story levels at 144 FPS with no particular issues.)

K0bin:
I think I can reproduce the issue but nowhere near as bad as it is for you. I tried the very first level and it runs at 600 FPS at 4k. If I open the pause menu, it stays at 600 FPS but when I then open the options menu in particular, it drops to 300 FPS.

K0bin (approved):

qinlili23333:
> I completed all story levels at 144 FPS with no particular issues

Thanks for the information. I'll try high fps later and update content on pcgw if it works as expected.

PR #4811, opened 2025-03-30
[util] Fix broken LZCNT&TZCNT on non-BMI platforms

### LINKED ISSUE #4808: `bit::lzcnt` does not work as intended on MSVC when running on a CPU without BMI1 instruction set
I have a DirectX game I build with MSVC, and I recently added support for Vulkan similar to how Valve did it in Source Engine by directly compiling DXVK sources into the binaries and not linking DirectX SDK libs. The problem is that one of my players reported a crash on startup somewhere in memory allocator, which I quickly traced to `bit::lzcnt`.
While trying to allocate 4096 bytes of memory, a `DxvkPoolAllocator::computeListIndex` call is made, which calls into `bit::lzcnt` and then `_lzcnt_u32`. On any CPU made in the last ~15 years, this call with 4096 as argument returns 3. For this user, it returns 4294967290 instead. Under MSVC, it simply calls the instruction directly, without checking if it's supported by the CPU or not. The player's CPU is i5 750, and it does not support that instruction. Release DLLs from this repository do not crash for them, since they are not built with MSVC and I assume either `__builtin_clz` is backwards-compatible with the old CPUs, or your compiler falls back to your own implementation of this instruction.
I didn't look into it much, but I think re-implementing the instruction using `_BitScanReverse` should fix the issue.

qinlili23333:
Now MSVC will always use BSR for LZCNT.
Also correct 0 input behavior for TZCNT when fallback to BSF.
Should fix #4808 .
Need test before merge.

### DIFF
--- a/src/util/util_bit.h
+++ b/src/util/util_bit.h
@@ -160,7 +160,7 @@ namespace dxvk::bit {
   }
 
   inline uint32_t lzcnt(uint32_t n) {
-    #if ((defined(_MSC_VER) && !defined(__clang__))) && !defined(__LZCNT__)
+    #if defined(_MSC_VER) && !defined(__clang__) && !defined(__LZCNT__)
     unsigned long bsr;
     if(n == 0)
       return 32;
@@ -186,7 +186,7 @@ namespace dxvk::bit {
   }
 
   inline uint32_t lzcnt(uint64_t n) {
-    #if ((defined(_MSC_VER) && !defined(__clang__))) && !defined(__LZCNT__) && defined(DXVK_ARCH_X86_64)
+    #if defined(_MSC_VER) && !defined(__clang__) && !defined(__LZCNT__) && defined(DXVK_ARCH_X86_64)
     unsigned long bsr;
     if(n == 0)
       return 64;
--- a/src/util/util_bit.h
+++ b/src/util/util_bit.h
@@ -178,8 +178,15 @@ namespace dxvk::bit {
   inline uint32_t lzcnt(uint64_t n) {
     #if defined(_MSC_VER) && !defined(__AVX2__)
     unsigned long bsr;
-    _BitScanReverse(&bsr, n);
-    return 31-bsr;
+    #if defined(DXVK_ARCH_X86_64)
+    _BitScanReverse64(&bsr, n);
+    #else
+    if  (_BitScanReverse(&bsr, n >> 32))
+        bsr += 32;
+    else
+        _BitScanReverse(&bsr, n & 0xffffffff);
+    #endif
+    return 63-bsr;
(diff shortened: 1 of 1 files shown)


src/util/util_bit.h
```diff
@@ -156,7 +156,11 @@ namespace dxvk::bit {
   }
 
   inline uint32_t lzcnt(uint32_t n) {
-    #if (defined(_MSC_VER) && !defined(__clang__)) || defined(__LZCNT__)
+    #if defined(_MSC_VER) && !defined(__AVX2__)
```
doitsujin, line 159: why AVX2?

src/util/util_bit.h
```diff
@@ -156,7 +156,11 @@ namespace dxvk::bit {
...
+    unsigned long bsr;
+    _BitScanReverse(&bsr, n);
```
doitsujin, line 161: This is undefined if n is 0, need to check the return value too.

src/util/util_bit.h
```diff
@@ -178,8 +182,15 @@ namespace dxvk::bit {
   inline uint32_t lzcnt(uint64_t n) {
     #if defined(_MSC_VER) && !defined(__AVX2__)
     unsigned long bsr;
-    _BitScanReverse(&bsr, n);
-    return 31-bsr;
+    #if defined(DXVK_ARCH_X86_64)
+    _BitScanReverse64(&bsr, n);
+    #else
```
doitsujin, line 187: We already have a fallback path below that uses the 32-bit implementation, no need to duplicate that here.

doitsujin (changes requested):

src/util/util_bit.h
qinlili23333: Bases on MSVC document, MSVC only predefine AVXs, there is no predefine for BMI or LZCNT. That's why I have to use AVX2 predefine to check it as all AVX2 processors support LZCNT. https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

doitsujin: The problem I have with this is that building DXVK with AVX2 is inherently unsafe, and locking lzcnt behind it will lock out non-msvc out of using the fast path too even when BMI and/or SSE4a (not 4.1) are enabled.

qinlili23333: Fixed, now MSVC will always use BSR, until they add support for LZCNT flags.
I have also raised a ticket for this in Microsoft community: https://developercommunity.visualstudio.com/t/Lacks-LZCNT-predefined-flag-and-LZCNT-wi/10882435

qinlili23333:
TZCNT also got fixed now.

Issue #2779, opened 2022-08-02
[D3D9] Implement Clamp Negative LOD-Bias


ViNi-Arco:
Hello, I was always afraid to ask if it would be possible to implement a way to Clamp the Lod, because I thought it would be something impossible to implement in DXVK, but this recent [commit](https://github.com/doitsujin/dxvk/commit/727fd7ac33ca8ba09f166bce04c0cbb77601fe98) from Mr. Philip showed me that it might be possible.

**Clamp Negative LOD-Bias** (which would prevent using negative LOD values for texture filtering and replace them with 0.0)

It would help a lot to fix the texture shimmering in old games.

Here are images and mentions of what I mean if it was not very clear, sorry for my bad English.
https://www.bsimracing.com/nvidia-image-quality-lower-on-newer-cards/
https://linustechtips.com/topic/499767-low-image-quality-on-newer-gpus/page/2/#comment-6675744

doitsujin:
> On the commit, I don't get why it uses float instead of an integer value?

Because the LOD bias is not an integer property, it's a float property, and values like -0.5 are rather common.

The motivation for that particular commit was that Nioh 2 does **not** use a negative LOD bias when DLSS enabled, which leads to blurry textures.

> does ***not*** mean it prevents games from having no textures at all (in case of using a high positive offset).

It prevents you from putting a very high number into a config file to do things like [this](https://www.youtube.com/watch?v=eixkJ_7VbQE). The option does not affect how any game works by default. The clamp is only for the config option, not for the value that the sampler is created with.

lextra2:
> 
> Because the LOD bias is not an integer property, it's a float property, and values like -0.5 are rather common.

Yeah. I expected this answer. Actually, I mistook MeshLODbias for MipMapLODbias. My bad.

> 
> It prevents you from putting a very high number into a config file and do things like [this](https://www.youtube.com/watch?v=eixkJ_7VbQE). The option does not affect how any game works by default. The clamp is only for the config option, not for the value that the sampler is created with.

And yes. That is exactly what I'm talking about. IMHO a blacklist for games that ban for this would be a better approach. (Or at least a way to overwrite the calmping limit "at your own risk" kind of thing)

doitsujin:
> And yes. That is exactly what I'm talking about. IMHO a blacklist for games that ban for this would be a better approach (Or at least a way to overwrite the clamping limit "at your own risk" kind of thing)

[...]

PR #4812, opened 2025-03-31
Update d3d9_options.cpp

babamonkey222:
dd


qinlili23333 (changes requested):
See https://github.com/doitsujin/dxvk/issues/2779#issuecomment-1202786896
This clamp is intentional to avoid imbalanced advantage in some games.

PR #4824, opened 2025-04-03
[d3d9] Add an option to allow biased correctness factor to fix some buggy games

### LINKED ISSUE #4618: Borderlands 2 / Pre-sequel: Horizontal Scan Lines with DXVK 2.5.2 and 2.5.3 when AO (Ambient Occlusion) enabled
Borderlands2 2025-01-16 12-42-59
Please describe your issue as accurately as possible.

Dear DXVK DEV! I want to report a problem which is also in Borderlands 2 without DXVK a general Problem: When using WQHD Resolution, in Borderlands 2 you get these Horizontal Scan lines graphic-issue. Just goes away when disabling AO. With 1920x1080 resolution this bug doesn't exist.

BUT: With DXVK 2.5.1 everything looks fine and there is this bug NOT exisitng! When i tried using 2.5.2 or 2.5.3 i get this issue.
Problem is with 2.5.1 my transparent twitch chat overlay does not work... and there is issue like missing lava which was fixed with 2.5.2.

Can you please look into it? Would be cool if these Horizontal Scan lines would be gone with the newest versions like in 2.5.1.

If you use Windows, please check the following page: https://github.com/doitsujin/dxvk/wiki/Windows

### Software information
Name of the game, settings used etc.

### System information
- GPU: RTX 2070 Super
- Driver: Latest
- Wine version: none / windows 11 Home
- DXVK version: 2.5.3 and 2.5.2

### Apitrace file(s)
- Put a link here

For instructions on how to use apitrace, see: https://github.com/doitsujin/dxvk/wiki/Using-Apitrace

### Log files
Please attach Proton or Wine logs as a text file:
- When using Proton, set the Steam launch options for your game to `PROTON_LOG=1 %command%` and attach the corresponding `steam-xxxxx.log` file in your home directory.
- When using regular Wine, use `wine game.exe > game.log 2>&1` and attach the resulting `game.log` file.

qinlili23333:
Fix #4618 
But should we add such an option just for a few buggy games? (Currently only Borderlands 2&TPS and FFXIII known to be buggy)
And I'm not sure whether remove the constexpr to make it configurable will result observable performance issue. (I could not see any performance issue during my own test since I didn't reach CPU bottleneck)

Put in draft here first. Needs some more discussion.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -773,6 +773,16 @@
 
 # d3d9.extraFrontbuffer = False
 
+# Bias correctness factor a bit
+#
+# Some buggy games may have horizontal scan lines on specified resolutions.
+# Enable this will bias correctness factor a bit to fix these scan lines.
+# This is only a workaround for some buggy games.
+#
+# Supported values: True, False
+
+# d3d9.correctnessBias = False
+
 # Dref scaling for DXS0/FVF
 #
 # Some early D3D8 games expect Dref (depth texcoord Z) to be on the range of
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -6741,8 +6741,10 @@ namespace dxvk {
     // corner so we can get away with flipping the viewport.
     const D3DVIEWPORT9& vp = m_state.viewport;
 
+    
     // Correctness Factor for 1/2 texel offset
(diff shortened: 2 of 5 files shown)


doitsujin:
Didn't we have this discussion before, with the conclusion that we're *already* doing the correct thing since it does the same thing on Windows?

PR #4825, opened 2025-04-03
[wsi] Handle SDL3 refresh rate not being available

flibitijibibo:
In SDL3 it's possible for the platform to refuse returning a good refresh rate, so we have to try as many options as we can. Numerator/Denominator is best, but sometimes we just get a float, and sometimes we don't even get that (I managed to repro this via Xwayland).

This ensures that if the WsiRational is ever invalid, it's almost certainly SDL's fault rather than dxvk's.


doitsujin (approved):
Is the refresh_rate float being non-zero while numerator is zero something that actually intended behaviour? I was assuming the float to be more of a convenience thing that would be there *in addition* to the rational, but this doesn't seem to be very well documented.

Either way, patch looks good.

flibitijibibo:
I believe so, and I think this is just inherited from SDL2 only having `refresh_rate` at first - taking a quick look it looks like macOS/iOS/tvOS, Android, and OpenVR use this behavior.

PR #4826, opened 2025-04-03
[wsi] For SDL3 use GetWindowSizeInPixels

flibitijibibo:
Another subtle thing in SDL3 is that "points" vs "pixels" became very separated; you could calculate this manually by using Points and getting the pixel density of the display, but this is a lot easier. (Interestingly this feature is _not_ available for the Set function, so [_some_ people](https://github.com/FNA-XNA/FNA/blob/master/src/FNAPlatform/SDL3_FNAPlatform.cs#L458) are still doing some scaling manually...)

Fixes Wayland windows getting a partial view of the swapchain instead of the full surface; X was not affected.


doitsujin:
Is this actually the correct thing to do for *windowed mode* swapchains? DXGI has no real notion of desktop vs display coordinates, and Windows seems to do scaling at a compositor level rather than giving the app a larger swapchain, but I'm not 100% sure what the exact behaviour is there.

For full-screen i guess the issue is that the window's desktop coordinates are smaller than the display mode.

doitsujin:
There's also the question of what to do with DXGI `ResizeTarget`. Not sure if that's supposed to be display or desktop coordinates if the app is high dpi aware (which, again, we don't really have a notion of, but Windows does).

flibitijibibo:
At least from SDL's point of view this _should_ be more correct; the calls line up with dxvk's win32 WSI:

https://github.com/doitsujin/dxvk/blob/master/src/wsi/win32/wsi_window_win32.cpp#L109

https://github.com/libsdl-org/SDL/blob/main/src/video/windows/SDL_windowswindow.c#L1044

SDL should probably do the left/top subtraction too... A bonus: Here's where SDL3 does the scaling for Set, which results in GetWindowSize being different on Windows too: https://github.com/libsdl-org/SDL/blob/main/src/video/windows/SDL_windowswindow.c#L264

PR #4835, opened 2025-04-07
build: Locate SDL by its case-sensitive name

smcv:
The pkg-config file describing SDL's API is `sdl2.pc` or `sdl3.pc`, so the name to look for is `sdl2` or `sdl3`. Looking for it with a different case combination will not necessarily work.

With the versions of `pkg-config` and `meson` in Steam Runtime 3 'sniper', it seems that SDL 2 can be found via either case combination, but SDL 3 requires the canonical case combination, `sdl3`.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -147,8 +147,8 @@ else
     './include/native/directx'
   ]
 
-  lib_sdl3 = dependency('SDL3', required: false)
-  lib_sdl2 = dependency('SDL2', required: false)
+  lib_sdl3 = dependency('sdl3', required: false)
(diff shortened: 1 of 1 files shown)


smcv:
I believe SDL2 already worked because Meson has a special case for it (to be able to fall back to `sdl2-config` if `sdl2.pc` is not found), and Meson looks up `dependency()` special cases with their lower-cased names; but SDL3 no longer has `sdl3-config` so there is no similar special case.

smcv:
Falling back from pkg-config to CMake might also have been case-insensitive, because CMake is generally case-insensitive.

PR #3795, opened 2024-01-12
[util] Fix e2k build

r-a-sattarov:
Ref: https://en.wikipedia.org/wiki/Elbrus_(computer)

PR #4853, opened 2025-04-16
Don't error out for otherwise unknown CPU architectures

smcv:
* [util] Fall back to busy-loop on spinlock for unknown CPU architectures
    
    If there's no known equivalent of _mm_pause() on an architecture,
    it seems to be most typical to fall back to not doing anything,
    resulting in busy-looping for a short time. Porters who don't like this
    behaviour can add an `#elif` here with their architecture's closest
    equivalent, if it has one.

* [util] Don't error out for otherwise unknown CPU architectures
    
    All of the conditionally-compiled parts seem to have a `#else` that is
    ordinary non-architecture-specific C++, except for the spinlock that
    was made generic in the previous commit, so there's no particular reason
    to forbid compiling for other CPUs.

---

The bug report that I believe this resolves:

Debian unstable currently has builds of DXVK as PE DLLs (only) for use with Wine, on the amd64, arm64, i386 and armhf architectures (so, 64- and 32-bit x86 and ARM, which matches the architectures for which Debian compiles Wine 10).

After upgrading to 2.6.1 and starting to build DXVK Native (currently only in experimental as a result of the Debian 13 freeze), I noticed that it no longer builds successfully on armhf: we get the `#error`. In principle it ought to be possible to compile DXVK Native for any CPU architecture, as far as I can see. I will admit that I haven't actually tested this on anything non-x86, hence the draft status (if necessary I can get an armhf environment going on a Raspberry Pi at some point).

While investigating this I noticed that the CPU detection is also pretending that Elbrus E2K is a kind of x86, which seems odd and quite possibly wrong, but for the moment I've assumed that #3795 was (sufficiently close to) correct, rather than sending it into the "generic/unknown CPU" code path.

### DIFF
--- a/src/util/sync/sync_spinlock.h
+++ b/src/util/sync/sync_spinlock.h
@@ -26,7 +26,8 @@ namespace dxvk::sync {
         #elif defined(DXVK_ARCH_ARM64)
         __asm__ __volatile__ ("yield");
         #else
-        #error "Pause/Yield not implemented for this architecture."
+        /* Do nothing (busy-loop). Please add more #elif above here if
+         * your CPU architecture has a suitable pause/yield instruction */
         #endif
         if (fn())
           return;
--- a/src/util/util_bit.h
+++ b/src/util/util_bit.h
@@ -8,8 +8,6 @@
   #endif
 #elif defined(__aarch64__) || defined(_M_ARM64) || defined(_M_ARM64EC)
   #define DXVK_ARCH_ARM64
-#else
-#error "Unknown CPU Architecture"
 #endif
 
 #ifdef DXVK_ARCH_X86


smcv:
This does at least compile successfully on Debian armhf (32-bit ARMv7 hardfloat, in a chroot on a Raspberry Pi). I wasn't able to test it further, because the RPi's Vulkan drivers don't meet DXVK's requirements.

WinterSnowfall:
> While investigating this I noticed that the CPU detection is also pretending that Elbrus E2K is a kind of x86, which seems odd and quite possibly wrong

If I understood/remember things correctly, and do take this with a grain of salt, that particular CPU architecture supports two instruction sets, and one of them is indeed x86.

smcv:
> When looking up ARM official document, I found that `void __yield(void)` should also be available on armhf.

If you can confirm this, you could open a tested pull request that might look something like this untested diff:

```diff
+#ifdef __ARM_ACLE
+#include <arm_acle.h>
+#endif

 ...

         #elif defined(DXVK_ARCH_ARM64)
         __asm__ __volatile__ ("yield");
+        #elif defined(__ARM_ACLE)
+        __yield();
          #else
         /* Do nothing (busy-loop). Please add more #elif above here if
          * your CPU architecture has a suitable pause/yield instruction */
         #endif
 ...
```

But I'm not going to open that PR myself, because my interest in armhf is limited to "it compiled successfully in the past, so if possible it should still compile in the future".

I haven't tested the diff above on any specific compiler - I believe it's what is suggested by the PDF document you linked, but I don't know whether commonly-used versions of gcc, clang or MSVC actually implement what's described in that document.

>  I think maybe it's possible to find some way to make 32bit arm fully compatible

I believe the changes in this PR should be enough to make DXVK compile and run successfully on 32-bit ARM, but with the limitation that its multi-threading performance might not necessarily be optimal. If users of 32-bit ARM systems want to improve its performance on 32-bit ARM, please do, but I'm not intending to work on that myself.

qinlili23333:
> I haven't tested the diff above on any specific compiler - I believe it's what is suggested by the PDF document you linked, but I don't know whether commonly-used versions of gcc, clang or MSVC actually implement what's described in that document.

I'll put it on todo list and try to figure out it once I can unbrick my arm device.

PR #4854, opened 2025-04-17
Fix bug: UAV image format has four channels.

water-chika:


doitsujin:
In which universe is this a bug? Atomics can only be done on R32 formats.

PR #4869, opened 2025-04-24
[d3d9] Implement hideVendorGPU logic for d3d9

### LINKED ISSUE #4860: [Request] Add d3d9.hideIntelGpu option and enable it by default if possible
Some d3d9/8 games turn off certain features or have unnecessary behavior when they see Intel vendor.

Some example games I tested on Intel Arc B580 are:
- Need for Speed: Most Wanted (2005) - MSAA unavailable
- Source Engine games - High shadow detail unavailable (that will require high amount of profiles)
- A Hat in Time - iGPU warning popup
- and maybe more

Faking AMD vendor by using `DXVK_CONFIG=d3d9.customVendorId=1002 %command%`  works around it, but a friendlier option like `d3d9.hideIntelGpu` would be better, and forcing VendorId can override profiles too if used globally.
But in my opinion, if it's possible, then enabling it by default will maybe not hurt, but I don't have entire d3d9/8 game catalog to confirm that.

WinterSnowfall:
Closes #4860 and goes a bit beyond that to:
- clean up some vendorid jank in d3d9 code (now we determine what we're using in a single place)
- report all non-Intel, non-Nvidia and Non-AMD GPUs as AMD in d3d8/9 (this affects mobile GPU vendors) - I have not hidden this under a,  say, `d3d9.hideGenericGpu` config option, but I can do that if anyone thinks it's worthwhile.

Needs quite a bit of testing on Intel, AMD and mobile, as I was only able to confirm the new `d3d9.hideNvidiaGpu` option works as expected.

~We could also potentially rework some of the existing config options now, but I haven't really looked into any of that yet.~ ~Mostly done now, save for Warhammer, which I'm sure sure what is about.~ Done deal.

### DIFF
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -18,14 +18,17 @@ namespace dxvk {
 
   const char* GetDriverDLL(DxvkGpuVendor vendor) {
     switch (vendor) {
-      default:
       case DxvkGpuVendor::Nvidia: return "nvd3dum.dll";
 
 #if defined(__x86_64__) || defined(_M_X64)
+      default:
       case DxvkGpuVendor::Amd:    return "aticfx64.dll";
+
       case DxvkGpuVendor::Intel:  return "igdumd64.dll";
 #else
+      default:
       case DxvkGpuVendor::Amd:    return "aticfx32.dll";
+
       case DxvkGpuVendor::Intel:  return "igdumd32.dll";
 #endif
     }
@@ -42,7 +45,7 @@ namespace dxvk {
     m_ordinal         (Ordinal),
     m_displayIndex    (DisplayIndex),
     m_modeCacheFormat (D3D9Format::Unknown),
-    m_d3d9Formats     (Adapter, m_parent->GetOptions()) {
+    m_d3d9Formats     (this, Adapter, m_parent->GetOptions()) {
     m_adapter->logAdapterInfo();
   }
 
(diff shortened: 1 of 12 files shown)


WinterSnowfall:
Appears to be mostly fine, so might as well land it in experimental and see how it goes. Thanks to @Blisto91 for testing.

qinlili23333 (approved):

src/d3d9/d3d9_options.h
```diff
@@ -22,6 +22,19 @@ namespace dxvk {
     int32_t customDeviceId;
     std::string customDeviceDesc;
 
+    /// Reports Nvidia GPUs running on the proprietary driver as a different
+    /// vendor (usually AMD)
+    bool hideNvidiaGpu;
+
+    /// Reports Nvidia GPUs running on NVK as a different vendor (usually AMD)
+    bool hideNvkGpu;
```
K0bin, line 30: Is there a good reason to have a separate option for NVK? The `hideXYGPU` options mostly change the GPU we report to the application.

We don't have an `hideAmdPropGPU` or `hideAmdVlkGPU` option either.

WinterSnowfall: I don't recall why it was added to dxgi in the first place, to be fair. I only included it for feature parity. It *could* potentially become more useful as NVK gains more traction/adoption.

src/d3d9/d3d9_adapter.cpp
```diff
@@ -71,15 +74,82 @@ namespace dxvk {
...
-    uint32_t deviceId  = options.customDeviceId == -1     ? props.deviceID   : uint32_t(options.customDeviceId);
-    const char*  desc  = options.customDeviceDesc.empty() ? props.deviceName : options.customDeviceDesc.c_str();
     const char* driver = GetDriverDLL(DxvkGpuVendor(vendorId));
 
     copyToStringArray(pIdentifier->Description, desc);
     copyToStringArray(pIdentifier->DeviceName,  displayName.c_str()); // The GDI device name. Not the actual device name.
-    copyToStringArray(pIdentifier->Driver,      driver);            // This is the driver's dll.
+    copyToStringArray(pIdentifier->Driver,      driver);              // This is the driver's dll.
 
     pIdentifier->DeviceIdentifier       = guid;
```
qinlili23333, line 154: Should we also hide DeviceIdentifier?

As [Microsoft documented](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3dadapter-identifier9), `DeviceIdentifier can be queried to check changes in the driver and chip set. This GUID is a unique identifier for the driver and chip set pair. Query this member to track changes to the driver and chip set in order to generate a new profile for the graphics subsystem. DeviceIdentifier can also be used to identify particular problematic drivers.` 

In artifacts of this PR, DeviceIdentifier is not changed when hideIntelGPU config changed. I'm not sure whether some game will reuse some cache it generated when hideGPU is not enabled.

I can confirm at least some game engine will check DeviceIdentifier somewhere (e.g. [krkrz visual novels](https://github.com/krkrz/krkrz/blob/fd5c4baa6a2ef5978db1bd043634351f48667daf/visual/win32/WindowImpl.cpp#L362)), but I cannot confirm whether this may lead to some issues.

WinterSnowfall: Since simply changing the vendor ID works just fine for most games (though handling Intel jankiness was a bit suboptimal until now), I'd only worry about it *if* it actually becomes a problem, but I sincerely doubt it's worth messing with. At worst, some games may chose to reset their graphical settings when detecting a new guid, but I don't think any are seriously mad enough to use it in any other way.

In fact, the only reason I also bothered with the device description on d3d8/9 side is that there are games known to use it in determining their rendering paths.

qinlili23333: > Perhaps, but in our case dxvk is the driver as far as the application is concerned, and its caches don't need any particular sort of special handling. The driver version reported by dxvk to the application also stays constant at all times.

So I think a more proper way to handle this is combining deviceUUID and dxvk version as DeviceIdentifier, although I agree currently it's not in priority. I'll figure it out and maybe make another PR about this separately later.

WinterSnowfall: Sure, although as far as I remember dxvk doesn't make its version visible to applications in any way, nor does it really have any reason to do so. But I guess that's a wider discussion which needs to take dxgi into account as well, and I'm not sure how that's being handled currently.

WinterSnowfall:
One last thing: VCache query support was getting rather tedious to manage as a config option now that we have a hideVendorGpu logic in place, and since it only ever was supported on Nvidia GPUs, I've simplified things quite a bit by removing the config option and determining support on device creation (based on our new vendorID masking logic).

src/d3d9/d3d9_adapter.cpp
```diff
@@ -71,15 +74,88 @@ namespace dxvk {
...
+          vendorId = fallbackVendor;
+
+          if (options.customDeviceId < 0)
+            deviceId = fallbackDevice;
+
+          if (options.customDeviceDesc.empty())
+            desc = fallbackDesc;
+        }
+      // Default to reporting ATI/AMD for other vendors
+      } else {
```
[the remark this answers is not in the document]
WinterSnowfall: > but on turnip we report amd anyway? What's the reason for this?

Ah, fair point. I was going for a catch-all type of approach and put this in place *before* having any actual hideAmdGpu options :facepalm:.

> Doesn't it make more sense to just set `hideGpu |= !intel && !amd && !nv` above?

And that's why we have a review process :sweat_smile:. It does indeed.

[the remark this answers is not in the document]
WinterSnowfall: Alright, hopefully it makes more sense now, thanks for setting me on the right path.

src/d3d9/d3d9_options.h
WinterSnowfall: @doitsujin prefers we keep it in place for dxgi, so I'll mirror it for d3d8/9 as well while it's a thing.

src/d3d9/d3d9_format.cpp
```diff
@@ -461,17 +462,19 @@ namespace dxvk {
...
+    D3DADAPTER_IDENTIFIER9 adapterId9;
+    HRESULT res = pParent->GetAdapterIdentifier(0, &adapterId9);
+    const uint32_t vendorId = SUCCEEDED(res) ? adapterId9.VendorId : 0;
 
     // NVIDIA does not natively support any DF formats
     m_dfSupport = vendorId == uint32_t(DxvkGpuVendor::Nvidia) ? false : options.supportDFFormats;
     m_x4r4g4b4Support = options.supportX4R4G4B4;
     // Only AMD supports D16_LOCKABLE natively
-    m_d16lockableSupport = vendorId == uint32_t(DxvkGpuVendor::Amd) ? true : options.supportD16Lockable;
+    m_d16lockableSupport = vendorId == uint32_t(DxvkGpuVendor::Amd) ? true : false;
```
doitsujin, line 477: `a ? true : false` is equivalent to `a`

src/d3d9/d3d9_device.cpp
```diff
@@ -193,6 +193,12 @@ namespace dxvk {
     m_activeRTsWhichAreTextures = 0;
     m_alphaSwizzleRTs = 0;
     m_lastHazardsRT = 0;
+
+    // Determine VCache query support
+    D3DADAPTER_IDENTIFIER9 adapterId9;
+    HRESULT res = m_adapter->GetAdapterIdentifier(0, &adapterId9);
+    const uint32_t vendorId = SUCCEEDED(res) ? adapterId9.VendorId : 0;
+    m_isVCacheQuerySupported = vendorId == uint32_t(DxvkGpuVendor::Nvidia) ? true : false;
```
doitsujin: same here

doitsujin (commented):
just a nit now, should be good otherwise.

src/d3d9/d3d9_format.cpp
WinterSnowfall: Ugh, missed this one.

src/d3d9/d3d9_device.cpp
WinterSnowfall: Sorted this one out while handing the ATOC hacks, but I'll modify the original commit too.

PR #4872, opened 2025-04-24
[d3d11] Support planar outputs in video processor

LizzyFleckenstein03:
This aims to support the game JR East Train Simulator, which uses the video processor with input and output streams both in the NV12 format.

Vulkan does not (usually?) support rendering to a G8_B8R8_2PLANE_420_UNORM (=NV12) view (it lacks the COLOR_ATTACHMENT_BIT in features - tho I've only checked on a modern AMD gpu).

Two passes are necessary since the Y and CbCr views may have different dimensions (in the case of NV12, CbCr is half the size of the Y view).

Since the creation of the output view now significantly overlaps with the input view's code (both have to iterate over the aspectMask to create views), I've introduced a superclass `VideoProcessorView` to share the behavior.

In case of planar output, the output view still creates a view in its "native" format, to support being cleared by `D3D11CommonContext::ClearView`.

### DIFF
--- a/src/d3d11/d3d11_video.cpp
+++ b/src/d3d11/d3d11_video.cpp
@@ -228,62 +228,14 @@ namespace dxvk {
           ID3D11Resource*         pResource,
     const D3D11_VIDEO_PROCESSOR_INPUT_VIEW_DESC& Desc)
   : D3D11DeviceChild<ID3D11VideoProcessorInputView>(pDevice),
-    m_resource(pResource), m_desc(Desc),
-    m_destructionNotifier(this) {
-    D3D11_COMMON_RESOURCE_DESC resourceDesc = { };
-    GetCommonResourceDesc(pResource, &resourceDesc);
+    m_common(pDevice, pResource, CreateViewInfo(Desc)),
+    m_desc(Desc), m_destructionNotifier(this) {
 
-    Rc<DxvkImage> dxvkImage = GetCommonTexture(pResource)->GetImage();
-
-    DXGI_VK_FORMAT_INFO formatInfo = pDevice->LookupFormat(resourceDesc.Format, DXGI_VK_FORMAT_MODE_COLOR);
-    DXGI_VK_FORMAT_FAMILY formatFamily = pDevice->LookupFamily(resourceDesc.Format, DXGI_VK_FORMAT_MODE_COLOR);
-
-    VkImageAspectFlags aspectMask = lookupFormatInfo(formatInfo.Format)->aspectMask;
-
-    DxvkImageViewKey viewInfo;
-    viewInfo.format = formatInfo.Format;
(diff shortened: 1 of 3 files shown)


doitsujin:
Would it be possible to split this into multiple commits? I'm having a very hard time working out what's going on when there's a relatively large refactor + adding new functionality all in one massive diff.

> Vulkan does not (usually?) support rendering to a G8_B8R8_2PLANE_420_UNORM (=NV12) view

Because rendering to planar images conceptually doesn't make much sense; especially when the chroma plane is subsampled.

As for ClearView, I don't know if this is supposed to do anything on planar views, would need some sort of test case to prove that it does, and *if* it does, work out how the color mapping works, and then clear the sub-views for each aspect one at a time. Views can only be created for one plane at a time.

doitsujin:
OK, looking at [ClearView](https://learn.microsoft.com/en-us/windows/win32/api/d3d11_1/nf-d3d11_1-id3d11devicecontext1-clearview) docs, the color mapping for yuv formats is straight-forward and well-defined, but we'd still need to deal with individual aspects, take subsampling into account (especially if there's clear rects involved) etc etc. Really going to have to write tests for this.

LizzyFleckenstein03:
sorry if it's a naive question, but are there tests located anywhere inside this repo or in a different one? I could contribute tests for this (I already some amount of testing code for this), I'm just having trouble figuring out where to put them

doitsujin:
Not really, no.

LizzyFleckenstein03:
I split it into different commits now - I hope it makes more sense like this.

src/d3d11/d3d11_video.cpp
```diff
@@ -309,41 +309,16 @@ namespace dxvk {
...
-        throw DxvkError("Invalid view dimension");
+    VideoProcessorView(pDevice, pResource, CreateViewInfo(Desc)),
+    m_desc(Desc), m_destructionNotifier(this) {
+    auto views = GetViews();
+    if (views[1] == nullptr) {
+      m_view = views[0];
+    } else {
+      DxvkImageViewKey viewInfo = m_viewInfo;
+      viewInfo.usage = VK_BUFFER_USAGE_TRANSFER_DST_BIT;
+      m_view = GetImage()->createView(viewInfo);
```
doitsujin, line 320: As discussed earlier, this seems somewhat nonsensical since we can't actually do anything with a multi-planar "view"; for video views, `ClearView` needs to iterate over available aspects like everyone else.

If the game doesn't use `ClearView` then i'll happily take a `TODO` comment in that function as well for the time being as well since that requires reworking the entire `ClearView` implementation.

doitsujin (changes requested):
Thanks, this is indeed much easier to reason about.

See comments; overall approach looks solid, but unless absolutely required I'd prefer to ignore the whole `ClearView` thing for now since there's going to be more refactoring needed.

PR #3435, opened 2023-05-20
[d3d9/d3d11] add samplerLodBias to d3d9 and clampNegativeLodBias for both d3d9 and d3d11

EndlesslyFlowering:
`d3d9.clampNegativeLodBias` addresses https://github.com/doitsujin/dxvk/issues/2779

a negative lod bias can help for some d3d9 games to look better
while clamping the negative load bias can also help some other games too 😄

Thanks to [@Kaldaien](https://github.com/Kaldaien) for giving me the code for clamping in D3D9!

PR #4875, opened 2025-04-26
[d3d9/d3d11] Negative/Positive/Both clampLodBias

Margen67:
Makes clampLodBias able to clamp positive and both instead of just negative.
Positive might prevent blurry textures. Not sure if any games do this, though.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -12,7 +12,7 @@
 # as determine it based on the DXGIOutput's current ColorSpace instead of
 # using CheckColorSpaceSupport.
 # This defaults to the value of the DXVK_HDR environment variable.
-# 
+#
 # Supported values: True, False
 
 # dxgi.enableHDR = True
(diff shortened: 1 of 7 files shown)


doitsujin:
Which problem is this trying to solve?

Margen67:
> Which real-world problem is this trying to solve?

To improve texture quality if a game uses positive LOD bias.

Blisto91:
I think it would be better to have a practical example of this being helpful to a application. If there are no known use cases then we might as well save having to keep a extra config option around.

Margen67:
> I think it would be better to first have a practical example of this being helpful to a application. If there are no known use cases then we might as well save having to keep a extra config option around.

It can be demonstrated using d3d-af-tester like from #3435:
* `clampLodBias = None` (Default)
none
* `clampLodBias = Positive`
positive

Blisto91:
I meant a game or application where it is like that on its own.

doitsujin:
yeah idk this doesn't seem useful for *real games*.

Issue #1307, opened 2019-12-25
[d3d9] Kane & Lynch: Dead Men (8080) Unhandled render state D3DRS_ADAPTIVETESS_Y


pchome:
The FOURCC value is **KanL**. I doubt it used somewhere in drivers, it looks like it was either for debugging purposes or to be treated as `D3D9Format::Unknown` (Edit: I mean "and applied with D3DRS_ADAPTIVETESS_Y". Name likely was "invented" to be unsupported format.).

Out of curiosity I created a hack to accept this format (where `D3D9Format::ATOC` check is in current code)
https://github.com/doitsujin/dxvk/blob/b4f2094c02bea9b241f558a1c1888ba28a88d3f0/src/d3d9/d3d9_device.cpp#L1695-L1697
and there is indeed noticeable difference (see screenshots). Actually I don't know what I'm doing, so this bug report is FYI.

#### Before
![KanL-default](https://user-images.githubusercontent.com/10661854/71434355-c5269980-26ec-11ea-9ecf-33c8ef697723.png)

#### After
![KanL-hack](https://user-images.githubusercontent.com/10661854/71434361-ce176b00-26ec-11ea-9ec2-4b2e1bf8a4c0.png)

### Software information
- Native dlls: d3dx9_31, d3dcompiler_47, d3dx9_34.dll (present in game's installdir)
- GFWL: xliveless (Bundled version failed to install)
- Maxed settings (upped to AA:8/AF:16).

### System information
- Wine version: Proton-4.11-11 (local build)

### Apitrace file(s)
- 'll try to do on request

### Log files
- d3d9.log: [kaneandlynch_d3d9.log](https://github.com/doitsujin/dxvk/files/3999667/kaneandlynch_d3d9.log)

Joshua-Ashton:
What you linked to was just my code for ADAPTIVENESS_Y and ATOC?

Joshua-Ashton:
IDK what KanL is meant to do, I imagine its meant to signify to the driver that this is **Kan**e and **L**ynch for drive hacks.

Is there an actual problem with the game btw?

pchome:
> What you linked to was just my code for ADAPTIVENESS_Y and ATOC?

Yes, for context.

> Is there an actual problem with the game btw?

No, the game works good.

Joshua-Ashton:
This isn't a bug then.

If there's any difference on native drivers that's noticable, let me know.
(ie. does it look like ATOC or another feature is enabled?)

[...]

PR #4888, opened 2025-05-02
[d3d9] Properly determine support for vendor hacks

WinterSnowfall:
In addition to gating vendor hacks behind their proper vendorIds, we should also properly report support for said hacks in `CheckDeviceFormat`, as querying for support is the first step in any guide that describes these hacks. I definitely think exposing some of these on all the vendors is asking for trouble, especially considering how poorly some d3d9 games actually behave when faced with unexpected support for certain features.

Also took the liberty to cache device specific identifier data in the adapter, since it won't change during its lifetime and some queries can get spammed both internally and by external calls. I'm also getting the (cached) vendorId for the device from the adapter now.

~~D3D8 doesn't seem to expose *any* vendor hacks (the entire madness probably came into being a bit later), so I've handled this accordingly too.~~ Actually, the NULL and RESZ (on AMD) hacks seem to be exposed in D3D8 as well, based on what modern native drivers report.

I will write tests to confirm all the `CheckDeviceFormat` behavior on all vendors, both on olden and modern drivers, otherwise I think @K0bin will murder me :sweat_smile:, so WIP for now.

P.S.: 
- Reference for R2VB: https://web.archive.org/web/20150517205947/http://developer.amd.com/wordpress/media/2012/10/R2VB_programming.pdf
- Reference for ATOC/SSAA: https://web.archive.org/web/20061006161816/http://download.nvidia.com/developer/SDK/Individual_Samples/DEMOS/Direct3D9/src/AntiAliasingWithTransparency/docs/AntiAliasingWithTransparency.pdf

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -193,11 +193,6 @@ namespace dxvk {
     m_activeRTsWhichAreTextures = 0;
     m_alphaSwizzleRTs = 0;
     m_lastHazardsRT = 0;
-
-    // Determine used vendor ID
-    D3DADAPTER_IDENTIFIER9 adapterId9;
-    HRESULT res = m_adapter->GetAdapterIdentifier(0, &adapterId9);
-    m_vendorId = SUCCEEDED(res) ? adapterId9.VendorId : 0;
   }
 
 
@@ -2254,25 +2249,28 @@ namespace dxvk {
     bool changed = old != Value;
 
     if (likely(changed)) {
-      const bool oldClipPlaneEnabled = IsClipPlaneEnabled();
+      const uint32_t vendorId            = m_adapter->GetVendorId();
+      const bool     isNvidia            = vendorId == uint32_t(DxvkGpuVendor::Nvidia);
+      const bool     isAmd               = vendorId == uint32_t(DxvkGpuVendor::Amd);
+      const bool     isIntel             = vendorId == uint32_t(DxvkGpuVendor::Intel);
 
-      const bool oldDepthBiasEnabled = IsDepthBiasEnabled();
-
-      const bool oldATOC = IsAlphaToCoverageEnabled();
-      const bool oldNVDB = states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB);
-      const bool oldAlphaTest = IsAlphaTestEnabled();
+      const bool     oldClipPlaneEnabled = IsClipPlaneEnabled();
+      const bool     oldDepthBiasEnabled = IsDepthBiasEnabled();
+      const bool     oldATOC             = !m_isD3D8Compatible ? IsAlphaToCoverageEnabled() : false;
+      const bool     oldNVDB             = !m_isD3D8Compatible ? states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB) : false;
+      const bool     oldAlphaTest        = IsAlphaTestEnabled();
 
       states[State] = Value;
 
-      // AMD's driver hack for ATOC and RESZ.
+      // AMD's driver hack for ATOC, RESZ, INST and CENT (also supported on Nvidia)
       if (unlikely(State == D3DRS_POINTSIZE &&
-                   m_vendorId == uint32_t(DxvkGpuVendor::Amd))) {
-        // ATOC
+                   !m_isD3D8Compatible && !isIntel)) {
+        // ATOC (AMD specific)
         constexpr uint32_t AlphaToCoverageEnable  = uint32_t(D3D9Format::A2M1);
         constexpr uint32_t AlphaToCoverageDisable = uint32_t(D3D9Format::A2M0);
 
-        if (Value == AlphaToCoverageEnable
-         || Value == AlphaToCoverageDisable) {
+        if ((Value == AlphaToCoverageEnable
+          || Value == AlphaToCoverageDisable) && isAmd) {
           m_amdATOC = Value == AlphaToCoverageEnable;
 
           bool newATOC = IsAlphaToCoverageEnabled();
@@ -2287,19 +2285,36 @@ namespace dxvk {
           return D3D_OK;
         }
 
-        // RESZ
+        // RESZ (AMD specific - once supported by Intel
+        // as well, however modern drivers do not expose it)
         constexpr uint32_t RESZ = 0x7fa05000;
-        if (Value == RESZ) {
+        if (Value == RESZ && isAmd) {
           ResolveZ();
+          return D3D_OK;
+        }
+
+        // INST (AMD specific)
+        if (unlikely(Value == uint32_t(D3D9Format::INST) && isAmd)) {
+          // Geometry instancing is supported by SM3, but ATI/AMD
+          // exposed this hack to retroactively enable it on their
+          // SM2-capable hardware. It's esentially a no-op.
+          return D3D_OK;
+        }
+
+        // CENT (AMD & Nvidia)
+        if (unlikely(Value == uint32_t(D3D9Format::CENT))) {
+          // Centroid (alternate pixel center) hack.
+          // Taken into account anyway, so yet another no-op.
           return D3D_OK;
         }
       }
 
-      // Nvidia/Intel's driver hack for ATOC.
+      // Nvidia's driver hack for ATOC (also supported on Intel), COPM and SSAA
       if (unlikely(State == D3DRS_ADAPTIVETESS_Y &&
-                   (m_vendorId == uint32_t(DxvkGpuVendor::Nvidia)
-                 || m_vendorId == uint32_t(DxvkGpuVendor::Intel)))) {
+                   !m_isD3D8Compatible && !isAmd)) {
+        // ATOC (Nvidia & Intel)
         constexpr uint32_t AlphaToCoverageEnable  = uint32_t(D3D9Format::ATOC);
+        // Disabling both ATOC and SSAA is done using D3DFMT_UNKNOWN (0)
         constexpr uint32_t AlphaToCoverageDisable = 0;
 
         if (Value == AlphaToCoverageEnable
@@ -2318,9 +2333,16 @@ namespace dxvk {
           return D3D_OK;
         }
 
-        if (unlikely(Value == uint32_t(D3D9Format::COPM))) {
+        // COPM (Nvidia specific)
+        if (unlikely(Value == uint32_t(D3D9Format::COPM) && isNvidia)) {
           // UE3 calls this MinimalNVIDIADriverShaderOptimization
           Logger::info("D3D9DeviceEx::SetRenderState: MinimalNVIDIADriverShaderOptimization is unsupported");
+          return D3D_OK;
+        }
+
+        // SSAA (Nvidia specific)
+        if (unlikely(Value == uint32_t(D3D9Format::SSAA) && isNvidia)) {
+          Logger::warn("D3D9DeviceEx::SetRenderState: Transparency supersampling is unsupported");
           return D3D_OK;
         }
       }
@@ -2553,14 +2575,16 @@ namespace dxvk {
         case D3DRS_ADAPTIVETESS_X:
         case D3DRS_ADAPTIVETESS_Z:
         case D3DRS_ADAPTIVETESS_W:
-          if (states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB) || oldNVDB) {
+          // Nvidia specific depth bounds test hack
+          if (!m_isD3D8Compatible &&
+              (states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB) || oldNVDB) &&
+              isNvidia) {
             m_flags.set(D3D9DeviceFlag::DirtyDepthBounds);
 
             if (m_state.depthStencil != nullptr && m_state.renderStates[D3DRS_ZENABLE])
               m_flags.set(D3D9DeviceFlag::DirtyFramebuffer);
-            break;
           }
-        [[fallthrough]];
+          break;
 
         default:
           static bool s_errorShown[256];
@@ -4538,7 +4562,7 @@ namespace dxvk {
 
 
   bool D3D9DeviceEx::SupportsVCacheQuery() const {
-    return m_vendorId == uint32_t(DxvkGpuVendor::Nvidia);
+    return m_adapter->GetVendorId() == uint32_t(DxvkGpuVendor::Nvidia);
   }
 
 
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -47,6 +47,7 @@ namespace dxvk {
     m_modeCacheFormat (D3D9Format::Unknown),
     m_d3d9Formats     (this, Adapter, m_parent->GetOptions()) {
     m_adapter->logAdapterInfo();
+    CacheIdentifierInfo();
   }
 
   template <size_t N>
@@ -61,10 +62,6 @@ namespace dxvk {
     if (unlikely(pIdentifier == nullptr))
       return D3DERR_INVALIDCALL;
 
-    auto& options = m_parent->GetOptions();
-    
-    const auto& props = m_adapter->deviceProperties();
-
     WCHAR wideDisplayName[32] = { };
     if (!wsi::getDisplayName(wsi::getDefaultMonitor(), wideDisplayName)) {
       Logger::err("D3D9Adapter::GetAdapterIdentifier: Failed to query monitor info");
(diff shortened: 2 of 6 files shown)


pchome:
Somehow related info: 
#1307 
https://aras-p.info/texts/D3D9GPUHacks.html

As for Kane & Lynch hack the patch is old and not applicable to current code, but tl;dr:
1. I added new format for KanL FOURCC
2. added check in State == D3DRS_ADAPTIVETESS_Y for D3D9Format::KanL to enable ATOC

_Also there is a chance that KanL wasn't a specific hack in d3d driver but D3DRS_ADAPTIVETESS_Y + Format::Unknown_

pchome:
Either way the fences looks better with that hack.
I know it against dxvk rules (philosophy) to add visual improvements which are not present in original game, but maybe worth to add an option to enable such hacks (disabled by default).

The game itself worked fine with and without hack.

WinterSnowfall:
Tests have been written and I've already adjusted some things based on AMD/Nvidia behavior on native. It was mostly in line with what I had done, with the exception of NULL/RESZ being available in D3D8 as well.

Will wait for @Blisto91 to generously provide test outputs on Intel before undrafting, as I'd like to get the vendor-specific behavior right, even if we report it as AMD by default.

> I know it against dxvk rules (philosophy) to add visual improvements which are not present in original game, but maybe worth to add an option to enable such hacks (disabled by default).

Well, then you know :wink:. It certainly enters modding territory, as it's non-standard behavior. And while dxvk does allow some tweaks here and there, they're mostly in line with what age-accurate drivers allowed as overrides. This certainly isn't one of those cases.

WinterSnowfall:
Incidentally, there's support for R2VB reported on my ATI HD 4650, but it doesn't seem to have lasted very long, as allegedly newer GPUs/drivers dropped it.

mirh:
> Also there is a chance that KanL wasn't a specific hack in d3d driver but D3DRS_ADAPTIVETESS_Y + Format::Unknown

I did check ati's 8.8 and nvidia's 179.48 drivers (intel's GMA ones from 2009 too but I couldn't even find normal formats so idk).
I could find a "KanL" string in `nvd3dum.dll` and `nvd3dumx.dll`. Though I don't have ghidra now in my hands to really check-check.

WinterSnowfall:
> I could find a "KanL" string in `nvd3dum.dll` and `nvd3dumx.dll`. Though I don't have ghidra now in my hands to really check-check.

Interesting. It may have been a game-specific hack/workaround on Nvidia at the time indeed, but I doubt modern drivers still cater for it. And also as to what it actually did is anyone's guess these days.

mirh:
Actually.. they are still there (I had 556.12 hanging around to check). 
Which in turn makes the one match per dll I had, too sus now to just have been a coincidence.
Still, of course with no disassembly, symbols or source code it's hard to take much of value from just its existence.

pchome:
In case of ATOC/SSAA do this even meter what was supported/dropped in d3d drivers? I mean you translate to Vulkan ATOC, so if you know when enable/disable ATOC for specific render state it should work?

For what the game required separate format is the other question, but it worked with simple patch in dxvk. Should maybe even work on AMD hardware.

The game seems was delisted from Steam some time ago, but should still be available for those who have it in library.
I'll try to install and test it.

WinterSnowfall:
> In case of ATOC/SSAA do this even meter what was supported/dropped in d3d drivers? I mean you translate to Vulkan ATOC, so if you know when enable/disable ATOC for specific render state it should work?

That is essentially correct, but I'm skeptical that this was even ever about ATOC or *only* ATOC, since there's literally no documentation for it anywhere. Of course anyone's free to patch this into being and it will work cross-vendors, but I doubt it will ever be mainlined. It's certainly outside the scope of the PR to do so, anyway.

mirh:
I was using ATOC as a string to double check (in case drivers didn't have "KanL" just because they were obfuscated) and while there are slightly fewer matches in the latest ones there are still some.
And I'm not sure why you think this should have been dropped? The whitepaper seems clear: enlightened games did check for that format support, and if so they allowed/enabled the thing. There's even this comment in the [sample](https://download.nvidia.com/developer/SDK/Individual_Samples/DEMOS/Direct3D9/AntiAliasingWithTransparency.zip):

> 	Per primitive MSAA and SSAA are turned on by setting D3DRS_ADAPTIVETESS_Y to either:
	(D3DFORMAT)MAKEFOURCC('A', 'T', 'O', 'C'))
	(D3DFORMAT)MAKEFOURCC('S', 'S', 'A', 'A'))
	Setting D3DRS_ADAPTIVETESS_Y back to D3DFMT_UNKNOWN turns both modes off

Of note that's probably gated behind a vendor check, while amd [has](https://web.archive.org/web/20150905183945/http://developer.amd.com/tools-and-sdks/graphics-development/amd-radeon-sdk/archive/#direct3d9):

> #define ALPHA_TO_COVERAGE_ENABLE (MAKEFOURCC(‘A’,‘2’,‘M’,‘1’))
#define ALPHA_TO_COVERAGE_DISABLE (MAKEFOURCC(‘A’,‘2’,‘M’,‘0’))

---
As for KanL, after having accessed higher planes of existence I found it has even a few brothers in the nvidia drivers: HL2A, CENT, TR70 and TIME (half life 2, tomb raider legends, and timeshift respectively). Of course I'm an inept with ghidra, and I couldn't get anything out of their exes (though for K&L I can tell you it's near direct3d initialization, so it's really not there by chance).
...but I still felt confident enough so after scavenging the net for clues, let's say I found that at a very specific point in time a decade and half ago a certain world renowned engine had what I could condense like so:
```C++
if ( g_pHardwareConfig->Caps().m_bNeedsATICentroidHack )
{
	SetSupportedRenderStateForce( D3DRS_POINTSIZE, MAKEFOURCC( 'C', 'E', 'N', 'T' ) );
}
else if ( g_pHardwareConfig->Caps().m_VendorID == VENDORID_NVIDIA )
{
	// This helps the driver to know to turn on fast z reject for NVidia dx80 hardware
	SetSupportedRenderStateForce( D3DRS_POINTSIZE, MAKEFOURCC( 'H', 'L', '2', 'A' ) );
}
if( g_pHardwareConfig->Caps().m_bDisableShaderOptimizations )
{
	SetSupportedRenderStateForce( D3DRS_ADAPTIVETESS_Y, MAKEFOURCC( 'C', 'O', 'P', 'M' ) );
}
```
The first must be [this](https://web.archive.org/web/20040715143734fw_/http://www.ati.com/developer/gdc/D3DTutorial_ShaderModels.pdf#page=41) or [this](https://web.archive.org/web/20030407232530fw_/http://www.ati.com/developer/gdc/D3DTutorial1_Shaders.pdf#page=84) and for a week it was even all the [rage](https://web.archive.org/web/20040213012938/http://www.halflife2.net/forums/showthread.php?s=&threadid=3071) in 2003 (if nvidia supported it apparently something changed).
The second seems pretty self-descriptive, while the third is a new one for me (vendor unknown) in what was probably a forefather of the modern option through nvapi/ags.
I assume the other codes were also something like that, probably arranged lovely together as a part of the TWIMTBP program.

WinterSnowfall:
> And I'm not sure why you think this should have been dropped? 

I was never unsure about ATOC/SSAA, since those are *well documented*. "KanL" and other application specific hacks probably pre-date Nvidia app profiles, which is why I assumed they'd be removed by now.

> As for KanL, after having accessed higher planes of existence I found it has even a few brothers in the nvidia drivers: HL2A, CENT, TR70 and TIME (half life 2, tomb raider legends, and timeshift respectively).

All of this is certainly interesting, but again, not a basis to handle any of them in any particular way in dxvk.

As for the ATI centroid hack (aka "Alternate Pixel Center"), I think we cater for that globally in dxvk. But that was indeed a driver option at the time (since removed):

Compatibility

pchome:
@WinterSnowfall 
> If we do decide to do something with these, it will definitely be included in a separate PR. For now I'll investigate to see if there's anything else other games use.

I guess you could add an potential place to extend code with those hacks.
Just to avoid refactoring and to simplify code extension. Maybe you able to run one of those examples and implement an simple working hack in this PR, but just don't "git stage" hack part.

Asking because I tried to recreate my old KanL patch and it currently involves messing with D3DRS_ADAPTIVETESS_Y + vendorId conditions.

WinterSnowfall:
> Asking because I tried to recreate my old KanL patch and it currently involves messing with D3DRS_ADAPTIVETESS_Y + vendorId conditions.

If you want to support it across vendors, simply remove the vendor check for the D3DRS_ADAPTIVETESS_Y section in this PR and then add a block handling the KanL fourcc in there.

Once this PR gets finalized and merged I'll follow it up with another PR that will handle these game-specific hack values as a no-op, which should be done anyway, since none of these should end up in tessellation territory.

Right now I'm looking to see if there's other values to take into account, and I've already stumbled upon `COJ1`, used by Call of Juarez.

mirh:
Cool, probably something related with whatever quick and dirty [improvements](https://www.anandtech.com/show/2267/3) they did during development.
Still couldn't find anywhere, the old nvidia dll I checked did indeed have more than just what the newest sports (COPM seems indeed a thing, and in the same conditional I could also find COPL and COPD) but no COJ1.
Also FWIW the 2008 `atiumdag.dll` only seemed to have CENT, A2M1, A2M0 and INST.

It would be nice if tracestats tool could even help spot whatever different behaviour (or pixel result?) resulted from those flags tho.
Like we don't know about KanL, but even for those we do like fast z (*dxlevel 80* should mean [this](https://www.anandtech.com/show/875/2)?) or sampling behavior not having to trust/rely on source code comments would go a long way forward.
And idk if maybe one day those couldn't be used to inform that you can take certain performance shortcuts.

WinterSnowfall:
Alright, finished parsing my collection of traces and didn't find any new values besides what Call of Juarez uses. Interestingly, Tomb Raider - Anniversary also sets `D3DRS_ADAPTIVETESS_Y = TR70`, so it's not just a Legends exclusive.

> CENT

I've found no indication this is used in any way in d3d8/9. No games use this FOURCC and all the references I could find online about alternate pixel center refer to older d3d7 game hacks, particularly related to Final Fantasy VII.

> It would be nice if tracestats tool could even help spot whatever different behaviour (or pixel result?) resulted from those flags tho.

Yeah, that's not a bit, but rather *a lot* out of scope :).

> we do like fast z (_dxlevel 80_ should mean [this](https://www.anandtech.com/show/875/2)?) or sampling behavior not having to trust/rely on source code comments would go a long way forward.

Fast Z clear, from what I recall, was a hardware feature added to "new" GPUs in the GeForce 3 series (?) and sometime later to ATI cards, not something we can take advantage of today. I expect all cards these days will perform well in that regard and there's nothing to squeeze out of modern GPUs there.

WinterSnowfall:
Made some final adjustments based on what modern Intel drivers report on Battlemage (thanks to @Blisto91). Seems like RESZ is in fact not supported on modern intel.

On a tangent, also noticed all vendors seem to report the following adapter/backbuffer format support:


```
Device back buffer format support (full screen):
  - The D3DFMT_A8R8G8B8 format is not supported
  + The D3DFMT_X8R8G8B8 format is supported, device creation SUCCEEDED
  + The D3DFMT_R5G6B5 format is supported, device creation SUCCEEDED
  - The D3DFMT_X1R5G5B5 format is not supported
  - The D3DFMT_A1R5G5B5 format is not supported
  + The D3DFMT_A2R10G10B10 format is supported, device creation SUCCEEDED

Device back buffer format support (windowed):
  - The D3DFMT_A8R8G8B8 format is not supported
  + The D3DFMT_X8R8G8B8 format is supported, device creation SUCCEEDED
  + The D3DFMT_R5G6B5 format is supported, device creation SUCCEEDED
  + The D3DFMT_X1R5G5B5 format is supported, device creation SUCCEEDED
  - The D3DFMT_A1R5G5B5 format is not supported
  - The D3DFMT_A2R10G10B10 format is not supported
```

However, none will list any modes for `D3DFMT_X1R5G5B5`, which is what was causing the problems with the Beyond Good & Evil settings app. But the format IS a valid adapter format, and a valid backbuffer format in windowed mode. I've adjusted our reporting logic to reflect the above test results.

Now the PR is ready, only needs a bit of testing.

WinterSnowfall:
Converting back to draft for a bit until the specifics of `D3DRS_POINTSIZE = CENT` become more clear. It is apparently set by HL2/Black Mesa on AMD, but allegedly also exists on Nvidia. I'll have to test this on age accurate hardware and drivers.

@mirh As for `D3DRS_POINTSIZE = HL2A`, it doesn't exist in modern HL2 at least. On Nvidia it's just the regular ATOC hack I see in apitrace.

mirh:
> I've found no indication this is used in any way in d3d8/9. 

It was always gated behind a cfg whitelist (a quick grep showed it's still there in `shaderapidx9.dll` and somewhat funnily even in `shaderapivk.dll`). So even though that covers pretty much every card from both teams between 1999 and 2011, it wouldn't show up unless you got somebody with <dx12 hardware trying the games or editing the hardware list file.

> As for D3DRS_POINTSIZE = HL2A, it doesn't exist in modern HL2 at least. 

Sure it doesn't show up in the recent branches. But it should still be there for example in L4D.

> There's isn't any use of COPM either.

It's still there actually, but as you see it needs to be manually enabled. ~~(probably used for debugging?)~~ 
Ok that's what the UE3 *UseMinimalNVIDIADriverShaderOptimization* variable used to be about too. Supposedly (at least on older gpus? not sure if because perhaps those didn't use to get many improvements, or because they used them just as a proxy of having a slower system?) it helped with compilation stuttering.
I would guess that dxvk being its own different bag of worms it shouldn't matter.

Of very fucking annoying note also, that while checking for it I also found out about *DisableATITextureFilterOptimizationChecks* using `SetRenderState(D3DRS_POINTSIZE, 0x7fa02001)` to disable what I believe was the catalyst AI opt analysis logic.
Needless to say that's very much *not* a character string.

WinterSnowfall:
> So even though that covers pretty much every card from both teams between 1999 and 2011, it wouldn't show up unless you got somebody with <dx12 hardware trying the games or editing the hardware list file.

I happen to have a collection of d3d8/9/10 era cards, but it will take a while to test.

> I also found out about _DisableATITextureFilterOptimizationChecks_ using `SetRenderState(D3DRS_POINTSIZE, 0x7fa02001)` to disable what I believe was the catalyst AI opt analysis logic. Needless to say that's very much _not_ a character string.

Enabling RESZ is done by setting D3DRS_POINTSIZE to 0x7fa05000, so I'm not all that surprised.

WinterSnowfall:
> Sure it doesn't show up in the recent branches. But it should still be there for example in L4D.

Nope. @Blisto91 was kind enough to get me a trace, but it isn't there either. I think it *might* have been part of an ancient Source version, but I don't think it's relevant in any way these days, so may have been dropped for some time now.

The CENT hack however is something I'll follow up on for sure, even if it's only for academical reasons (in dxvk, it will be a no-op, since that's handled anyway). Kind of surprised Aras's website has no mention of it whatsoever - once I figure out more about it, will drop an email their way.

mirh:
> I happen to have a collection of d3d8/9/10 era cards, but it will take a while to test.

I'm pretty confident you can just edit the cfg file then.

> Nope. @Blisto91 was kind enough to get me a trace, but it isn't there either.

I mean, it is _there_. Maybe the logic to trigger it is iffy (to be sure it's vendor gated, and mutually exclusive with CENT) but still.

> Edit: Managed to find a game that sets it, namely SiN Episodes: Emergence, which has a Source build from around 2013.

AFAIK that's using source 2004 actually (being the probable only thing still officially available on that branch). But it shouldn't be necessary to go all \*that\* back.

> Enabling RESZ is done by setting D3DRS_POINTSIZE to 0x7fa05000, so I'm not all that surprised.

What surprises me though is that I cannot find a "record" of either of them into any ati dll ;(
So my fishing expedition for more of them had to be dropped.

WinterSnowfall:
Alright, so CENT isn't actually advertised by any modern drivers, but some games (older Source titles especially) do set it on AMD, so we should still no-op it.

mirh:
Surely it shouldn't be a big deal, but HL2A, CENT, TR70, TIME and KanL are all still present even in 576.40.

WinterSnowfall:
Yes, and more importantly we've determined said games actually use them when presented with the Nvidia vendorId, but closed source drivers won't help us much here. Even if we were to decide to factor in those game-specific formats, it would be to NOP them. I have already done that for CENT, since it has a potentially larger scope.

src/d3d9/d3d9_device.cpp
```diff
@@ -2254,25 +2249,28 @@ namespace dxvk {
...
+      const bool     isIntel             = vendorId == uint32_t(DxvkGpuVendor::Intel);
 
-      const bool oldATOC = IsAlphaToCoverageEnabled();
-      const bool oldNVDB = states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB);
-      const bool oldAlphaTest = IsAlphaTestEnabled();
+      const bool     oldClipPlaneEnabled = IsClipPlaneEnabled();
+      const bool     oldDepthBiasEnabled = IsDepthBiasEnabled();
+      const bool     oldATOC             = !m_isD3D8Compatible ? IsAlphaToCoverageEnabled() : false;
+      const bool     oldNVDB             = !m_isD3D8Compatible ? states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB) : false;
+      const bool     oldAlphaTest        = IsAlphaTestEnabled();
```
doitsujin, line 2261: I think this entire thing is due for a refactor in the future, this entire block before the `switch` incurs overhead for *every single state change* even when most state changes aren't affected by vendor hacks at all.

PR #4891, opened 2025-05-02
Fix image relocation on interop interfaces

doitsujin:
Untested, but relevant for VR workloads.


Patola:
Do you know which games this can affect? I'll try and test them

PR #4908, opened 2025-05-12
[d3d11] Fix planar video output when dst rect is enabled

LizzyFleckenstein03:
Fixes an oversight from #4872


doitsujin:
What does this fix exactly? If the view only contains `ASPECT_1_BIT`, then `mipLevelExtent` should already only return the subsampled size.

PR #4939, opened 2025-05-24
[d3d9] Improve SetRenderState state tracking

WinterSnowfall:
Should speed up `SetRenderState` calls where state changes by a bit. I don't expect anything dramatic, but that being said `SetRenderState` is a *very* hot path, albeit not that heavy, and these things do tend to add up.

### DIFF
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -136,14 +136,14 @@ namespace dxvk {
     if (unlikely(rt && CheckFormat == D3D9Format::A8 && m_parent->GetOptions().disableA8RT))
       return D3DERR_NOTAVAILABLE;
 
-    // NULL RT format hack (supported across all vendors,
-    // and also advertised in D3D8 by modern drivers)
+    // NULL RT format hack (supported across all
+    // vendors, and also advertised in D3D8)
     if (unlikely(rt && CheckFormat == D3D9Format::NULL_FORMAT && twoDimensional))
       return D3D_OK;
 
-    // AMD/Intel's driver hack for RESZ (also advertised
-    // in D3D8 by modern AMD drivers, not advertised
-    // at all by modern Intel drivers)
+    // AMD/Intel's driver hack for RESZ
+    // (also advertised in D3D8 by AMD drivers,
+    // not advertised at all by modern Intel drivers)
     if (unlikely(rt && CheckFormat == D3D9Format::RESZ && surface))
       return isAmd
         ? D3D_OK
@@ -155,8 +155,8 @@ namespace dxvk {
         ? D3D_OK
         : D3DERR_NOTAVAILABLE;
 
-    // Nvidia's driver hack for SSAA
-    // (supported on modern Nvidia drivers)
+    // Nvidia's driver hack for SSAA (supported on Nvidia
+    // drivers, ever since the GeForce 6 series)
     if (unlikely(CheckFormat == D3D9Format::SSAA && surface)) {
       if (!isD3D8Compatible && isNvidia)
         Logger::warn("D3D9Adapter::CheckDeviceFormat: Transparency supersampling (SSAA) is unsupported");
@@ -164,6 +164,7 @@ namespace dxvk {
     }
 
     // Nvidia specific depth bounds test hack
+    // (supported ever since the GeForce 6 series)
     if (unlikely(CheckFormat == D3D9Format::NVDB && surface))
       return (!isD3D8Compatible &&
               m_adapter->features().core.features.depthBounds && isNvidia)
@@ -178,14 +179,15 @@ namespace dxvk {
       return D3DERR_NOTAVAILABLE;
     }
 
-    // AMD specific INST hack
+    // AMD specific INST (geometry instancing)
+    // hack for SM2-only capable cards
     if (unlikely(CheckFormat == D3D9Format::INST && surface))
       return (!isD3D8Compatible && isAmd)
         ? D3D_OK
         : D3DERR_NOTAVAILABLE;
 
-    // AMD/Nvidia CENT(roid) hack (not advertised by
-    // either AMD or Nvidia modern drivers)
+    // AMD/Nvidia CENT(roid) hack (not advertised
+    // by either AMD or Nvidia drivers)
     if (unlikely(CheckFormat == D3D9Format::CENT && surface))
       return D3DERR_NOTAVAILABLE;
 
@@ -250,7 +252,7 @@ namespace dxvk {
     // Check if this is a power of two...
     if (sampleCount & (sampleCount - 1))
       return D3DERR_NOTAVAILABLE;
-    
+
     // Therefore...
     VkSampleCountFlags sampleFlags = VkSampleCountFlags(sampleCount);
 
@@ -643,7 +645,7 @@ namespace dxvk {
     pCaps->MaxNpatchTessellationLevel = 0.0f;
     // Reserved for... something
     pCaps->Reserved5                  = 0;
-    // Master adapter for us is adapter 0, atm... 
+    // Master adapter for us is adapter 0, atm...
     pCaps->MasterAdapterOrdinal       = 0;
     // The group of adapters this one is in
     pCaps->AdapterOrdinalInGroup      = 0;
@@ -892,11 +894,11 @@ namespace dxvk {
       [](const D3DDISPLAYMODEEX& a, const D3DDISPLAYMODEEX& b) {
         if (a.Width < b.Width)   return true;
         if (a.Width > b.Width)   return false;
-        
+
         if (a.Height < b.Height) return true;
         if (a.Height > b.Height) return false;
-        
-        return b.RefreshRate < a.RefreshRate;
+
+        return a.RefreshRate > b.RefreshRate;
     });
   }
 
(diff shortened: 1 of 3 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -2574,17 +2586,21 @@ namespace dxvk {
 
         case D3DRS_ADAPTIVETESS_X:
         case D3DRS_ADAPTIVETESS_Z:
-        case D3DRS_ADAPTIVETESS_W:
+        case D3DRS_ADAPTIVETESS_W: {
           // Nvidia specific depth bounds test hack
-          if (!m_isD3D8Compatible &&
-              (states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB) || oldNVDB) &&
-              isNvidia) {
+          const bool nvdbEnabled = isNvidia ? IsNVDepthBoundsTestEnabled() : false;
```
K0bin, line 2591: Do we really need to add a field to the device for something that depends on a single render state? That seems unnecessary.

K0bin: Besides that, this can be simplified to

`isNvidia && IsNvidiaDepthBoundsTestEnabled()`

WinterSnowfall: Not sure I understand the concern. If you look further down the dirty-ing needs to happen even on state switches too (from enabled to disabled), so yes, this does need to be tracked, same as everything else which previously was calculated on all SetRenderState calls.

src/d3d9/d3d9_device.cpp
```diff
@@ -8521,6 +8541,16 @@ namespace dxvk {
     for (uint32_t i = 0; i < caps::MaxStreams; i++) {
       SetStreamSource(i, nullptr, 0, 0);
     }
+
+    m_amdATOC          = false;
+    m_nvATOC           = false;
+    m_ffZTest          = false;
```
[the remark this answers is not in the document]
WinterSnowfall:
```
    inline bool IsZTestEnabled() {
      return m_state.renderStates[D3DRS_ZENABLE] && m_state.depthStencil != nullptr;
    }
```

Surely this isn't guaranteed to remain the same on a device reset? I should update it to IsZTestEnabled() here as well at the very least.

src/d3d9/d3d9_device.cpp
```diff
@@ -2465,8 +2475,10 @@ namespace dxvk {
         case D3DRS_CLIPPLANEENABLE: {
           const bool clipPlaneEnabled = IsClipPlaneEnabled();
 
-          if (clipPlaneEnabled != oldClipPlaneEnabled)
+          if (clipPlaneEnabled != m_clipPlaneEnabled) {
```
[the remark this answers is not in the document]
WinterSnowfall: Fair enough, this one can indeed be tracked directly. Nice catch.

src/d3d9/d3d9_adapter.cpp
```diff
@@ -137,14 +137,14 @@ namespace dxvk {
...
+    // vendors, and also advertised in D3D8)
     if (unlikely(rt && CheckFormat == D3D9Format::NULL_FORMAT && twoDimensional))
       return D3D_OK;
 
-    // AMD/Intel's driver hack for RESZ (also advertised
-    // in D3D8 by modern AMD drivers, not advertised
-    // at all by modern Intel drivers)
+    // AMD/Intel's driver hack for RESZ (also
+    // advertised in D3D8 by AMD drivers, not
+    // advertised at all by modern Intel drivers)
```
K0bin, line 147: I think having not in the line before and having both lines start with advertised makes this seem like its advertised on Intel but whatever. Thats too much of a nitpick.

src/d3d9/d3d9_device.cpp
```diff
@@ -2244,66 +2250,67 @@ namespace dxvk {
...
 
-    if (likely(changed)) {
-      const uint32_t vendorId            = m_adapter->GetVendorId();
-      const bool     isNvidia            = vendorId == uint32_t(DxvkGpuVendor::Nvidia);
-      const bool     isAmd               = vendorId == uint32_t(DxvkGpuVendor::Amd);
-      const bool     isIntel             = vendorId == uint32_t(DxvkGpuVendor::Intel);
+    if (likely(Value != oldValue)) {
+      constexpr uint32_t nvidiaVendorId = uint32_t(DxvkGpuVendor::Nvidia);
+      constexpr uint32_t amdVendorId    = uint32_t(DxvkGpuVendor::Amd);
+      constexpr uint32_t intelVendorId  = uint32_t(DxvkGpuVendor::Intel);
```
K0bin, line 2259: Why did you change those?

IMO the version with bools was cleaner.

src/d3d9/d3d9_device.cpp
```diff
@@ -2244,66 +2250,67 @@ namespace dxvk {
...
-      const bool     oldNVDB             = !m_isD3D8Compatible ? states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB) : false;
-      const bool     oldAlphaTest        = IsAlphaTestEnabled();
+      const uint32_t vendorId = m_adapter->GetVendorId();
 
       states[State] = Value;
 
       // AMD's driver hack for ATOC, RESZ, INST and CENT (also supported on Nvidia)
-      if (unlikely(State == D3DRS_POINTSIZE &&
-                   !m_isD3D8Compatible && !isIntel)) {
+      if (unlikely(State == D3DRS_POINTSIZE && vendorId != intelVendorId)) {
```
K0bin, line 2266: I guess RESZ is supported on D3D8 then?

src/d3d9/d3d9_device.cpp
```diff
@@ -2575,17 +2591,24 @@ namespace dxvk {
 
         case D3DRS_ADAPTIVETESS_X:
         case D3DRS_ADAPTIVETESS_Z:
-        case D3DRS_ADAPTIVETESS_W:
+        case D3DRS_ADAPTIVETESS_W: {
           // Nvidia specific depth bounds test hack
-          if (!m_isD3D8Compatible &&
-              (states[D3DRS_ADAPTIVETESS_X] == uint32_t(D3D9Format::NVDB) || oldNVDB) &&
-              isNvidia) {
+          const bool nvdbEnabled = vendorId == nvidiaVendorId && IsNVDepthBoundsTestEnabled();
```
K0bin, line 2596: So NVDB is supported on D3D8 too?

src/d3d9/d3d9_device.h
```diff
@@ -1564,10 +1568,14 @@ namespace dxvk {
 
     bool                            m_isSWVP;
     bool                            m_isD3D8Compatible;
-    bool                            m_amdATOC          = false;
-    bool                            m_nvATOC           = false;
     bool                            m_ffZTest          = false;
-    
+
+    bool                            m_depthBiasEnabled = false;
```
K0bin, line 1573: Please add a comment in the code explaining why we have those fields when the helper functions exist.

K0bin (commented):
It's fine, I just have a few small questions.

src/d3d9/d3d9_adapter.cpp
WinterSnowfall: No, that's fair. I'll rework the line split so it makes more sense.

src/d3d9/d3d9_device.cpp
WinterSnowfall: It's not, but we check for that inside of `IsNVDepthBoundsTestEnabled()`, so there's no point in double checking.

src/d3d9/d3d9_device.cpp
```diff
@@ -2244,66 +2250,67 @@ namespace dxvk {
...
 
         if ((Value == AlphaToCoverageEnable
-          || Value == AlphaToCoverageDisable) && isAmd) {
-          m_amdATOC = Value == AlphaToCoverageEnable;
-
-          bool newATOC = IsAlphaToCoverageEnabled();
-          bool newAlphaTest = IsAlphaTestEnabled();
+          || Value == AlphaToCoverageDisable) &&
+             vendorId == amdVendorId &&
+             !m_isD3D8Compatible) {
```
K0bin, line 2274: `m_isD3D8Compatible` is redundant here because `IsAlphaToCoverageEnabled` checks it anyway. So both `atocEnabled` and `m_atocEnabled` are always false.

src/d3d9/d3d9_device.cpp
```diff
@@ -2312,38 +2319,43 @@ namespace dxvk {
 
       // Nvidia's driver hack for ATOC (also supported on Intel), COPM and SSAA
       if (unlikely(State == D3DRS_ADAPTIVETESS_Y &&
-                   !m_isD3D8Compatible && !isAmd)) {
+                   vendorId != amdVendorId && !m_isD3D8Compatible)) {
```
K0bin, line 2322: Same thing as above, checking `m_isD3D8Compatible` is redundant.

src/d3d9/d3d9_device.cpp
K0bin: Alright.

src/d3d9/d3d9_device.cpp
K0bin: Okay but then keep it consistent please. Most of the other `m_isD3D8Compatible` checks are equally redundant.

WinterSnowfall:
As discussed on Discord, I've moved all the vendor hack logic within the render state update switch statement, which has allowed some further optimization. I've tested ATOC in a few games and it seems fine, but I'll be doing some further testing in the coming days, only to confirm nothing has regressed.

K0bin (approved):

PR #4942, opened 2025-05-24
[dxgi] Flip inverted refresh rate modes

WinterSnowfall:
Rather cursed, but in practice seems to mimic what native drivers do, at least with Titan Quest: Anniversary Edition.


WinterSnowfall:
> I'd prefer that, at least that way we'd get consistent behaviour across different systems.

Done. Let me know if this is good enough or too derpy. To be fair I kind of get lost in dxgi code, so would prefer to keep the detour logic in this spot, but that being said I will deffer to your thoughts on the matter.

src/dxgi/dxgi_output.cpp
```diff
@@ -597,21 +597,29 @@ namespace dxvk {
     uint32_t minDiffRefreshRate = 0;
 
     if (TargetMode.RefreshRate.Numerator && TargetMode.RefreshRate.Denominator) {
```
doitsujin, line 599: can't we literally just do an extra check `Numerator >= Denominator` here and get the desired behaviour for free? not sure why we need all the extra code

WinterSnowfall: Haven't looked too closely at what happens outside of it, but can give it a shot like that, sure.

WinterSnowfall: Yeah, the order is pretty much the same even on native Nvidia (thanks to @Blisto91 for confirming). I'll leave it as is.

WinterSnowfall:
Since this isn't quite native behavior, such cases should be handled with the new refresh rate override config option on setups where it's indeed a problem.

PR #4956, opened 2025-05-30
Improve device UUID filtering logic

gilvbp:
This is a replacement for PR #2408.

This PR improves and corrects the UUID filtering logic in `DxvkDeviceFilter`,
ensuring proper selection based on Vulkan `deviceUUID`. It also prevents memory
leaks by cleaning up rejected adapters and simplifies the adapter query logic.

Closes #2408.

### DIFF
--- a/src/dxvk/dxvk_device_filter.cpp
+++ b/src/dxvk/dxvk_device_filter.cpp
@@ -1,11 +1,12 @@
 #include "dxvk_device_filter.h"
 #include <iomanip>
+#include <sstream>
 
 namespace dxvk {
 
     DxvkDeviceFilter::DxvkDeviceFilter(
       DxvkDeviceFilterFlags flags,
-      const DxvkOptions&          options)
+      const DxvkOptions& options)
       : m_flags(flags) {
       m_matchDeviceName = env::getEnvVar("DXVK_FILTER_DEVICE_NAME");
       m_matchDeviceUUID = env::getEnvVar("DXVK_FILTER_DEVICE_UUID");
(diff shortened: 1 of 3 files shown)


doitsujin:
is there any reason why this changes the indentation of the entire file?

gilvbp:
> Since device UUID's aren't guaranteed to be unique I don't think this is good as a general solution. Even if it helped in your case.
> https://github.com/doitsujin/dxvk/issues/2036#issuecomment-1254339655
> 
> Edit: actually I see now that the example shown is unique so maybe that isn't true ? Then it would be fine I guess.

"Implementations should return deviceUUID values which are likely to be unique even in the presence of multiple Vulkan implementations (such as a GPU driver and a software renderer; two drivers for different GPUs; or the same Vulkan driver running on two logically different devices)."

Reference: https://registry.khronos.org/vulkan/specs/latest/man/html/VkPhysicalDeviceIDProperties.html?utm_source=chatgpt.com

WinterSnowfall:
> Since device UUID's aren't guaranteed to be unique I don't think this is good as a general solution.

I don't think they're guaranteed to be unique across the entire population of cards and vendors, however the odds of someone having two cards with the same UUID in the same system should be quite astronomical. At least that's the gist I'm getting from the documentation.

src/dxvk/dxvk_instance.cpp
```diff
@@ -269,12 +269,18 @@ namespace dxvk {
 
     for (uint32_t i = 0; i < numAdapters; i++) {
       if (filter.testAdapter(deviceProperties[i])) {
-        result.push_back(new DxvkAdapter(m_vki, adapters[i]));
+        auto adapter = new DxvkAdapter(m_vki, adapters[i]);
```
doitsujin, line 272: `Rc<DxvkAdapter> adatper = new ...` so we can skip the explicit `delete` at the end.

Yeah I know our homebrew refcount thing kind of sucks, but it is what it is.

doitsujin (commented):
Just a nit now.

I will likely refactor this whole area in the near future so we can skip devices that don't support the required feature set beyond just the core Vulkan version, but in the meantime this will do.

src/dxvk/dxvk_device_filter.cpp
```diff
@@ -42,5 +56,19 @@ namespace dxvk {
 
     return true;
   }
+
   
+  bool DxvkDeviceFilter::testCreatedAdapter(const DxvkDeviceInfo& deviceInfo) const {
+    if (m_flags.test(DxvkDeviceFilterFlag::MatchDeviceUUID)) {
+      std::string uuidStr = convertUUID(deviceInfo.vk11.deviceUUID);
+      Logger::info(str::format("UUID usado para filtro: ", uuidStr));
```
Blisto91, line 64: English

PR #4969, opened 2025-06-03
Clean up device and instance creation

doitsujin:
Gets rid of hundreds of lines of unmaintainable garbage and should make adding extensions and features a lot easier and less bug-prone in the future.

Feature enablement is now moved entirely out of the front-ends, and the backend filters out any devices that don't support the required feature set, so that games won't try to create Vulkan devices on outdated drivers or otherwise cursed setups.

Hopefully I didn't forget to enable device features that we're using. Could do with some testing across different drivers.


src/d3d9/d3d9_adapter.cpp
```diff
@@ -907,10 +906,10 @@ namespace dxvk {
 
     const auto& props = m_adapter->deviceProperties();
 
-    m_deviceGuid   = bit::cast<GUID>(m_adapter->devicePropertiesExt().vk11.deviceUUID);
-    m_vendorId     = props.vendorID;
-    m_deviceId     = props.deviceID;
-    m_deviceDesc   = props.deviceName;
+    m_deviceGuid   = bit::cast<GUID>(m_adapter->deviceProperties().vk11.deviceUUID);
```
WinterSnowfall, line 909: `m_adapter->deviceProperties()` is now basically `props`, if I read this correctly.

src/d3d9/d3d9_adapter.cpp
```diff
@@ -946,7 +945,7 @@ namespace dxvk {
         fallbackDesc   = "NVIDIA GeForce RTX 3060";
       }
 
-      bool hideNvidiaGpu = m_adapter->devicePropertiesExt().vk12.driverID == VK_DRIVER_ID_NVIDIA_PROPRIETARY
+      bool hideNvidiaGpu = m_adapter->deviceProperties().vk12.driverID == VK_DRIVER_ID_NVIDIA_PROPRIETARY
```
WinterSnowfall, line 948: Same here.

WinterSnowfall (changes requested):
Some minor nits I found looking over the d3d9 part.

WinterSnowfall (approved):
Did a quick test pass and it seems fine on Nvidia proprietary at least.

PR #4980, opened 2025-06-07
D3D9 Tracking mask cleanup

K0bin:
- Removes a few unnecessary bit masks
- Fixes the DRef clamping for textures used in vertex shaders
- Fixes a bug where per-slot RT hazard mask is getting compared to the previous one `m_lastHazardsRT`, which  only had 4 bits despite it being used to track texture slots rather than RT slots
- Moves texture tracking bit masks into a struct and documents every single one


K0bin:
I've tested this with a bunch of games and all my apitraces and it's fine so far.

Before merging, I'd appreciate it if someone with an AMD GPU could throw this at GTA IV because of the feedback loops in that game. It's ready for reviews nonetheless.

src/d3d9/d3d9_device.h
```diff
@@ -195,6 +195,22 @@ namespace dxvk {
...
+    /* D3D9 allows using 16 vertex buffers ('streams'). */
+
+    /** Whether there's a vertex buffer bound to the slot */
+    uint16_t bound = 0;
+
+    /** Whether the vertex buffer at each slot needs to be uploaded at draw time */
+    uint16_t needsUpload = 0;
+
+    /** Whether the vertex buffer for each slot gets copied at draw time to act like DrawUP */
+    uint16_t uploadPerDraw = 0;
```
doitsujin, line 208: what does this do vs `needsUpload`?

PR #4981, opened 2025-06-07
Fix OSU crash due to null pointer in the backend

### LINKED ISSUE #4769: osu!stable black screens on launch using compatibility mode (DX9)
As soon as osu! is launched using compatibility mode, it just black screens.
After some bisecting, I found out commit https://github.com/doitsujin/dxvk/commit/49b18f03fe7c6dd29e9aa296c5cd072d3765a1bb is breaking it, more specifically just the following lines:

```
// Unbind all buffers that were still bound to the backend to avoid leaks.
    EmitCs([](DxvkContext* ctx) {
      ctx->bindIndexBuffer(DxvkBufferSlice(), VK_INDEX_TYPE_UINT32);
      for (uint32_t i = 0; i < caps::MaxStreams; i++) {
        ctx->bindVertexBuffer(i, DxvkBufferSlice(), 0);
      }
    });
```

Steps to reproduce (osu!stable needs dotnet40 and gdiplus to work properly):
```
WINEPREFIX=~/.osutry winetricks -q -f dotnet20 dotnet48 gdiplus_winxp dxvk
WINEPREFIX=~/.osutry wine osu\!.exe &> logs.txt
```
and enable Compatibility Mode in settings.

### Software information
osu! with Compatibility Mode (DX9) enabled

### System information
- GPU: RX 6600
- Driver: mesa 25.0.1
- Wine version: 10.3
- DXVK version: 2.6

### Log files
[logs.txt](https://github.com/user-attachments/files/19252787/logs.txt)

K0bin:
Doesn't fi.x #4769 but still has some good stuff.

The backend commit should probably be backported to the v2.6.x branch.
I'll do that after the merge.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3645,7 +3645,6 @@ namespace dxvk {
       // D3D9 doesn't actually unbind any vertex buffer when passing null.
       // Operation Flashpoint: Red River relies on this behavior.
       needsUpdate = false;
-      vbo.offset = 0;
     }
 
     if (needsUpdate)
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -493,6 +493,12 @@ namespace dxvk {
       ResetState(pPresentationParameters);
       m_implicitSwapchain->DestroyBackBuffers();
       m_autoDepthStencil = nullptr;
+
+      // Tests show that regular D3D9 ends the scene in Reset
+      // while D3D9Ex doesn't.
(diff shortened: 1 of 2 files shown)


WinterSnowfall (changes requested):
I've checked BeginScene -> Reset -> EndScene on Windows XP and it fails, so removing this is most likely not a good idea. Not sure where that trace was captured, or if a Windows 11 update broke the behavior or whatnot.

Edit: After further testing it seems the in scene state is not reset on device reset on a D3D9Ex device. So just an added check will do.

K0bin:
Good point.

PR #4983, opened 2025-06-07
DXBC: Zero-initialize all private/workgroup variables

doitsujin:
May need some testing for perf regressions, but since the alternative is UB when the game is broken...

Removes a config option to zero workgroup memory and unconditionally relies on the corresponding Vulkan feature.


doitsujin:
Did some perf testing on amd and nv, doesn't seem to have any noticeable impact.

PR #4992, opened 2025-06-09
[d3d9] Relax present interval validations for apitrace replays

WinterSnowfall:
Fixing some apitrace replay issues when traces attempt to do windowed replays with technically invalid presentation intervals. I'm not aware of any games relying on the winnowed mode validations, so we should be good to relax things a bit.


K0bin:
Hopefully we can just close this because the Apitrace fix will get merged.
https://github.com/apitrace/apitrace/pull/948

WinterSnowfall:
Fixed in apitrace.

PR #4998, opened 2025-06-11
[d3d9] Add a modeCountCompatibility config option

### LINKED ISSUE #3255: AquaNox 2 crash on resolution change

WinterSnowfall:
I've resurrected this, because the problem with the AquaNox games has come back to bite me even on my most recent Lunar Lake laptop. It seems that laptops in general (their screens in particular) simply expose too many modes for some games to be happy.

That can usually manifest as:
- higher resolutions are missing from in-game mode selection options
- lower resolutions get cut, which sometimes breaks the first-run configurations of games
- games overflow in their mode count logic and crash or refuse to start

There's always the option of coming up with an allow list of resolutions, similarly to how certain drivers apparently reduce excessive mode counts, but I'd rather have something user configurable, as we can't possibly account for all preferences and hardware combinations.

Edit: Revised version fixes #3255

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -707,6 +707,23 @@
 # dxgi.forceRefreshRate = 0
 # d3d9.forceRefreshRate = 0
 
+# Mode Count Compatibility
+#
+# Attempts to aggressively limit the advertised mode count
+# by only listing the current desktop resolution and refresh
+# rate, along with a minimal list of fallback resolutions.
+#
+# Useful for titles that encounter issues when too many modes
+# are reported, e.g., AquaNox, AquaNox 2: Revelation.
+#
+# Can be used in conjunction with forceAspectRatio
+# and forceRefreshRate to further restrict reported modes.
+#
+# Supported values:
+# - True/False
+
+# d3d9.modeCountCompatibility = False
+
 # Enumerate by Displays
 #
 # Whether we should enumerate D3D9 adapters by display (windows behaviour)
--- a/src/d3d9/d3d9_adapter.cpp
+++ b/src/d3d9/d3d9_adapter.cpp
@@ -10,8 +10,6 @@
 #include "../util/util_ratio.h"
 #include "../util/util_string.h"
 
-#include "../wsi/wsi_monitor.h"
-
 #include <cfloat>
 
 namespace dxvk {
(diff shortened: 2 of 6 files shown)


doitsujin:
To finally give some input here, I think the preferable way to go about this would be to have a list of high-priority display modes, something like
```
native@60hz (±1hz since displays love doing 59.97hz or some shit)
native@native refresh
640x480@60
640x480@highest refresh
800x600@60
800x600@highest refresh
1024x768@...
1280x1024@...
1280x720@...
1920x1080@...
```
and a config to limit the mode count based on that? Just need to make sure we filter first and *then* sort.

If we do this, I'd want it to be done in such a way that we can ship it as app profiles.

WinterSnowfall:
> To finally give some input here, I think the preferable way to go about this would be to have a list of high-priority display modes, something like
> 
> ```
> native@60hz (±1hz since displays love doing 59.97hz or some shit)
> native@native refresh
> 640x480@60
> 640x480@highest refresh
> 800x600@60
> 800x600@highest refresh
> 1024x768@...
> 1280x1024@...
> 1280x720@...
> 1920x1080@...
> ```
> 
> and a config to limit the mode count based on that? Just need to make sure we filter first and _then_ sort.

I'd build that list dynamically, considering if the refresh rate filter is used as well. But sure, sounds like a plan.

> If we do this, I'd want it to be done in such a way that we can ship it as app profiles.

Fair enough. I guess with this approach, if people really want to use a certain obscure resolution, they'd have the option to mode switch to it first. Do we have a way of reliably getting the native resolution/refresh rate, btw? :sweat_smile:

doitsujin:
`wsi::getDesktopDisplayMode` is as reliable as it gets.

WinterSnowfall:
@doitsujin Alright, I followed mostly what you were hinting at, but I kept the current filtering and sorting order intact, since games will expect that anyway. Perhaps it's not the most elegant way of filtering, but it works with the least amount of disruption.

The only concern I have now is that if the current desktop refresh rate gets hits (supported modes) with any of the "compatible" resolutions, that will inflate the overall mode count. An overall maximum would be: (1 current resolution + 6 compatibility resolutions) * 2 refresh rates * 3 formats = `42`. The answer to life, the universe and everything, perhaps, but above a hopeful 32 max I was aiming for.

Do you have any objection to only keeping 60 Hz for the compatibility modes? That would lower the max count to (1 * 2 + 6 * 1) * 3 = `24`. If people want higher/different refresh rates on any of the compatibility modes, they need only mode switch to them before starting the game.

24 modes at least seems to make the first two AquaNox games happy (will add built-in configs for them too soon).

Edit: Did the 60 Hz only for compatibility modes change already, since the entire point of the option is to reduce the mode count as aggressively as possible, while not messing with the initial configuration/expectations of games.

PR #5000, opened 2025-06-11
[util] Disable direct image mapping for LEGO City Undercover

### LINKED ISSUE #4997: Video playback regression on LEGO City Undercover
On the game [LEGO City Undercover](https://store.steampowered.com/app/578330/LEGO_City_Undercover/), cutscenes are rendered with half of the screen being green due to some bug. This is the first cutscene of the game:

Image

It seems that it also happens on native Windows, because this guide mentions DXVK to fix the bug: https://steamcommunity.com/sharedfiles/filedetails/?id=2889353226

It also seems to be exclusive to AMD.

I can't neither confirm if it also happen on Windows nor if it's exclusive to AMD, but I can confirm it's a regression because downgrading DXVK versions fix it.

### System information
- GPU: AMD Radeon RX 6600
- Driver: Mesa RADV 25.0.7
- Wine version: [GE-Proton10-4](https://github.com/GloriousEggroll/proton-ge-custom/tree/ef7ea60c1c8adfdd4eec622569671ad7882838d9)
- DXVK version: commit https://github.com/doitsujin/dxvk/commit/c1aef5b4f9f324f47f6d5e893c51781b2138c4f9

I then tested some stable DXVK versions (by putting dxgi and d3d11 DLLs in the game folder) and found that on version 2.4.1 the green cutscene bug does not occur, but it occurs on version 2.5.

On version 2.5 with `dxvk.enableMemoryDefrag = False` the problem is still present.
Also tried `dxgi.hideAmdGpu = True` but no success either.

### Apitrace file(s)
Unable to create Apitrace file because the game crashes when trying to load the cutscene on WineD3D.

### Log files
[lego_logs_dxvk.tar.gz](https://github.com/user-attachments/files/20684437/lego_logs_dxvk.tar.gz)

Tiagoquix:
Closes #4997

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -60,7 +60,7 @@ namespace dxvk {
     /* D3D11 GAMES                                */
     /**********************************************/
 
-    /* Batman Arkham Knight - doesn't like intel vendor id 
+    /* Batman Arkham Knight - doesn't like intel vendor id
       (refuses to boot if vendor isn't 0x10de or 0x1002)  */
     { R"(\\BatmanAK\.exe$)", {{
       { "dxgi.hideIntelGpu",                "True" },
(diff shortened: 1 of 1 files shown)


doitsujin:
For future reference, please avoid unrelated formatting changes in the same commit since that makes things very annoying to backport / revert.

PR #5001, opened 2025-06-11
Fix crash with in destroySwapchain with nullptr m_swapchain

canadacow:
Fix crash (in Bandicam/other Vulkan layers) when destroySwapchain is  called but m_swapchain is already null. Hit this pretty quickly with bandicam and the Superposition demo.


doitsujin:
This is not a bug, passing null handles to `vkDestroy*` methods is legal. I *guess* we can merge this, but in that case, please add a comment that this is a workaround for broken third-party Vulkan layers.

canadacow:
Roger wilco. Will add the comment. Yeah, annoying as heck to see bandicam yet again in my stack trace. :-(

PR #5009, opened 2025-06-13
Enable descriptor buffers on ANV

doitsujin:
Draft because we really need more data here.


HunterCZ122:
I tested few games on B580 with descriptor buffers:

A Hat in Time, Max settings, Dead Bird Studio Act 1 - from 218fps to 209fps, more GPU bound with descriptor buffers
The Witcher 3, Ultra High, White Orchard - from 53fps to 52fps, can't really tell which runs better, GPU bound
LEGO Star Wars The Skywalker Saga - from 163fps to 159fps, GPU bound
GTA 4, High settings - fps is all over the place (ranging from 100fps to 250fps), but it rarely drops below 100fps with descriptor buffers, CS thread is constantly at 100% and mostly CPU bound
Need for Speed Payback, Ultra settings - from 119fps to 115fps, GPU bound

Mesa 25.2.0-devel latest git as of today + [34872](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/34872)

doitsujin:
Thanks for the data. Bit unfortunate about the GPU-bound regressions, it's a bit much for my liking at least in some cases, but as far as I understand, that's also not going away when we move away from the legacy path.

PR #5015, opened 2025-06-14
[d3d8/9] Clean up CheckDeviceFormat and CreateImageSurface formats

WinterSnowfall:
Clean up mostly of d3d8 format support on `CheckDeviceFormat` and `CreateImageSurface`, which addresses the following corner cases:
- d3d8 `CheckDeviceFormat` shouldn't report support for d3d9-exclusive formats
- `CreateImageSurface` should work even with unsupported formats, including lockable depth formats ~(to be confirmed on native)~, similarly to how `CreateOffscreenPlainSurface` with `D3DPOOL_SCRATCH` behaves in d3d9
- volume textures can't be used as render targets, yet we were advertising support (fixed)... some native drivers seem to do that too for some reason, although creation will fail on them as well
- `W11V11U10` is not available in d3d9, so should not be advertised outside of d3d8 (this brings us in line with native and WineD3D)
- `A2B10G10R10_XR_BIAS` does not appear to be used or supported on native, so removed support for it accordingly
- a more limited selection of render target formats for d3d8, to be in line with age-accurate drivers (modern drivers seem to expose all that d3d9 exposes, which is... hardly ideal).

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -331,6 +331,9 @@ namespace dxvk {
     if (unlikely(ppTexture == nullptr))
       return D3DERR_INVALIDCALL;
 
+    if (unlikely(isD3D9ExclusiveFormat(Format)))
+      return D3DERR_INVALIDCALL;
+
     // Nvidia & Intel workaround for The Lord of the Rings: The Fellowship of the Ring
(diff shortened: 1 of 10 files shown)


WinterSnowfall:
Found a derp I left in with vendorId specific format checks that affected all adapter operations (ergo format check calls), however actual creation calls got the right info. *Sigh*, but fixed now.

Issue #4895, opened 2025-05-04
Source Engine Performance Loss under DXVK


ward201185:
DXVK doesnt seem to be performing 1 to 1 with native d3d9 under windows with Source Engine games, the same performance loss is present under linux.
im getting a consistent 60-100+ fps loss under TOGL and DXVK with this card whereas when i tried a different nvidia card (gtx 1650), TOGL was MUCH closer to native windows performance while dxvk still performed worse than TOGL and native windows.
i have an i5 4460 and 16 gb ddr3 ram, i would like to know why this occurs and if this problem applies to other gtx960 users as well?
TOGL only does slightly better under the gtx 960 but theres still a huge performance loss.
this was always a problem i had under linux, i never had it running as well as on windows natively.
unfortunately even on ubuntu 14.10 with the first driver released for my card has this problem.

i tried other games on dxvk under linux to see if it was an issue on my end and performance seems to be on parity with windows.
### games in below that i have only tried with dxvk currently:

GTA San Andreas v1.0
Call Of Duty 4 Modern Warfare Demo
Battlefield 3

any clues as to why this is happening?

### Software information
Likely Any Source engine game on Maxed Out settings
1920x1080 60HZ

### System information
- GPU: NVIDIA GeForce GTX 960 2GB
- Driver: 572.16 (under windows)
- DXVK version: Likely all are affected (i already tried 3 years ago and i dont think this has changed)

i will give other necessary information if requested or need be.

K0bin:
Maybe start by naming a particular game. Then tell us whether the problem is CPU performance of GPU performance (is GPU usage 100%?)

In my experience DXVK runs Source games excellently and that's what you'd expect considering most of the DXVK d3d9 code was written by someone with source engine experience.

ward201185:
i currently have screenshots when i tried it with the gtx 1650 but the cpu and gpu usage is not any different from the gtx 960.
Windows:

![Image](https://github.com/user-attachments/assets/ca9f8a7d-a684-4e83-b0b0-2a554dbb3109)

![Image](https://github.com/user-attachments/assets/edd6da63-fe55-4a7f-bfe9-6a56d6f4377f)
Linux:

![Image](https://github.com/user-attachments/assets/ef975819-83eb-4c13-90df-931fade2355b)

![Image](https://github.com/user-attachments/assets/cc97b84d-e853-40c2-a7a0-7f7f59b35776)

Blisto91:
[...]

PR #5023, opened 2025-06-17
[d3d9] Optimize StretchRect stretching resolves

K0bin:
Improves #4895 when using MSAA in Source.

HL2 frequently resolves a 2048x1536 screenshot into a 1024x1024 image which is then used for refraction when rendering glass.

The game does these resolves (using StretchRect) quite frequently and unfortunately a lot of them are redundant.
image

Current DXVK master would first resolve the original texture to a temporary image and then blit that to the actual target texture.
While that's actually what the D3D9 docs suggest happens behind the scenes, it's a bit slow:
> Note that use of the extra surface involved in using StretchRect to downsample a Multisample Rendertarget will result in a performance hit.

This PR changes it so the blit and resolve are done in one step rather than in two. It improves performance in the HL2 main menu at 4k Max settings with 8x MSAA from 240 fps to 265 fps.

Besides that, the PR also forbids creating a lockable texture with more than 1 sample. This matches what I've seen in tests done on Windows.

In the future I'd like to get rid of the resolve step in `LockImage` but for now we still need it because of the GDI-based presentation fallback.

### DIFF
--- a/src/d3d9/d3d9_common_texture.cpp
+++ b/src/d3d9/d3d9_common_texture.cpp
@@ -166,8 +166,9 @@ namespace dxvk {
        || (blockSize.Height && (pDesc->Height & (blockSize.Height - 1))))
         return D3DERR_INVALIDCALL;
     }
-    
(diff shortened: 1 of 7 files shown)


Isaac-Clarke:
Thank you for your work.

I am happy to have helped.

PR #5031, opened 2025-06-19
Fix usage of copyFeature

canadacow:
Usage of `copyFeature` is inconsistent in `initSupportedFeatures` when `deviceInfo` is provided. This happens in D3D11on12 scenarios where the device is reused. `copyFeature`'s signature is `const void*` not `const void**` as the ref to `deviceInfo->pNext` suggests. The first `copyFeature` succeeds because the usage is correct. The `CORE_VERSIONS` define crashes immediately.

Found testing on Windows with the  https://github.com/microsoft/directx-graphics-samples/tree/master/Samples/Desktop/D3D12VariableRateShading sample. It uses D3D11on12 and crashes in the `D3D11On12CreateDevice` call.

### DIFF
--- a/src/dxvk/dxvk_device_info.cpp
+++ b/src/dxvk/dxvk_device_info.cpp
@@ -343,8 +343,8 @@ namespace dxvk {
 
       copyFeature(deviceInfo->pNext, nullptr, &m_featuresSupported.core);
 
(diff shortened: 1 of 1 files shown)


doitsujin:
nice catch, 11on12 is basically untested on our end since there's like next to nothing using it in the wild.

Issue #4834, opened 2025-04-06
forceSampleRateShading blurs 2D sprites


Arthandas:
I wanted to use DXVK to force supersampling in Persona 4 Arena Ultimax. I set d3d9.forceSampleRateShading = True and that's the only setting I changed.
The supersampling works as expected by clearing aliasing from the 3D backgrounds:

![Image](https://github.com/user-attachments/assets/8a451766-b643-4d39-8279-c38c72712f9b)
![Image](https://github.com/user-attachments/assets/c758fb8e-9c91-4f3d-b7b1-db75cf12157a)

Sadly it also blurs the 2D sprites, which as far as normal supersampling goes (like setting r.ScreenPercentage=200 in Unreal Engine games), shouldn't be affected (just like the UI):

![Image](https://github.com/user-attachments/assets/38e12cc3-78f9-4a40-9b44-30320f11a2db)
![Image](https://github.com/user-attachments/assets/e88654a8-60ce-403e-a4cb-07ed483fb171)

Is there any setting I can use to keep the supersampled 3D background without touching sprites/UI?

Btw, I'm using DXVK 2.6 on Win11.

doitsujin:
The whole point of supersampling is to compute *everything* per sample, including textures, so this is expected.

Arthandas:
I don't understand, I used supersampling (via Nvidia Profile Inspector) in many other games, including other 2.5D titles with 2D sprites on 3D backgrounds like Melty Blood Type Lumina or Under Night In-Birth and the sprites/UI were untouched, actually this is the first time I see it affecting the 2D elements.
Here's an example of me forcing it in Melty Blood:
https://cdn.knightlab.com/libs/juxtapose/latest/embed/index.html?uid=cb72aed8-21ac-11ed-b5bb-6595d9b17862

K0bin:
> I used supersampling (either via Nvidia Profile Inspector or r.ScreenPercentage in Unreal Engine) 

Let's start with Unreal Engine. It's a game engine so it has a lot more information about what each draw does. Because of that it can easily do super sampling only for the stuff where it's beneficial.

The Nvidia profile inspector likely has a way more sophisticated, hacky solution than DXVK. Those profiles behave differently for different games so it's both a ton of effort to build something like that and a giant mess.

We're not interested in making this any more complicated than it is right now. That should be left to mods.

Arthandas:
[...]

PR #5032, opened 2025-06-19
[util] Adjust the behavior of the forceSwapchainMSAA option

WinterSnowfall:
A more sane approach to the forceSwapchainMSAA debacle, at least in my humble opinion. In addition to actually checking for hardware support of the provided sample counts, it only does anything if MSAA is enabled in the first place, acting more like a sample count override than a MSAA injector. Because there are games that *will* break if the config option is used globally/indiscriminately, and we've certainly seen people do just that.

@K0bin FYI, hopefully this is a fair compromise to removing it altogether.

Needs a bit of testing, especially when used to disable MSAA, after which I will undraft.


K0bin:
> it only does anything if MSAA is enabled in the first place

That makes it pretty much useless. The option was used to force MSAA in games that don't already have the option. The ones that do almost always let you pick the sample count.

WinterSnowfall:
> That makes it pretty much useless. The option was used to force MSAA in games that don't already have the option. The ones that do almost always let you pick the sample count.

Older d3d8/9 games usually cap out at 4x (which was a hardware limitation at the time), but if you feel outright removing it makes more sense, I can do that.

Edit: Done, and agreed most games will allow some form of control, which we should ultimately respect. The difference in effect is mostly negligible on modern resolutions. We're not playing games @ 1024x768 any more, thankfully.

ChthonVII:
> Older d3d8/9 games usually cap out at 4x (which was a hardware limitation at the time),

Um... I was using this precisely for that purpose, to override the MSAA to a higher factor than the in-game setting allowed.

I don't think "some people might misuse it" is a good reason to get rid of a working feature that some people were using. If anything, just document that it's meant as a "sample count override."

K0bin:
The problem is that this option, as it was before, only worked with games made before around 2004.

If the game did _any_ off screen rendering, the option would just break things. So it only worked in games that you could easily just run with 4x super sampling anyway.

ChthonVII:
> The problem is that this option only worked with games made before around 2004.

So what? That was exactly my use case. You had a feature that worked with some games and did exactly what people wanted for those games. And now you've removed it for, in my view, a very poor reason.

Again, I strongly feel that the correct solution to people potentially misusing a feature is to document it better, not to remove it.

ChthonVII:
> So it only worked in games that you could easily just run with 4x super sampling anyway.

You edited your comment after I responded.

By "super sampling" I assume you mean `forceSampleRateShading`. That [has issues with 2D elements](https://github.com/doitsujin/dxvk/issues/4834), particularly small fonts. It's also prohibitively demanding for a lot of hardware. So, no, it's not an equally good alternative.

WinterSnowfall:
> Again, I strongly feel that the correct solution to people potentially misusing a feature is to document it better, not to remove it.

The option was objectively breaking some games, and we've seen people applying it globally then raising issues with the expectation it should always be supported and working properly, which was never the case. Unfortunately you can't expect users to not abuse what you expose these days.

I'll remind you the main goal of dxvk is to ensure correct rendering behavior in games, not to act as an injector of features which were never supported in the first place. That is the place of mods and other tooling.

I'd also keep in mind MSAA support is very much present even in most d3d8 games, and if it is missing there's usually a strong reason for it.

My recommendation would be to use post-processing AA through the "magic" of Vulkan layers in its stead. That would work universally, with all games, regardless of MSAA support, and is generally even cheaper computationally. Some games will also have the option of unofficial patches/mods to add in MSAA and modern resolution support, but that usually requires changes to the game's engine which are simply not possible to replicate with dxvk without causing some form of collateral damage.

> Um... I was using this precisely for that purpose, to override the MSAA to a higher factor than the in-game setting allowed.

Unless you're playing those games at very low resolutions, the differences between 4x and 8x are rather marginal from 1080p upwards. I guess there are a few cases where games only have an AA on/off toggle that lands you on 2x, but that's pretty rare.

K0bin:
> By "super sampling" I assume you mean forceSampleRateShading. That has issues with 2D elements, particularly small fonts. It's also prohibitively demanding for a lot of hardware. So, no, it's not an equally good alternative.

No, I meant running the entire game at a higher resolution using something like Gamescope.

ChthonVII:
> No, I meant running the entire game at a higher resolution using something like Gamescope.

Gamescope's got a whole raft of its own bugs and limitations.

K0bin:
That doesn't change the fact that DXVK is not meant to be a modding tool.

TRPB:
This is a very annoying change.

It seems pointless to remove it, many ~2003 games don't have an MSAA option but work it when it's forced on. This option used to work fine for a lot of games and I don't expect it to be flawless given the technologies involved but why remove it entirely?

Now there's no way (AFAIK) to force MSAA in these games where previously this option just worked most of the time

> Because there are games that will break if the config option is used globally/indiscriminately, and we've certainly seen people do just that.

That seems like a silly reason to remove it. People have to manually enable this option so if stuff breaks it's on them

Can we make it an environment variable instead? That way you have to put it in the launch command so it's more obviously enabled per-game.

It's a shame to get a feature that does work fine in a lot of cases

doitsujin:
We can only lose here, removing the option is going to annoy people and keeping the option is going to annoy a different set of users and also lead to annoying bug reports. The correct decision would have been to never implement it in the first place.

Either way, just use gamescope with a custom resolution instead. If the game UI doesn't scale you're kinda screwed at anything higher than 800x600 anyway.

WinterSnowfall:
> That seems like a silly reason to remove it. People have to manually enable this option so if stuff breaks it's on them

There have been several bug reports raised by people with this option enabled (globally). I've explained it in a post above. While I would have been open to keep it as a sample override perhaps, keeping it as is was not on the table. It causes more harm than good in plenty of games.

I understand how this is annoying to people who have come to rely on it, but it was wrong to expose it in the first place IMHO.

> Can we make it an environment variable instead?

That wouldn't solve anything. People would simply make it a global environment variable and we'd be back where we started. The painful but correct option is switching to either some form of super resolution or post-processing AA if you can't/won't rely on the former.

TRPB:
It works well for games like the original 3d era GTAs and Freelancer.

Is forceSwapChainMSAA =4 identical to rendering at 4x and scaling down though? I don't know the internals but that seems like it would be super slow even for a nearly 25 year old game, I'm playing the games above with it set to 8 and at 4k (looks great), I can't imagine rendering at that resolution is really feasible, though maybe that is how it works?

That argument could be made for any of the options in dxvk.conf though, a lot of which will cause (albeit probably less severe) side-effects. Wouldn't the better option to be to only load settings from a config when `DXVK_CONF` is actually set? 

I _suppose_ it doesn't really matter because it's a compiled DLL I can forever use 2.6.2 with `DXVK_CONF` set to a file with the option set but it's a bit of a pain to remember which version is needed for the feature in several year's time and won't remember what the option is called when I try to set something up again in 3 years.

Could we keep a note in the sample `dxvk.conf` and have it say something like:

```
# Overrides the application's MSAA level on the swapchain
# 
# Supported values: -1 (application) and 0 to 16 (user override)
#
# REMOVED. See https://github.com/doitsujin/dxvk/pull/5032
# Use 2.6.2 or earlier if you wish to use this feature
# 
# d3d9.forceSwapchainMSAA = -1
```

At least then anyone wondering why it suddenly stopped working will have some context and a workaround

TRPB:
one other option, what about a registry setting? That would have to be done at the prefix level rather than globally for anyone using Wine but would have the same global effect on Windows

WinterSnowfall:
> At least then anyone wondering why it suddenly stopped working will have some context and a workaround

It will be mentioned in the changelog of the next release. I know some people will be very vocal about it, but the reality is the majority of users will be unaffected.

> one other option, what about a registry setting? That would have to be done at the prefix level rather than globally for anyone using Wine but would have the same global effect on Windows

dxvk doesn't use the Windows registry, nor does it officially support Windows.

> Is forceSwapChainMSAA =4 identical to rendering at 4x and scaling down though?

No AA method is identical to any other AA method, they all have their benefits and limitations. If you are playing those games at 4K, which certainly isn't supported in the vanilla releases, I expect you have other options for AA on the table anyway. If you even feel the need for AA at that resolution - I certainly can't say I do.

TRPB:
> If you even feel the need for AA at that resolution - I certainly can't say I do.

It's night and day even at 4k, especially in these older games. Notice the stepping around the round part of the space station:





That's not zoomed at all just part of a screenshot.  Worse is motion as you get a lot of the shimmering effect that AA removes. Requires an ini file change to set the resolution in Freelancer but otherwise works.

the `forceSwapchainMSAA` solution was by far the simplest. It may be possible to find some enb dll that works for some games but `forceSwapchainMSAA` just worked whenever I tried it. Removing it entirely definitely feels like a step backwards, especially given dxvk supports it outside of an option being exposed and that there's no obvious alternative

WinterSnowfall:
> there's no obvious alternative

More than one alternatives have already been suggested, not sure why you feel you are out of options. Perhaps none are as straight forward as `forceSwapchainMSAA`, but keeping a half-broken option around just because it works *sometimes* is not exactly a step forward either.

TRPB:
Can we document the alternatives here?

- Rendering at 8k and downsampling to 4k using gamescope, looks significantly worse than MSAA and performance suffers (obviously). Gamescope doesn't really support downscaling properly which is why it doesn't look great: https://github.com/ValveSoftware/gamescope/issues/692

- vkbasalt supports SMAA and FXAA both of which failed to remove the stepping at all in the scene above and generally look worse than MSAA

Surely something that works _sometimes_ is an improvement on something that works _never_?

edit: I have found a working solution: https://github.com/elishacloud/dxwrapper

Install the dxvk dlls into the prefix's system32 and the dxwrapper dlls in the game's directory. For dx8 games, enable the 8to9 wrapper in dxwrapper and enable antialiasing for dx9 I don't know what you'd do on windows if you wanted to use both though

mirh:
Whatever your linux environment is not allowing for good downsampling (like VSR/DSR/gedosato or whatnot do on windows) doesn't seem like an *us* problem.
On the other hand though I also wanted to point out that:

> Older d3d8/9 games usually cap out at 4x (which was a hardware limitation at the time), but if you feel outright removing it makes more sense, I can do that.

Older games were always overridable via the driver (which in fact, even to this day, have a global forcing option).
Both with a setting to altogether supplant the game AA, but also to augment/enhance it with just more samples (or with SSAA for regions with alpha). So in a sense I see why it doesn't belong in dxvk. 

Though in another I'm also skeptical that it could/should be included in mesa (maybe when it still had nine, but now it doesn't have an interface with the native code). And sure, many games required compatibility bits.. but still? It was not only doable but also often done.
Assume I wanted to reinvent the wheel with TrSGSSAA.. where else could it even be implemented?

WinterSnowfall:
> Both with a setting to altogether supplant the game AA, but also to augment/enhance it with just more samples (or with SSAA for regions with alpha).

That is true, however on most hardware that was still limited to 4x for quite some time (up to the 2010s, I'd say). Drivers also had way more advanced heuristics for overriding MSAA than just force enabling it on a swapchain, so any comparison of drivers vs dxvk here ends up in apples to oranges territory.

> Assume I wanted to reinvent the wheel with TrSGSSAA.. where else could it even be implemented?

The easy answer, when dxvk is in the picture at least, is in a Vulkan layer. Generally, if a game is missing MSAA it usually can be (and most likely already has been) added in with a game-specific mod, with various degrees of difficulty.

ChthonVII:
> We can only lose here, removing the option is going to annoy people and keeping the option is going to annoy a different set of users and also lead to annoying bug reports.

> There have been several bug reports raised by people with this option enabled (globally)... It causes more harm than good in plenty of games.

> I know some people will be very vocal about it, but the reality is...

I think this attitude reflects a skewed sample. You personally see the bug reports when some dipstick enables forced MSAA globally when they shouldn't and mistakes it for a bug. You generally do not hear anything from the people who use this feature correctly and without problems. Without telemetry to tell you otherwise, the first group looms larger in your thinking -- especially because they are annoying to you personally -- while the second group is probably much, much bigger in reality. "I don't want to deal with bug reports from a minority of fools who misuse the feature" remains a terrible reason to remove a feature for everyone else.

Since reducing bug reports from fools is the real motivation behind this change, I fully support TRPB's idea to just stick the feature behind an extra gate like an env variable.

> People would simply make it a global environment variable and we'd be back where we started. 

If people are forced to read the docs to figure out how to get past the gate, then they should hopefully read the part explaining the limited use case for this feature. Moreover, you can name the env variable something like `YES_I_REALLY_WANT_TO_FORCE_MSAA_IN_THIS_PARTICULAR_INSTANCE_AND_I_PROMISE_I_WONT_FILE_A_BUG_REPORT=1`. No one is going to be stupid enough to make that global without realizing they're not supposed to. For that matter, you could just gate it behind an extra conf option with a similar name, with the documentation explaining the limited use case right there next to it in dxvk.conf.

> More than one alternatives have already been suggested

And all of them are more broken than forcing MSAA in DXVK. Yes, this feature has a narrow use case, but, for that use case, it is the perfect solution.

WinterSnowfall:
> You generally do not hear anything from the people who use this feature correctly and without problems. Without telemetry to tell you otherwise, the first group looms larger in your thinking -- especially because they are annoying to you personally -- while the second group is probably much, much bigger in reality. "I don't want to deal with bug reports from a minority of fools who misuse the feature" remains a terrible reason to remove a feature for everyone else.

I expect most dxvk users will not actually care much about this option, but again, this isn't really a numbers game, the problem is the option outright _breaks_ rendering in some games. Sure, this can be attributed to user misuse, however one could equally argue that if it's half-broken then it shouldn't be exposed in the first place. If you can't understand this, I don't know what much else there is left to say.

> And all of them are more broken than forcing MSAA in DXVK.

Sorry, but no, that's simply not true. They may not *look* as good as MSAA in certain situations, but post-processing AA at least will NOT break games that mess with depth in other ways, for example.

TRPB:
>  the problem is the option outright breaks rendering in some games. 

- `dxvk.deviceFilter` - Will stop games launching entirely if set improperly
- `dxgi.enableDummyCompositionSwapchain` - Will likely break some games if set globally
- `dxvk.allowFse` - Breaks alt tab
- `d3d11.relaxedBarriers ` - can break rendering
- `d3d11.relaxedGraphicsBarriers ` - can break rendering and even says "Please do not report bugs with this option enabled"
- `d3d11.forceComputeUavBarriers` - Reduces performance if enabled, will it be removed when someone creates a performance related but report with it enabled globally?
- `d3d11.disableDirectImageMapping` - Can break rendering

Each of these (and probably others) can lead to broken rendering and bug reports if enabled.

Since a lot of them are to fix issues with specific games, having them as global options is likely to have the same problem you're trying to solve by removing `forceSwapchainMSAA` yet they get to stay?

Having a global config for a lot of these options seems like the issue here as they're there to fix game specific issues. It's easy to set one and then forget it's set when you play something else

WinterSnowfall:
> Since a lot of them are to fix issues with specific games, having them as global options is likely to have the same problem you're trying to solve by removing `forceSwapchainMSAA` yet they get to stay?

For now at least. Ideally we should not expose problematic options, however sometimes that is required for *proper game functionality*. MSAA is not a feature that is required for functionality, it is cosmetic.

> Having a global config for a lot of these options seems like the issue here as they're there to fix game specific issues. It's easy to set one and then forget it's set when you play something else

Many of the above are actively used as built-in configs for some titles, so removing them outright is not really possible at the moment, while `forceSwapchainMSAA` was not.

ChthonVII:
> MSAA is not a feature that is required for functionality, it is cosmetic.

You're seriously proposing a drawing a line between "functional" and "cosmetic" with respect to visual media? That's ridiculous.

> Many of the above are actively used as built-in configs for some titles, so removing them outright is not really possible at the moment

How is it that "someone is using that" a sufficient reason to refrain from removing those conf options, but not this one?

WinterSnowfall:
> You're seriously proposing a drawing a line between "functional" and "cosmetic" with respect to visual media? That's ridiculous.

No, it's called engineering.

> How is it that "someone is using that" a sufficient reason to refrain from removing those conf options, but not this one?

It's not "someone is using that", it's "a game requires it to perform adequately and we couldn't come up with a better option yet".

K0bin:
DXVK is not a modding tool, it's a compatibility tool to make games playable on Linux.

The option only ever worked with very old games (early 2000s). This makes it misleading because users expect it to have the same level of compatibility as the forced MSAA options of the Nvidia or AMD Windows driver. On top of that rather than just doing nothing in the worst case, it can actually break games.

So it's both very limited in its usefulness and confusing to users. That is why we decided to remove it.

mirh:
> Sorry, but no, that's simply not true. They may not look as good as MSAA in certain situations, but post-processing AA at least will NOT break games that mess with depth in other ways, for example.

Sorry but no that's such a skewed view. I do agree that there are methods that will always unquestionably be better than MSAA (I suppose there is also SGSSAA requiring it, and the performance aspect.. but let's put that aside) namely downsampling/SSAA... but post-processing AA ***really*** isn't it. Even the highest quality SMAA (which somehow is rare to get a hold of) still has questionable edge detection quality and whatnot.

> You're seriously proposing a drawing a line between "functional" and "cosmetic" with respect to visual media? That's ridiculous.

I think there are reasonable alternative to wholesale removal (like, at worst it could just be banned from being enabled globally), but that distinction seems very sensible to me still.

WinterSnowfall:
> Sorry but no that's such a skewed view. 

What I meant to convey is that post-processing AA will never *break* rendering in a game. That is purely factual...

> I do agree that there are methods that will always unquestionably be better than MSAA (I suppose there is also SGSSAA requiring it, and the performance aspect.. but let's put that aside) namely downsampling/SSAA... but post-processing AA _**really**_ isn't it. Even the highest quality SMAA (which somehow is rare to get a hold of) still has questionable edge detection quality and whatnot.

... but now we are getting into subjective territory. I also never said post-processing AA is better looking than MSAA, I said it won't ever cause rendering problems, so it can be used reliably with all games, unlike forcing MSAA.

> I think there are reasonable alternative to wholesale removal (like, at worst it could just be banned from being enabled globally).

Unless you can come up with a way that always makes it behave reliably and not break any games, I fail to see that reasonable alternative. The only one I could come up with was changing it to act like a sample count override, but @K0bin made the fair point that it made very little sense to keep it only for that, which is why it got removed.

We've also had people force sample counts that are not supported by the Vulkan driver (e.g. 16x on Nvidia), which was causing crashing, but arguably that was preventable (and I had addressed it in the first iteration of the PR). Ergo, relying on all users to behave responsibly in also never a "reasonable alternative".

ChthonVII:
> The only one I could come up with was changing it to act like a sample count override,

While I think the right course of action is to restore the prior functionality entirely, a sample count override would satisfy my most common use case personally.

TRPB:
For me the vast majority of games I use this feature on are DX8.

Could it be a DX8 only option? They're significantly less likely to use offscreen rendering (is it even possible in DX8?)  and I don't ever remember seeing an AA option in a DX8 game

WinterSnowfall:
> Could it be a DX8 only option?

D3D8 is not exempt from deferred shading techniques and therefore MSAA issues. Hitman: Contracts is actually one of the games several people have reported issues with, consisting of overlapping geometry and clipping. They turned out to be caused by, you guessed it, `forceSwapchainMSAA` use.

> I don't ever remember seeing an AA option in a DX8 game

Huh? It's not uncommon at all, albeit a bit rarer than in d3d9. I've attached even an example of a d3d8 game supporting 8x AA (which is indeed rare, most cap out at 4x).

<details>
<summary>4x</summary>

DFBHD

</details>

<details>
<summary>8x</summary>

8X

</details>

I could go on posting, actually, but hope this is enough to prove the point.

mirh:
> Unless you can come up with a way that always makes it behave reliably and not break any games, I fail to see that reasonable alternative

Now that seems some odd standard.
I do agree that you can't compare a "fancy feature" with a "technical requirement" (even though it's debatable how much performance is either) but that's also what happens with half of the other dxvk.conf knobs.

And once you ban it from applying globally.. what is even left? People \*specifically\* adding it per-game, and then forgetting it?
Let's even say somebody is this dumb, but at that point even setting `d3d9.floatEmulation=True` because "faster" is written there could fly (and you have a dozen options like that). So whether the average joe is borderline close to being this brainless (and a sea of pointless bug reports happens every week/month anyway) or the baseline is people with a modicum of sense I struggle to see why this would create a noticeable change in garbage feedback.

ElTioRata:
As a replacement please consider adding [this](https://github.com/GameTechDev/CMAA2)

WinterSnowfall:
> As a replacement please consider adding [this](https://github.com/GameTechDev/CMAA2)

You can use various such AA techniques with layers such as ReShade and vkBasalt. As we've mentioned before, this is out of scope in dxvk.

ElTioRata:
I disagree. ReShade doesn't work for all games (specifically the one that I'm running), plus a built-in post-process AA solution is handy.

K0bin:
DXVK is not a modding tool, it's a compatibility tool. Having built in shader AA is out of scope.

ElTioRata:
Alright then suggest alternatives that aren't either ReShade or some Linux-exclusive application to inject CMAA.

WinterSnowfall:
@ElTioRata I've answered this already, see here: https://github.com/doitsujin/dxvk/issues/5075#issuecomment-3048521779

K0bin:
> Alright then suggest alternatives

It's not our job to find mods for you.

PR #5034, opened 2025-06-20
[util] Don't count losable resources for Astebreed

WinterSnowfall:
Otherwise any graphical change gets you this lovely thing:

Astebreed


K0bin:
Does the game run into this error on Windows?

If not, we shouldn't use the option as there might be a genuine DXVK bug in here which we should investigate instead.

WinterSnowfall:
> Does the game run into this error on Windows?

Allegedly it does, yet I can't confirm it myself.

PR #5038, opened 2025-06-21
[dxvk] Error out at build time if AVX is enabled

doitsujin:
@WinterSnowfall made the good point that we should at least give users a warning when using unsupported build flags.

AVX is a no-no because it GCC will generate code that requires 32-byte stack alignment, but we're never going to get more than 16 (or 4 on 32-bit builds). Native builds and MSVC *should* be unaffected, but it's not tested.

### DIFF
--- a/src/util/util_bit.h
+++ b/src/util/util_bit.h
@@ -12,6 +12,9 @@
 
 #ifdef DXVK_ARCH_X86
   #ifndef _MSC_VER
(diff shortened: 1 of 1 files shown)


Ammar-Sadaoui:
how about clang ?

WinterSnowfall:
> how about clang ?

Clang and MSVC *should* be fine. That being said, nobody here is testing clang/MSVC builds.

PR #5044, opened 2025-06-24
[dxvk] Disable descriptor buffer on older Nvidia

### LINKED ISSUE #5043: [Regression] Massive FPS drop in dxvk with the latest commits
I was testing a few proton forks for some game fixes and I noticed a huge frame drops across all of my tested games (Zenless Zone Zero, Control...). After some digging, when I switch back to Release 2.6.2 of dxvk, the FPS returns to normal values, however when using the latest commit builds, I lose around 1/3rd of the baseline FPS (regardless of the wine/proton version).

Sample of FPS Drops on tested games:

- ZZZ: from 70 to 50
- Control: from 40 to 28

### Software information
Control: Medium settings, no motion blur
ZZZ: Low, x0.8, no motion blur, no bloom

### System information
- GPU: Quadro M4000
- Driver: 570.153.02 (proprietary)
- Wine version: Wine10.10, Proton10
- DXVK version: ~~2.6.2 and e4989d1~~ master

### Apitrace file(s) and Log files
ZZZ Apitrace https://files.catbox.moe/qlcqiu.log 
Control would lock up kde/plasma when it reaches the 3D scene when using WineD3D.

doitsujin:
Apparently this is not really an option when there's 40% perf regressions.

Fixes #5043.

### DIFF
--- a/src/dxvk/dxvk_device_info.cpp
+++ b/src/dxvk/dxvk_device_info.cpp
@@ -312,6 +312,13 @@ namespace dxvk {
         }), extensions.end());
     }
 
+    // HACK: Use mesh shader extension support to determine whether we're
+    // running on older (pre-Turing) Nvidia GPUs.
+    m_hasMeshShader = std::find_if(extensions.begin(), extensions.end(),
+      [] (const VkExtensionProperties& ext) {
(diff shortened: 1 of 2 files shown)


K0bin:
Looking for mesh shader support is pretty clever!

doitsujin:
I hate it, but I don't think we have anything better.

PR #5045, opened 2025-06-24
[dxvk] Disable descriptor buffer on AMDVLK on RDNA2 and older

doitsujin:
So yeah apparently amdvlk just goes -90% perf with this in Half-Life 2. Thanks to @Blisto91 for spotting this.


K0bin (approved):
The logic for deciding whether we use descriptor buffers is getting annoying complex. :(

doitsujin:
Once we can stop caring about these GPUs we can nuke all of this along with the legacy path, but yeah for now this is going to be *very* annoying to maintain.

PR #5062, opened 2025-06-29
[d3d9] Only use readonly depth layout if necessary

K0bin:
In current master, we use `LAYOUT_DEPTH_READONLY` views every time writing the depth buffer is disabled. This causes unnecessary layout transitions. It also doesn't really have any advantages because we never use this layout when also sampling the depth buffer.

The PR changes it so we always use the read-write layout unless the depth texture is also bound for sampling and depth write is disabled. I'm not sure if the extra complexity is worth it compared to just resorting to FEEDBACK_LOOP_LAYOUT when the depth texture is bound for depth testing and sampling. Let me know, what you think.

Either way, not using DEPTH_READONLY for depth attachments significantly reduces the number of render passes in Half Life 2. One random hallway in one of my apitraces goes from 17 render passes down to just 10. The City 17 main menu scene goes from 51 render passes down to just 28.

Hopefully helps a bit for #4895


WinterSnowfall (approved):
LGTM

PR #5070, opened 2025-07-04
[d3d9] Interop: support creating stable-address and stable-layout images

AlpyneDreams:
One challenge of D3D9 interop is that images can be relocated at any time. DXVK already has a simple solution to this problem, as images can be created with relocation disabled via stable GPU address mode, for shared images. We simply expose a flag in the interop interface for it. This means that there is finally a case where `GetVulkanImageInfo` can be supported (whereas previously, it was never guaranteed to be safe).

Additionally, dealing with layout transitions can be a challenging part of D3D9 interop. One way to handle it (what I've been using) has been to call `GetVulkanImageInfo` to check the layout and transition it in your own command buffer if needed. Another way would be to call `TransitionTextureLayout`. However, DXVK has the ability to automatically transition images to their default layout after each submission, again also for shared images, so this PR exposes that as a flag too. This also helps prevent cases where a clear is deferred until after the next submit, so any deferred clears will be completed by the time the interop user adds commands.


AlpyneDreams:
Rebased to 2.7

PR #5072, opened 2025-07-04
[util] Add configs for GTR - FIA GT Racing Game

Blisto91:
Otherwise the games config tool will complain about vram unless we restrict it and also wont allow up to 1080p if set too high. Also helps CPU performance.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1253,6 +1253,14 @@ namespace dxvk {
     { R"(\\SFC3\.exe$)", {{
       { "d3d9.countLosableResources",      "False" },
     }} },
+    /* GTR - FIA GT Racing Game                   *                  
+     * Vram complaint & restricted resolutions    *
+     * Performance                                */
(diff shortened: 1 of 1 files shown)


src/util/config/config.cpp
```diff
@@ -1253,6 +1253,15 @@ namespace dxvk {
     { R"(\\SFC3\.exe$)", {{
       { "d3d9.countLosableResources",      "False" },
     }} },
+    /* GTR - FIA GT Racing Game                   *                  
+     * Vram complaint & restricted resolutions    *
+     * Performance                                */
+    { R"(\\(GTR - FIA GT Rac(e)?ing Game|GTR Demo))"
+      R"(\\(GTR(Demo)?|(3D)?Config)\.exe$)", {{
```
doitsujin, line 1260: is the intention here that the first line is the install path?

Bit confusing and weird to write it out like this.

doitsujin: Would be more readable for sure.

WinterSnowfall: @Blisto91 To make it less kilometric, you can also compact it a bit with: `GTR (- FIA GT Rac(e)?ing Game|Demo)`

WinterSnowfall (approved):

PR #5078, opened 2025-07-09
[d3d8/9] 🐸

WinterSnowfall:
Is it a map trap, or is it a way to fingerprint dxvk in d3d8/9? Ask 100 frogs and you'll get 100 gribbits.


doitsujin:
ribbit

WinterSnowfall:
_In-game dialogue option feedback_: The frogs will remember your decision.

PR #5092, opened 2025-07-13
Additional missing D3D9 Texture validations + Another framebuffer binding optimization

K0bin:
The validation changes are fairly straightforward.
It's particularly about the USAGE flags in the CreateXYZSurfaceEx functions that 9Ex added.
Thankfully they simply reject almost all usage flags instead of magically turning an offscreen surface
into a render target.

Besides that, it cleans up the actual image creation.
For example offscreen surfaces always got VK_IMAGE_USAGE_COLOR_ATTACHMENT
because they can be used with D3D9Device::ColorFill. That function however
uses `DxvkContext::ensureImageCompatibility` anyway and it's not super commonly
used with plain offscreen surfaces.

The last point is the optimization for transitioning out of hazards.
In the last couple of months, I optimized the framebuffer binding logic to bind
the render targets that the game tells us to bind instead of trying to be smart
and only binding the ones that the shader writes to and that have a non-zero write mask.
The goal behind this was to avoid having to break up render passes based on state or
shader changes. Besides that, we don't use VK_IMAGE_LAYOUT_DEPTH_READONLY_STENCIL_ATTACHMENT_OPTIMAL
anymore.
Those changes massively reduced the amount of render passes (and barriers) in some games.
The exception to either of those two is that the same texture is bound for sampling and as an RT or DS.
In that case we first try to unbind it in the case of RTs or bind it as DEPTH_READONLY in the case of DS
if possible (shader write masks, blend state write masks/depth write state allow it) before pulling out
the big guns with LAYOUT_FEEDBACK_LOOP.

Dead Space 2 has a deferred lighting pass where it ping pongs between two shaders.
One shader is very simple and only used to update the stencil buffer.
The other shader samples the depth buffer and renders something to a render target.
Neither of those writes to the depth buffer.
Both of those use the same depth stencil surface but because of the different shader
sampler masks, we transition the depth buffer to READONLY when the shader, that samples it,
gets bound. Then when the next shader gets bound, the "hazard" is cleared because the sampler mask
indicates it's not sampled anymore. This makes us rebind the depth buffer as RW because the assumption is that binding as RW by default will allow us to avoid splitting the render pass for every change to the depth write state.
The game does ping-pong between those two shaders after a single draw call each, so we also switch between RW and READONLY after a single draw call each.

The PR solves this by storing whether we bound each attachment last time BindFramebuffer got called and in the case of the depth buffer, whether it was bound as readonly. We then try to match that if possible to minimize render pass changes caused by state or shader changes.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -1104,6 +1104,7 @@ namespace dxvk {
       MarkTextureMipsDirty(dstTexInfo);
 
     ConsiderFlush(GpuFlushType::ImplicitWeakHint);
+    ResetFrameBufferBindingInfo();
 
     return D3D_OK;
   }
@@ -1170,6 +1171,7 @@ namespace dxvk {
 
     dstTexInfo->SetNeedsReadback(dst->GetSubresource(), true);
     TrackTextureMappingBufferSequenceNumber(dstTexInfo, dst->GetSubresource());
+    ResetFrameBufferBindingInfo();
 
     return D3D_OK;
   }
@@ -1439,6 +1441,8 @@ namespace dxvk {
     if (dstTextureInfo->IsAutomaticMip())
       MarkTextureMipsDirty(dstTextureInfo);
 
+    ResetFrameBufferBindingInfo();
+
     return D3D_OK;
   }
 
@@ -1666,6 +1670,13 @@ namespace dxvk {
     if (m_state.renderTargets[RenderTargetIndex] == rt)
       return D3D_OK;
 
+    // We're changing the RT anyway, so make sure we don't attempt to
+    // skip binding it in an attempt to match what we did with the
+    // previous RT at that slot.
+    // Sampling a masked-out RT is rare, so assume that a bound render target
+    // will also end up getting used.
+    m_fbBindingInfo.rtsBound |= (1u << RenderTargetIndex);
+
     // Do a strong flush if the first render target is changed.
     ConsiderFlush(RenderTargetIndex == 0
       ? GpuFlushType::ImplicitStrongHint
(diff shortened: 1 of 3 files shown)


src/d3d9/d3d9_util.cpp
```diff
@@ -47,7 +47,7 @@ namespace dxvk {
 
     // Check if this is a power of two...
     if (sampleCount & (sampleCount - 1))
-      return D3DERR_INVALIDCALL;
+      return D3DERR_NOTAVAILABLE;
```
WinterSnowfall, line 50: I don't think this is entirely safe to do, because `D3DERR_NOTAVAILABLE` should only be returned from CreateRenderTarget in this case. It will probably mess up response codes in other situations.

src/d3d9/d3d9_util.cpp
```diff
@@ -56,7 +56,7 @@ namespace dxvk {
     VkSampleCountFlags supportedSampleCounts = limits.framebufferColorSampleCounts & limits.framebufferDepthSampleCounts;
 
     if ((sampleCount & supportedSampleCounts) == 0)
-      return D3DERR_INVALIDCALL;
+      return D3DERR_NOTAVAILABLE;
```
WinterSnowfall, line 59: Same here. I would not change these haphazardly.

src/d3d9/d3d9_util.cpp
K0bin: So creating multisampled textures with invalid sample codes  (and USAGE_RT) behaves differently?

WinterSnowfall: According to Wine tests (which happen to test these quite extensively), yes.

src/d3d9/d3d9_device.cpp
```diff
@@ -6100,6 +6172,11 @@ namespace dxvk {
     // context, then flush the command list
     uint64_t submissionId = ++m_submissionId;
 
+    // Flushing ends the render pass anyway so the previous binding info no longer matters.
+    m_fbBindingInfo.ds = D3D9DSBindingMode::ReadOnly;
+    m_fbBindingInfo.rtsBound = 0b1111;
```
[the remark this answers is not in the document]
K0bin: We also have it in d3d9_caps.h

K0bin:
I re-tested it on Windows and it seems like the BindFramebuffer logic can be simplified quite a bit.

I'll also split up this PR to make it easier to review and merge.

PR #5103, opened 2025-07-19
[d3d9] Fix several Wine 9Ex tests

WinterSnowfall:
WIP to fix some of the Wine 9Ex tests:
- Also validate present parameters on ResetEx() calls (we were only doing it on Reset() calls)

Before this PR: `0020:d3d9ex: 2960 tests executed (51 marked as todo, 4 as flaky, 176 failures), 1 skipped.`
With this PR: `0020:d3d9ex: 2970 tests executed (51 marked as todo, 4 as flaky, 166 failures), 1 skipped.`


WinterSnowfall:
This has grown a bit beyond what I originally intended (but at least I fixed all the pesky Wine shader validation tests across d3d8/9/9Ex, yay), so I'll undraft it now. I've tested ~30 games, and it looks good enough to throw into experimental at least. 

There are also a few more usage flag / pool tests to address in theory, but I'll do a follow up PR once #5092 is merged, since that one fixes a few tests as well, so it has to be factored in.

Issue #5074, opened 2025-07-06
Morrowind antialiasing artifacts regression in v2.7


Greatness7:
As of version 2.7 some new rendering artifacts have appeared. A video of these artifacts is posted below. Turning off anti-aliasing seems to remove the artifacts. Downgrading to previous version of dxvk also solves the issue.

Others reported the same issue in the Morrowind discord server. It is seemingly affecting users of both NVIDIA and AMD hardware.

https://github.com/user-attachments/assets/7d432451-1682-41bd-8a83-4da8547ac1a7

### Software information
Morrowind + MGE XE 0.18.0

### System information
- GPU: NVIDIA GeForce GTX 1080 TI
- Driver: 5760.2
- DXVK version: 2.7
- Windows 11 24H2

Blisto91:
Hi there and thank you for the report. Are you able to make a apitrace of the game?

Blisto91:
Bisected to start with commit https://github.com/doitsujin/dxvk/commit/1b0d2e80c6e4273917c8e49464d51ea57e679b2d

Apitrace that reproduces the issue: https://drive.proton.me/urls/FGCYGJRNZW#zVWvfNiXdWPy
Notice how the trees change appearance depending on if they are on the right or left side of the camera. At least on Nvidia as i haven't checked amd.

PR #5108, opened 2025-07-22
[dxvk] Support average resolve in blit shader

### LINKED ISSUE #5107: [d3d9] World of Warcraft 3.3.5: antialiasing no longer works
As of dxvk 2.7, the antialiasing feature of World of Warcraft 3.3.5 no longer works. No matter what option you choose antialiasing remains disabled. It worked fine with dxvk 2.6.2.

### Software information
World of Warcraft 3.3.5 (d3d9) on Windows

### System information
- GPU: RTX 4090
- Driver: 576.88
- Wine version: n/a 
- DXVK version: 2.7

doitsujin:
Fixes #5107, #5074. Needs some testing that it doesn't regress Source Engine again.

TLDR if there is no stretching happening, just do a normal shader-based resolve. WoW hits the blit path here for... some reason (presumably view swizzles), but this also allows the resolve to work properly in scenarios where the regular resolve functions wouldn't work, e.g. if the image gets mirrored at the same time.

### DIFF
--- a/src/dxvk/dxvk_context.cpp
+++ b/src/dxvk/dxvk_context.cpp
@@ -1328,9 +1328,13 @@ namespace dxvk {
     renderingInfo.pColorAttachments = &attachmentInfo;
     
     // Retrieve a compatible pipeline to use for rendering
+    auto resolveMode = filter == VK_FILTER_NEAREST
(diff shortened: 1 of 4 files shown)


K0bin (approved):
Looks good. I did some quick testing (though nothing thorough) and didn't see any issues either.

PR #5109, opened 2025-07-22
[util] Enable d3d9.hideNvidiaGpu for Perilous Warp

Blisto91:
In dxvk 2.6.2 and 2.7 changes were made so we only expose support for various bits and pieces as they are also supported in the vendors native d3d9 drivers. 
Prior to those two releases the game used its AMD rendering path on Nvidia, which worked without visual issue. Now that the game can correctly select the Nvidia path we've discovered it depends on `NvAPI_D3D9_StretchRectEx` being implemented. Without it visual effects effects such as blood splatters will appear buggy. 
It was decided that it was much simpler to revert the game back to its AMD path rather than going through with implementing the above function fully.

Edit: I was mixing issues together. This does not need a backport to v2.6.x as it only affects 2.7
Edit 2: No i am dumb. It does need a backport

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1077,6 +1077,13 @@ namespace dxvk {
     { R"(\\TestDrive2\.exe$)", {{
       { "d3d9.deviceLossOnFocusLoss",       "True" },
     }} },
+    /* Perilous Warp - Nvidia path depends on     *
+     * unimplemented NvAPI_D3D9_StretchRectEx.    *
+     * Without it screen effects such as blood    *
+     * splatter are bugged.                       */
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
> In dxvk 2.7 changes were made so we only expose support for various bits and pieces as they are also supported in the vendors native d3d9 drivers.

The "various bits and pieces" here are vendor hacks support. They're not really "vendor" hacks if they're exposed globally, now are they? :P In case of this particular game it looks like it determines if it's on AMD based on RESZ support, otherwise it will proceed to load nvapi and try to use `NvAPI_D3D9_StretchRectEx`.

PR #5110, opened 2025-07-23
[util] Enable d3d9.hideNvidiaGpu for Perilous Warp

Blisto91:
Quick cherry as it works around a regression


WinterSnowfall:
Strawberries, (quick) cherries and an angel's kiss in spring
Your backports really are made from all these things

PR #5111, opened 2025-07-24
[d3d9] Texture creation cleanup + missing D3D9Ex texture creation validations

K0bin:
Part 1 of what was originally #5092.

The validation changes are fairly straightforward.
It's particularly about the USAGE flags in the CreateXYZSurfaceEx functions that 9Ex added.
Thankfully they simply reject almost all usage flags instead of magically turning an offscreen surface
into a render target.

Besides that, it cleans up the actual image creation.
For example offscreen surfaces always got VK_IMAGE_USAGE_COLOR_ATTACHMENT
because they can be used with D3D9Device::ColorFill. That function however
uses DxvkContext::ensureImageCompatibility anyway and it's not super commonly
used with plain offscreen surfaces.


K0bin:
I can't assign @WinterSnowfall as a reviewer so I'll do it informally: Please review it.

WinterSnowfall (approved):

PR #5112, opened 2025-07-25
Bind framebuffer opt

K0bin:
Successor to the scary half of #5092

First of all it rewrites the `D3D9DeviceEx::BindFramebuffer` logic based on what I found during testing:
Ignoring binds based on the write mask is an exception for 1x1 RTs.
If there's a size mismatch across render targets (and/or with the DS) the draw just gets skipped on Windows.
I didn't implement the skipping because it's another thing that would be checked on every draw and not a single game has hit that case so far. So I'll just ignore that until it becomes necessary.

After that there's another optimization to avoid needless layout transitions.
In the last couple of months, I optimized the framebuffer binding logic to bind
the render targets that the game tells us to bind instead of trying to be smart
and only binding the ones that the shader writes to and that have a non-zero write mask.
The goal behind this was to avoid having to break up render passes based on state or
shader changes. Besides that, we don't use VK_IMAGE_LAYOUT_DEPTH_READONLY_STENCIL_ATTACHMENT_OPTIMAL
anymore.
Those changes massively reduced the amount of render passes (and barriers) in some games.
The exception to either of those two is that the same texture is bound for sampling and as an RT or DS.
In that case we first try to unbind it in the case of RTs or bind it as DEPTH_READONLY in the case of DS
if possible (shader write masks, blend state write masks/depth write state allow it) before pulling out
the big guns with LAYOUT_FEEDBACK_LOOP.

Dead Space 2 has a deferred lighting pass where it ping pongs between two shaders.
One shader is very simple and only used to update the stencil buffer.
The other shader samples the depth buffer and renders something to a render target.
Neither of those writes to the depth buffer.
Both of those use the same depth stencil surface but because of the different shader
sampler masks, we transition the depth buffer to READONLY when the shader, that samples it,
gets bound. Then when the next shader gets bound, the "hazard" is cleared because the sampler mask
indicates it's not sampled anymore. This makes us rebind the depth buffer as RW because the assumption is that binding as RW by default will allow us to avoid splitting the render pass for every change to the depth write state.
The game does ping-pong between those two shaders after a single draw call each, so we also switch between RW and READONLY after a single draw call each.

The PR solves this by storing whether we bound each attachment last time BindFramebuffer got called and in the case of the depth buffer, whether it was bound as readonly. We then try to match that if possible to minimize render pass changes caused by state or shader changes.


K0bin:
Briefly tested with:

- Risen
- Dead Space 1 (with the 1x1 shadows mode)
- Dead Space 1 (with the NULL format shadows mode)
- Dead Space 2
- Assassins Creed 2
- Fallout New Vegas
- GTA IV
- Half Life 2
- Mirrors Edge
- Remember Me
- Silent Hill 2 (with Enhanced edition mod)
- Skyrim
- Star Craft 2
- Star Wars Republic Commando
- Team Fortress 2
- Tomb Raider Legend
- Track Mania Nations Forever
- Witcher 2

src/d3d9/d3d9_device.cpp
```diff
@@ -6302,14 +6299,14 @@ namespace dxvk {
...
-  inline void D3D9DeviceEx::UpdateActiveHazardsRT(uint32_t rtMask, uint32_t texMask) {
+  inline void D3D9DeviceEx::UpdateActiveHazardsRT(uint32_t texMask) {
     uint32_t oldHazardMask = m_textureSlotTracking.hazardRT;
     m_textureSlotTracking.hazardRT             &= ~texMask;
     m_textureSlotTracking.unresolvableHazardRT &= ~texMask;
 
     auto psMasks = PSShaderMasks();
-    rtMask  &= m_rtSlotTracking.canBeSampled & ((1u << caps::MaxTextures) - 1);
-    texMask &= m_textureSlotTracking.rtUsage & psMasks.samplerMask & ((1u << caps::MaxSimultaneousTextures) - 1);
+    uint32_t rtMask = m_rtSlotTracking.canBeSampled & ((1u << caps::MaxSimultaneousTextures) - 1);
```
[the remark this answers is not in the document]
misyltoad: Why mask this anyway? Surely it's already only bits 0-3

src/d3d9/d3d9_device.cpp
```diff
@@ -2335,13 +2335,24 @@ namespace dxvk {
           break;
 
         case D3DRS_ZWRITEENABLE:
-          if (likely(!Value != !oldValue))
+          if (likely(m_state.depthStencil != nullptr
+            && m_state.renderStates[D3DRS_ZENABLE])) {
+            // Whether we write the depth has been changed => check for hazards
             UpdateActiveHazardsDS(std::numeric_limits<uint32_t>::max());
```
misyltoad, line 2341: Don't we need to handle the disable case here? That's why I had `if (likely(!Value != !oldValue))`

src/d3d9/d3d9_device.cpp
```diff
@@ -6328,61 +6337,56 @@ namespace dxvk {
...
       return;
 
-    auto masks = PSShaderMasks();
-    masks.samplerMask &= m_textureSlotTracking.dsUsage & texMask & ((1u << caps::MaxTextures) - 1);
-
-    const bool usesDepthBuffer = m_state.renderStates[D3DRS_ZENABLE]
-      || m_state.renderStates[D3DRS_STENCILENABLE]
-      || m_nvdbEnabled;
+    auto psMasks = PSShaderMasks();
+    texMask &= psMasks.samplerMask & m_textureSlotTracking.dsUsage & ((1u << caps::MaxTextures) - 1);
```
misyltoad: Again, I don't get the MaxTextures mask here

src/d3d9/d3d9_device.cpp
```diff
@@ -6696,34 +6700,99 @@ namespace dxvk {
...
     bool srgb = m_state.renderStates[D3DRS_SRGBWRITEENABLE];
 
-    // Some games break if render targets that get disabled using the color write mask
-    // end up shrinking the render area. So we don't bind those.
-    // (This impacted Dead Space 1.)
+    // The extents and sample counts of all render targets need to match.
+    // There's a few exceptions for mismatching extents of depth stencil surfaces:
+    //   - It is allowed to be larger than RT0.
+    //   - It is allowed to be smaller than RT0 IF only one render target is bound, RT0 has the NULL format
+    //     or the extents 1x1 and the color write mask for RT0 is 0.
```
misyltoad, line 6731: I wish we actually had a test suite for these weird cases... I regret not doing that years ago. :(

src/d3d9/d3d9_device.cpp
```diff
@@ -6328,61 +6337,56 @@ namespace dxvk {
...
     }
   }
 
 
   inline void D3D9DeviceEx::UpdateActiveHazardsDS(uint32_t texMask) {
-    uint32_t oldHazardMask = m_textureSlotTracking.hazardDS;
+    uint32_t oldHazardMask             = m_textureSlotTracking.hazardDS;
+    uint32_t oldUnresolvableHazardMask = m_textureSlotTracking.unresolvableHazardDS;
     m_textureSlotTracking.hazardDS             &= ~texMask;
     m_textureSlotTracking.unresolvableHazardDS &= ~texMask;
```
misyltoad, line 6369: With the changes, any reason we shouldn't just be setting to 0 now? Isn't this what it's effectively doing...

misyltoad (changes requested):

src/d3d9/d3d9_device.cpp
```diff
@@ -6302,14 +6299,14 @@ namespace dxvk {
...
+  inline void D3D9DeviceEx::UpdateActiveHazardsRT(uint32_t texMask) {
     uint32_t oldHazardMask = m_textureSlotTracking.hazardRT;
     m_textureSlotTracking.hazardRT             &= ~texMask;
     m_textureSlotTracking.unresolvableHazardRT &= ~texMask;
 
     auto psMasks = PSShaderMasks();
-    rtMask  &= m_rtSlotTracking.canBeSampled & ((1u << caps::MaxTextures) - 1);
-    texMask &= m_textureSlotTracking.rtUsage & psMasks.samplerMask & ((1u << caps::MaxSimultaneousTextures) - 1);
+    uint32_t rtMask = m_rtSlotTracking.canBeSampled & ((1u << caps::MaxSimultaneousTextures) - 1);
+    texMask &= m_textureSlotTracking.rtUsage & psMasks.samplerMask & ((1u << caps::MaxTextures) - 1);
```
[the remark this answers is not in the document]
K0bin: The PS Mask is only considered late if we find hazards that otherwise can't be resolved by unbinding the render target. 

The goal of that is to avoid breaking up the render pass on shader changes.
Dead Space 2 runs into this (with depth stencil but it's the same general idea):
- Draw with shader that samples texture => DS gets bound as readonly
- Draw with shader that does not sample texture => DS gets bound as RW
- Repeat

Same thing with RTs would be:
- Write mask for RT1 is 0
- Draw with shader that samples RT1 texture => RT1 gets unbound
- Draw with shader that does not sample RT1 texture => RT1 gets bound
- Repeat

Ignoring the `psMasks.samplerMask` means we'd still have a resolvable hazard there with the second shader, so it would keep RT1 unbound instead of binding it again and thus get rid of all those barriers.

src/d3d9/d3d9_device.cpp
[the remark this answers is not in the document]
K0bin: Yup, that was just me being overly careful and can be removed.

src/d3d9/d3d9_device.cpp
misyltoad: Oh derp.

src/d3d9/d3d9_device.cpp
misyltoad: Ah yeah, coming back to me now. :)

src/d3d9/d3d9_device.cpp
K0bin: Yeah I should've tested the exact behavior years ago.
I *should* also write and contribute some Wine tests.

misyltoad (approved):
Looks fine in theory. Definitely needs more stress testing though

Issue #4322, opened 2024-10-04
[d3d8/9] Support for n-patches tesselation / ATI TruForm


WinterSnowfall:
There are a few d3d8 games that can apparently leverage n-patches (originally in conjunction with ATI cards), namely:

-> BorderZone
-> Unreal Tournament 2003/2004
-> C&C: Renegade
<details>
<summary>C&C: Renegade configuration screen (Curved Surfaces)</summary>

![image](https://github.com/AlpyneDreams/d8vk/assets/6306593/0e87c00d-4a19-47e0-8014-d0f0fbc3b0da)
</details>

-> Serious Sam (or maybe it's exclusive to the OpenGL renderer? Will have to check)

Here's a list of other titles, though most of these are OpenGL based: https://en.wikipedia.org/wiki/ATI_TruForm#Games_with_TruForm_support

Would be nice to see if we can add support for it at some point.

[Also, here's a very interesting read](https://www.anandtech.com/show/773) about TruForm in general and in d3d8/OpenGL. Also included as part of the Wikipedia page, as external links, (see above) are archive.org references to the ATI TruForm Whitepaper and Developer FAQ.

P.S.: @Blisto91 needs this feature URGENTLY! Don't dawdle.

mrdeathjr28:
Another title support n-patches is hellforces

![4404219-hellforces-windows-front-cover](https://github.com/user-attachments/assets/271951b1-2f1d-4e94-b9e2-693f47bc2c0b)

WinterSnowfall:
> Another title support n-patches is hellforces

It's using the same engine as BorderZone. What a small world :).

mrdeathjr28:
Hell forces have many graphic options, a few similar with tomb raider angel of darkness launcher (many graphic options)

![hllfopt](https://github.com/user-attachments/assets/0361a790-b8be-409e-92c5-8adeb37179b7)

and n-patches option appear but dont possible activate in my case

RiasatSalminSami:
Are there any wrappers for openGL tesselation? I know DgVoodoo2 can do the dx8, 9 tesselation but after googling for hours, I haven't found any solution for opengl games.

niobium93:
Maybe open a Mesa issue asking for the extension to be implemented? Zink and/or RadeonSI might do it.

mirh:
Wasn't tessellation supposed to be the glorified modern version? Even then, cpu emulation is probably fast enough today.

[...]

PR #5114, opened 2025-07-26
[d3d8/9] Assorted fixes and nits

WinterSnowfall:
Noticed a distinct difference between dxvk and native d3d9 when looking at several traces, namely `GetAvailableTextureMem()` consistently reports lower numbers on native (this is for newly created devices). The exact delta varies between 2 MB and 8 MB depending on chosen resolution, so I can only assume it's memory reserved for back buffer use or other internal (Windows) purposes.

This PR might help us get rid of the caps for some games currently using a 2048 limit, but that needs further testing.

Other than that, I've taken the liberty to attempt a somewhat (more) consistent formatting/alignment of the config.cpp profiles. The only actual change in there besides formatting/alignment is that I've bumped the memory limit for NFS4 to 1024 MB, to account for potential modding, since the game will actually get in trouble close to 2048 MB.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5777,10 +5777,17 @@ namespace dxvk {
       availableTextureMemory += memoryProp.memoryHeaps[i].size;
 
     constexpr VkDeviceSize Megabytes = 1024 * 1024;
+    // Windows will typically "reserve" some amount of video memory,
+    // presumably for back buffers, which gets subtracted from the
+    // reported size, e.g. in case of 4 GB it will report a total of
+    // 4286578687 available bytes. The reserved amount varies depending
+    // on the number of back buffers and the back buffer resolution,
+    // however 8 MB has been generally observed for 1080p.
+    constexpr VkDeviceSize ReservedMemory = 8 * Megabytes;
 
     // The value returned is a 32-bit value, so we need to clamp it.
     VkDeviceSize maxMemory = (VkDeviceSize(m_d3d9Options.maxAvailableMemory) * Megabytes) - 1;
-    availableTextureMemory = std::min(availableTextureMemory, maxMemory);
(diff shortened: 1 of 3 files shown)


WinterSnowfall:
Confirmed to obsolete the need for a memory cap for:
- Heroes of Annihilated Empires
- Majesty 2 Collection (which was a UMA-only problem anyway)

@Blisto91, if you happen to still have Pirate Hunter around, would appreciate a test.

Blisto91:
> @Blisto91, if you happen to still have Pirate Hunter around, would appreciate a test.

Pirate Hunter still needs configs to not crash on startup. 
On current master `d3d9.maxAvailableMemory` needs to be set to 3983 or below and with this PR it needs to be 3991 or below. So the upper limit has been raised by 8 MB.

WinterSnowfall:
> Pirate Hunter still needs configs to not crash on startup. On current master `d3d9.maxAvailableMemory` needs to be set to 3983 or below and with this PR it needs to be 3991 or below. So the upper limit has been raised by 8 MB.

Thanks for checking it out. Oh well, two less config profiles is still a (small) win.

I'm not sure if it makes any difference for Sims 2, but I'd rather not touch that dumpster fire at all :sweat_smile:.

WinterSnowfall:
Added one more minor thing to this PR, just because I'm too lazy to start another one: calls to DrawRectPatch/DrawTriPatch always return D3D_OK even on drivers that don't implement TruForm. There's allegedly one game that depends on this to work properly, see: https://github.com/doitsujin/dxvk/issues/4322#issuecomment-2907820508 .

WinterSnowfall:
Additionally, blanked the return values and silenced the logger for `IDirect3DSwapChain9Ex::GetPresentStats()`, which is spammed by Blades of Time.

misyltoad (approved):

PR #5136, opened 2025-08-08
[util] Hide Nvidia in Fallout New Vegas

### LINKED ISSUE #5134: [Fallout: New Vegas] One of the recent versions (assuming since 2.6.2+) broke depth buffer resolves
One of the latest versions (I am assuming 2.6.2) broke [depth buffer resolves](https://github.com/pr0bability/TESReloaded10/blob/master/src/core/RenderManager.cpp#L301) used in the New Vegas Reloaded mod.

There are two things that were broken.

On older versions (like 2.4), NVR relied on RESZ being available on Nvidia cards with DXVK for the resolves. I assume the change which limited vendor hacks to only the original vendors broke this.

As a fallback option for Nvidia cards, which was used without DXVK and which NVR tries to use now with the above change, NVAPI was used. This also doesn't work on the latest version - the depth buffer is just not getting resolved at all.

I am not sure what exactly broke NVAPI, could have been the recent StretchRect changes, but I haven't digged around the code.

The simplest solution that comes to mind would be a configuration option to let user enable back vendor hacks across all vendors.

### Software information
Fallout: New Vegas, with New Vegas Reloaded mod.

### System information
- GPU: Nvidia RTX 3090
- Driver: 572.16
- DXVK version: 2.7

### Apitrace file(s)
Don't have apitrace, can supply renderdoc captures before and after if needed.

### Log files

[FalloutNV_d3d9.log](https://github.com/user-attachments/files/21675401/FalloutNV_d3d9.log)

Blisto91:
The Nvidia path in the New Vegas Reloaded mod depends on `NvAPI_D3D9_StretchRectEx` being implemented in nvapi for depth buffer resolves.
Previously it took a different path because `RESZ` was exposed on Nvidia. This is no longer the case because we match what the native drivers expose now.

Closes #5134

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -918,9 +918,12 @@ namespace dxvk {
       { "d3d9.deviceLossOnFocusLoss",       "True" },
     }} },
     /* Fallout New Vegas - Various visual issues  *
-     * with mods such as New Vegas Reloaded       */
-    { R"(\\FalloutNV\.exe$)", {{
+     * in mod New Vegas Reloaded. Nvidia path in  *
+     * same mod use NvAPI_D3D9_StretchRectEx for  *
+     * depth buffer resolves                      */
+    { R"(\\FalloutNV(Launcher)?\.exe$)", {{
       { "d3d9.floatEmulation",            "Strict" },
+      { "d3d9.hideNvidiaGpu",               "True" },
     }} },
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
As a reminder, the launcher needs to be included too, so the regex needs to be: `FalloutNV(Launcher)?`

PR #5141, opened 2025-08-10
[d3d9] Additional d3d9Ex Wine test fixes

WinterSnowfall:
Fixes a bunch of d3d9Ex specific behavior, as follows:
- mode size checks on `GetAdapterDisplayModeEx` and `GetDisplayModeEx`
- reject D3DPOOL_MANAGED use for Index/Vertex buffers as well
- retain MinZ/MaxZ viewport values during device Reset (I guess 9Ex considers them part of the overall state, which is preserved)
- validations around `pFullscreenDisplayMode` and mode / backbuffer dimensions on device creation and reset

In addition to that, something that applies to d3d9 as well:
- restrict ATI2 texture use - various RT tests for ATI2 were already failing with D3DERR_OUTOFVIDEOMEMORY because creation failed in the backend, so this prempts those situations and more.

Current master: `0020:d3d9ex: 2749 tests executed (46 marked as todo, 4 as flaky, 141 failures), 1 skipped.`
This PR: `0020:d3d9ex: 2662 tests executed (41 marked as todo, 4 as flaky, 105 failures), 1 skipped.`

Will undraft after I've had a chance to test everything properly.


WinterSnowfall:
Seems fine. Most of the above is rather academical anyway, I guess, since there aren't all that many games relying on these things out there in the wild.

PR #5153, opened 2025-08-16
[d3d9] Handling more state-related setters/getters

WinterSnowfall:
Apparently there are a few games (mostly d3d8) that use `Set/GetPaletteEntries` and `Set/GetCurrentTexturePalette`, although, strangely, they do not employ any pelleted textures whatsoever...

`SetNPatchMode` calls are also somewhat common, even though the games do not make use of NPatch. It might be the games explicitly disabling it, however there are some doing thousands/hundred of thousands of calls :frog:...

`GetGPUThreadPriority` has no known users, but I've returned 0 for good measure since we D3D_OK it, so it would expect *something*.

Drafting until I do a bit more testing and investigate potential corner cases around palettes.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -4071,6 +4071,11 @@ namespace dxvk {
     if (!std::exchange(s_errorShown, true))
       Logger::warn("D3D9DeviceEx::GetGPUThreadPriority: Stub");
 
+    if (unlikely(pPriority == nullptr))
+      return D3DERR_INVALIDCALL;
+
+    *pPriority = 0;
+
     return D3D_OK;
   }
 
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
Should be good to go. The texture palette setters & getters seem to work just fine on XP, even with cards/driver that don't support P8, although they won't do anything useful in that case, of course (and yes, some games still call them, because why not).

I've also piped in `D3DRS_PATCHSEGMENTS` from D3D8 into Set/GetNPatchMode. Of course there's no documentation around it, but since `D3DRS_PATCHSEGMENTS` is missing in D3D9, if we ever do get around to adding support for "TruForm", I'm guessing that's needed to get things working properly anyway, since values > 1.0f are what enable n-patches tessellation in both D3D8 and D3D9.

PR #5154, opened 2025-08-16
[d3d9] Add feedback loop usage to RT image view

K0bin:
> err:   VUID-VkRenderingAttachmentInfo-imageView-10780: 
err:   vkCmdBeginRendering(): pRenderingInfo->pColorAttachments[0].imageLayout is VK_IMAGE_LAYOUT_ATTACHMENT_FEEDBACK_LOOP_OPTIMAL_EXT, but image view usage is VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT.
err:   The Vulkan spec states: If feedback loop is enabled for the attachment identified by imageView, then imageView must have been created with a usage value including VK_IMAGE_USAGE_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT, either VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT or VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT, and either VK_IMAGE_USAGE_INPUT_ATTACHMENT_BIT or VK_IMAGE_USAGE_SAMPLED_BIT (https://docs.vulkan.org/spec/latest/chapters/renderpass.html#VUID-VkRenderingAttachmentInfo-imageView-10780)

### DIFF
--- a/src/d3d9/d3d9_common_texture.cpp
+++ b/src/d3d9/d3d9_common_texture.cpp
@@ -649,7 +649,7 @@ namespace dxvk {
   Rc<DxvkImageView> D3D9CommonTexture::CreateView(
           UINT                   Layer,
           UINT                   Lod,
-          VkImageUsageFlagBits   UsageFlags,
+          VkImageUsageFlags      UsageFlags,
           VkImageLayout          Layout,
           bool                   Srgb) {
     DxvkImageViewKey viewInfo;
--- a/src/d3d9/d3d9_common_texture.h
+++ b/src/d3d9/d3d9_common_texture.h
@@ -360,7 +360,7 @@ namespace dxvk {
     Rc<DxvkImageView> CreateView(
             UINT                   Layer,
             UINT                   Lod,
-            VkImageUsageFlagBits   UsageFlags,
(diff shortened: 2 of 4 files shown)


doitsujin:
Shouldn't be much actually, there's like one thing checking for storage image usage that affects descriptor types and some code dealing with view swizzles, which is likely broken with the current code.

K0bin:
We need to make sure there's no performance hit for images that haven't been transitioned to the feedback loop layout before merging this.

If there is, I can destroy the views after the image gets transitioned and recreate them with the now required feedback loop usages.

doitsujin:
What's the perf concern about specifically? Not sure I see a universe where having the old views around would cause any issues.

K0bin:
The way I've implemented it right now, we'd always have USAGE_SAMPLED + USAGE_FEEDBACK_LOOP on the RT/DS views even when the image hasn't been transitioned to the feedback loop layout yet.

I assume that's fine in practice but I should probably ask around.

K0bin:
Okay I did ask around and it doesn't cause any problems as expected.

PR #5155, opened 2025-08-18
[d3d8] Top Spin fixes

### LINKED ISSUE #4214: [D3D8]Top Spin 2004: career mode graphics corruptions
Top Spin 2004, default configuration, 1024x768, 1920x1080, 1680x1050 all @ 60 Hz. XP SP2 compatibility mode on, run as admin. Both original and with official game patch.

- GPU: nVidia GeForce RTX 3070 8 Gb (but also with a GeForce GTX 750Ti and a GeForce RTX 2080Ti)
- Driver: 560.70
- DXVK version: 2.4
- Windows 10 and 11 latest versions

Apitrace file: https://mega.nz/file/1jZHnb6Z#fUkahYqxAwjGRH74I2sVoBtM2nOCvREDe54kcMan_1s

As you can see there is an issue rendering some poligons: totally missing or with crazy vertex positions. 

I've tried the old D8VK 1.0 with the conf option **d3d8.managedBufferPlacement = True** and it solves the problem, BUT the game exits after 2-3 minutes of gameplay without a specific error: Windows Event Manager says generically that either Topspin.exe or msvcirt.dll crashed.

Is this a possible regression from D8VK? Where can I find an equivalent conf option for DXVK 2.4 for d3d8.managedBufferPlacement = True?

Thanks in advance for your help,
Nicola

WinterSnowfall:
Fixes #4214

Turns out the game had two issues:
- the missing/wonky geometry problems were caused by the game relying on legacy DISCARD behavior, like some other d3d8 games at the time (see Rayman 3)
- memory leaks were caused by us double reffing d3d9 state blocks during d3d8 state block creation (specifically on calls to D3D8Device::CreateStateBlock). Since d3d9 already refs the return during state block creation, this was adding an extra ref that was preventing d3d9 state blocks created like that from ever being released. And since this was a contained/nested object, the issue wasn't obvious in apitraces, so was not immediately noticeable...

@Blisto91 would appreciate a test on your side too. ~~Note that it has to be on Nvidia if I recall correctly, because on AMD the game apparently takes a different path.~~ Nevermind, looks like it was broken with dxvk regardless of vendor.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1289,6 +1289,12 @@ namespace dxvk {
     { R"(\\Aqua\.exe$)", {{
       { "d3d9.modeCountCompatibility",      "True" },
     }} },
+    /* Top Spin (2005)                            *
+     * Missing geometry and textures without      *
+     * legacy DISCARD behavior                    */
+    { R"(\\TopSpin\.exe$)", {{
+      { "d3d8.forceLegacyDiscard",          "True" },
+    }} },
   };
 
 


WinterSnowfall:
For posterity, let it be known that the Com changes fix a crash on level load seen in Sea Dogs (aka Age of Pirates) 2: City of Abandoned Ship, during `m_indices = buffer;` in `3D8Device::SetIndices`.

Blisto91:
~~Looks good on AMD too~~ nvm that test wasn't needed 😝

PR #5158, opened 2025-08-20
[util] Track memory use for Alone in the Dark

WinterSnowfall:
Technically, there already is a "mod fix" out there for the crash, but might as well fix it ourselves in dxvk too and turn it to gold/platinum: https://www.protondb.com/app/259170 .


mrdeathjr28:
Many thanks this game stay many years with this error in my case but now working ok, resolution can change without crash (input issues are solved too, wine issue)

Also work with game executable edit for remove blackbars (HxD)

aitd2k8a

aitd2k8b

PR #5159, opened 2025-08-20
[d3d9] Bring back allowDiscard hack

Blisto91:
Makes Scarface graphics render properly

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5372,6 +5372,9 @@ namespace dxvk {
     if (unlikely(ppbData == nullptr))
       return D3DERR_INVALIDCALL;
 
+    if (unlikely(!m_d3d9Options.allowDiscard))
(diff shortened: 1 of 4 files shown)


WinterSnowfall (approved):

doitsujin:
do we understand *why* this is an issue?

Blisto91:
Kinda never replied to this on GitHub. So for completeness sake.

> do we understand why this is an issue?

The game depends on some legacy behavior on Windows that doesn't hold true anymore. It is also broken on modern d3d9 Windows, so it is a game issue that happened to work once upon a time when it was originally released.
After poking it a bit i asked @CookiePLMonster if this was correct and he confirmed. 
He does something similar in his SilentPatch for the game https://github.com/CookiePLMonster/SilentPatchScarface/blob/395b01f48121c7518500ad0fd97ff4e4dfdfac67/source/SilentPatchScarface.cpp#L190

PR #5163, opened 2025-08-22
[d3d9] Fix crash in GetVulkanImageInfo for SYSMEM or NULL textures

AlpyneDreams:
If you call `GetVulkanImageInfo` on a texture that does not have a `DxvkImage`, then DXVK will immediately segfault. This usually happens for `NULL` format textures and textures in `SYSMEM` or `SCRATCH`, but all the specific cases are opaque to the user. This PR fixes that by making it fail gracefully and return `D3DERR_NOTFOUND`.

### DIFF
--- a/src/d3d9/d3d9_interop.cpp
+++ b/src/d3d9/d3d9_interop.cpp
@@ -108,6 +108,17 @@ namespace dxvk {
           VkImageLayout*        pLayout,
           VkImageCreateInfo*    pInfo) {
     const Rc<DxvkImage> image = m_texture->GetImage();
+    
+    if (unlikely(!image)) {
+      if (pHandle != nullptr)
+        *pHandle = VK_NULL_HANDLE;
+
+      if (pLayout != nullptr)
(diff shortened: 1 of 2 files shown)


K0bin (approved):

src/d3d9/d3d9_interop.cpp
```diff
@@ -108,6 +108,17 @@ namespace dxvk {
           VkImageLayout*        pLayout,
           VkImageCreateInfo*    pInfo) {
     const Rc<DxvkImage> image = m_texture->GetImage();
+    
+    if (!image) [[unlikely]] {
```
WinterSnowfall, line 112: dxvk has an unlikely() macro, which is used everywhere else.

WinterSnowfall (changes requested):
Just a minor styling nit.

src/d3d9/d3d9_interop.cpp
AlpyneDreams: Ah, yeah, forgot about that. I'm used to C++20

WinterSnowfall (approved):

PR #5164, opened 2025-08-22
[d3d8/9] Yet another motley crew of fixes

WinterSnowfall:
Fixes the last remaining failing Wine cursor test.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -278,6 +278,11 @@
 # Overrides anisotropic filtering for all samplers. Set this to a positive
 # value to enable AF for all samplers in the game, or to 0 in order to
 # disable AF entirely. Negative values will have no effect.
+#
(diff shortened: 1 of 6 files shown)


WinterSnowfall:
Derped in some minor restructuring of how d3d8 present parameters validation is done, since we had some duplicate code for no reason.

WinterSnowfall:
In today's unrelated fixes:
- D3DDEVTYPE_REF are apparently supported on OSes that ship a d3dref9.dll , which seems to be the case on Win 10/11 at least. The documentation seems to suggest that has been the case ever since Windows 8, and there are traces for newer d3d9 games (2009+), which indeed use D3DDEVTYPE_REF devices for querying caps. I've made sure we expose only SWVP for them, at least, as would be the case on native.
- Added a note to samplerAnisotropy, as a follow-up to #5172

PR #5179, opened 2025-08-31
[d3d8/9] Adjust invalid viewport handling

WinterSnowfall:
Using a NULL pViewport in SetViewport() will crash the native runtime, but we should handle it a bit more gracefully and not dereference it in that case.

More interestingly, neither d3d8 or 9 will allow the viewport MaxZ to be smaller than or equal to MinZ, and will automatically set MaxZ to MinZ + 0.001f if that's the case.

Drafting until it's confirmed to behave the same on modern OSes, though I don't think there will be any surprises there.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -1152,31 +1152,34 @@ namespace dxvk {
   HRESULT STDMETHODCALLTYPE D3D8Device::SetViewport(const D3DVIEWPORT8* pViewport) {
     D3D8DeviceLock lock = LockDevice();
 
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
Confirmed to behave the same on modern Windows, thanks to @Blisto91.

WinterSnowfall:
Squeezed in a minor behavior correction due to an incomplete test. CheckDeviceMultiSampleType will only D3D_OK D3DMULTISAMPLE_NONE checks on NULL format, and will still return D3DERR_NOTAVAILABLE for anything else. Narrows down the behavior introduced in #5039.

K0bin (approved):

Issue #5172, opened 2025-08-25
[d3d9] Warhammer 40,000: Space Marine Graphical Corruption when forcing Anisotropic Filtering


ShinChaosBahamut:
Noticed while playing Space Marine that if I force Anisotropic Filtering via dxvk.conf the screen will be "garbled" with green noisy artifacts. (as shown in the attached image)

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/77d13022-3120-486e-9fbd-1658ca5a2f94" />

Setting the Post Process FX setting (which also controls post process AA; which I think might be the cause of this, sadly there's no way to disable JUST the AA part of the setting to confirm) to Off works around this issue, but I figured I'd bring it up just in case a fix is possible.

OS: Windows 11
DXVK version: 2.7 (also tried the build provided in https://github.com/doitsujin/dxvk/issues/5134#issuecomment-3169243085 just to see if that already fixed it; no dice)

WinterSnowfall:
The game is already using varying levels of anisotropy, from 4x to 8x... why do you feel you need to override that? Forcing a certain global level, although usually fine, is known to break games at times. This is one such case, where the game was designed around various expected levels on certain textures, and will break otherwise.

doitsujin:
Forcing anisotropy is known to break passes that rely on bilinear filtering, overriding game behaviour isn't epxected to work in all cases.

ShinChaosBahamut:
> The game is already using varying levels of anisotropy, from 4x to 8x...

In my defense, I did not know that. (experience has kinda taught me that if a relatively older game does not have AF settings, it's using just basic Bilinear/Trilinear Filtering)

Since this ain't a bug I'll just close this.

WinterSnowfall:
> In my defense, I did not know that. (experience has kinda taught me that if a relatively older game does not have AF settings, it's using just basic Bilinear/Trilinear Filtering)

Believe you me you *will know* when a game only uses bi/trilinear. For example the original Mafia looks awful if you don't force some form of AF. Everything more than 10m away gradually turns into texture barf.

ShinChaosBahamut:
Well, might as well ask since this is up (even though closed); any reason why forcing AF via DXVK.conf would have this issue but forcing AF via the drivers (when using the native API) wouldn't? No snark intended, I'm genuinely curious.

WinterSnowfall:
[...]

PR #5180, opened 2025-09-01
[d3d9] Override sampler anisotropy only on multi-mip textures

### LINKED ISSUE #2400: "Limit Custom Anisotropic Filtering to mipmapped textures with many levels"
This isn't so much a bug but simply a solution to a problem DXVK has that has also recently been found and fixed in the emulator Ryujinx. Here's the PR for this change: https://github.com/Ryujinx/Ryujinx/pull/2832

The same issue is present with DXVK in at least one game, Dead Rising 2. Here is a screenshot of the issue: https://i.imgur.com/yHTc9Fv.png

You'll notice that it's the exact same issue as visible in the emulator progress report, both being triggered by applying anisotropic filtering to all samplers in the scene, even those that it should not be applied to.

The emulator developers created a simple trick to ensure anisotropic filtering is not applied to these types of shaders and post process effects by only applying it to 2D textures with mipmap levels, and that are set to be filtered with bilinear or trilinear texture filtering. Implementing this same type of fix into DXVK would allow us to set all games to a proper 16x AF level globally without any flaws, exactly the same as it's done through the GPU driver for older APIs like DX8-12 and OpenGL.

WinterSnowfall:
Fixes the use of the option with #5172 (so it seems that _was_ actually a bug of sorts), and I'm assuming also fixes #2400 (the issue mentioned in it).

@K0bin please review, ~~I'm not entirely sure if I'm piping in things properly here, though the principle seems sound~~ (nevermind, after looking at it in more detail, pretty sure the piping is a good as it can get).

I've checked that other games where the option is useful (e.g. Mafia) still work properly in terms of the intended effect.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -7324,10 +7324,10 @@ namespace dxvk {
       key.setDepthCompare(cIsDepth, VK_COMPARE_OP_LESS_OR_EQUAL);
 
       if (cState.mipFilter) {
-        // Anisotropic filtering doesn't make any sense with only one mip
         uint32_t anisotropy = cState.maxAnisotropy;
 
-        if (cState.minFilter != D3DTEXF_ANISOTROPIC)
+        // Anisotropic filtering doesn't make any sense with only one mip
+        if (cState.minFilter != D3DTEXF_ANISOTROPIC || !cIsMultiMip)
           anisotropy = 0u;
 
         // Forcing anisotropic filtering doesn't make any sense with only one mip
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -7293,11 +7293,12 @@ namespace dxvk {
     const D3D9CommonTexture* tex = GetCommonTexture(m_state.textures[Sampler]);
 
     EmitCs([this,
-      cSlot     = slot,
-      cState    = D3D9SamplerInfo(m_state.samplerStates[Sampler]),
-      cIsCube   = tex && tex->IsCube(),
-      cIsDepth  = bool(m_textureSlotTracking.depth & (1u << Sampler)),
-      cBindId   = m_samplerBindCount
+      cSlot       = slot,
+      cState      = D3D9SamplerInfo(m_state.samplerStates[Sampler]),
+      cIsCube     = tex && tex->IsCube(),
+      cIsMultiMip = tex && (tex->Desc()->MipLevels > 1u),
+      cIsDepth    = bool(m_textureSlotTracking.depth & (1u << Sampler)),
+      cBindId     = m_samplerBindCount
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
It's unclear to me if I should also:

```
      if (cState.mipFilter) {
        uint32_t anisotropy = cState.maxAnisotropy;

        // Anisotropic filtering doesn't make any sense with only one mip
        if (cState.minFilter != D3DTEXF_ANISOTROPIC || !cIsMultiMip)
          anisotropy = 0u;
```

Because as things stand, anisotropy (whatever cState.maxAnisotropy stores) *is* applied on textures with single mips too (if they have a mipFilter specified).

P.S.: Did that for now, and it seems fine. Maybe there's some extra perf gained in such cases?

doitsujin:
> P.S.: Did that for now, and it seems fine. Maybe there's some extra perf gained in such cases?

If anything it's the right thing to do and there's content out there (albeit D3D11, where we have no good way to work around it) relying on that behaviour, so it's not going to hurt to do this on our end..

KuraiShidosha:
Confirmed this fixed Dead Rising 2: Off the Record for me. Great job.
Before:
yHTc9Fv

After:
deadrising2otr_2025_09_02_09_46_53_141

I'm guessing this isn't possible to do in D3D11? Blackwell GPUs coincidentally have the same problem natively in several games. Nvidia is working on it and say they'll have a fix down the line but I was really hoping DXVK would circumvent it in the here and now. I'm guessing if this fix could have been applied to D3D11 games it would have.

The real question is, when Nvidia fixes it on their end will that translate to DXVK? An example game I have that has this anisotropic filtering corruption issue in D3D11 on a 5090 is Metal Gear Solid V: Ground Zeroes. Simply loading up the main menu with DXVK set to 16x anisotropic filtering will showcase the issue on a Blackwell GPU. Example pic here:
MgsGroundZeroes_2025_09_02_09_44_06_006


Could someone with a 40 series or older Nvidia card test this game with DXVK forcing anisotropic filtering and see if it occurs? If not then it's a purely Nvidia 50 series issue and we'll have to wait for them to fix it in drivers.

WinterSnowfall:
> I'm guessing this isn't possible to do in D3D11? [...] I'm guessing if this fix could have been applied to D3D11 games it would have.

The way samplers work is very different between d3d8/9 and d3d11, so even if this could be fixed (which I doubt), it would have to be fixed in an entirely different way.

> The real question is, when Nvidia fixes it on their end will that translate to DXVK?

The Nvidia proprietary driver is a black box, so it's not like anyone can know what the problem is or how it will be fixed, but if it affects their Vulkan implementation, then it will "translate" to dxvk because its backend uses Vulkan. As for that particular game, I don't have it, so can't test, but it looks like another case of something expecting purely linear filtering, where the sample override won't work correctly with d3d11.

Issue #3534, opened 2023-06-24
Unreal Tournament 2004: Shader compilation stutter


K0bin:
Unreal Tournament 2004 (with D3D8to9 or D8VK) suffers from shader compilation stutters.
DXVK generates a lot of fixed function vertex shaders. Additionally the game has a queue sync every frame, which naturally means that we also synchronize with the CS thread. So each shader compilation stalls both threads making it even worse.

A possible solution would be to move our generated fixed function shaders more towards uber-shaders and handle multiple versions in one shader.

### Software information
Unreal Tournament 2004

### System information
- GPU: Nvidia RTX 3090
- DXVK version: 2.2+ (doesn't really matter, happens on all versions)

### Apitrace file(s)
https://drive.google.com/file/d/1dTbqCiJL8AJESJWZZsJNliCZvV0tc4Cl/view?usp=sharing

### Log files
I made it print out the 4 uint32_t words of the FF VS key.
[shader_compilation_log.txt](https://github.com/doitsujin/dxvk/files/11857257/shader_compilation_log.txt)

WinterSnowfall:
Most likely related: https://github.com/AlpyneDreams/d8vk/issues/165

It's also using UE2, and arguably the stuttering is nowhere near as bad as it is in UT2004, but it's there.

K0bin:
According to Josh it's mostly down to TexcoordFlags, TransformFlags and TexcoordDeclMask.

netborg-afps:
This eliminates stutters almost completely, but has rendering artifacts on the HUD:
https://github.com/doitsujin/dxvk/compare/master...netborg-afps:dxvk:2023_06_fixing_ut2004_stutters

Not sure if this is a fix or a hack?

netborg-afps:
Because my previous hack made me confident the game is actually largely stutter-free when moving some variables to the shader, I made a real fix last week. In case someone wants to play UT2004, or any UE2 game via dxvk, the fix is available here:
https://github.com/netborg-afps/dxvk/tree/2023_06_fixing_ut2004_stutters

The disadvantage of the fix though is that vertex shader compile times roughly doubled, so the only true solution which works for all games would be a single uber-shader which wouldn't need to be compiled at runtime.

adolfintel:
Still an issue unfortunately :disappointed:

netborg-afps:
> Still an issue unfortunately 😞

Have you tried my fix?
[...]

Issue #4091, opened 2024-06-26
[d3d9 Win10\11] Stuttering in Lineage2 and other UE2 games


qefyr:
### Software information
Lineage 2 very low settings
(Additionally, I tested the UT2004 demo, it had the same problem.)

### System information №1 RIG:
- GPU: GeForce GTX1660 super
- Driver: 555.99
- Wine version: Windows 11
- DXVK version: 2.3.1

### System information №2 RIG:
Lenovo IdeaPad 1 15IAU7
- GPU: Intel UHD (i3-1215U)
- Driver: 31.0.101.5592
- Wine version: Windows 10
- DXVK version: 2.3.1

### Apitrace file(s)
- Put a link here

For instructions on how to use apitrace, see: https://github.com/doitsujin/dxvk/wiki/Using-Apitrace

Lineage2 game has the worst optimization and performance of any existing game on earth. DXVK allows you to increase performance and the biggest difference is on an integrated Intel video card, because... The initial Intel UHD performance in this game is very poor compared to nvidia video cards. **Unfortunately, a side effect is a huge number of stutters.**
Today I found out about another wrapper dgVoodoo2_82_5 and decided to test it on Intel UHD (i3-1215u), performance increased but not to the level of DX9-DXVK, and on 1660super the effect is completely opposite - performance drops. But what surprised me was that when translate DX9=>DX11=>DXVK there are absolutely no stutters. Therefore, I had a question: Why do stutters appear specifically when translate from DX9 to DXVK?

K0bin:
That's a known issue with how we handle D3D9 fixed function rendering.

guglovich:
Now I've seen for myself how problematic DX9 > DXVK is in Lineage 2 when I tried to make an AO. There are so many pitfalls and compatibility issues. I'm thinking about trying DX9 > DX11 > DXVK, too, to see if it improves the inject.

Blisto91:
Try with current master https://github.com/doitsujin/dxvk/actions/runs/24126385952
Improvements have been made in regards to stuttering with fixed function.

guglovich:
> Try with current master https://github.com/doitsujin/dxvk/actions/runs/24126385952 Improvements have been made in regards to stuttering with fixed function.

Thank you. I have Lineage 2 with +- stable FPS. I wrote more about the difficulty of working with the wrapper and this game. I'm currently trying to make a DXVK fork with AMD CACAO for this game.

Blisto91:
[...]

PR #5192, opened 2025-09-03
D3D9 Fixed function ubershader

K0bin:
This creates a single shader that handles the D3D9 fixed function vertex stage and another one that handles the fixed function pixel stage.

This initial PR keeps the old FF code generator around. The ubershaders are enabled by default but can be disabled using `d3d9.ffUbershaderVS=False` and `d3d9.ffUbershaderFS=False` respectively.
The long term plan is to get rid of the FF code generator.

To get performance up to par with the generated shaders (mostly), we have to optimize the fragment shader using spec constants. That works like this:
The D3D9 fixed function pixel pipeline stage consists of 8 texture stages. Each of those has both a color operation with 3 arguments and an alpha operation with 3 arguments. The arguments decide which value to use and the operation decides what to do with it. For good performance we need to turn most of those into specialization constants.
At first this seemed like a huge problem because we also store a bunch of other information for each texture stage like: the texture type, whether it's a depth ("shadow") sampler and whether it's projected. Thankfully we already had spec constants for those because shader model 1 shaders need those three bits of data too. So the first step was to shrink each texture stage by making it reuse the existing spec constants (9bd9cb2bf92201b9232d83f678e868e1da564b0c). Those changes are done for the generated shaders too, so the PR changes those as well.
On top of that we can ignore putting argument 0 into the spec constant because that's only used very rarely. (yes, argument 0 is the rarely used one, not 1 or 2. D3D9...). If a game uses one of the two operations that use argument 0, that will just get loaded from a buffer.
Besides that, the arguments could be repacked into 5 instead of 6 bits because they only contain values from 0 to 7 and 2 flags.
With those changes one texture stage takes up 31 bits. So one uint32_t specialization constant can cover one texture stage.
The DXVK backend is configured to support 12 specialization constants right now and in current master the D3D9 frontend uses 6 of those.
I went with optimizing the first 4 texture stages. This covers almost all draws in the games I've tested.
In Star Wars Republic Commando I saw a 95% rate of draws that only use 4 texture stages and dont use argument 0 for either color or alpha. In UT2004 it was even 99%.
Besides that, I also managed to squeeze in the number of active texture stages into one of the existing spec constant dwords. That allows the shader to optimize away all stages past that.
In most cases this should hopefully yield shaders that are basically what we manually generated before. I looked into a few shaders with RGP captures and this was indeed the case as far as I can tell.

Vertex shaders get no such optimization treatment because games from 2004 only have tiny amounts of geometry anyway.

Performance in Star Wars Republic Commando on the Steam Deck is about 0-10% worse than the generated shaders (difficult to test because the framerate varies a little even on a static scene).
It completely fixes any shader compilation stutter for fixed function shaders so it also makes Unreal Tournament 2004 (and other UE2 games presumably) finally playable. UT2004 in particular has a combinatorics problem with FF vertex shaders on master. The compiled output for fully optimized fragment shaders is practically the same as on master with some very minor exceptions that require the compiler to have better knowledge of the vertex shader (which we don't bother optimizing).

The ubershaders pass all Wine tests that master passes.

### DIFF
--- a/src/d3d9/shaders/d3d9_fixed_function_frag.frag
+++ b/src/d3d9/shaders/d3d9_fixed_function_frag.frag
@@ -1,0 +1,685 @@
+#version 450
+#extension GL_GOOGLE_include_directive : enable
+#extension GL_EXT_scalar_block_layout : require
+#extension GL_EXT_spirv_intrinsics : require
+#extension GL_EXT_demote_to_helper_invocation : require
+#extension GL_ARB_derivative_control : require
+#extension GL_EXT_control_flow_attributes : require
+#extension GL_EXT_nonuniform_qualifier : require
+
+
+// The locations need to match with RegisterLinkerSlot in dxso_util.cpp
+layout(location = 0) in vec4 in_Normal;
+layout(location = 1) in vec4 in_Texcoord0;
+layout(location = 2) in vec4 in_Texcoord1;
+layout(location = 3) in vec4 in_Texcoord2;
+layout(location = 4) in vec4 in_Texcoord3;
+layout(location = 5) in vec4 in_Texcoord4;
+layout(location = 6) in vec4 in_Texcoord5;
+layout(location = 7) in vec4 in_Texcoord6;
+layout(location = 8) in vec4 in_Texcoord7;
+layout(location = 9) in vec4 in_Color0;
+layout(location = 10) in vec4 in_Color1;
+layout(location = 11) in float in_Fog;
+
+layout(location = 0) out vec4 out_Color0;
+
+
+const uint SpecConstOptimizedTextureStageCount = 4;
+const uint TextureArgCount = 3;
+const uint MaxSharedPushDataSize = 64;
+
+#include "d3d9_fixed_function_common.glsl"
+
+struct D3D9FFTextureStage {
+    uint Primitive[2];
+};
+
+struct D3D9FixedFunctionPS {
+    vec4 textureFactor;
+    D3D9FFTextureStage Stages[8];
+};
+
+struct D3D9SharedPSStage {
+    float Constant[4];
+    float BumpEnvMat[2][2];
+    float BumpEnvLScale;
+    float BumpEnvLOffset;
+    float Padding[2];
+};
+
+struct D3D9SharedPS {
+    D3D9SharedPSStage Stages[TextureStageCount];
+};
+
+const uint D3DTOP_DISABLE                   = 1;
+const uint D3DTOP_SELECTARG1                = 2;
+const uint D3DTOP_SELECTARG2                = 3;
+const uint D3DTOP_MODULATE                  = 4;
+const uint D3DTOP_MODULATE2X                = 5;
+const uint D3DTOP_MODULATE4X                = 6;
+const uint D3DTOP_ADD                       = 7;
+const uint D3DTOP_ADDSIGNED                 = 8;
+const uint D3DTOP_ADDSIGNED2X               = 9;
+const uint D3DTOP_SUBTRACT                  = 10;
+const uint D3DTOP_ADDSMOOTH                 = 11;
+const uint D3DTOP_BLENDDIFFUSEALPHA         = 12;
+const uint D3DTOP_BLENDTEXTUREALPHA         = 13;
+const uint D3DTOP_BLENDFACTORALPHA          = 14;
+const uint D3DTOP_BLENDTEXTUREALPHAPM       = 15;
+const uint D3DTOP_BLENDCURRENTALPHA         = 16;
+const uint D3DTOP_PREMODULATE               = 17;
+const uint D3DTOP_MODULATEALPHA_ADDCOLOR    = 18;
+const uint D3DTOP_MODULATECOLOR_ADDALPHA    = 19;
+const uint D3DTOP_MODULATEINVALPHA_ADDCOLOR = 20;
+const uint D3DTOP_MODULATEINVCOLOR_ADDALPHA = 21;
+const uint D3DTOP_BUMPENVMAP                = 22;
+const uint D3DTOP_BUMPENVMAPLUMINANCE       = 23;
+const uint D3DTOP_DOTPRODUCT3               = 24;
+const uint D3DTOP_MULTIPLYADD               = 25;
+const uint D3DTOP_LERP                      = 26;
+
+const uint D3DTA_SELECTMASK     = 0x0000000f;
+const uint D3DTA_DIFFUSE        = 0x00000000;
+const uint D3DTA_CURRENT        = 0x00000001;
+const uint D3DTA_TEXTURE        = 0x00000002;
+const uint D3DTA_TFACTOR        = 0x00000003;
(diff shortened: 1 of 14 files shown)


src/d3d9/shaders/d3d9_fixed_function_vert.vert
```diff
@@ -0,0 +1,731 @@
...
+            normal = nrmSum;
+        } else {
+            vtx = vtx * data.WorldView;
+
+            mat3 nrmMtx = mat3(data.NormalMatrix);
+
+            normal = nrmMtx * normal;
+        }
+
+        // Some games rely no normals not being normal.
```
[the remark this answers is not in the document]
K0bin: That's your typo. :P

https://github.com/doitsujin/dxvk/blame/master/src/d3d9/d3d9_fixed_function.cpp#L1124

WinterSnowfall:
As of the most recent build, it looks good and fixes: #3534 , #4091 and https://github.com/AlpyneDreams/d8vk/issues/165 .

werman:
"Star Wars Republic Commando" is a d3d8 game, how it was tested with d3d9?

WinterSnowfall:
> "Star Wars Republic Commando" is a d3d8 game, how it was tested with d3d9?

"d3d9" here refers to dxvk's d3d9 backend, which is shared between the d3d8 and d3d9 frontends. That's a nice way of saying that in d3d8 dxvk land we mostly rely on existing d3d9 dxvk code, so the same paths are hit.

K0bin:
Draft while I clean up the commit history and split off parts into separate PRs.

K0bin:
Cleaned up the commit history, cleaned up the code and made sure it works.
It passes all Wine tests that master passes and it works in all FF apitraces I have.

It would of course be great if @Blisto91 & @WinterSnowfall could take it for a spin.

src/d3d9/shaders/d3d9_fixed_function_frag.frag
```diff
@@ -0,0 +1,667 @@
...
+    layout(offset = MaxSharedPushDataSize) uint packedSamplerIndices[TextureStageCount / 2];
+};
+
+layout(set = 0, binding = 13) uniform texture2D t2d[TextureStageCount];
+layout(set = 0, binding = 13) uniform textureCube tcube[TextureStageCount];
+layout(set = 0, binding = 13) uniform texture3D t3d[TextureStageCount];
+
+layout(origin_upper_left) in vec4 gl_FragCoord;
+
+layout(set = 15, binding = 0) uniform sampler sampler_heap[];
```
doitsujin, line 145: I know we do this in the legacy compilers as well, but there's some tooling that shits the bed with `set = 15`, so might be a good idea to just set this to 1 if doesn't mess with the plumbing code too much.

src/d3d9/shaders/d3d9_fixed_function_frag.frag
```diff
@@ -0,0 +1,667 @@
...
+    vec4 color = oColor;
+    vec3 color3 = color.rgb;
+    vec3 fogFact3 = vec3(fogFactor);
+    vec3 lerpedFrog = mix(fogColor, color3, fogFact3);
+    return vec4(lerpedFrog.r, lerpedFrog.g, lerpedFrog.b, color.a);
+}
+
+
+// [D3D8] Scale Dref to [0..(2^N - 1)] for D24S8 and D16 if Dref scaling is enabled
+vec4 adjustDref(vec4 texCoord, uint referenceComponentIndex, uint samplerIndex) {
```
doitsujin, line 234: readability / potential code gen meme nit, i'd prefer if this just consumed and returned a `float` so usage code just does something like `a.z = adjustDref(coord.z, ...)`

src/d3d9/shaders/d3d9_fixed_function_frag.frag
```diff
@@ -0,0 +1,667 @@
...
+        vec2 t = previousStageTextureVal.xy;
+        float result = tc_m_n + dot(bm, t);
+        coords[i] = result;
+    }
+    return coords;
+}
+
+
+uint loadSamplerHeapIndex(uint samplerBindingIndex) {
+    uint packedSamplerIndex = packedSamplerIndices[samplerBindingIndex / 2u];
```
doitsujin, line 266: i swear one day amd's windows driver will support 16-bit push constants so we can get rid of this stupid packing

src/d3d9/shaders/d3d9_fixed_function_frag.frag
```diff
@@ -0,0 +1,667 @@
...
+        previousStageColorOp = colorOp(stage - 1);
+    }
+
+    if (stage != 0 && (
+        previousStageColorOp == D3DTOP_BUMPENVMAP
+        || previousStageColorOp == D3DTOP_BUMPENVMAPLUMINANCE)) {
+        texcoord = calculateBumpmapCoords(stage, texcoord, previousStageTextureVal);
+    }
+
+    vec4 texVal;
```
doitsujin, line 288: is there a universe where this could be left undefined?

src/d3d9/shaders/d3d9_fixed_function_frag.frag
```diff
@@ -0,0 +1,667 @@
...
+        reg = vec4(1.0) - reg;
+
+    // reg = reg.wwww
+    if ((arg & D3DTA_ALPHAREPLICATE) != 0)
+        reg = reg.aaaa;
+
+    return reg;
+}
+
+vec4[TextureArgCount] readArgValues(uint stage, uint args[TextureArgCount], vec4 current, vec4 temp, vec4 textureVal) {
```
doitsujin, line 364: I'm still rather allergic to this being an array, there doesn't seem to be any actual dynamic indexing going on for these anyway so can this just be changed into a struct?

src/d3d9/shaders/d3d9_fixed_function_frag.frag
```diff
@@ -0,0 +1,667 @@
...
+
+        default:
+        case VK_COMPARE_OP_ALWAYS:
+            atestResult = true;
+            break;
+    }
+
+    bool atestDiscard = !atestResult;
+    if (atestDiscard) {
+        discard;
```
doitsujin, line 533: `demote`, discard is deprecated.

doitsujin (changes requested):
some minor complaints about FFPS, other than that this is shockingly "clean".

WinterSnowfall:
> It would of course be great if @Blisto91 & @WinterSnowfall could take it for a spin.

Thrown it at about 20-ish known d3d8 FVF games and it looks good. All UE2 titles I've tried are buttery smooth now, I can't believe my :eyes:. Can do another pass before it's ready to merge if needed.

src/d3d9/shaders/d3d9_fixed_function_frag.frag
K0bin: We'd need some major bug in the frontend that somehow sets the texture type to 3 (the highest possible value given we're reading 2 bits).
I guess I can just add a default case to be sure.
It'll get optimized out anyway.

doitsujin (approved):

PR #5208, opened 2025-09-13
[d3d9] Show ubershader in FF HUD element

K0bin:
image


K0bin:
@WinterSnowfall

WinterSnowfall:
I'd recommend something shorter like "1*", so that it doesn't take up so much of the screen, but that being said it's up to you.

K0bin:
image

Issue #3753, opened 2023-12-18
[chore] update VP_DXVK_requirements.json


Blisto91:
Reminder issue to date [VP_DXVK_requirements.json](https://github.com/doitsujin/dxvk/blob/master/VP_DXVK_requirements.json) with the new extensions and features used so it isn't forgotten. Would do it myself but don't have enough knowledge about all the specific extension use and their features.

Example extensions and their features i found have been made use of since last update of the file. (Might have missed some or added irrelevant ones)
```
VK_EXT_line_rasterization
VK_KHR_maintenance5
VK_KHR_present_id
VK_KHR_present_wait
VK_EXT_swapchain_maintenance1
VK_EXT_surface_maintenance1 (maybe? not completely sure)
VK_KHR_win32_keyed_mutex
VK_EXT_depth_bias_control
```

Blisto91:
Added `VK_NV_descriptor_pool_overallocation`

Blisto91:
Added `VK_EXT_pageable_device_local_memory`

Blisto91:
Added `VK_EXT_multi_draw`

Blisto91:
Added `VK_KHR_maintenance7`

Blisto91:
Removed the following that have now been added to the `VP_DXVK_requirements.json`
```
VK_EXT_line_rasterization
VK_KHR_maintenance5
VK_KHR_present_id
VK_KHR_present_wait
VK_EXT_swapchain_maintenance1
VK_EXT_depth_bias_control
VK_EXT_pageable_device_local_memory
VK_EXT_multi_draw
```

artemyto:
I also see VK_EXT_memory_budget, VK_EXT_shader_module_identifier, VK_EXT_extended_dynamic_state3, VK_EXT_graphics_pipeline_library mentioned here: https://github.com/doitsujin/dxvk/wiki/Driver-support

and VK_EXT_conservative_rasterization, VK_EXT_fragment_shader_interlock, VK_EXT_shader_stencil_export mentioned here: https://github.com/doitsujin/dxvk/wiki/Feature-support

but missing here: https://github.com/doitsujin/dxvk/blob/master/VP_DXVK_requirements.json

P. S. Forgive me if I misunderstood something

K0bin:
@artemyto Most of those aren't implementable by MoltenVK.

artemyto:
[...]

PR #5211, opened 2025-09-16
Require some extensions and simplify legacy binding model

doitsujin:
TL;DR `maintenance6` and `depthClipEnable` features are now a hard requirement; everything that can even remotely run D3D content should already support the latter anyway, and dxbc-spirv no longer implements any of the ancient and untested workarounds to deal with the feature not being present.

This also greatly simplifies the legacy descriptor set allocation and turns `VkDescriptorPool` objects into linear allocators that are swapped and reset every couple of frames; reasoning being here that the old code was incredibly hard to reason about *and* problematic with tooling, validation layers frequently got confused because we'd reuse descriptor sets without resetting the pool first, etc.

This **will** come at the cost of some CPU performance, but all relevant Linux drivers have `VK_EXT_descriptor_buffer` enabled anyway which isn't affected.

Needs some testing with descriptor_buffer explicitly disabled.

### DIFF
--- a/VP_DXVK_requirements.json
+++ b/VP_DXVK_requirements.json
@@ -76,6 +76,8 @@
         "dxvk_common_required": {
             "extensions": {
                 "VK_KHR_maintenance5": 1,
+                "VK_KHR_maintenance6": 1,
+                "VK_EXT_depth_clip_enable": 1
             },
             "features": {
                 "VkPhysicalDeviceFeatures": {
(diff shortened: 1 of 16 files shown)


artemyto:
Could you please update VP_DXVK_requirements.json as well?

doitsujin:
Done. I don't know how to really keep track of the profile though, we have nothing that uses it.

artemyto:
> Done. I don't know how to really keep track of the profile though, we have nothing that uses it.

A simple script could help, which would go through the codebase and find all the enabled extensions and features comparing them with the profile. I could try to contribute one if you are interested. This is not the proper solution, but at least it could catch what is missing and give you the place where it is enabled/used.

Btw here is an issue with some extension which are probably missing in the profile: https://github.com/doitsujin/dxvk/issues/3753

doitsujin:
I did update the profile against (most of) those extensions at some point.

There's also the problem of how to deal with extensions that are in use but non-essential, e.g. `NV_raw_access_chains`. We don't exactly expect anyone but NV to support that given that it addresses a hardware-specific issue. Or things like keyed mutex support which makes no sense whatsoever on non-Windows platforms. Or things like stencil export which are optional in D3D and simply impossible for some vendors to support. Not sure how much sense it makes to have those in a generic profile.

PR #5212, opened 2025-09-17
[d3d9] Advertise multisample NULL format checks

WinterSnowfall:
Note to self: don't validate vendor hack tests on ancient cards that don't actually support those vendor hacks. Looks like modern cards expose all supported sample counts with NULL format as well.

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -47,7 +47,7 @@ namespace dxvk {
     , m_behaviorFlags(BehaviorFlags)
     , m_multithread(BehaviorFlags & D3DCREATE_MULTITHREADED) {
     // Get the bridge interface to D3D9.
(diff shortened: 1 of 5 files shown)


doitsujin:
What did this break and when?

WinterSnowfall:
It didn't break anything, since Wargame: European Escalation only checked D3DMULTISAMPLE_NONE, but the correct behavior ever since the GeForce 6 generation (and other modern ATI/Intel equivalents) is to treat it as a regular format and report everything that's supported in terms of samples. I had done my testing on a GeForce 4 Ti, so it was more of an ancient history derp.

But anyway, I'll push another commit to fix some d3d8 error message derpiness soon, so don't merge it yet please :frog:.

WinterSnowfall:
Thanks to @CkNoSFeRaTU for bringing the d3d9 bridge related exception handling to my attention, I had forgotten they even existed. They were indeed simply causing a crash and not outputting anything in the logs, which could very well be confusing for first time users forgetting to use dxvk's d3d9.dll as well.

PR #5213, opened 2025-09-18
[d3d9] Fast skip DrawIndexedPrimitive(UP) calls with 0 NumVertices

### LINKED ISSUE #4803: [d3d9] `DrawIndexedPrimitive` with an underflowed primitive count "works" natively, but breaks DXVK
### Software information

The issue is general, but I reproduced it on Colin McRae Rally 3 with an **earlier** version [of my patch](https://github.com/CookiePLMonster/SilentPatchCMR3). I used to have a bug where I [attempted to draw 0 rectangles](https://github.com/CookiePLMonster/SilentPatchCMR3/commit/49461383275e9df38d75a3af13d668a14ced3c35) using the in-game drawing functions, but the problem is that this always "worked" fine on native D3D9, but resulted in issues in DXVK, hence it took me this long to notice my mistake.

In the case of this game, a 0-rectangle draw issues a call to `DrawIndexedPrimitive` with `2 * (3 * 0 - 2)) = 4294967292` primitive count. This is obviously invalid, but in native D3D9 this doesn't produce visible side effects, while with DXVK it wrecks performance and results in random corruption showing on screen:

Image

Image

### System information
- GPU: NVIDIA GTX 1070
- Driver: 560.94
- Wine version: Native, on Windows 10 22H2
- DXVK version: 2.6

### Apitrace file(s)

Something to note - this issue crashes both PIX and apitrace. I managed to capture a trace but it crashes the moment you try to preview the last frame of the capture. Nonetheless, I'm attaching the file in case it is useful in this broken state. `.7z` compresses it much better than `.zip` but it cannot be attached on GitHub, hence the double packing:

[Rally_3PC.7z.zip](https://github.com/user-attachments/files/19483573/Rally_3PC.7z.zip)

WinterSnowfall:
Should fix #4803. @CookiePLMonster If you happen to still have that broken patch version around, would appreciate if you can confirm things work and look fine with this PR.

Native d3d9, unfortunately, does not validate either NumVertices, nor PrimitiveCount in any way, however it does appear to take into account the NumVertices value to fast skip calls.

Based on observed behavior on Windows XP:
`DrawPrimitive(D3DPT_TRIANGLELIST, 0, 8388608);` -> hard locks the OS
`DrawIndexedPrimitive(D3DPT_TRIANGLELIST, 0, 0, 5000, 0, 8388608);` -> hard locks the OS and crashes the GPU
`DrawIndexedPrimitive(D3DPT_TRIANGLELIST, 0, 0, 8, 0, 8388608);` -> works just fine (with the same valid index/vertex buffers)
`DrawIndexedPrimitive(D3DPT_TRIANGLELIST, 0, 0, 0, 0, 8388608);` -> also works just fine and returns immediately

### DIFF
--- a/src/d3d9/d3d9_caps.h
+++ b/src/d3d9/d3d9_caps.h
@@ -27,7 +27,7 @@ namespace dxvk::caps {
 
   constexpr uint32_t MaxTransforms                = 10 + 256;
 
-  constexpr uint32_t TextureStageCount           = MaxSimultaneousTextures;
+  constexpr uint32_t TextureStageCount            = MaxSimultaneousTextures;
 
   constexpr uint32_t MaxEnabledLights             = 8;
 
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -3019,7 +3019,7 @@ namespace dxvk {
       return D3DERR_INVALIDCALL;
 
     if (unlikely(!PrimitiveCount))
-      return S_OK;
+      return D3D_OK;
 
     bool dynamicSysmemVBOs;
     uint32_t firstIndex     = 0;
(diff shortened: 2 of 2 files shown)


CookiePLMonster:
Confirmed to fix the test case revealed by modded Colin McRae Rally 3:
image

WinterSnowfall:
Should be good to go then, thanks for testing. I checked our code and it's not like things are expected to work properly with 0 NumVertices and non-zero PrimitiveCount anyway (or the other way around).

K0bin (approved):

PR #5220, opened 2025-09-23
[d3d9] Do sequence number tracking for directly mapped buffer

K0bin:
Use SyncAll while it is bound and replace that with the current sequence number when a different buffer gets bound.

Removes a lot of CS syncs in GTA 4. In some frames I was seeing 1ms+ wasted on CS syncs.


K0bin:
Superseded by #5350 and I haven't found another game where this matters.

PR #5223, opened 2025-09-24
[d3d9] Lock the device when applying or capturing state blocks

WinterSnowfall:
Probably not an issue in most cases, since I expect games capturing/applying state blocks and modifying state in close succession and on separate threads would be somewhat rare, but a rather big ooof nonetheless.


WinterSnowfall:
As it usually goes with my PRs, something small has turned into a rather unexpected refactoring :sweat_smile:. In this case d3d8 state blocks have also been affected. Other than covering some corner cases and optimizing how we apply state blocks (we won't iterate over the entire streams & texture arrays now needlessly), it mostly aligns what we're doing in d3d8 with what's already been done in d3d9.

Needs a bit of testing to ensure nothing has regressed.

misyltoad:
I agree with the DX9 locking change so thats fine. Maybe split that into another MR.

WinterSnowfall:
> Maybe split that into another MR.

If it hasn't caused issues so far, don't think it's that urgent. I'd rather have them both go together once I've tested the d3d8 side of things (will do that tomorrow). I'm pretty confident it's all fine though.

WinterSnowfall:
Looks good after some testing. Haven't seen anything break and haven't actually come across any games that abuse state block types in the wild, but those checks are true to native.

P.S.:
- d3d8 state block Wine tests: `0020:stateblock: 9283 tests executed (0 marked as todo, 0 as flaky, 0 failures), 0 skipped.`
- d3d9 state block Wine tests: `0020:stateblock: 14738 tests executed (0 marked as todo, 0 as flaky, 0 failures), 0 skipped.`

PR #5227, opened 2025-09-26
[d3d9] Relax lock dimension checks on SYSTEMMEM surfaces

WinterSnowfall:
Another day in the land of the corner-case API... long story short, I was a bit too aggressive (or maybe lazy) when I implemented the dimensions checks on surface locks, because, as it turns out SYSTEMMEM surfaces can be locked with zero-length pRects. And of course there's a game that does that, namely "Legends of Eisenwald".

It simply locks and unlocks the same SYSTEMMEM surface with a `pRect = [{left = 481, top = 443, right = 481, bottom = 443}]`, twice at the start of every frame when on the world map. It doesn't copy anything, nor does it use the locks otherwise, just a lock and unlock in quick succession... twice. Yeah...

Zero-length locks are invalid however on block-aligned formats and there are Wine tests for that, so I've taken the route of relaxing the specific validation which the game cares about, and hopefully we're now fully in line with native.

### DIFF
--- a/src/d3d9/d3d9_surface.cpp
+++ b/src/d3d9/d3d9_surface.cpp
@@ -116,7 +116,7 @@ namespace dxvk {
     pDesc->Type               = D3DRTYPE_SURFACE;
     pDesc->Usage              = desc.Usage;
     pDesc->Pool               = desc.Pool;
-    
+
     pDesc->MultiSampleType    = desc.MultiSample;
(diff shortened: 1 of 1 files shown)


llde:
thansk Wintersnowfall, this is basically required for this game to work without being a crashfest

K0bin (approved):

PR #5251, opened 2025-10-06
Export some D3DKMT functions from dxgi/d3d11 and forward them to gdi32

rbernon:
Similar thing is on Windows, and the MinGW runtime as well, causing linker to resolve any of those D3DKMT functions with the dxgi/d3d11 one, which then fail at runtime if the module doesn't actually export them.

The dxgi module on Windows 10 doesn't export these anymore but they were up to Windows 8 and are still in the MinGW runtime, so they still get linked with dxgi.dll when building with MinGW (and doesn't seem worth requiring toolchain updates just for that).

### DIFF
--- a/src/dxgi/dxgi.def
+++ b/src/dxgi/dxgi.def
@@ -5,3 +5,37 @@ EXPORTS
     CreateDXGIFactory2 @11
     DXGIDeclareAdapterRemovalSupport @16
     DXGIGetDebugInterface1 @17
+    D3DKMTCloseAdapter = gdi32.D3DKMTCloseAdapter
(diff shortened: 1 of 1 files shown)


rbernon:
There's an ongoing discussion on Wine side, about whether this should be forwards or exported imports (however that would need to be done with .def files).

rbernon:
Somehow this turned out to be rather controversial. The dxgi dll doesn't export these entry points anymore since Windows 10, so arguably this is incorrect and would be better fixed by removing them from the MinGW import lib (but that's much more annoying).

I'll fix it differently instead, by linking the dlls to gdi32 explicitly before dxgi.

PR #5257, opened 2025-10-08
Implement shared resources using the D3DKMT API.

rbernon:
This is an updated version of the shared resources, meant to work with upstream Wine, as I'm working on getting shared resources upstream. It should be in a future Proton 11 rebase, and will likely replace the old implementation (or we'll suffer having both, but that might get even more painful).

The old version should probably be kept around for as long as possible, if we want newer DXVK releases to be kept compatible with older Proton releases, as it's doesn't look realistic to backport the D3DKMT changes there.

As discussed, I've dropped the idea of generating the D3D runtime descriptors directly within Wine vulkan, and instead this uses a Wine-specific D3DKMTEscape code to update a shared resource private runtime data. There's no such mechanism on Windows, but there's also no way to set the private runtime data of a shared resource from user space either [*].

This PR first introduces the new API, while keeping the old DXVK metadata format, then optionally also introduces a new format that has been (partially) reversed from Windows D3D private runtime data. It might be useful for compatibility with importing shared resources that have been created from native D3D runtime, or for compatibility in general, but it also perhaps doesn't yet cover the exact same use cases as the old metadata.

It's also a bit more complicated as there's conversions to be done between different runtime versions, and exact interop behavior might require additional testing. I've based the conversion on the current interop tests I have written so far (https://gitlab.winehq.org/wine/wine/-/blob/master/dlls/win32u/tests/d3dkmt.c?ref_type=heads#L4005), more could be added.

This is still a draft as not everything has been upstreamed in Wine yet, but I'm creating this for visibility and discussion (about the descriptor format mostly). I'll create a similar PR on the vkd3d-proton side, as it requires synchronized changes.

[*] The only option on Windows seems to be to call into native D3D to allocate the resource and let it set the runtime private data, export it, and import it into Vulkan, although NVIDIA Vulkan driver still seem to generate a private runtime data descriptor in some cases.


rbernon:
Here as well I implemented a cleaner D3DKMT integration. The relevant objects also try to open their local D3DKMT handle, to use it for the shared resource implementation, calling D3DKMT API directly where appropriate.

I moved to using the native-compatible shared resource descriptors directly, as it makes it possible to share keyed mutexes with KMT handles. I added some conversion logic for cases that seem appropriate, not every case that is supposed to be working is implemented yet (esp. non-2D texture resources), there are warning messages for missing ones.

The last commit implements keyed mutexes with the D3DKMT API as well, although I'm not completely sure whether it's done right. The acquire / release functions are not yet exported by Wine gdi32.dll, and they aren't in older Proton versions either, so I'm loading them dynamically.

The same is actually mostly true with other D3DKMT API, which only have been added to Wine since ~9.0, so if it's considered desirable to keep support for older Wine versions I could keep loading them dynamically too.

rbernon:
MSCV build should hopefully be fixed now, the issue with operator==(RECT, RECT) redefinition error comes from MinGW headers which lack the operator that MSVC defines in d3d11 in C++ mode, and as D3D11_RECT is typedefed to RECT. I ifdefed the d3d9_util.h one, as fixing the MinGW headers is going to take a long time.

rbernon:
Fwiw everything needed for this except for the keyed mutex acquire / release functions is now in Wine 10.17.

rbernon:
Rebased and added a boolean atomic flag to avoid race conditions when releasing / acquiring keyed mutexes.

The keyed mutex support is now implemented in Wine since b7ab3bdef443cca417e29ff2d98640385cda08bd (will be in Wine 10.18) so I'm undrafting this. (Support for VK_KHR_win32_keyed_mutex is still pending but it'll be in Wine 10.18 and DXVK doesn't use it anyway)

SingleInfiniteLoop:
I'm getting errors like these when videos are playing using AVPro player in VRChat:
```
019c:fixme:d3dkmt:NtGdiDdDDICreateKeyedMutex2 params 0xfd150 semi-stub!
err:   Failed to create timeline semaphore
```
I'd like to ask for an advice on how to diagnose cause of the issue.
I'm using custom `wine-tkg-staging` which includes the latest changes from MRs related to keyed mutexes together with DXVK built from this branch.
~~Userpatches that I'm using:
https://github.com/SingleInfiniteLoop/wine-tkg-git/blob/master/wine-tkg-git/wine-tkg-userpatches/d3dkmt-keyed-mutex.mypatch
https://github.com/SingleInfiniteLoop/wine-tkg-git/blob/master/wine-tkg-git/wine-tkg-userpatches/d3dkmt-timeline-semaphore.mypatch~~
Thanks in advance!
EDIT: Now that `wine-staging` has been updated, the userpatches are not needed anymore as the changes are part of mainline `wine`, but the errors still occur.

SingleInfiniteLoop:
The errors are fixed with the following patch:
```
diff --git a/dlls/win32u/vulkan.c b/dlls/win32u/vulkan.c
index 47ceee74afb..22c5febcd1d 100644
--- a/dlls/win32u/vulkan.c
+++ b/dlls/win32u/vulkan.c
@@ -1788,7 +1788,6 @@ static VkResult win32u_vkCreateSemaphore( VkDevice client_device, const VkSemaph
                 nt_shared = !(export_info->handleTypes & VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_OPAQUE_WIN32_KMT_BIT);
                 export_info->handleTypes = get_host_external_semaphore_type();
             }
-            *next = (*next)->pNext; next = &prev;
             break;
         case VK_STRUCTURE_TYPE_EXPORT_SEMAPHORE_WIN32_HANDLE_INFO_KHR:
             export_win32 = *(VkExportSemaphoreWin32HandleInfoKHR *)*next;
@@ -1992,7 +1992,6 @@ static VkResult win32u_vkCreateFence( VkDevice client_device, const VkFenceCreat
                 nt_shared = !(export_info->handleTypes & VK_EXTERNAL_FENCE_HANDLE_TYPE_OPAQUE_WIN32_KMT_BIT);
                 export_info->handleTypes = get_host_external_fence_type();
             }
-            *next = (*next)->pNext; next = &prev;
             break;
         case VK_STRUCTURE_TYPE_EXPORT_FENCE_WIN32_HANDLE_INFO_KHR:
         {
```

src/d3d11/d3d11_resource.cpp
```diff
@@ -92,8 +92,32 @@ namespace dxvk {
...
+      m_fenceValue = acquire.FenceValue;
+      m_owned.store(true, std::memory_order_release);
+      return S_OK;
+    }
+
+    /* try legacy Proton shared resource implementation */
+
     Rc<DxvkDevice> dxvkDevice = m_device->GetDXVKDevice();
 
     VkResult vr = dxvkDevice->vkd()->wine_vkAcquireKeyedMutex(
```
SingleInfiniteLoop: The check for `wine_vkAcquireKeyedMutex` have to be moved from the constructor to here as well as check for `wine_vkReleaseKeyedMutex` to `ReleaseSync`.
The mainline wine doesn't implement those.

SingleInfiniteLoop:
After some testing it became apparent that whatever AVPro player does requires using the shared synchronisation object as well as keyed mutex handles when opening the shared texture in `D3D11Device::OpenSharedResource1`, which current implementation doesn't support.

src/dxvk/dxvk_adapter.cpp
```diff
@@ -25,12 +25,26 @@ namespace dxvk {
   : m_instance      (&instance),
     m_handle        (handle),
     m_capabilities  (instance, handle, nullptr) {
+    const auto& properties = m_capabilities.getProperties();
+
+    if (properties.vk11.deviceLUIDValid) {
```
doitsujin: Should probably gate this behind a wine / winevulkan check? Not sure if this is going to break Windows or not, but I suppose this whole thing isn't going to work there anyway.

src/dxvk/dxvk_memory.cpp
```diff
@@ -1119,6 +1126,67 @@ namespace dxvk {
       }
     }
 
+#ifdef _WIN32
+    auto sharedExport = (const VkExportMemoryAllocateInfo*)next;
+    if (sharedExport && sharedExport->sType == VK_STRUCTURE_TYPE_EXPORT_MEMORY_ALLOCATE_INFO) {
+      VkMemoryGetWin32HandleInfoKHR handleInfo = { VK_STRUCTURE_TYPE_MEMORY_GET_WIN32_HANDLE_INFO_KHR };
+      handleInfo.handleType = (VkExternalMemoryHandleTypeFlagBits)sharedExport->handleTypes;
```
doitsujin: `VkExternalMemoryHandleTypeFlagBits(...)`, we don't do C-style casts

doitsujin:
This also seems to be in need of a rebase.

src/dxvk/dxvk_adapter.cpp
rbernon: Every D3DKMT call introduced here should work as intended on Windows, except for the D3DKMTEscape calls which are Wine-specific. This means that importing a shared resource should work on Windows, but creating and exporting one wouldn't as there's no way to associate the resource runtime descriptor.

This only the case for resources and importing and exporting fences should work in both ways. Also, on Windows with NVIDIA drivers I believe a resource runtime descriptor is sometimes automatically being created somehow by the Vulkan driver, so exporting might work in that case as well.

src/dxvk/dxvk_fence.h
```diff
@@ -65,6 +66,24 @@ namespace dxvk {
     VkSemaphore handle() const {
       return m_semaphore;
     }
+    
+    /**
+     * \brief D3DKMT sync object local handle
```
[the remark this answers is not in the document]
rbernon: D3DKMT handles may be local or global. The local handles are process private and represent either a non-shared D3DKMT object or a shared global object opened in a process. Each local handle owns a reference to the global object in the latter case.

Global D3DKMT handles are shareable directly across process and they are the "old" way of sharing D3DKMT objects (the "KMT" Vulkan handle types).

The "new" way of sharing D3DKMT (D3D11_1+) is through NT shared handles (the non-KMT Vulkan handle types), which also own a reference to the global D3DKMT object. The NT shared handles are private to a process but can be duplicated or closed using the standard Windows DuplicateHandle / CloseHandle APIs.

These NT shared handles are allocated by calling D3DKMTShareObjects (which allows to provide a name, possibly different one on each call and share the same object multiple times under a different name), and that function takes D3DKMT local handles.

src/d3d11/d3d11_resource.cpp
```diff
@@ -126,6 +152,34 @@ namespace dxvk {
...
 
+    auto keyedMutex = texture->GetImage()->getKeyedMutex();
+    if (keyedMutex && keyedMutex->kmtLocal()) {
+      auto syncObject = texture->GetImage()->getSyncObject();
+      if (syncObject) {
+        VkSemaphoreSignalInfo info = { VK_STRUCTURE_TYPE_SEMAPHORE_SIGNAL_INFO };
+        info.semaphore = syncObject->handle();
+        info.value = m_fenceValue + 1;
+
+        if (dxvkDevice->vkd()->vkSignalSemaphore(dxvkDevice->handle(), &info)) {
```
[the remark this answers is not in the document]
rbernon: The resource synchronization is one thing I'm not completely sure about, and it doesn't seem to be working very well yet, which I am looking into at the moment.

At the very least, this interacts with the Vulkan keyed mutex support which is implemented in winevulkan: The keyed mutex acquire from https://docs.vulkan.org/refpages/latest/refpages/source/VkWin32KeyedMutexAcquireReleaseInfoKHR.html is first acquiring the mutex on the CPU side, then translates to an additional timeline semaphore wait of the keyed mutex acquired value to synchronize on the GPU side. Similarly the release is translated to an additional timeline semaphore signal submit of the released value, and a release of the mutex on the CPU side.

And I'm not completely sure how this should be implemented in DXVK as we don't use VkWin32KeyedMutexAcquireReleaseInfoKHR. Is it okay to signal the semaphore on the CPU? Should we then also call vkWaitSemaphores in D3D11DXGIKeyedMutex::AcquireSync?

src/d3d11/d3d11_resource.cpp
rbernon: Thanks, I did that.

rbernon:
Rebased and addressed the comments. @SingleInfiniteLoop I've also sent the keyed mutex fixes to Wine in https://gitlab.winehq.org/wine/wine/-/merge_requests/9381, thanks!

Fwiw it seems to be calling everything as expected, but I'm still experiencing playback issues after a few seconds in some games, although I'm not sure if this is related to the new shared resources implementation or to some other sync regression in upstream Wine. I'll need to find a way to compare the old and new implementation, but it's a bit complicated as Wine and Proton diverged a lot at this point.

rbernon:
I backported the changes and they seem to be working alright in Proton 10. The issue I was seeing probably comes from some regressions or missing patches in MF or some other places in upstream Wine. I think we can consider this to be ready.

SingleInfiniteLoop:
Can confirm that AVPro player in VRChat works with the recent changes. I tested it using my custom `wine-tkg-staging` build, which contains some userpatches taken from Proton to deal with the `winegstreamer`/`mfplat` issues.
Thank you for your great work! :heart:
P. S. Not sure if this is indication of an error, but I'm getting a warning message from `DxvkResourceAllocation::initKmtHandles`:
```
08ec:fixme:d3dkmt:NtGdiDdDDIOpenResourceFromNtHandle params 0xfdbc0 semi-stub!
08ec:fixme:d3dkmt:NtGdiDdDDIOpenSyncObjectFromNtHandle params 0xfdf40 semi-stub!
08ec:fixme:d3dkmt:NtGdiDdDDICreateKeyedMutex2 params 0xfe310 semi-stub!
08ec:fixme:d3dkmt:NtGdiDdDDIOpenResourceFromNtHandle params 0xfe620 semi-stub!
08ec:fixme:d3dkmt:NtGdiDdDDIOpenSyncObjectFromNtHandle params 0xfe180 semi-stub!
08ec:fixme:d3dkmt:NtGdiDdDDIOpenResourceFromNtHandle params 0xfd920 semi-stub!
warn:  DxvkMemoryAllocator::createImageResource: Failed to open shared NT handle
```
Videos play fine regardless.

PR #5283, opened 2025-10-21
[meta] Issue template updates

Blisto91:
Expand the bug report template slightly in the hopes it will save at least a few of the common requests we have to make when a new issue is made.
Also disable the blank template and replace it with a 'other' template. The only function this serves is again with the hopes that it will catch at least a few of the issues created that aren't using the bug report template when they should.


K0bin:
Nice idea. I doubt it'll help in practice but it also can't hurt.

PR #5293, opened 2025-10-25
[d3d9] Move non-dirty flags out of D3D9DeviceFlag

K0bin:
I think it's pretty ugly to have some random non-dirty-flags packed in with what is almost exclusively dirty flags.
So this PR moves those out of the enum and then renames the enum and its members to reflect the more well clearly defined purpose.

After that, it cleans up `SetRenderTargetInternal` a bit.

I prefer it with those changes made but feel free to yell at me about how unnecessary this is.


K0bin:
Yeah it's not a huge change even if the changed naming impacts a lot of lines.

doitsujin:
needs a rebase apparently

K0bin:
Rebased

Issue #5291, opened 2025-10-24
[d3d9] Halo: CE (2003) some shaders not drawing with Ubershader


Cambidd:
A number of shaders such as shadows for dynamic objects,  jackal shields, covenant deployable cover, pelican thrusters, probably more, are failing to draw on the latest master.

<img width="1402" height="1119" alt="Image" src="https://github.com/user-attachments/assets/8c78015c-c246-48ea-8887-b719dafce7fc" />

2.7.1
<img width="1402" height="1119" alt="Image" src="https://github.com/user-attachments/assets/03c3f376-aaa7-46bf-825a-105c62f8f89d" />

<img width="1402" height="1119" alt="Image" src="https://github.com/user-attachments/assets/48c35b03-15c1-473b-86ae-902f0b5a959a" />

2.7.1
<img width="1402" height="1119" alt="Image" src="https://github.com/user-attachments/assets/60f63284-df8e-4563-a749-46e70305a9f3" />

### Software information
Halo: CE (2003), Max setting.

### System information
- GPU: AMD 5700XT
- Driver: Mesa radv 25.1.9
- Wine version: wine-10.15 (Staging)
- DXVK version:  v2.7.1-195-g2f930978

### Apitrace file(s)
https://drive.google.com/file/d/1FmVzG58W9klQWSBF-TV0wV7woEDTQOq6/view?usp=drive_link

### Log files
[halo_d3d9.log](https://github.com/user-attachments/files/23117936/halo_d3d9.log)

Blisto91:
Thank you for the well made issue report.
While the missing shadows in the apitrace does indeed reappear when disabling ubershader the missing ship thruster effect seem to be a separate issue. I have bisected that one to commit https://github.com/doitsujin/dxvk/commit/3d2ee752b457fc79ff91f195975395e886f91d2a

K0bin:
This will need some testing:

- VS FF, PS shader: Read color1 value with specular enabled and disabled. Is it 0,0,0,0? Is it the color1 value of the vertex?
-  VS shader, FF PS: Write color1 value with specular disabled, Write no color1 value with specular enabled
- VS FF: Use color0/color1 as source but vertex doesn't have it. We do 1,1,1,1 for color0 and 0,0,0,0 for color1. Docs say it should use the material color.
- Why do we only add the rgb values of the specular color to the final color and not alpha as well?

K0bin:
[...]

Issue #4976, opened 2025-06-05
Battle for Middle Earth 1 - Glitch Glowing Textures


arntmk:
[Battle for Middle Earth Patch 2.22](https://www.moddb.com/mods/battle-for-middle-earth-patch-222/downloads/patch-222)

![Image](https://github.com/user-attachments/assets/e462abf9-c6e7-4dea-a016-e633ee53fa87)
![Image](https://github.com/user-attachments/assets/4efe74d2-8168-460e-a575-c99dad027acb)
![Image](https://github.com/user-attachments/assets/701ebdfa-5769-419d-a623-d5dbf032b5d2)
![Image](https://github.com/user-attachments/assets/47db584d-e29f-404e-aba2-39183de789f3)

Blisto91:
Hi there and thank you for the report. Please fill out the rest of the bug report template. GPU, driver, OS and a dxvk log. If you are on Windows it should appear next to the games exe.
Also if you are able then try to [make a apitrace](https://github.com/doitsujin/dxvk/wiki/Using-Apitrace) of the game

arntmk:
**GPU**: RX580 2048sp 
**Driver**: AMD Adrenalin 25.3.1
**OS**: Windows Pro x64 version 22H2 (OS Build 19045.5845)
[**DXVK Log**](https://github.com/user-attachments/files/20616327/lotrbfme_d3d9.log)
[DUMP](https://github.com/user-attachments/files/20616229/DUMP-20250605-173259-7296-14868.zip)

arntmk:
> Hi there and thank you for the report. Please fill out the rest of the bug report template. GPU, driver, OS and a dxvk log. If you are on Windows it should appear next to the games exe. Also if you are able then try to [make a apitrace](https://github.com/doitsujin/dxvk/wiki/Using-Apitrace) of the game

**GPU**: RX580 2048sp 
**Driver**: AMD Adrenalin 25.3.1
**OS**: Windows 10 Pro x64 version 22H2 (OS Build 19045.5854)
[**DXVK Log**](https://github.com/user-attachments/files/20616327/lotrbfme_d3d9.log)
[DUMP](https://github.com/user-attachments/files/20616229/DUMP-20250605-173259-7296-14868.zip)

K0bin:
Please make an apitrace.

arntmk:
> Please make an apitrace.

### // DebugView
```
[...]

PR #5296, opened 2025-10-25
[d3d9] FF: Fix for D3DRS_SPECULARENABLE

K0bin:
Works with both #4976 and #5291.

Fine as an interim solution but I need to investigate what Windows does in a couple of weird edge cases (https://github.com/doitsujin/dxvk/issues/5291#issuecomment-3447811440) once I have some time.


K0bin:
Mhmm, actually, maybe we should pass `in_Color1` through when specular is disabled, similar to what we do for both color0 and color1 when lighting in general is disabled.

K0bin:
I did test it and my guess was correct. I also tested 2 other things and added commits to fix where we differ from that in DXVK.

PR #5301, opened 2025-10-30
Fix: Add null check in CreateShaderResourceView

DealsBeam:
Adds a null check for the `ppSRView` parameter in the `CreateShaderResourceView` function to prevent a potential crash when a null pointer is passed.

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -363,7 +363,9 @@ namespace dxvk {
     if (hr != S_OK)
       return hr;
     
-    *ppSRView = view;
+    if (ppSRView)
+      *ppSRView = view;
+
     return S_OK;
   }
   
--- a/tests/test_srv.cpp
+++ b/tests/test_srv.cpp
@@ -1,0 +1,49 @@
+#include "d3d11_device.h"
+#include <iostream>
+
+int main() {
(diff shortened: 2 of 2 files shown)


src/d3d11/d3d11_device.cpp
```diff
@@ -363,7 +363,9 @@ namespace dxvk {
     if (hr != S_OK)
       return hr;
     
-    *ppSRView = view;
```
K0bin: Why go through all the work of creating the SRV only to destroy it immediately again?
Check the pointer at the beginning of the function and return early.

Judging by `CreateShaderResourceViewBase`, it should also return `S_FALSE`.

K0bin:
Does it actually crash? Because this change looks redundant:

`CreateShaderResourceView` calls `CreateShaderResourceViewBase`:
```cpp
 HRESULT hr = CreateShaderResourceViewBase(pResource,
      pDesc    ? &desc : nullptr,
      ppSRView ? &view : nullptr);
```

If that call fails, it returns the return code of it:
```cpp
  if (hr != S_OK)
      return hr;
```

`CreateShaderResourceViewBase` checks for a `nullptr`: 
```cpp
if (!ppSRView)
      return S_FALSE;
```

doitsujin:
is this some sort of ai slop? As was mentioned, this shouldn't be crashing as-is.

Issue #5330, opened 2025-11-17
DXVK_FRAME_RATE produces unstable VRR Hz counter on monitor compared to MangoHud


nokia8801:
When playing on a VRR monitor with VSYNC set to Mailbox and FPS limited below refresh rate externally with MangoHud and/or DXVK, the DXVK_FRAME_RATE option produces worse results compared to MangoHud.

MangoHud with `fps_limit=232` and `fps_limit_method=early` (produces the most stable VRR Hz counter and the smoothest frame times):
https://github.com/user-attachments/assets/6cfb592c-28c4-4880-95e1-a7486980202a

MangoHud with `fps_limit=232` and `fps_limit_method=late` (1-2ms min-max frame time variance but still a very stable VRR Hz counter):
https://github.com/user-attachments/assets/265288cc-10fd-469d-9a91-d8a5843e3b3b

`DXVK_FRAME_RATE=232`:
https://github.com/user-attachments/assets/cbfd00a0-24bd-480d-bf00-f57a12b9ff9b

(The videos are recorded in 60 FPS so it might not really show the results accurately, but it fluctuates a lot with `DXVK_FRAME_RATE` and is much worse in reality)

In-game FPS limiter is disabled. In-game VSYNC is disabled. VSYNC set externally to Mailbox with `dxvk.tearFree = True` in `dxvk.conf` and `vsync=2` in `mangohud.conf`. FPS is limited with DXVK and MangoHud to test. This is the recommended competitive setup on both Windows and Linux. You can read these to see why this is the case: <sup>[1](https://zamundaaa.github.io/wayland/2021/12/14/about-gaming-on-wayland.html)</sup> <sup>[2](https://blurbusters.com/gsync/gsync101-input-lag-tests-and-settings/14/)</sup>.

### Software information
Any game. But I'm focusing on Call of Duty games that [Plutonium](https://plutonium.pw/) supports.

### System Information
- **OS:** Arch Linux x86_64
- **Kernel:** Linux 6.17.8-arch1-1
- **Mesa:** 25.2.7-arch1.1
- **Proton:** GE-Proton10-25
- **DXVK:** v2.7.1-207-ge9ad90562cf4f78 (commit e9ad905)
- **GPU:** AMD RX 9060 XT
- **DE/WM:** GNOME 49.1 on Wayland

### dxvk.conf
```
dxgi.enableHDR = False
d3d11.maxFeatureLevel = 12_1
d3d9.shaderModel = 3
[...]

PR #5331, opened 2025-11-17
Remove useless features

doitsujin:
Nukes the built-in latency thing (which has only ever really been useful for bringing up Reflex support almost a year ago), and nukes `DXVK_FRAME_RATE` so that the limiter will now only engage via app profiles or the display mode heuristic, and only if no external limiter is already limiting us to the desired refresh rate (or less).

Users should use mangohud or other external solutions.


shelterx:
FWIW I think DXVK_FRAME_RATE was quite handy for games that didn't have a framerate limiter and you don't want to mess around with mangohud. 
Anyway, if anyone reads this PR and want the same feature with mangohud (just limit the fps with no visible stats), this should work in steam:
`MANGOHUD_CONFIG=no_display,fps_limit=60 mangohud %command%`

Tiagoquix:
MangoHUD is getting a new env. var. `MANGOHUD_FPS_LIMIT` in the next release. So, if you are reading this from the future, `MANGOHUD_FPS_LIMIT=60 mangohud %command%` may also work.

Eckoa:
great...this limiter provided the smoothest frame times for me across the board but now ill have to deal with micro stutters

oh well

Tiagoquix:
@Eckoa you can still use it via dxvk.conf file. just not the env. var. the idea is to discourage users from using it, though. use mangohud's, it's fine.

Eckoa:
> @Eckoa you can still use it via dxvk.conf file. just not the env. var. the idea is to discourage users from using it, though. use mangohud's, it's fine.

mangohuds frame pacing isnt fine which is why i used this so im left with nothing else but stuttering :|

i cant be certain the conf method will continue to to work in the future anymore either.

EDIT: anyway, this is done so Ill just either find another way or just deal with it.

EDIT2: to clarify my position, I have an Nvidia Optimus Laptop and Mangohuds pacing is atrocious in this situation and outside of this manghuds limiter causes issues for some engines (witcher 3 for example and MH Wilds as another). I dont have 3-4k$ to drop on a mediocre AMD machine as they basically dont exist outside the HP G1a or the Framework for gaming laptops and there is no assurance that would fix it as my previous AMD laptop had the same issue with Mangohud. Im not gonna try and convince anyone to keep it, but i cant help but at least voice a bit of frustration that my setup is getting crapped on pretty heavily.

netborg-afps:
Applying external Vulkan-based fps-limiters like Mangohud on dxvk is nonsense. It will result in one additional frame of input lag and possibly other issues.

The key issue here is that a frame is starting at "d3d.present()" and ending at "Vulkan.present()". For an external limiter to work effectively, it needs access to the d3d-domain.

Digger1955:
@doitsujin

Proposing mangohud as an FPS limiter is a blatant disrespect for DXVK users.
MangoHud is an external solution - the result of it will **ALWAYS** be worse than that of DXVK. And MangoHud, like any other external overlay, cannot be universally used because of compatibility problems, which results in lower GPU perfomance at best or preventing application proper functioning at worst. You don't have to look far for examples - @Eckoa and @netborg-afps have already explained everything.

Removing `DXVK_FRAME_RATE` environment variable makes no sense and only creates maintenance burden for applications that rely on DXVK, by which I mean all Proton forks and Lutris-like applications - this decision is not user-friendly.

Removing built-in latency tracker solution, which has potential uses and forcing to use proprietary vendor-locked techniques, which cannot work without both game and driver support is not user-friendly too.

You have a chance to create a competitive advantage vs Windows. Instead you snatch defeat from the jaws of victory.

WinterSnowfall:
Have you heard about its evil twin brother: `DXVK_CONFIG="dxgi.maxFrameRate = 60; d3d9.maxFrameRate = 60"` ?

Digger1955:
@WinterSnowfall

Have you ever tried to explain to a kid how to configue DXVK? User-friendly means - the simpler the better, even if it means duplicating logic.

WinterSnowfall:
Pretty sure I saw a "6 years+" label on dxvk's packaging, but anyway, we can debate this for eternity. I've only suggested a viable option, albeit one that's a bit more involved.

Eckoa:
> Have you heard about its evil twin brother: `DXVK_CONFIG="dxgi.maxFrameRate = 60; d3d9.maxFrameRate = 60"` ?

the question is, how long till this no longer works either? 

The intent seems to be to push people to other tools, Mangohud is an option for some but not for all and in the case where you speedrun games having not only the latency (im not as concerned about 1 frame) but the incosistent frame rate/time/pace that comes with Mangohud in a number of situations makes it absolutely unusable and the only other option i know of was libstrangle which was abandoned 4yrs ago. 

The lack of consistency in tooling for gaming, speedrunning, etc. only linux just gets worse now. Its fine if you game from the couch but being in the competative space is more and more frustrating. Its hard to feel like you just dont want to bother with linux anymore when youve been using it for 10yrs exclusively and every tool you used starts to evaporate.

Blisto91:
The config editions are staying as they are intended to work around buggy games and are used for several already.

WinterSnowfall:
> the question is, how long till this no longer works either?

It will work as long as the code for it stays around. And as explained in the issue that the PR addresses, it *is* needed to handle games that tie physics to framerate, which sadly wasn't at all uncommon in the d3d9 era at least. If you insist on using it, you can do so via the DXVK_CONFIG env variable.

Eckoa:
> > the question is, how long till this no longer works either?
> 
> It will work as long as the code for it stays around. And as explained in the issue that the PR addresses, it _is_ needed to handle games that tie physics to framerate, which sadly wasn't at all uncommon in the d3d9 era at least. If you insist on using it, you can do so via the DXVK_CONFIG env variable.

testing that option right now on bleeding edge experimental in MH Wilds yeilds the same frame rate instability as Mangohud, is it supposed to be identical to the previous method? because it doesnt appear to be.

WinterSnowfall:
> testing that option right now on bleeding edge experimental in MH Wilds yeilds the same frame rate instability as Mangohud, is it supposed to be identical to the previous method because it doesnt appear to be.

Both control the very same logic, yes, you can see that in the code touched by this PR. You can activate logging to make sure the options actually get applied properly, but as long as they are, the effect can only be identical.

Eckoa:
> > testing that option right now on bleeding edge experimental in MH Wilds yeilds the same frame rate instability as Mangohud, is it supposed to be identical to the previous method because it doesnt appear to be.
> 
> Both control the very same logic, yes, you can see that in the code touched by this PR. You can activate logging to make sure the options actually get applied properly, but as long as they are, the effect can only be identical.

well Im currently denuvo drm locked out of Wilds after doing testing, but preliminary test on latest bleeding edge showed that to not be the case in Monster Hunter Wilds with VkD3D 3.0. Mangohud, the DXVK_CONFIG option, libstrangle, and the in game limiter all yeild worse results than previously. 

Maybe its a VKD3D regression, maybe its this not working right with VKD3D? idk, cant test it further for this game for 24hrs.

dobosken:
So the feature remains intact, but you choose to break existing end-user configurations and arbitrarily limit how the feature is exposed? Why?

How has DXVK_FRAME_RATE become useless, and why _should_ I use "mangohud or other external solutions"?

Edit: https://github.com/doitsujin/dxvk/issues/5330#issuecomment-3540774621
This has been rock-solid for me over the years, but hey, alright.

doitsujin:
Keep in mind that vkd3d-proton's frame rate limiter is largely its own thing, the only dxvk code involved there is related to passing the configured limit to the D3D12 swapchain, which hasn't changed.

Kiotrw25:
> So the feature remains intact, but you choose to break existing end-user configurations and arbitrarily limit how the feature is exposed? Why?
> 
> How has DXVK_FRAME_RATE become useless, and why _should_ I use "mangohud or other external solutions"?
> 
> Edit: [#5330 (comment)](https://github.com/doitsujin/dxvk/issues/5330#issuecomment-3540774621) This has been rock-solid for me over the years, but hey, alright.

Don't ever use mangohud fps limiter, especially with smooth motion, those who suggests this might not care about input lag. 

The amount of inputlag will give your hand cancer. Has worked great with DXVK_FRAME_RATE, but figured out new one thanks to doitsujin's last reply here (VKD3D_FRAME_RATE=) which works as great in terms of inputlag as DXVK, so far from what i have tried.

Considering this, it could be considered useless feature for dx12 games. 

But when nvidia staff recommends DXVK_FRAME_RATE might not be so useless i think unless they catch up on the new VKD3D_FRAME_RATE, they might keep recomending DXVK limiter. This helped me aswell https://forums.developer.nvidia.com/t/smooth-motion-with-40-series-has-really-bad-input-lag/343915.

 Only found this commit cause DXVK limiter suddenly stopped working and as many others using it daily.

This commit probably gonna get tons of hate so maybe make it clear you can use VKD3D_FRAME_RATE instead fine for DX12 games. Don't suggest mangohud until you try it, especially with smooth motion :P

Why was it removed to begin with, did it break the code or cause bloat, just curious?

Tiagoquix:
I've made a proper issue to discuss this - please use it. Thanks.

https://github.com/doitsujin/dxvk/issues/5340

PR #5335, opened 2025-11-19
Add Vulkan throughput-optimized config: max CPU/GPU saturation

thee3rdplayer:
Introduces a DXVK configuration aimed at saturating Vulkan pipelines for maximum CPU/GPU throughput. 
Enables async compilation, relaxed barriers, persistent caches, and direct SSBO access, trading some stability for raw performance. 
Intended for performance testing and benchmarking on NVIDIA 4GB devices.

### DIFF
--- a/config/dxvk-throughput.conf
+++ b/config/dxvk-throughput.conf
@@ -1,0 +1,50 @@
+# DXVK THROUGHPUT-OPTIMIZED CONFIG — Vulkan saturation build
+
+# === Threading & Compilation ===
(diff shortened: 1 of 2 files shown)


mbriar:
Which AI generated this nonsense?

Blisto91:
These are all a mix of things that don't exist (in this project at least), are set to their default or are unnecessary to touch

doitsujin:
this is missing `dxvk.snakeOil = True`

PR #5348, opened 2025-11-27
Adjust behaviour of maxFrameRate options

doitsujin:
TL;DR this makes it so that positive values behave like `DXVK_FRAME_RATE` did before in that they *always* force the limiter to engage, while negative values do the thing where it only engages if the heuristic detects a higher frame rate than configured for a short time. Less annoying to handle than the env var was because that was kinda just tacked on and had special behaviour.

The downside is that *most* app profiles want the negative numbers now, except the Prince of Persia thing since I guess frame rates shooting up briefly can still cause problems. Also, -1 continues to exist as a special value, which is a bit awkward but we need a way to deal with Crysis 1.

@Blisto91 ping for heads-up.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -77,11 +77,16 @@
 # rate, but do not provide their own limiter.
 #
 # Supported values
-# -1: Always disables the limiter
 #  0: Default behaviour. Limits the frame rate to the selected display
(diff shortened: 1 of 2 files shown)


Blisto91:
Remember to change the description in the example config

Issue #5067, opened 2025-07-03
Initial load/calculation time on D3D9 demo much slower than WineD3D


TheEssem:
When starting the scene demo Median by Ümlaüt Design (https://www.pouet.net/prod.php?which=67384) with WineD3D, it takes around 3-5 seconds between the window opening and the progress bar starting to fade out. However, when starting the same demo with DXVK, it takes around 4 minutes and 20 seconds for loading to finish and the progress bar to start fading out. The load time on Windows itself also appears to be closer to the WineD3D time.

### Software information
Uses Direct3D 9. Launched at 1280x720 (16:9) resolution with fullscreen off and multisampling enabled.

### System information
- GPU: AMD Radeon RX 6700 XT
- Driver: Mesa/RADV 25.1.5
- Wine version: wine-staging 10.11 (commit 22d837f)
- DXVK version: eb0d67a0111f4263b0fbb84948eb875ec225d848

### Apitrace file(s)

https://downtown.wetdry.world/UDS3.trace

### Log files

[game.log](https://github.com/user-attachments/files/21030754/game.log)

Blisto91:
Hi there and thank you for the report.
Does running it with `DXVK_CONFIG="d3d9.cachedDynamicBuffers = True"` make a difference?

TheEssem:
> Hi there and thank you for the report. Does running it with `DXVK_CONFIG="d3d9.cachedDynamicBuffers = True"` make a difference?

It appears that it does, reducing the time closer to that of WineD3D/Windows.

Blisto91:
Can see in your log the exe is named `UDS3.exe`. Do you know if that is a generic name or unique to the game? 
Just wondering why a game named Median have an exe called UDS3

TheEssem:
> Can see in your log the exe is named `UDS3.exe`. Do you know if that is a generic name or unique to the game? 
> Just wondering why a game named Median have an exe called UDS3

It appears to be a shared, in-house engine used by multiple Ümlaüt productions. I haven't tested any other prods using their engine yet, but there should be some more listed under their Pouet group page.

WinterSnowfall:
> Hi there and thank you for the report. Does running it with `DXVK_CONFIG="d3d9.cachedDynamicBuffers = True"` make a difference?

d3d8/9 demos & benchmarking tools have been known to need it to perform adequately, at least in some scenes, so no surprise there. I would not add a built-in profile in such cases, unless it's known to affect actual games in some manner.

[...]

PR #5350, opened 2025-11-27
Change Buffer mapping strategy

K0bin:
- Don't directly map `D3DUSAGE_WRITEONLY` buffers (unless they also have `D3DUSAGE_DYNAMIC`)
- Don't put `D3DUSAGE_DYNAMIC` buffers into host visible VRAM (unless they also have `D3DUSAGE_WRITEONLY`. Technically tests show that those are in regular RAM too but let's keep those in VRAM for now.)

Both of those matches what I'm seeing in tests.

Also solves GTA IV performance on systems without resizable bar because GTA IV has 200+ MB of `D3DUSAGE_WRITEONLY` buffers.

Besides that, I port over the discard allocation throttling from the D3D11 frontend. Hopefully that allows us to get rid of the `allowDirectBufferMapping` config options because we either no longer do direct buffer mapping for the problematic buffers or because we stall to avoid running OOM in cases where we have the option to avoid `DISCARD` memory spam.

This will require lots of testing, particularly the games that have the aforementioned config option.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5546,7 +5546,8 @@ namespace dxvk {
     // (TODO: Apparently this is meant to happen for DYNAMIC too but I am not sure
     //  how that works given it is meant to be a DIRECT access..?)
     const bool respectUserBounds = !(Flags & D3DLOCK_DISCARD) &&
-                                    SizeToLock != 0;
+                                    SizeToLock != 0 &&
+                                    (desc.Pool == D3DPOOL_MANAGED || (desc.Usage & D3DUSAGE_DYNAMIC));
 
     // If we don't respect the bounds, encompass it all in our tests/checks
     // These values may be out of range and don't get clamped.


WinterSnowfall:
@K0bin 

```/* Red Faction                                *
     * Fixes crashing when starting a new game    */
    { R"(\\RF\.exe$)", {{
      { "d3d9.allowDirectBufferMapping",   "False" },
    }} },
```

Is also fixed by this PR and can be removed.

P.S.:

```
    /* Commandos 3                                *
     * The game doesn't use NOOVERWRITE properly  *
     * and reads from actively modified buffers,  *
     * which causes graphical glitches at times   */
    { R"(\\Commandos3\.exe$)", {{
      { "d3d9.allowDirectBufferMapping",   "False" },
    }} },
```

This one also appears to be fixed, not sure if by this PR or just prior d3d8 changes, but it's no longer needed.

Blisto91:
PR have not been found to regress some GPU bound games like https://github.com/doitsujin/dxvk/pull/4103 did. Tested on a R9 380x

dreams-of-formaldehyde:
This does in fact solve small BAR performance on GTA IV.

dreams-of-formaldehyde:
Late but i tested several games and none of them exhibited issues. AMD on windows and on radv.
GTA IV
Half-Life 2 (source engine in general works fine)
Fallout: New Vegas
Star Wars Battlefront II (2005)
Black Mesa
Call Of Duty Black Ops 1
Tom Clancy's Splinter Cell: Chaos Theory
F.E.A.R.
no new regressions were caused by the PR.

K0bin:
Thanks @dreams-of-formaldehyde

WinterSnowfall:
Haven't observed any ill effects either in my testing. However, I have found quite a few d3d8 `cachedWriteOnlyBuffers` config options which are no longer needed, but I think the PR can be merged as is, and I'll adjust things in a follow-up.

Blisto91:
Needs to be checked out before being merged https://github.com/doitsujin/dxvk/issues/5067#issuecomment-3614848321

Edit: sorry for the closure. Phone page was misbehaving :p

ViNi-Arco:
@dreams-of-formaldehyde 

Hi there, have you noticed any improvement in fps or frametime stability in Black Mesa?
In the rocket silo, it's unstable, even though the fps is constant.

dreams-of-formaldehyde:
> @dreams-of-formaldehyde
> 
> Hi there, have you noticed any improvement in fps or frametime stability in Black Mesa? In the rocket silo, it's unstable, even though the fps is constant.

I can redownload and test rocket silo if you want me to. I test forget about freeman and the intro. I did not notice an increase in anything other than gta iv, nor a decrease.

Issue #5314, opened 2025-11-06
Mixed Flip/Blt presenters on same HWND flicker


clshortfuse:
### Software information
Dual DX9Ex present and DX11Present on same HWND on DXVK causes flickering between presenters

### System information
N / A

### Apitrace file(s)
N / A

### Log files
N / A

Hi, everyone! I'm the author of https://github.com/clshortfuse/renodx and to support HDR in DX9 games, I do the following:

* Upgrade DX9 to DX9Ex
* Create DX11 Device
* Create DX11 Swapchain with Flip Model
* Create DX11 shared resources
* Clone DX9 swapchain buffer
* On DX9 swapchain write, write to clone
* On DX9 present, copy clone to shared DX11 resource and call DX11 Present

This actually works well in Windows (and works similarly with OpenGL => DX11). There is a problem in that I am relying on the following Windows behavior (emphasis mine):

> Use flip model in an [HWND](https://learn.microsoft.com/en-us/windows/win32/winprog/windows-data-types) that is not also targeted by other APIs, including DXGI bitblt presentation model, other versions of Direct3D, or GDI. Because bitblt model maintains an additional copy of the surface, you can add GDI and other Direct3D contents to the same HWND through piecemeal updates from Direct3D and GDI. When you use the flip model, only Direct3D content in flip model swap chains that the runtime passes to DWM are visible. **The runtime ignores all other bitblt model Direct3D or GDI content updates.**

https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-flip-model

Basically, when a HWND has a DXGI Flip Model presenter, any other blt models effectively are blocked/null-op'd. I have reports (don't have a device to confirm) that with DXVK both the DX9 and DX11 layer both try to paint the same window causing flickering. Analyzing the code, it seems any special consideration for this honestly awkward setup doesn't seem to exist.

I'd imagine tagging a HWND would be a solution though I'd hope adding support for this doesn't become too difficult. Whether falls into the scope of the project is ultimately up to you (it's a weird mix of DWM and DXGI), but I thought I'd at least bring it to your attention.

doitsujin:
Yes, we don't really support this. Wine does a bunch of stuff to make this sort-of work in that it can mux multiple Vulkan swapchains onto the same window and display the most recently presented swapchain, but honestly I ***really*** don't want to deal with any of this garbage. It's not an intended use case.

doitsujin:
[...]

PR #5357, opened 2025-12-02
[d3d9] Add API for unlocking Swapchain formats before device creation.

MapleHinata:
Hi, I'm one of the contributors for https://github.com/clshortfuse/renodx, a project that seeks to mod HDR support for games that lack it, and I received a suggestion from autumn on Discord to use the existing `ID3D9VkExtSwapchain` extension as an workaround for #5314, and during the implementation of a solution utilizing that extension, I came across some issues that required further extending dxvk. 

As it currently is, while dxvk does allow you to set the swapchain format to formats traditionally disallowed by the official Direct3D9 runtime, by the use of `ID3D9VkExtSwapchain->UnlockAdditionalFormats()`, this can only be called after the implicit swapchain is created. 

And while i can create the implicit swapchain with it's original format, call `ID3D9VkExtSwapchain->UnlockAdditionalFormats()`, and then call `IDirect3DDevice9::Reset()`to recreate the swapchain with a different format, this...doesn't always work. 

On games that use Direct3D9Ex and support recovering from a device loss, this seems to work fine, but on games that don't, the device reset fails due to still having alive losable resources, and I do not have source code for the game to safely release those resources.

While I could in theory further hook into the game in order to make that Reset operation safer, I find that adding an API to unlock the swapchain formats for a given Direct3D9 interface makes more sense, because, in my opinion, it feels a bit of an oversight to not be able to create the implicit swapchain with those unlocked formats, and this would save an unnecessary Reset for consumers of `ID3D9VkExtSwapchain`.


doitsujin:
Urg, *even more* D3D9 interop APIs. Well at least this one doesn't try to cement some obscure backend implementation detail for all eternity I guess.

The problem with all these interop APIs is that, at this point, it's a borderline unmaintainable mess of random things that people kept adding with nobody ever really thinking about the design and long-term impacts. If we add this then I at least want to get rid of the extended swapchain interface, I *already* don't know if it even works as intended due to the 2.6 swapchain rewrite, we don't really have any way to test those things either.

@misyltoad do you happen to have any idea who/what uses `ID3D9VkExtSwapchain`?

MapleHinata:
Uh, welp, sorry. 

I understand your point of view here, getting an endless stream of D3D9 Interop API PRs must be annoying. 
While i don't know what currently uses `ID3D9VkExtSwapchain`, I was trying to use it to workaround that issue, and it does still work from what i tested, it presented a `VK_FORMAT_R16G16B16A16_SFLOAT` swapchain with the `VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT` colorspace, just as I had asked from the D3D9 side.

The only reason i submitted this, was because from my perspective, this was just a small addition that addresses some usability issues I had with `ID3D9VkExtSwapchain`, without this, I'd be forced to ship a forked dxvk build with some small changes, which I *really* did not want to do, if it was possible. 

Do note that as it stands, this PR doesn't replace `ID3D9VkExtSwapchain`, it only implements the UnlockAdditionalFormats function from it, I just wanted to have a way to create an `D3D9Format::A16B16G16R16F` swapchain from it's creation, instead of doing it from an Reset.

If you'd prefer, I could move this behavior to an environment variable to reduce the code added, and keep `ID3D9VkExtSwapchain`as is, I implemented this through an interface, because I wanted an easier way to programmatically enable it, while also stopping people from enabling things they don't understand, this doesn't fully address the maintainability concerns as this is only really useful in conjunction with `ID3D9VkExtSwapchain`, but it's at least less code.

doitsujin:
> Do note that as it stands, this PR doesn't replace ID3D9VkExtSwapchain, it only implements the UnlockAdditionalFormats function from it, I just wanted to have a way to create an D3D9Format::A16B16G16R16F swapchain from it's creation, instead of doing it from an Reset.

Yes, and this looks cleaner than what we currently have. Which is why I would like to re-evaluate the current D3D9 interop situation and see what we need and which interfaces / functions we can get rid of, document the things we have better, and adjust things as necessary, even if it means breaking (or at the very least deprecating) existing APIs. Now's a good time to do so as well since we're working towards a 3.0 release early next year.

The current APIs have been biting us with every major rework lately, the 2.5 memory management stuff has been a pain in the arse with texture sharing APIs since resources are no longer guaranteed to be kept in one and the same memory location, the 2.6 swapchain rework has affected, well, all the swapchain code including this, 2.7 has been all sorts of fun with CUDA-related APIs (not relevant here), and the synchronization rework from last week once again has implications for how texture interop works. Something really has to change here.

Again, I'm not opposed to merging this, but I'd like to understand who uses the swapchain->reset method first and if that can be changed.

doitsujin:
Merging for now, sounded like Autumn has no objections to getting rid of this particular legacy method at least.

PR #5373, opened 2025-12-08
[d3d9] Introduce proper Unwrap/Return API for interop purposes

doitsujin:
TL;DR this deprecates the entirety of `ID3D9VkInteropTexture`, as well as `D3D9VkInteropDevice::TransitionTextureLayout`. These APIs no longer work conceptually due to changes in memory management and (more recently) synchronization and image layout-related stuff, and ideally I'd want to completely remove these since they simply cannot do the right thing in all cases anymore.

@AlpyneDreams @misyltoad gonna need some feedback here since you're actually using these.

Think once this is sorted out, I'll revisit #4506.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -240,7 +240,8 @@ namespace dxvk {
       return S_OK;
     }
 
-    if (riid == __uuidof(ID3D9VkInteropDevice)) {
+    if (riid == __uuidof(ID3D9VkInteropDevice)
+     || riid == __uuidof(ID3D9VkInteropDevice1)) {
       *ppvObject = ref(&m_d3d9Interop);
       return S_OK;
     }
--- a/src/d3d9/d3d9_interface.h
+++ b/src/d3d9/d3d9_interface.h
@@ -172,10 +172,9 @@ namespace dxvk {
 
     std::vector<D3D9Adapter>      m_adapters;
 
-    D3D9VkInteropInterface        m_d3d9Interop;
-
     bool m_unlockAdditionalFormats = false;
 
+    D3D9VkInteropInterface        m_d3d9Interop;
     D3D9VkExtInterface            m_d3d9ExtInterface;
 
     static const D3D9ON12_ARGS* Find9On12Args(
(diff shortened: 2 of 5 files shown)


AlpyneDreams:
This looks very promising! While it would be fine as a replacement for `TransitionTextureLayout`, unfortunately I do not think it could replace `ID3D9VkInteropTexture` for our renderer as it is for a few reasons:

### Texture Relocation

Like many modern renderers, we use bindless textures. We put _all_ of our textures in a giant descriptor set with descriptor indexing. In order to make that work, currently we have to disable memory defragmentation (we use the stable GPU address flag in #5070, but we should probably just disable it in the config instead) so that textures are not reloacted, which is not ideal but has been our only option.

While this PR addresses this issue, I'm not sure it would be practical for bindless. Particularly, it feels like it would be painful to have to call `UnwrapTexture` for every texture in our texture table (2000+ textures) before doing any bindless work and then `ReturnTexture` after we're done, and multiple times per frame at that.

Of course, I would prefer if we could keep memory defragmentation. I think the simplest way that this could be achieved would be to add an interface to disable automatic defragmentation but allow the interop user to request defragmentation manually that would report which textures were relocated, so we can update them in our descriptor sets.

### Image Layouts

For bindless textures, we also have to make sure DXVK keeps resources in their default layout and flushes deferred clears. While this PR addresses that, it has the same problem as texture relocation: it feels like it would impractical to unwrap every single texture before doing bindless work.

Something along the lines of the stable layout flag in #5070 seems like the best solution. It just marks the image as shared in the `DxvkImageCreateInfo` so that it will always be returned to its default layout and have deferred clears flushed automatically.

### Parallel Command Building

This API assumes there is a total separation of Vulkan calls from DX9 calls, which our renderer does not have. We build a Vulkan command buffer while we issue DX9 commands and then we flush the DXVK command stream and submit our command buffer after it. This is possible because we are able to handle image relocations and layout changes (which we need to be able to do for bindless anyway).

In theory it might be possible to refactor our renderer to keep Vulkan calls separate from DX9 calls (by adding another layers of indirection, for example), but the solutions I proposed above would make it unnecessary to do that.

Issue #3762, opened 2023-12-22
Painkiller Black Edition (the original, not the HD remaster) doesn't size it's game window properly


Deus-nsf:
This is fixable with PROTON_USE_WINED3D=1 %command% but the problem is that it uses direct3d to opengl, it doesn't use dxvk anymore this way, so all the dxvk.conf options are inaccessible. OpenGL is "fine" but can have severe frame pacing and input latency issues with Vsync on (even when there's no action, GPU and CPU underutilized), and SteamOS forces Vsync ON even when "allow tearing" is enabled (it does nothing).

Would be great to have good support for this game, as it's one of the best shooters ever made :)

Blisto91:
You should post more details like a screenshot since it isn't very clear what is meant. Also remember logs.

The issue presents itself like below when you start the game
<details>
  <summary>Screenshot</summary>
  
![image](https://github.com/doitsujin/dxvk/assets/47954800/338a5be5-43aa-44bf-af2f-cff04b9934a0)
</details>

It can be worked around by forcing a aspect ratio e.g. `d3d9.forceAspectRatio = 16:9` for my monitor.

Log without config (quite spammy)
[Painkiller_d3d9-no-config.log](https://github.com/doitsujin/dxvk/files/13757198/Painkiller_d3d9-no-config.log)
Log with config
[Painkiller_d3d9-yes-config.log](https://github.com/doitsujin/dxvk/files/13757200/Painkiller_d3d9-yes-config.log)

Deus-nsf:
Thank you for the answer! I should have precised that I'm on the Steam Deck, but I was able to reproduce this issue in desktop mode as well.
Switching back to DXVK (WINED3D=0), unfortunately force aspect ratio 16:9 doens't help, both in desktop an deck ui, unless the lowest resolution is selected. it's working because only 16:9 resolutions are available

waspennator:
I can also confirm the same issue on my oled deck too, plus it's not just Painkiller Black Edition that's effected, but the majority of the Painkiller games save for hell and damnation, found the same issue with the game's below.

Painkiller overdose
Painkiller Reoccurring evil
Painkiller Redemption
Painkiller resurrection

waspennator:
Tried it out Black Edition on my windows laptop, game wouldn't boot up with DXVK, sharing a link of my Trace file with it incase it helps.

https://www.mediafire.com/file/uc2pwrf5k5fzg5r/Painkiller.zip/file

[...]

PR #5387, opened 2025-12-15
[wsi] Implement exclusive fullscreen with Wine specific D3DKMT escape

rbernon:
This mechanism will be in Wine 11 (it's available since wine-11.0-rc1, though will probably only work with Vulkan after wine-11.0-rc3 not yet released, as I just sent some missing bits), and can be used to hint Wine host WSI, to fix games that are doing fancy things with their windows while in fullscreen mode. One example is Fallout 3, which incorrectly calls SetWindowPos with an upside down rectangle after restoring from minimized state.

It also allows to make windows fullscreen without having to change the window style, which is known to confuse some applications too. That becomes optional on Wine however it is still going to be necessary on Windows, as this D3DKMTEscape is Wine-specific. A more compatible solution would probably require using the D3DKMT API for presentation, but that doesn't look very realistic in the short term.


doitsujin:
Do we have a (reasonably strong) guarantee that the call will be ignored / do nothing on Windows? If I'm reading the docs correctly then this seems to be an extension mechanism where vendors can do whatever they want, so the concern would be that we may run into weird / unwanted behaviour on some setups without a wine check.

If we *do* need a wine check then I think we can resurrect the helper function that once existed that just checks whether winevulkan.dll is loaded, rather than the ntdll thing.

rbernon:
I think it should be fine as we leave hAdapter and hDevice unset, though I haven't done much testing. Doc says `If the UMD/client driver sets hContext to a non-NULL value, the driver must have also set hDevice to a non-NULL value, and hDevice must correspond to the device that owns the context.`, which makes these call actually invalid by spec at least.

I can and will add some tests in Wine conformance test suite just to make sure it doesn't regress.

DarkSwan86:
Sorry to hijack the PR here but this might help with #3762 ? Thanks!

rbernon:
The tests have been merged in Wine with https://gitlab.winehq.org/wine/wine/-/merge_requests/9783, and everything should be in place with upcoming today's Wine RC (11.0-rc3), so I think this could be ready to be merged too.

PR #5392, opened 2025-12-17
[meta] Update dxbc-spirv

doitsujin:
Changes FMad code gen in SPIR-V, as well as some debug names.

Needs some D3D11 testing.


doitsujin:
Or, hm, maybe this is a stupid idea, looking at radv codegen.

Issue #4307, opened 2024-09-30
[d3d9] The Big Secret of a Small Town cursor issues.


Blisto91:
Game have an issue with dxvk where as soon as you move the cursor it turns invisible. If you then go in to options and enable "custom cursors" it will show a game cursor with the desktop cursor on top of it.
With wined3d the cursor shows fine in the first scenario and in the second it will shrink the cursor to about the desktop ones size and show it with the game graphic.

A interesting detail is that, before you move the cursor with dxvk so it disappears, it will appear small without the "custom cursors" option. With wined3d it appears big. 
With the option enabled it will appear small for wined3d and big for dxvk.

Haven't found configs to affect the issue and the game also doesn't report using a software cursor.

### Software information
[The Big Secret of a Small Town](https://store.steampowered.com/app/409090/The_Big_Secret_of_a_Small_Town/)
Settigns like fullscreen or windowed doesn't matter

### System information
- GPU: RTX 4080
- Driver: 550.40.71
- Wine version: Proton Experimental
- DXVK version: 1.5.5 -> master. 1.5.5 is earliest that runs the game

### Apitrace file(s)
https://drive.proton.me/urls/M5VK2ZR6QW#AenBlD6ua3u7
Traces with wined3d and dxvk. Custom cursors is enabled to being with and then disabled.

### Log files
[steam-409090.log](https://github.com/user-attachments/files/17191003/steam-409090.log)

WinterSnowfall:
@Blisto91 Might want to give this a quick recheck with latest master. If the behavior hasn't changed, I'll eventually explore the dubious avenue I had in mind.

Blisto91:
Forgot to reply for all to see that i did give this a spin on latest master but with no change.
Also checked on Windows with dxvk and there it behaves the same.

damnkrat:
Original bug footage

https://github.com/user-attachments/assets/33ff5bac-15c2-4c6f-9010-2cd7298ded3c

WinterSnowfall:
@Blisto91 Friendly reminder to give it a go on Windows, then most likely close, as it's fixed on master.

Blisto91:
Seems to work fine now on master

PR #5452, opened 2026-01-17
[d3d9] Allow cursor clears on hidden HW cursors

### LINKED ISSUE #5449: D3D9 Regression: Cursor stuck after transition from CEF overlay Arizona RP(GTA:SA/SAMP)
Please describe your issue as accurately as possible. Include screenshots or videos when relevant.

Cursor gets stuck when open phone(CEF) -> open any textdraw -> close textdraw

https://github.com/user-attachments/assets/ffa79ba3-574a-4cba-851f-96ba869140a1

I figured out this bug is happening because of 66ac77a79b0cf03fa021dd8080de138ed7a140c6

The issue is likely caused by the new m_hCursor == nullptr check in d3d9_cursor.cpp, which returns early and prevents the cursor state from being restored when the focus returns to the game from the Chromium process

### Software information
Arizona RP(GTA SA:MP): lastest

### System information
- GPU: RTX 3060
- Driver: 590.48.01
- Wine version: GE-Proton-10-16
- DXVK version: 2.7.1

### Apitrace file(s)
Recording an apitrace is difficult due to the Arizona Launcher's integrity checks and multi-process nature, but the regression is 100% reproducible by toggling the mentioned commit.

### Log files
[Arizona Games Launcher_dxgi.log](https://github.com/user-attachments/files/24664254/Arizona.Games.Launcher_dxgi.log)
[Arizona Games Launcher_d3d11.log](https://github.com/user-attachments/files/24664256/Arizona.Games.Launcher_d3d11.log)
[gta_sa_d3d9.log](https://github.com/user-attachments/files/24664255/gta_sa_d3d9.log)
[gta_sa_dxgi.log](https://github.com/user-attachments/files/24664253/gta_sa_dxgi.log)
[cef_process_dxgi.log](https://github.com/user-attachments/files/24664257/cef_process_dxgi.log)

WinterSnowfall:
Addresses the sort of jank described in #5449, however it needs blessing from @Blisto91, since it will cause a regression with #4307. That game has broken cursor handling anyway, to be fair, but this "guard" at least gave users a visible cursor to work with, although its bitmap was wrong. Closes #5449.

### DIFF
--- a/src/d3d9/d3d9_cursor.cpp
+++ b/src/d3d9/d3d9_cursor.cpp
@@ -52,15 +52,10 @@ namespace dxvk {
     if (unlikely(!IsHardwareCursor() && !IsActiveSoftwareCursor()))
       return m_visible;
 
-    if (IsHardwareCursor()) {
-      // Prevents the win32 cursor from being overwritten with nullptr
-      // in situations when a hardware cursor is set, but not shown.
-      if (!m_visible && !bShow)
-        return m_visible;
+    if (IsHardwareCursor())
       ::SetCursor(bShow ? m_hCursor : nullptr);
-    } else {
+    else
       m_sCursor.DrawCursor = bShow;
-    }
 
     return std::exchange(m_visible, bShow);
   }


damnkrat:
I've been rereading through the context of #4307 to understand why that fix was needed

In that thread, you mentioned that "The Big Secret" relies on the system cursor and likely never calls SetCursorProperties (or at least doesn't define a custom hardware cursor), so blind ShowCursor calls were hiding the GDI cursor

In the case of Arizona, we clearly have a custom hardware cursor texture, which implies SetCursorProperties is definitely being called to set that bitmap

Perhaps we can distinguish the two cases by checking if the cursor has actually been initialized?
If SetCursorProperties hasn't been called yet (state is undefined), we could ignore ShowCursor requests satisfying #4307.
Once it is called (state is defined), we respect the visibility toggles normally

This would make the guard logic dependent on whether the app is actually using the D3D9 cursor features, rather than just checking if it's currently hidden

WinterSnowfall:
> In that thread, you mentioned that "The Big Secret" relies on the system cursor and likely never calls SetCursorProperties (or at least doesn't define a custom hardware cursor), so blind ShowCursor calls were hiding the GDI cursor

Thanks for thinking about this, but honestly don't bother. I've already lost about two days of my life trying to figure out what that game needs to behave properly... that "guard" simply prevented the system cursor from being hidden in this particular case, but it was only a partial fix.

The game uses both a D3D9 hardware cursor (it does create one by calling SetCursorProperties initially and sets the position exclusively using D3D9 calls) and Win32 cursor calls at the same time. So it controls the position of an invisible D3D9 cursor (which it hides every frame) and expects the system cursor to be displayed, in short.

But the real main problem is that the cursor bitmap in actuality seems to be shared between the two. What *I believe* is happening is that the game is using direct memcpy to override the cursor bitmap, but those changes obviously won't get reflected in dxvk... and polling the Win32 bitmap for changes every frame is simply too obnoxious to implement for this one game. WineD3D gets away with it because it implements both APIs.

I'm fine with simply reverting the offending lines for the moment, as in the vast majority of cases it won't matter anyway, even if it's not "accurate" behavior per se.

damnkrat:
Where and how exactly does #4307 break? I just tested this patch in this game and... It just works

https://github.com/user-attachments/assets/aee8190d-0c67-4f14-b705-63043b332057

WinterSnowfall:
Well that's... interesting. I might have wasted two days of my life for nothing then. I don't have the game, but @Blisto91 can have a look when he has the time, it's his bug report after all.

All the more reason to merge this, if it behaves properly now. Maybe the breakage was on the Win32 side.

damnkrat:
I guess the breakage was in rest code that this patch fixed. Am now compressing a video of breakage without the original patch, and soon will upload it to big secret issue

While compressing, I discovered even more current behavior depends on a state of "game cursor" state on game start(from exe), this patch fixes that

damnkrat:
Thanks for pushing. Also, a question. How exactly did you test in those 2 days if you never started the big secret?

WinterSnowfall:
> Thanks for pushing. Also, a question. How exactly did you test in those 2 days if you never started the big secret?

I came up with what I *thought* was right behavior, and Blisto would tests things. But IIRC nothing really ended up working properly... anyway, I can only be glad if it's finally sorted, though I'm still not sure how that bitmap ends up being set as the game's not using `SetCursorProperties`. Or at least it wasn't in the trace I had.

PR #5453, opened 2026-01-18
Implement unified image layouts

doitsujin:
Uses `VK_KHR_unified_image_layouts` to avoid image layout transitions and all sorts of related tracking. Has a special path to allow RDNA2 on RADV to benefit from this as well, at least for D3D11.

This is required for more fine-grained render pass synchronization in the future since transitions to/from `ATTACHMENT_OPTIMAL` currently prevent any optimizations in that area. On the other hand, this re-enables the use of `ATTACHMENT_OPTIMAL` layouts even for storage images and should fix some perf regressions that we seemingly introduced on older Intel.

Requires somewhat extensive testing, especially with:
- Games that use feedback loops (GTA 4 and friends), and
- Drivers that **don't** expose unified_layouts (Intel?), or at least with `dxvk.enableUnifiedImageLayouts=False` with validation enabled.

~~**Edit**: Marking as draft because feedback loops are completely broken right now. Having assumptions in the D3D9 frontend that are no longer true sure is fun.~~

### DIFF
--- a/src/dxvk/dxvk_context.cpp
+++ b/src/dxvk/dxvk_context.cpp
@@ -3086,6 +3086,7 @@ namespace dxvk {
   void DxvkContext::beginRenderPassDebugRegion() {
     VkSampleCountFlagBits sampleCount = VK_SAMPLE_COUNT_1_BIT;
 
+    bool hasFeedbackLoop = false;
     bool hasColorAttachments = false;
     bool hasDepthAttachment = m_state.om.renderTargets.depth.view != nullptr;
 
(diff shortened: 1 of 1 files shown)


src/dxvk/dxvk_image.cpp
```diff
@@ -387,6 +390,16 @@ namespace dxvk {
       m_info.viewFormats = m_viewFormats.data();
     }
 
+    // If feedback loops are enabled and the unified layout extension is
+    // not natively supported, we need to disable the unified layout path
+    if (m_info.usage & VK_IMAGE_USAGE_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT)
+      m_unifiedLayoutEnabled = m_unifiedLayoutEnabled && m_unifiedLayoutAvailable;
```
K0bin: Isn't this redundant because `m_unifiedLayoutEnabled` is set from the return value of `canUseUnifiedLayout()` which checks `VK_IMAGE_USAGE_ATTACHMENT_FEEDBACK_LOOP_BIT_EXT`as well?

PR #5456, opened 2026-01-19
Support for MW:R HMW-Mod

reqlay:
Add support for hmw-mod.exe
A continuation mod of the H2M-Mod.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -177,7 +177,8 @@ namespace dxvk {
     { R"(\\h2m-mod\.exe$)", {{
       { "dxgi.customVendorId",              "10de" },
     }} },
-    /* HMW-Mod - Modern Warfare Remastered        */
+    /* HMW-Mod - Modern Warfare Remastered        *
+     * AMD AGS crash                              */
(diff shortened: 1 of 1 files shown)


K0bin:
Why is this config option necessary? What does that mod do when it detects the AMD vendor ID?

reqlay:
> Why is this config option necessary? What does that mod do when it detects the AMD vendor ID?

We noticed in the mods discord community that the same config option that the mods before (h1,h2m) and the game itself had used circumvents the AMD AGS crash on startup that was happening on those mods/games fixes it here as well.
Should've stated that in the pull request my bad.

doitsujin:
> We noticed in the mods discord community that the same config option that the mods before (h1,h2m) and the game itself had used circumvents the AMD AGS crash on startup that was happening on those mods/games fixes it here as well.
Should've stated that in the pull request my bad.

Please add the reasoning to the code comment as well so that we don't end up scratching our heads over the app profile in the future.

Should be good otherwise.

JVlahovic:
IW7-mod also came out and will probably require support as well, so there might be another pr for that XD

PR #5461, opened 2026-01-21
config: Disable descriptor buffer for NFS Most Wanted (2005)

### LINKED ISSUE #5460: [Intel Arc Battlemage] NFS Most Wanted (2005) - VK_ERROR_DEVICE_LOST in v2.7.1 (Worked in v2.6.x)
### Software information

- Game Name: Need for Speed: Most Wanted (2005)
- Settings: 2560x1440, Max Settings.
- Mods used: NFSMW Widescreen Fix, NFSMW Extra Options, HD Reflections.
- Note: The issue occurs even with a clean install. The game freezes on the transition from intro movies to the main menu. Audio continues to play, menu controls still seem to work, but the visual frame is stuck.

### System information

- GPU: Intel(R) Arc(TM) B580 Graphics (Battlemage)
- Driver: 32.0.101.8331
- Wine version: N/A (Windows 11)
- DXVK version: v2.7.1 (Worked correctly in v2.6.x)

### Issue Description

On DXVK 2.7.1, the game experiences a VK_ERROR_DEVICE_LOST crash during the transition between the EA/Intro movies and the main menu.

Criseda:
Fixes #5460

The VK_EXT_descriptor_buffer extension (enabled by default in 2.7.1) causes VK_ERROR_DEVICE_LOST crashes on Intel Arc drivers for Need for Speed: Most Wanted (2005).

This PR adds a configuration profile for speed.exe with the following stable settings:

- `dxvk.enableDescriptorBuffer = False` - Primary Fix. Disables the extension responsible for the device loss/crash.
- `d3d9.cachedWriteOnlyBuffers = True` - Mitigates severe stuttering caused by the game reading from write-only buffers (a common issue in this engine).
- `d3d9.hideIntelGpu = False` - Ensures the game and DXVK identify the GPU correctly as Intel, preventing fallbacks to unstable vendor-specific paths (like the AMD profile).

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -662,7 +662,6 @@ namespace dxvk {
      * Descriptor buffer extension causes         *
      * crashes on Intel Arc drivers.              */
     { R"(\\speed\.exe$)", {{
-      { "d3d9.hideIntelGpu",                "False" },
       { "dxvk.enableDescriptorBuffer",      "False" },
       { "d3d9.cachedWriteOnlyBuffers",      "True" },
     }} },


doitsujin (changes requested):
Disabling Vulkan extension on a per game basis is **not** a solution, if descriptor buffers are causing hangs then that needs to be debugged properly.

I'll take the other two options, but we can't just start working around either our own bugs or driver bugs (whatever this is) with app profiles.

Also, entirely unrelated to your code but why the absolute fuck am I getting spammed with copilot "reviews" now that don't do anything but copy and paste the fucking PR description? *This* is why we can't buy RAM anymore?

WinterSnowfall:
>     * `d3d9.hideIntelGpu = False` - Ensures the game and DXVK identify the GPU correctly as Intel, preventing fallbacks to unstable vendor-specific paths (like the AMD profile).

If the game's AMD path is truly problematic, we should use `d3d9.hideAmdGpu = True` to get an Nvidia masquerade. Reporting Intel for a game that was right in the middle of the Intel meme iGPU age has the potential to get you sub-par graphics without you even being aware of it.

> Also, entirely unrelated to your code but why the absolute fuck am I getting spammed with copilot "reviews" now that don't do anything but copy and paste the fucking PR description? _This_ is why we can't buy RAM anymore?

Think you can disable it from the repository configuration options. Looks like it's been default enabled everywhere, because why not, everyone wants AI, right?

Edit: It's in the account settings section, not a per-repo thing. Well, there's at least that.

Criseda:
I have removed the hideIntelGpu line as this was not the cause of the issue. The AMD path is fine and it works correctly with spoofing, I apologise for making you think otherwise.

From my own testing, apitrace confirms:
- DXVK 2.7.1 d3d9.dll: Hangs after the last cutscene
- DXVK 2.6.1 d3d9.dll: Completely stable, finishes apitrace gracefully
- DXVK 2.7.1 + `dxvk.enableDescriptorBuffer = False`: Stable, finishes apitrace

That being said, I am happy to keep this PR open  with `dxvk.enableDescriptorBuffer = false` until this can be further debugged and confirmed that it may be an issue bigger than just the game on Battlemage with Windows. I'm also keeping `d3d9.cachedWriteOnlyBuffers = True` since it's a nice optimisation for the game

Download apitrace: https://drive.google.com/file/d/1F6eY9Y7jo9Sqbqt_JFl6EEGY-s2aJm2j/view?usp=sharing

pchome:
FYI: some games in the series share `speed.exe` file name.
NFS Underground (?)

Criseda:
> FYI: some games in the series share `speed.exe` file name. NFS Underground (?)

Considering how old NFS Underground is I doubt there is any negative impact if this fix ends up disabling it on both. After all, the system requirements were a 700MHz CPU, and 256MB ram 😂

artemyto:
> > FYI: some games in the series share `speed.exe` file name. NFS Underground (?)
> 
> Considering how old NFS Underground is I doubt there is any negative impact if this fix ends up disabling it on both. After all, the system requirements were a 700MHz CPU, and 256MB ram 😂

You would better add the dir of the exe to the regular expression. This is not a good thing making app profile which affects other untested games. Both because of current possible effects and because of the furure maintenance when something new could be possibly added to the profile without testing in all the affected games.

Criseda:
> > > FYI: some games in the series share `speed.exe` file name. NFS Underground (?)
> > 
> > 
> > Considering how old NFS Underground is I doubt there is any negative impact if this fix ends up disabling it on both. After all, the system requirements were a 700MHz CPU, and 256MB ram 😂
> 
> You would better add the dir of the exe to the regular expression. This is not a good thing making app profile which affects other untested games. Both because of current possible effects and because of the furure maintenance when something new could be possibly added to the profile without testing in all the affected games.

https://github.com/doitsujin/dxvk/issues/5460#issuecomment-3784654792 I test here Underground 1 along with other games in the franchise around the same period. It runs fine and well on dxvk 2.7.1, both with these changes and without. I do agree that it is good practice to avoid what you said though.

The issue with this is that the dir of the game can be renamed to whatever the user wants, so it's a really fragile way to fix this. In my case, for example, i have a fresh install version and a modded version with different dir names, so if a mod breaks, I can use the files from the vanilla game. But if the modded dir is the one with the changed name, the regex wouldn't pick it up, and crash the game.

I could add a change to `getAppConfig` to add a check for a specific file name, specifically `server.cfg`, as it is only present in NFSMW, and use this to distinguish between the two. But is it really worth it to add this workaround just for this game? I guess it's up to the maintainers to see if it's worth it

doitsujin:
`EXT_descriptor_buffer` is now globally disabled for Intel Windows.

PR #5465, opened 2026-01-23
Descriptor Heaps

doitsujin:
Adds yet another binding model implementation based on `VK_EXT_descriptor_heap` and enables it by default.

Obviously this is going to need support from both Wine and graphics drivers, but if everything is in place, then there should be a log message saying `Binding model: Descriptor heap` (as opposed to descriptor buffer or legacy).

This builds on the existing `VK_EXT_descriptor_buffer` implementation that was added in 2.7 and most of the code is shared between the two, so there aren't really any major functional changes, perhaps the biggest differences are that fetching constant buffer descriptors can now be batched rather than requiring one API call per buffer binding (happy 32-bit noises maybe), and that DXVK has to manually compute the memory layout of the descriptors now rather than querying it from the driver. Other than that it's mostly just a **much** cleaner API.

I'm hoping that this fixes the small performance regressions that we have seen on Nvidia with descriptor buffers.

The descriptor buffer path will likely be removed in the (somewhat distant) future once we can expect most users to run drivers with descriptor heap support.

### DIFF
--- a/src/dxvk/dxvk_cmdlist.cpp
+++ b/src/dxvk/dxvk_cmdlist.cpp
@@ -450,6 +450,10 @@ namespace dxvk {
 
   
   void DxvkCommandList::reset() {
+    // We will re-apply heap bindings first thing in a
+    // new command list, so reset this flag here
+    m_descriptorHeapInvalidated = false;
+
     // Free resources and other objects
     // that are no longer in use
     m_objectTracker.clear();
@@ -599,6 +603,10 @@ namespace dxvk {
           size_t                        pushDataSize,
     const void*                         pushData) {
     auto setLayout = layout->getDescriptorSetLayout(0u);
+
+    // Whether heaps are valid is command list state, not context state,
+    // to facilitate interactions with external rendering
+    this->ensureDescriptorHeapBinding();
 
     // For built-in pipelines, the push data layout will have shader-defined
     // consants first, then a byte offset to the descriptor set, in contrast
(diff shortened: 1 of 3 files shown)


FoulPlay:
This pull request breaks GTA IV's lighting for me.

- GPU: NVIDIA GeForce RTX 5060 Ti/RTX 3050 Laptop GPU
- Driver: 582.32.0
- Wine version: N/A
- DXVK version:  v2.7.1-440-gd462c009
- OS: Windows 11 Home Version 26200.7705

Screenshot 2026-01-31 235029

[GTAIV_d3d9.log](https://github.com/user-attachments/files/24987069/GTAIV_d3d9.log)

Sorry, there's no trace available because I have no where to upload it. If you have a place to upload, let me know.

Basically I went in the 3D world from the main menu and once loaded, no lighting. Happens on my laptop as well.

FoulPlay:
DXVK version: v2.7.1-441-gc302007c

[GTAIV_d3d9.log](https://github.com/user-attachments/files/24987127/GTAIV_d3d9.log)

Screenshot 2026-02-01 002239

The latest build still has the same issue.

doitsujin:
I have an old GTA IV trace lying around that replays perfectly fine here on NV (RTX 4070 on driver 580.94.17), so no idea what's going on here.

**Edit:** As @Blisto91 found out, something is broken specifically with 32-bit builds here.

doitsujin:
Should be noted that DLSS stuff is currently borked and **will** result in a black screen. Non-trivial amount of work to fix.

doitsujin:
Hacked around the DLSS thing for now since there's no clean solution.

doitsujin:
I'm merging this for now with the extension **disabled** by default. Will enable stuff bit by bit once driver support improves, but it's becoming a bit obnoxious to keep this branch separate.

doitsujin:
Or, well, I would if Github wasn't dying.

Issue #3142, opened 2022-12-20
Regression: Blazblue: Centralfiction (586140) crashes when selecting specific character (Mai Natsume)


alasky17:
### Software information
Blazblue: Centralfiction (586140)
Reported by folks here: https://github.com/ValveSoftware/Proton/issues/1031#issuecomment-1355716116

Blazblue: Centralfiction crashes when selecting Mai Natsume.  The easiest way to test this is: 
1) Launch game and get to main menu
2) Under "Practice", select Training Mode
3) Select Mai Natsume as a character - she is in the very top right corner of the selection screen with purple hair

### System information
Reproduced on Steam deck
- GPU: AMD custom GPU 0405 (vangogh, LLVM 13.0.0, DRM 3.45, 5.13.0-valve21,3-1-neptune)
- Driver: Mesa 22.0.2 (git-71648faf15) - on SteamOS Stable
- Wine version: Broken in current Proton Experimental-7.0-20221208 and bleeding edge (experimental-bleeding-edge-7.0-31585-20221220-p1a1b4a-w30ed35-d12901b-ve8d00f).  Working with Proton 7.0-6 on frozen DXVK branch
- DXVK version: HEAD at 12901b52f19ea1d3eece450a009f4602b5f2adc3

### Apitrace file(s)
Instead of getting this, I bisected the regression.  It appeared to bisect cleanly and blamed: 
commit 6ca6554452543cca14daf017437c5b7b3e848909
Author: Robin Kertels <robin.kertels@gmail.com>
Date:   Wed Mar 16 02:28:22 2022 +0100

    [d3d9] Use memory mapped files for textures

Please let me know if apitrace would still be useful to try to obtain.

### Log files
- d3d9.log:[BBCF_d3d9.log](https://github.com/doitsujin/dxvk/files/10272152/BBCF_d3d9.log)

alasky17:
I got an apitrace of crash on experimental and working dxvk version (7.0-6).  Thank you for the help Blisto91! The files are way too big to upload, so I uploaded them here with unlock code "apitrace" : https://www.codeweavers.com/xfer/alasky/blazblue-traces/

K0bin:
Unfortunately both apitraces replay just fine. I don't have access to the game either.

PhantomX:
`d3d9.textureMemory = 0` fixed for me.

Blisto91:
[...]

PR #5466, opened 2026-01-23
[d3d9] Add a tiny amount of padding to texture mapping data

K0bin:
I'm of course open to alternative ideas. I'm however not a fan of disabling the unmappable managed memory for Total War 1 given that it's a 32bit game and it would still write out of bounds into who knows what. Just adding that tiny bit of padding is super simple and adding a app option specifically for that felt like overkill.

Padding is only needed for the last mipmap. The other mip maps are tightly packed one after the other. I did test that on Windows (at least for a power of two sized texture in D3DPOOL_MANAGED).

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -727,7 +727,6 @@ namespace dxvk {
     /* BlazBlue Centralfiction                    */
     { R"(\\BBCF\.exe$)", {{
       { "d3d9.floatEmulation",            "Strict" },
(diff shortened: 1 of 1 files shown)


K0bin:
Turns out this is also not the only game that has this issue: #3142

K0bin:
Turns out not doing this as game specific option was a good idea: Shogun: Total War gets fixed by the PR too.

PR #5488, opened 2026-02-03
[util] added support POSIX mmap() for config and log

GermanAizek:
Use mmap() for Zero-copy and Demand Paging (https://en.wikipedia.org/wiki/Mmap)

References:
- https://unix.stackexchange.com/questions/474926/how-does-memory-mapping-a-file-have-significant-performance-increases-over-the-s

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -5,6 +5,11 @@
 #include <regex>
 #include <utility>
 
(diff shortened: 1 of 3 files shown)


doitsujin:
and... why?

I don't see a single good reason to introduce platform-specific code for something that isn't even remotely perf-critical.

PR #5513, opened 2026-02-17
Switch from MinGW-m64 to UCRT64

NTULINUX:
UCRT64 is a new drop-in replacement for MinGW essentially.

Tested with The Sims 4, works great!

### DIFF
--- a/build-win64.txt
+++ b/build-win64.txt
@@ -1,9 +1,9 @@
 [binaries]
-c = 'x86_64-w64-mingw32-gcc'
-cpp = 'x86_64-w64-mingw32-g++'
-ar = 'x86_64-w64-mingw32-ar'
-strip = 'x86_64-w64-mingw32-strip'
-windres = 'x86_64-w64-mingw32-windres'
+c = 'x86_64-w64-mingw32ucrt-gcc'
+cpp = 'x86_64-w64-mingw32ucrt-g++'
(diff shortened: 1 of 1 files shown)


doitsujin:
and what exactly are the very good reasons to do this while breaking everyone's build setup?

NTULINUX:
Breaking everyone's build setup?

Please see:

https://packages.debian.org/trixie/g++-mingw-w64-ucrt64

https://packages.fedoraproject.org/pkgs/mingw-crt/ucrt64-crt/

UCRT64 offers many benefits over MinGW such as improved UTF-8 support, threading, portability, etc. Projects should be migrating to UCRT64 anyway due to improvements across the board. Just trying to help out here. I am happy to do further regression testing with UCRT64 as well.

doitsujin:
MinGW has switched its default to ucrt over msvcrt for *quite* some time now (since version 12 iirc), what am I missing exactly here?

> Breaking everyone's build setup?

All Arch-based setups (including my own) for starters, because the given executable names simply don't exist in the mingw packages that *already target* ucrt.

If this is some quirk on your distro where the default executables still build against msvcrt for whatever reason then you're going to have to work around that locally.

NTULINUX:
I suppose some distros haven't switched to UCRT64 by default, I guess Arch did. Gentoo btw gives you the option via USE flag.

NTULINUX:
Disregard.

PR #5518, opened 2026-02-19
Bump to Vulkan 1.4

doitsujin:
We have required a large portion of the features promoted to Vulkan 1.4 for a while now, and we have been silently relying on device limits matching D3D11 requirements which 1.4 also enforces now, so we might as well bump the required core version and clean up the extension soup a bit.

One motivation is to use `dynamicRenderingLocalRead` for some tiler optimizations in the near future, where the core feature is required to be supported and differs somewhat from the older KHR extension, another is that this essentially enforces running on mostly known-good drivers.

Draft for now, and will probably remain as such for a while, because:
- [ ] AMD GFX6-7 GPUs do not expose Vulkan 1.4 on RADV currently, needs an app profile or something.
- [ ] Need to figure out how to make this work with vkd3d-proton to not break 11on12.
- [ ] Need to update the profile.

### DIFF
--- a/src/dxvk/dxvk_memory.cpp
+++ b/src/dxvk/dxvk_memory.cpp
@@ -1821,8 +1821,12 @@ namespace dxvk {
 
       auto vk = m_device->vkd();
 
-      VkResult vr = vk->vkMapMemory(vk->device(),
-        memory.memory, 0, memory.size, 0, &memory.mapPtr);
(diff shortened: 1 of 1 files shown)


Venemo:
> AMD GFX6-7 GPUs do not expose Vulkan 1.4 on RADV currently, needs an app profile or something.

I am planning to address this in Mesa, hopefully in the next release. AFAIK what is missing is 8-bit index type, which DXVK doesn't need but the spec requires. I have a MR of a naive implementation of that, but it performs poorly so I'd like to expore a different alternative.

PR #5529, opened 2026-02-23
[dxvk] Fix OOB indexing by queue.family when queue type is absent

werman:
Happened on device that doesn't have sparse queue exposed.

Fixes: 9157dcc0 ("[dxvk] Expose queue ownership transfer properties in device")


doitsujin:
Is there actually no way to test configurations like this *without* running such a device? This kind of stuff keeps happening.

werman:
> Is there actually no way to test configurations like this _without_ running such a device? This kind of stuff keeps happening.

I think Vulkan Profiles may do the trick

PR #5531, opened 2026-02-25
Port built-in shaders to dxbc-spv IR

doitsujin:
Based on #5528, hence the draft (but also because there's some minor cleanups to be done).

Not really something I had planned on doing, but the tiler experiments revealed some issues with the old code, specifically, low launch rate on AMD hardware due to everything reading `gl_Layer` or `gl_FragCoord.xy` (yeah apparently they managed to make *that* slow), which *does* actually cause a nontrivial amount of overhead when there are lots of blits or weird resolves happening within a frame (which mostly affecty D3D9 since D3D11 stuff will typically hit the native Vulkan functions anyway).

There was also a case of `gl_FragDepth` export not being optimized out despite the depth test being turned off, which made things export-bound for no good reason.

Emitting dxbc-spirv IR directly has the advantages that we can generate unique shaders for each format, view type etc being fed into these passes, as opposed to having dozens of GLSL files around. The fallback paths previously weren't even complete, with some potential real-world impact for D7VK (@WinterSnowfall), now at least the `copyImage` and `copyImageToBuffer` paths can almost fully replace the native Vulkan functions (tested AC:Odyssey with that, which is a fairly complex workload for texture copies), only real exceptions would be some cursed color formats that have no texel buffer support, and even that could technically be worked around.

Needs a bunch of testing across various games, if anything breaks then it's *probably* going to happen in a D3D9 game.


WinterSnowfall:
> only real exceptions would be some cursed color formats that have no texel buffer support, and even that could technically be worked around.

If by cursed you mean 16-bit, then yes, I've had complaints of black screens during intro playback on MSAA swapchains and the like from people using D7VK on radv, though the same path seems workable on both Nvidia and Intel. Otherwise, I'll happily take more :b:erf.

doitsujin:
Well, guess that's another thing for the todo list then.

PR #5535, opened 2026-03-03
Remove redundant DxvkShaderKey

K0bin:
... and replace it with `DxvkShaderHash`.

Not strictly needed for the new compiler but it's simply nicer.


K0bin:
Also, it's not just about the D3D9 shader format being terrible. We don't really have a code size for fixed function shaders either because we create the key before creating the code for those.

PR #5537, opened 2026-03-03
[d3d9+dxso] Remove a bunch of shader options that are mostly unused

K0bin:
Most of those are pretty much never changed out of the box and I don't think we need to maintain micro-optimization options for end users.

The new compiler won't have those either.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -315,14 +315,6 @@
 # d3d9.clampNegativeLodBias = False
 
 
-# Declares vertex positions as invariant in order to solve
-# potential Z-fighting issues at a small performance cost.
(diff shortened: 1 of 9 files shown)


K0bin:
I think that was a bit of a micro-optimization to disable that for Hat in Time.

The option was implemented as one of the very first d3d9 ones.
https://github.com/misyltoad/d9vk/commit/1f61d2ffd6e7ccf7ee0a8e87849d65f7b2c8cfe6
Back then Hat in Time was the only game in the config.
I could imagine that the plan was to disable this for games that don't need it. We never really did that though.

PR #5538, opened 2026-03-04
[d3d9] Refactor shader creation

K0bin:
Moves the validation code from `D3D9CommonShader::GetShaderModule` to `D3D9DeviceEx::CreateShaderModule`.
On top of that, it reduces the reliance on `DxsoModule` and `DxsoModuleInfo`.
It also moves the current shader creation code into a function `CreateLegacyShader`, strongly inspired by how it's done in the D3D11 frontend.

### DIFF
--- a/src/d3d9/d3d9_shader.cpp
+++ b/src/d3d9/d3d9_shader.cpp
@@ -26,21 +26,6 @@ namespace dxvk {
 
       std::ofstream file(str::topath(str::format(dumpPath, "/", name, ".sm3_dxbc").c_str()).c_str(), std::ios_base::binary | std::ios_base::trunc);
       file.write(reinterpret_cast<const char*>(pShaderBytecode), bytecodeLength);
-
-      char comment[2048];
-      Com<ID3DBlob> blob;
-      HRESULT hr = DisassembleShader(
-        pShaderBytecode,
-        TRUE,
-        comment,
-        &blob);
-
-      if (SUCCEEDED(hr)) {
-        std::ofstream disassembledOut(str::topath(str::format(dumpPath, "/", name, ".sm3_dxbc.dis").c_str()).c_str(), std::ios_base::binary | std::ios_base::trunc);
-        disassembledOut.write(
-          reinterpret_cast<const char*>(blob->GetBufferPointer()),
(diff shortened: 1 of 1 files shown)


liu547161153:
I merged it into my branch.

src/d3d9/d3d9_shader.cpp
```diff
@@ -41,177 +36,104 @@ namespace dxvk {
         &blob);
       
       if (SUCCEEDED(hr)) {
-        std::ofstream disassembledOut(str::topath(str::format(dumpPath, "/", name, ".dxso.dis").c_str()).c_str(), std::ios_base::binary | std::ios_base::trunc);
+        std::ofstream disassembledOut(str::topath(str::format(dumpPath, "/", name, ".sm3_dxbc.dis").c_str()).c_str(), std::ios_base::binary | std::ios_base::trunc);
```
doitsujin: Do we actually want to keep the disasm dump? Especially with dxbc-spv shipping a convenient disassembler now I'm tempted to just drop it, d3d11 never wrote out the disassembled code either.

src/d3d9/d3d9_shader.cpp
```diff
@@ -41,177 +36,104 @@ namespace dxvk {
...
+
     DxsoReader reader(
       reinterpret_cast<const char*>(pShaderBytecode));
-
     DxsoModule module(reader);
 
-    if (unlikely(module.info().shaderStage() != ShaderStage))
-      throw DxvkError("GetShaderModule: Bytecode does not match shader stage");
+    // Decide whether we need to create a pass-through
+    // geometry shader for vertex shader stream output
```
doitsujin: I somehow doubt this is what's actually happening here

src/d3d9/d3d9_shader.cpp
K0bin: Fine by me. I can modify the PR to drop it later.

src/d3d9/d3d9_shader.cpp
K0bin: That comment already existed before and was simply moved with the rest of the code. But yes, it's not really fitting there. I'll change that.

PR #5550, opened 2026-03-12
[dxvk] Support VK_PRESENT_MODE_FIFO_LATEST_READY_KHR

silvanshade:
This PR adds support for `VK_PRESENT_MODE_FIFO_LATEST_READY_KHR`.

FIFO Latest Ready will be chosen (if supported) before Mailbox when selecting a tear-free present mode.

Tested working on an NVIDIA GeForce RTX 5090 with driver 595.45.04.


doitsujin (changes requested):
Needs an extension enable for `VK_KHR_present_mode_fifo_latest_ready` in `dxvk_device_info.h/.cpp`.

I guess the main question here is what exactly this is doing better than Mailbox to be higher up in the priority list.

silvanshade:
> Needs an extension enable for `VK_KHR_present_mode_fifo_latest_ready` in `dxvk_device_info.h/.cpp`.

Okay, I added an extension enable.

I'm not sure how meaningful of a change it is to have that though since, as I understand it, the logic here should just cause the mode to be skipped over if unsupported, since it shouldn't appear in `pSupported`:

```c++
    // Just pick the first desired and supported mode
    for (uint32_t i = 0; i < numDesired; i++) {
      for (uint32_t j = 0; j < numSupported; j++) {
        if (pSupported[j] == desired[i])
          return pSupported[j];
      }
    }
```

> I guess the main question here is what exactly this is doing better than Mailbox to be higher up in the priority list.

This is from the [documention](https://docs.vulkan.org/refpages/latest/refpages/source/VK_KHR_present_mode_fifo_latest_ready.html):

> While this seems similar in concept to [VK_PRESENT_MODE_MAILBOX_KHR](https://docs.vulkan.org/refpages/latest/refpages/source/VkPresentModeKHR.html), the fundamental difference is that the processing of the present requests is done during vblank. From the application perspective, this means for example, that in a flip-based model, a single vblank may cause multiple swapchain images to be released at once, while [VK_PRESENT_MODE_MAILBOX_KHR](https://docs.vulkan.org/refpages/latest/refpages/source/VkPresentModeKHR.html) may continuously be releasing images as new requests become ready.
>
> This additional present mode is useful when using a time-based present API.

I guess there are two points of consideration which motivate this.

First, FIFO Latest Ready is more forward looking due to potential interaction with `VK_EXT_present_timing` and similar functionality.

The second is less clear but I would kind of expect that if a driver supports both Mailbox and FIFO Latest Ready, it is possible or maybe even likely that the FIFO Latest Ready implementation is already more well behaved with regard to pacing even without explicit integration with application-specific present timing functionality.

This is just my perspective as an end-user interested in the feature though.

doitsujin:
> I'm not sure how meaningful of a change it is to have that though since, as I understand it, the logic here should just cause the mode to be skipped over if unsupported, since it shouldn't appear in pSupported:

It's required by the Vulkan spec to explicitly enable any feature + extension that's being used, and drivers / layers / tools are allowed to not even expose the present mode if the feature bit isn't enabled.

As for the present mode itself, the documentation is unfortunately not quite clear on what's happening, I'm also not sure what the Mailbox situation is like on Nvidia right now but the thing that makes Mailbox useful for us is that it's similar to `IMMEDIATE` without any delays or tearing. `FIFO_LATEST_READY` sounds like it would require more additional back buffers, but I'm not sure if it does.

Definitely going to need to test this at some point with various different setups, but I'm extremely swamped with higher-priority work at the moment.

silvanshade:
> > I'm not sure how meaningful of a change it is to have that though since, as I understand it, the logic here should just cause the mode to be skipped over if unsupported, since it shouldn't appear in pSupported:
> 
> It's required by the Vulkan spec to explicitly enable any feature + extension that's being used, and drivers / layers / tools are allowed to not even expose the present mode if the feature bit isn't enabled.

Fair enough.
 
> As for the present mode itself, the documentation is unfortunately not quite clear on what's happening, I'm also not sure what the Mailbox situation is like on Nvidia right now but the thing that makes Mailbox useful for us is that it's similar to `IMMEDIATE` without any delays or tearing. `FIFO_LATEST_READY` sounds like it would require more additional back buffers, but I'm not sure if it does.

Right. Mailbox behaves as expected on NVIDIA as far as I am aware.

The FIFO_LATEST_READY should behave similarly to MAILBOX even without additional configuration. It appears to based on my limited experimentation.

It is possible that it has slightly more latency than MAILBOX due to its orientation toward more consistent frame pacing and how the image dequeuing is localized during Vblank.

For some additional context, this is from the [original extension proposal](https://github.com/KhronosGroup/Vulkan-Docs/blob/main/proposals/VK_EXT_present_mode_fifo_latest_ready.adoc#3-proposal):

> In the absence of a present timing extension, a "ready" request is effectively any request issued after a `VkPresentInfoKHR’s` `pWaitSemaphores` have been signaled. In that context, this extension does not provide more than `VK_PRESENT_MODE_MAILBOX_KHR` already offers. However, when using a target present time, the presentation engine would only dequeue all present requests for which the specified present time has lapsed.

and (`VK_EXT_present_timing`: [3.5 Presentation timings feedback](https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_present_timing.html#statistics)):

> The semantics of specifying a target present time only apply to FIFO present modes (`VK_PRESENT_MODE_FIFO_KHR, VK_PRESENT_MODE_FIFO_RELAXED_KHR` and `VK_PRESENT_MODE_FIFO_LATEST_READY_KHR`). When attempting to dequeue a presentation request from the FIFO queue, the presentation engine checks the current time against the target time.

and (`VK_EXT_present_timing`: [3.6 Scheduling presents](https://docs.vulkan.org/features/latest/features/proposals/VK_EXT_present_timing.html#scheduling)):

> `presentStageCount` only reports the number of stages which contain definitive results. However, time values in completed `pPresentStages` can still be 0 for multiple reasons. Most notably, it is possible for a presentation request to never reach some present stages, for example if using a present mode that allows images to be replaced in the queue, such as `VK_PRESENT_MODE_FIFO_LATEST_READY_KHR`. Platform-specific events can also cause results for some present stages to be unavailable for a specific presentation request.

Finally there's some hints about the intended behavior from the Holoscan SDK docs about [presentation modes](https://docs.nvidia.com/holoscan/sdk-user-guide/api/cpp/enum_present__mode_8hpp_1aa9622576b4044e44067b94fe16c59620.html):

>  enumerator MAILBOX
>
>  the presentation engine waits for the next vertical blanking period to update the current image. Tearing cannot be observed. An internal single-entry queue is used to hold pending presentation requests. If the queue is full when a new presentation request is received, the new request replaces the existing entry, and any images associated with the prior entry become available for reuse by the application. One request is removed from the queue and processed during each vertical blanking period in which the queue is non-empty.

and

>  enumerator FIFO_LATEST_READY
>
> specifies that the presentation engine waits for the next vertical blanking period to update the current image. Tearing cannot be observed. An internal queue is used to hold pending presentation requests. New requests are appended to the end of the queue. At each vertical blanking period, the presentation engine dequeues all successive requests that are ready to be presented from the beginning of the queue.

As I suggested before, I think this mode may preferred over MAILBOX due to other less obvious factors involved in the implementation.

But I agree that it is a good idea to test this.

Either way, if you believe there is a case for continuing to use MAILBOX when both modes are supported, I would be happy to modify this PR to allow an explicit choice between the two modes. I considered that originally but opted not to since I wasn't sure what would be the right way to expose such an option, I mean like what to call it, etc.

> Definitely going to need to test this at some point with various different setups, but I'm extremely swamped with higher-priority work at the moment.

I'm willing to help with this if you can describe what information you need.

silvanshade:
Going to close this. I might submit an alternate implementation that doesn't override Mailbox.

Anuskuss:
While giving the user the option to choose sounds good in practice, I don't think your average user (including me) can make an informed decision between the two.
FWIW I've been experimenting with [Smooth Frog](https://www.aperturegrille.com/software) in "Black Bars White" mode (scrolling vertical lines) and `IMMEDIATE` (V-Sync off) tears and looks like shit (which is expected ofc) and `MAILBOX` (`dxvk.tearFree=True`) doesn't tear and looks okay, although the lines aren't quite correctly spaced like `FIFO` (V-Sync on). I would've loved to also compare `FIFO_LATEST_READY` but I have no idea how to even force that mode. And if the lines don't tear and are correctly spaced (= good frame pacing), I think that it should definitely be preferred over `MAILBOX`, even *if* it increases latency (if you want lowest latency you'll use `IMMEDIATE` anyway).

**Edit:** Saw your PR over at https://github.com/flightlessmango/MangoHud/pull/1989 and I also think it makes more sense there than here. Anyway, both `MAILBOX` and `FIFO_LATEST_READY` look basically the same. The only edge it has over `MAILBOX` is the potential interaction with `VK_EXT_present_timing`. This may be reconsidered only once the DXVK makes use of that extension though.

PR #5551, opened 2026-03-14
[util] Tweak the regex for VTM - Bloodlines

WinterSnowfall:
`vampire.exe` seems to be too popular for its own good, as Vampire: The Masquerade - Redemption also uses the same executable name. I'm fairly sure some visual novels have appropriated it as well.

For Bloodlines, GOG installs it in `Vampire - The Masquerade - Bloodlines`, Steam apparently in `Vampire The Masquerade - Bloodlines`, no idea about retail, but I'll make a reasonable assumption that whatever it is, it's also captured by the new regex.

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -599,8 +599,8 @@ namespace dxvk {
     { R"(\\csgo\.exe$)", {{
       { "d3d9.hideNvidiaGpu",               "True" },
     }} },
-    /* Vampire - The Masquerade Bloodlines        */
-    { R"(\\vampire\.exe$)", {{
+    /* Vampire: The Masquerade - Bloodlines       */
+    { R"(\\Vampire.*Bloodlines\\vampire\.exe$)", {{
       { "d3d9.deferSurfaceCreation",        "True" },
       { "d3d9.memoryTrackTest",             "True" },
       { "d3d9.maxAvailableMemory",          "1024" },


Blisto91:
Bloody hell

WinterSnowfall:
> Bloody hell

No, this is about bloody lines :drum:

pchome:
Cursed.

> For Bloodlines, GOG installs it in `Vampire - The Masquerade - Bloodlines`, Steam apparently in `Vampire The Masquerade - Bloodlines`, no idea about retail, but I'll make a reasonable assumption that whatever it is, it's also captured by the new regex.

Meanwhile user renaming directory to something `VTMB` for easy access from terminal/script. Or
`$ lgogdownloader --subdir-galaxy-install=VTMB --galaxy-install vampire`

WinterSnowfall:
> Meanwhile user renaming directory to something `VTMB` for easy access from terminal/script. Or `$ lgogdownloader --subdir-galaxy-install=VTMB --galaxy-install vampire`

User renaming can happen in any of the cases where we have builtin configs with directory matching. This isn't exactly a new problem.

The alternative would be to leave it as is and match other things as collateral damage, I guess.

pchome:
Alternatively you could use "**not**" (`[^]`) to exclude `Redemption`
something like this: `'\\[^R\\]+\\vampire\.exe$'`
but I guess matches are case-insensitive.

Or `'\\[^B\\]+\\vampire\.exe$'` to exclude `Bloodlines` as dummy match above `Bloodlines` match.

WinterSnowfall:
> Alternatively you could use "**not**" (`[^]`) to exclude `Redemption` something like this: `'\\[^R\\]+\\vampire\.exe$'` but I guess matches are case-insensitive.
> 
> Or `'\\[^B\\]+\\vampire\.exe$'` to exclude `Bloodlines` as dummy match above `Bloodlines` match.

And how exactly does this handle cases of user renaming better than the above... ?

pchome:
> And how exactly does this handle cases of user renaming better than the above... ?

At least it will match `\vtmb\`, `\bloodlines\`, etc.
Still will not match `\game\` but a bit better...

... oh, you mean those games use different api? Then nevermind.

doitsujin:
question is, does the app profile have a negative effect on any of the other games it matches? If not then I'd suggest we leave it as-is due to valid ergonomics concerns.

WinterSnowfall:
> question is, does the app profile have a negative effect on any of the other games it matches? If not then I'd suggest we leave it as-is due to valid ergonomics concerns.

Fair enough. I think it's generally harmless for Redemption at least.

PR #5552, opened 2026-03-14
add arm build

john-peterson:
it should find the include dir from compiler

echo | aarch64-w64-mingw32-clang++ -E -Wp,-v -xc++ -

but I left some code anyway if some one need  more dirs

I ended up linking the necessary folders

ln -s $P/include/spirv $P/opt/llvm-mingw-w64/generic-w64-mingw32/include/
ln -s $P/include/vulkan $P/opt/llvm-mingw-w64/generic-w64-mingw32/include/
ln -s $P/include/vk_video $P/opt/llvm-mingw-w64/generic-w64-mingw32/include/

I am building in termux with

meson setup out --wipe --cross-file build-arm64.txt --prefix=$P

### DIFF
--- a/build-arm64.txt
+++ b/build-arm64.txt
@@ -1,0 +1,17 @@
+# vim:ft=meson
+[binaries]
+c = 'aarch64-w64-mingw32-gcc'
+cpp = 'aarch64-w64-mingw32-g++'
+ar = 'aarch64-w64-mingw32-ar'
+strip = 'aarch64-w64-mingw32-strip'
+windres = 'aarch64-w64-mingw32-windres'
+
+[properties]
+needs_exe_wrapper = true
+# sys_root = '/data/data/com.termux/files/usr/opt/llvm-mingw-w64/generic-w64-mingw32'
+
+[host_machine]
+system = 'windows'
+cpu_family = 'aarch64'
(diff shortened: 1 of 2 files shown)


pchome:
Cross-files are pretty much standard, no need to pull all possible variants into the project.
Maybe some examples could be in wiki.

https://mesonbuild.com/Cross-compilation.html#cross-file-locations

> Distributions are encouraged to ship cross files either with their cross compiler toolchain packages or as a standalone package, and put them in one of the system paths referenced above.
>
> These files can be loaded automatically without adding a path to the cross file. For example, if a `~/.local/share/meson/cross` contains a file called **x86-linux**, then the following command would start a cross build using that cross files:
```
meson setup builddir/ --cross-file x86-linux
```

mati865:
Shouldn't this use Clang rather than GCC considering it's the only supported compiler?

JustCallMeJade:
I think it's better to use arm64ec than aarch64 little directx games use aarch64

mati865:
Proton now builds DXVK as ARM64X https://github.com/ValveSoftware/Proton/commit/a5ca30fed05f1e90b795358b958813338b41b7cc
However that requires a prerelease of llvm-mingw.

PR #5564, opened 2026-03-25
[d3d11] Promote D16 depth textures to D32 when also used as shader resource

### LINKED ISSUE #3583: Star Trek Resurgence: striped shadows in several areas
[Star Trek Resurgence](https://store.epicgames.com/en-US/p/star-trek-resurgence) (UE4, DX11) has weird striped shadows in several areas, similar but probably not related to the depth bias issues observed in some d3d9 games that's been fixed a few weeks ago.

### Software information
Star Trek Resurgence v1.05 from EGS, the game has no graphical settings except resolution.

### System information
- GPU: AMD 6900XT
- Driver: Mesa, latest master (23.3), VK_EXT_depth_bias_control is supported
- Wine version: Wine-GE-Proton8-12
- DXVK version: latest master

### Apitrace file(s)
* [Apitrace file](https://downloads.fdossena.com/issues/stresurgence_apitrace.7z)
* [Renderdoc trace](https://downloads.fdossena.com/issues/stresurgence_renderdoc.7z)

### Screenshots
Notice the striped pattern on his face and uniform  
Untitled

Looking at the Renderdoc trace, from what I can tell, the artifact appears around 2930, where it takes in input seemingly valid data (depth and normal) and outputs this striped mess.

diegoalejandrogomez:
## Summary
- Promote D16_UNORM to D32_SFLOAT for textures bound as both depth stencil and shader resource
- Update DSV and SRV view creation to match the promoted image format

## Problem
Games that use D16 shadow maps with `SampleCmp` (PCF filtering) exhibit visible banding/moiré artifacts on Vulkan that are not present on native D3D11 or wined3d (OpenGL).

Vulkan's `VK_FORMAT_D16_UNORM` provides exactly 16 bits of depth precision. D3D11 and OpenGL drivers likely use higher internal precision for D16 depth buffers, as the OpenGL spec explicitly allows implementations to use higher bitdepth than requested for internal formats.

## Fix
When a texture is created with both `D3D11_BIND_DEPTH_STENCIL` and `D3D11_BIND_SHADER_RESOURCE` and the format resolves to `VK_FORMAT_D16_UNORM`, promote it to `VK_FORMAT_D32_SFLOAT`. This matches the effective precision behavior of native D3D11/OpenGL, eliminating the shadow banding artifacts.

## Testing
- Verified fix with Star Trek Resurgence apitrace (issue #3583) — shadow banding eliminated
- Regression tested with Age of Empires: Definitive Edition — no visual issues

Closes #3583

### DIFF
--- a/src/d3d11/d3d11_texture.cpp
+++ b/src/d3d11/d3d11_texture.cpp
@@ -25,6 +25,16 @@ namespace dxvk {
     DXGI_VK_FORMAT_INFO   formatPacked = m_device->LookupPackedFormat(m_desc.Format, formatMode);
     m_packedFormat = formatPacked.Format;
 
+    // Promote D16 to D32 when the texture is used as both depth and shader resource.
+    // D16 shadow maps sampled with SampleCmp produce visible banding artifacts on
+    // Vulkan that are not present on native D3D11 or OpenGL, likely because those
+    // APIs or their drivers use higher internal precision for D16 depth buffers.
+    if (formatInfo.Format == VK_FORMAT_D16_UNORM &&
+        (m_desc.BindFlags & D3D11_BIND_DEPTH_STENCIL) &&
+        (m_desc.BindFlags & D3D11_BIND_SHADER_RESOURCE)) {
+      formatInfo.Format = VK_FORMAT_D32_SFLOAT;
+    }
+
     DxvkImageCreateInfo imageInfo;
     imageInfo.type            = GetVkImageType();
     imageInfo.format          = formatInfo.Format;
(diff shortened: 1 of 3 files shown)


doitsujin:
Is there a test proving that native drivers actually promote D16 to a higher-precision format?

Even if it does, we'd need to special-case copies in a million different places so this isn't really mergeable as-is.

diegoalejandrogomez:
I replayed the same apitrace on a Windows 10 VM through native D3D11 (WARP software renderer) and the moire is present there as well.

image

This suggests the banding is inherent to D16 shadow map precision and not a DXVK-specific issue. The original reporter mentioned the artifact was "fainter" on native Windows, which may be due to differences in how the NVIDIA D3D11 driver handles PCF filtering on D16, but the underlying artifact is the same.

Given this, the format promotion approach is probably not the right fix. Happy to close this PR if you agree, or adjust the approach if you think there's still something worth addressing here.

doitsujin:
Yeah promoting depth formats willy nilly just seems like a really bad idea, it's more likely that there are stiill differences in the way depth bias is implemented between drivers (which *is*, unfortunately, allowed).

PR #5567, opened 2026-03-27
Optimize DXVK context hot paths

Pppp1116:
use codex xhigh to optimize hotpaths, recommend proper testing. it compiles.


K0bin:
> recommend proper testing. it compiles

So you didn't even test it?

doitsujin:
yeah i'm not reviewing hundreds of lines of AI vomit. Sure, some of these transforms look harmless, but they are also exactly the kind of thing that any compiler from 1990 onwards will cover anyway (CSE etc), but things like the dynamic state check are just pointless.

PR #5577, opened 2026-04-09
macOS/native: improve WSI fallback and related fixes

fbraz3:
## Summary

This PR proposes a small macOS-focused patchset on top of `v2.6.x`, validated while bringing **Command & Conquer: Generals Zero Hour (2003)** to modern native platforms in the **GeneralsX** project.

GeneralsX repository:
- https://github.com/fbraz3/GeneralsX

The project uses DXVK-native as the graphics backend for an SDL3-based game runtime on Linux/macOS/Windows.

## Why this patchset exists

While integrating DXVK into a native macOS game runtime, we found a few startup/runtime issues that are easy to reproduce outside Wine/proton-style launch flows:

1. Native app launches where `DXVK_WSI_DRIVER` is not preset may fail early instead of selecting an available backend.
2. On macOS-native runs, DXSO wrapper emission could generate shadow-sampler wrapper paths that are unnecessary/problematic for this pipeline.
3. Missing include for `size_t` in utility headers can break stricter toolchains.

## Changes in this PR

Commits included:
- `ffcdbcaf` macOS: bake GeneralsX DXVK patchset into source
- `73d9e9f3` macOS: avoid shadow sampler wrapper emission in DXSO path
- `35965877` Fix DXVK_WSI_DRIVER crash on macOS native builds
- `46a3bc01` fix(util): include cstddef for size_t

Main behavioral changes:
- Improve native startup behavior on non-Win32 platforms when `DXVK_WSI_DRIVER` is not set:
  - keep explicit env var precedence when present
  - otherwise auto-select available backend (`SDL3` > `SDL2` > `GLFW`) instead of throwing immediately
- Improve exception visibility by making `DxvkError` propagate useful text via `std::exception::what()` in common handlers
- Avoid emitting problematic/unused shadow-sampler wrapper path in DXSO compilation for this macOS-native route
- Add missing `<cstddef>` include for `size_t` correctness on stricter builds

## Why target `v2.6.x` instead of newer lines

This contribution is intentionally based on `v2.6.x` because GeneralsX currently aims to keep broad macOS coverage, including older systems and driver stacks.

DXVK `2.7+` tightened baseline requirements (see release notes/wiki), notably:
- Vulkan 1.3 baseline
- required `VK_KHR_maintenance5`
- required `VK_KHR_maintenance6`

References:
- https://github.com/doitsujin/dxvk/releases/tag/v2.7
- https://github.com/doitsujin/dxvk/wiki/Driver-support\#dxvk-27-and-later

In practice, that baseline can exclude part of older macOS + MoltenVK combinations used by our user base, so `v2.6.x` is currently the compatibility target for this porting effort.

## Validation context

Validated in native macOS workflow for GeneralsX:
- configuration/build/deploy/run loops using SDL3 + DXVK-native
- startup path without requiring users to manually export `DXVK_WSI_DRIVER`

If preferred, I can split this into smaller PRs (WSI selection, DXSO change, and util/header cleanup) to ease review.

### DIFF
--- a/src/d3d10/meson.build
+++ b/src/d3d10/meson.build
@@ -10,7 +10,9 @@ d3d10_core_link_depends = []
 if platform == 'windows'
   d3d10_d3d11_dep = lib_d3d11
 else
+  if platform != 'darwin'
   d3d10_core_ld_args      += [ '-Wl,--version-script', join_paths(meson.current_source_dir(), 'd3d10core.sym') ]
+  endif
   d3d10_core_link_depends += files('d3d10core.sym')
   d3d10_d3d11_dep = d3d11_dep
 endif
--- a/src/d3d11/meson.build
+++ b/src/d3d11/meson.build
@@ -72,7 +72,9 @@ d3d11_link_depends = []
 if platform == 'windows'
   d3d11_dxgi_dep = lib_dxgi
 else
+  if platform != 'darwin'
   d3d11_ld_args      += [ '-Wl,--version-script', join_paths(meson.current_source_dir(), 'd3d11.sym') ]
+  endif
   d3d11_link_depends += files('d3d11.sym')
   d3d11_dxgi_dep = dxgi_dep
 endif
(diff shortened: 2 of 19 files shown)


K0bin:
We're not interested in maintaining support for an incomplete Vulkan implementation just so one particular game mod works well.

PR #5595, opened 2026-04-22
[dxgi] Update UMD version

gofman:


gofman:
This gets rid of GPU warning in Crimson Desert on NVidia. Current hardcoded UMD version translates to public version 155.35 while the game wants 581.29 minimum.

I also bumped the major number, 35 is what we currently have elsewhere in Wine.

doitsujin:
this is getting silly. I guess it'll do for now but I feel like we need a way to fake driver versions going forward?

src/dxgi/dxgi_adapter.cpp
```diff
@@ -134,12 +134,24 @@ namespace dxvk {
...
+    // On Nvidia UMD driver version translates to public version using the lower decimal digits, e. g., 32.0.15.7628
+    // corresponds to 576.28 and some games will check public version extracted from UMD version extracted, e. g., as
+    // (10000 * v1 % 100000 + v2) % 100000, where uint32_t v1 = LowPart >> 32, uint32_t v2 = LowPart && 0xffff
+    // (so LowPart 0xffffffff gives too low version of 155.35). 0x0013270f corresponds to [35.0.]1999.99 which converts
+    // to 999.99 public version.
     if (SUCCEEDED(hr) && pUMDVersion) {
-      pUMDVersion->HighPart = 0x00200000u;
-      pUMDVersion->LowPart  = 0xffffffffu;
+      pUMDVersion->HighPart = 0x00230000u;
+      switch (static_cast<DxvkGpuVendor>(m_desc.VendorId)) {
```
doitsujin:
```suggestion
      switch (DxvkGpuVendor(m_desc.VendorId)) {
```

gofman:
> this is getting silly. I guess it'll do for now but I feel like we need a way to fake driver versions going forward?

But we are already faking? FWIW UMD version exists in Wine, as binary under \\Registry\\Machine\\Software\\Microsoft\\DirectX\\<luid> DriverVersion and in a text a.b.c.d form as a devprop key (accessible in registry in a harder way or through setupapi). Those currently have different values. Maybe it makes sense to read from DirectX key, but also this one in dxvk wasn't probably touched for years without issues and I thought that maybe it doesn't worth the mess if the new value can also be fine for the next few years. But if it is going to need another bump dxvk can probably read it from registry and leave bumps to Wine in one place?

pchome:
Similar thing here:
https://github.com/doitsujin/dxvk/blob/dd4b432111528927738cabcc5642ab36c42adfde/src/d3d9/d3d9_adapter.cpp#L89
Definitely caused some issues on nvidia but it was a while and I don't remember which games.

PR #5597, opened 2026-04-22
[dxvk] Work around depth-stencil copies involving weird formats

### LINKED ISSUE #5596: Regression with depth format reads/writes
Ever since #5531 was merged, I've had failures (granted in D7VK) whenever D24X8 depth was read from or written to. D16 appears fine. Depth read/write support is needed in cursed early D3D because some games do their own depth on the CPU... yes, lovely indeed.

As to the crashes, I get the classic:

`Assertion failed: !status && "vkCreateGraphicsPipelines", file dlls/winevulkan/loader_thunks.c, line 3578`

And just before that the backend is trying to do:

```
debug:                 %meta_vs_copy_buffer_to_image_2d_x8_d24_unorm_pack32 = EntryPoint %main Vertex
[...]
debug:                 %meta_ps_copy_buffer_to_image_2d_x8_d24_unorm_pack32 = EntryPoint %main Pixel
```

Here are the full logs: [dxvk.log](https://github.com/user-attachments/files/26956766/dxvk.log)

Suppressing any D24X8 depth writes/reads magically makes things work again, at least to the point where affected games start just fine instead of crashing, however that will break rendering in some cases, for obvious reasons.

P.S.: This isn't happening in a MSAA context, just with plain old regular depth.

doitsujin:
@WinterSnowfall please verify that this fixes #5596 without regressing anything else.


WinterSnowfall:
@doitsujin That appears to do the trick indeed. The games I had crashing before all work properly now.

PR #5598, opened 2026-04-23
[dxvk] Always re-route mapped allocations to cached memory on tilers

doitsujin:
Sigh.

### DIFF
--- a/src/dxvk/dxvk_device.cpp
+++ b/src/dxvk/dxvk_device.cpp
@@ -731,6 +731,7 @@ namespace dxvk {
 
     applyTristate(tilerMode, m_options.tilerMode);
     hints.preferRenderPassOps = tilerMode;
+    hints.preferCachedMemory = tilerMode;
 
     // Honeykrisp does not have native support for secondary command buffers
     // and would suffer from added CPU overhead, so be less aggressive.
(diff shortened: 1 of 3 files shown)


harry24681012-blip:
Testing on i7-7700K (4C/8T) - Huge improvement for quad-core CPUs

I’ve tested this PR and the results are outstanding for older hardware:

    Latency/Stuttering: The "Red Peaks" (VAR spikes) in TF2 are significantly suppressed. Frame-time consistency is much better once in-game.

    Compilation Flow: This change seems to optimize asynchronous resource handling. I was able to enter the game and play smoothly when shader compilation was only at 12% (it usually takes ~20% to avoid heavy stuttering).

    Observation: There is a brief 100% CPU load for about 10 seconds upon initial launch (likely the re-routing process), but it stabilizes immediately into a much lighter and "snappier" experience.

This is a massive win for efficiency on tile-based rendering paths. Thank you for this!

doitsujin:
Someone please take your AI away, this doesn't affect your system at all.

Blisto91:
To clarify. This is only enabled (by default) on tiler based GPUs which neither, Nvidia, AMD or Intel are.

KostasGangstar-AI:
> To clarify. This is only enabled (by default) on tiler based GPUs which neither, Nvidia, AMD or Intel are.

What about Adreno such as A830 from Snapdragon Elite if you know

Blisto91:
That has tiler mode enabled by default yes.

PR #5607, opened 2026-04-29
[d3d9] Refactor calculating of binding slots.

K0bin:
Next preparation PR for the new shader compiler: refactor how we calculate resource bindings to look similar to how it's done in the D3D11 frontend. That'll allow us to get rid of `dxso_util.h` in the future.

### DIFF
--- a/src/d3d9/d3d9_shader.h
+++ b/src/d3d9/d3d9_shader.h
@@ -51,6 +51,56 @@ namespace dxvk {
 
 #endif
 
+  /**
+   * \brief Shader resource mapping
+   *
+   * Helper class to compute backend resource
+   * indices for D3D9 binding slots.
+   */
+  struct D3D9ShaderResourceMapping {
+    enum ConstantBuffers : uint32_t {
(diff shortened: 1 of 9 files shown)


src/d3d9/d3d9_shader.h
```diff
@@ -51,6 +51,58 @@ namespace dxvk {
...
+      VSCount,
+
+      PSConstantBuffer = 0,
+      PSFixedFunction  = 1,
+      PSShared         = 2,
+      PSCount
+    };
+
+    template<typename T>
+    static constexpr uint32_t computeCbvBinding(T stage, uint32_t index) {
```
doitsujin: Why is stage a template parameter and not just `D3D9ShaderType`?

I guess it's more or less copy-pasted from D3D11, but the reasoning there was that we actually had two different enums while porting to dxbc-spv (and yes, it should be changed there too), but I'm not really seeing that here.

doitsujin (commented):
nit

K0bin:
You're right, I got rid of the templates.

PR #5608, opened 2026-04-30
[d3d9] Changed `VK_FORMAT_R8_SRGB` to `VK_FORMAT_UNDEFINED` for `D3DFMT_L8`

Akabenko:
Changing `VK_FORMAT_R8_SRGB` to `VK_FORMAT_UNDEFINED` will allow to fix `IMAGE_FORMAT_I8` (`D3DFMT_L8`) for Garry's mod. `D3DFMT_L8` was as black.

Result of pull request — format works fine as greyscale.
image

https://github.com/ValveSoftware/Proton/issues/2848#issuecomment-4319967489
Btw `D3DFMT_A8L8` works fine by default on DXVK.

Thats look of `D3DFMT_L8` with `VK_FORMAT_R8_SRGB`. Black render target.
image

### DIFF
--- a/src/d3d9/d3d9_format.cpp
+++ b/src/d3d9/d3d9_format.cpp
@@ -101,7 +101,7 @@ namespace dxvk {
 
       case D3D9Format::L8: return {
         VK_FORMAT_R8_UNORM,
-        VK_FORMAT_R8_SRGB,
+        VK_FORMAT_UNDEFINED,
         VK_IMAGE_ASPECT_COLOR_BIT,
         { VK_COMPONENT_SWIZZLE_R, VK_COMPONENT_SWIZZLE_R,
           VK_COMPONENT_SWIZZLE_R, VK_COMPONENT_SWIZZLE_ONE }};


Akabenko:
RenderDoc. `IMAGE_FORMAT_I8` aka `D3DFMT_L8` with `VK_FORMAT_UNDEFINED`. Showed as `R8_UNORM`.
image

And addition test of default code of DXVK, `D3DFMT_A8L8`. Showed as `R8G8_UNORM`.
image

Akabenko:
`D3DFMT_L8` with `VK_FORMAT_R8_SRGB` (Current enums of DXVK for `L8`).

RGB:
image

Alpha is empty too:
image

doitsujin:
@misyltoad is there any reason we pretend to have sRGB support for L8? That doesn't really make much sense conceptually, given that L8 isn't really meant to hold color data.

Not sure what kind of AI was at work here, but the change *does* actually look useful at a first glance. I guess the reason it's black (rather than just having gamma applied incorrectly) is because your driver probably doesn't really support `R8_SRGB`.

Akabenko:
@doitsujin L8 has always been a problem on Proton. Also It doesn't work for me on Windows with DXVK, and not for my testers on Proton. Both for my GTX 1650 and Intel ARC.

I tried to solve the problem by brute force. At first, I just wanted to get the format working and reassigned it as `VK_FORMAT_B8G8R8A8_UNORM`, so it works. Then I changed the layer assignments (not helps). Then I removed sRGB. And last is probably the best solution.

misyltoad:
Someone should probably check the effect of sRGB stuff on native L8

PR #5611, opened 2026-05-02
[util] Allow wildcard option prefixes in user configs

### LINKED ISSUE #5604: [DI request] Make me `DXVK_CONFIG="*.maxFrameRate=60"` configs possible
This will add a little complexity to rather simple config parser, but will make regular user's life easier.

For example setting `DXVK_CONFIG="*.maxFrameRate = 60"` should populate all known "prefixes" w/ `maxFrameRate = 60`. Expected to be `dxgi.maxFrameRate=60` and `d3d9.maxFrameRate=60` but I guess options have no different meaning (and not existing options just ignored) so `dxvk.maxFrameRate=60`  will do nothing.

If later in config (env or file) there is `d3d9.maxFrameRate=30` it will override prev. setting (and vice-versa).

Similar solutions in a wild: `VK_LOADER_LAYERS_ENABLE=*KHRONOS_validation`.

Maybe I'll do PR myself if there are not some not obvious CONS.

pchome:
This will accept `*.optionName=value` options in user config files and treat them as any of `dxgi.`, `d3d11.`, `d3d9.`, `d3d8.` prefixed options.

Usable for dxgi/d3d9 or d3d11/d3d9 option pairs e.g. `*.maxFrameRate` or `.hideAmdGpu`, etc. Also when game supports (or uses) several different apis (d3d11 and d3d9 are common combinations) and user must periodically switch between them due to game bugs. Or when user have no clue what api game uses, etc. etc.

Tested with DXVK_CONFIG=...
```
info:  Found config env: bad*.bad=false;*bad.badbad=1;*.good=true;bad*bad.baddd=2;bad.bad*bad=3
info:    d3d9.good = true
info:    d3d11.good = true
info:    d3d8.good = true
info:    dxgi.good = true
```

Closes: #5604

### DIFF
--- a/src/util/config/config.cpp
+++ b/src/util/config/config.cpp
@@ -1474,6 +1474,14 @@ namespace dxvk {
 
       ctx.active = key.str() == env::getExeName();
     } else {
+      // Allow '*.OptionName=value' type options
+      // n + 4 here is a minimum valid option size, e.g. '*.a=b'
+      // and '*' not in isValidKeyChar to prevent some *b*u*l*l*s*h*i*t=true
+      if (line.size() >= (n + 4) && line[n] == '*' && line[n+1] == '.') {
+        key << "*.";
+        n += 2;
+      }
+
       while (n < line.size() && isValidKeyChar(line[n]))
         key << line[n++];
 
@@ -1498,7 +1506,13 @@ namespace dxvk {
       }
 
       if (ctx.active)
-        config.setOption(key.str(), value.str());
+        if (key.str().substr(0, 2) == "*.") {
+          // Use actual valid key prefixes in place of '*'
+          // exept "dxvk" because it's core options
+          for (auto pre : {"dxgi", "d3d11", "d3d9", "d3d8"})
+            config.setOption(pre + key.str().substr(1), value.str());
+        } else
+          config.setOption(key.str(), value.str());
     }
   }
 


doitsujin:
I'd much rather introduce a common config that we can interpret in both places rather than do this mess. Main reason we can't replace the separate configs entirely is because some games support both APIs and only need the frame rate limiting for one.

pchome:
Just to be clear: this is not intended to replace the separate configs or to be used in built-in configs/profiles.
This is for user configs only.

But common config options would be great.

src/util/config/config.cpp
```diff
@@ -1498,7 +1506,13 @@ namespace dxvk {
       }
 
       if (ctx.active)
-        config.setOption(key.str(), value.str());
+        if (key.str().substr(0, 2) == "*.") {
+          // Use actual valid key prefixes in place of '*'
+          // exept "dxvk" because it's core options
+          for (auto pre : {"dxgi", "d3d11", "d3d9", "d3d8"})
```
qinlili23333, line 1512: So this will set some non-existant options and increase unnecessary memory usage, as not all options exist in all prefix.

pchome: What memory usage it will increase?
It will not set anything if you not using it. DXVK will not use any non-existant options.

Besides, people using unknown options all the time in their configs.
You could open a random game page on protondb or something and see how dozens of commenters posting a copy/paste from something like
`DXVK_CONFIG="dxgi.pleaseMakeItWork=true;d3d11.thisIncreaseFPS=500"` (not literally).

qinlili23333: Each time you call `setOption`, you actually called `insert_or_assign` to `m_options` which is a `std:: unordered_map`, the more objects you pushed into `m_options` the more memory used. So pushing unused configs will definitely use unnecessary memory.

People do use some options that not working for this game, but I never seen people use options that not exist for DXVK. Your implementation does introduce options not exist like `d3d8.enableHDR` if user typed `*.enableHDR` although d3d8 does not have HDR at all.

I would suggest to check whether the prefix exist for the item before calling `setOption` to avoid unnecessary memory usage.

doitsujin:
Just gonna do the one extra option for frame rate and call it a day. I *really* can't be bothered with overengineered solutions to a non-issue, nor do I want to have to babysit users who inevitably shoot themselves in the foot with an overcomplicated config system.

PR #5618, opened 2026-05-04
[d3d11]: Optimize GetMaxFeatureLevel and shader memory efficiency

dt10812:
Changes :
- Decoupled GetMaxFeatureLevel from class instantiation to reduce overhead
- Implemented zero-copy shader bytecode handling in D3D11ShaderConverter
- Added dynamic GPU virtual address bit calculation based on hardware limits
- Improved multisample logic using independent multisampling probes

Impact :
- RAM: Saves 10MB–80MB of memory in shader-heavy games.
- Performance: Faster device initialization and reduced CPU stutter during shader translation.
- Hardware: Accurate memory reporting across NVIDIA, AMD, and Intel GPUs.

### DIFF
--- a/src/d3d11/d3d11_features.cpp
+++ b/src/d3d11/d3d11_features.cpp
@@ -1,4 +1,6 @@
 #include <array>
+#include <algorithm>
+#include <cmath>
 
 #include "d3d11_features.h"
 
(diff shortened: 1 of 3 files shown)


doitsujin:
This smells like AI garbage and the code wouldn't even compile. The entire concept is completely broken as well because apps can free the provided shader pointer immediately after Create*Shader returns.

PR #5627, opened 2026-05-10
d3d9: Export 0,0,0,0 as default for color1 when using SM3 pixel shaders

SoongVilda:
Implements the requirement to use `0,0,0,0` as the default value for the secondary color target (`color1`) when a Shader Model 3 pixel shader is active. For older shader models and fixed-function, the default remains `0,0,0,1`.

Since the required default value depends on the pixel shader version rather than the vertex shader itself, I've implemented this using a new specialization constant `SpecIsPSShaderModel3`.

Changes:
- Added `SpecIsPSShaderModel3` specialization constant to `d3d9_spec_constants.h`.
- Updated `D3D9DeviceEx` to track the PS version and update the spec constant during state changes.
- Modified `DxsoCompiler` to use `opSelect` for the `color1` default value based on the specialization constant.
- Updated SM3 VS output setup to always export a default for `color1` if missing from the signature, matching existing behavior for `color0`.

This improves compatibility for games that rely on these specific default values when using SM3.

---
*PR created automatically by Jules for task [6883176983437566686](https://jules.google.com/task/6883176983437566686) started by @SoongVilda*

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -7734,6 +7734,8 @@ namespace dxvk {
 
       UpdatePixelShaderSamplerSpec(textureTypes, fetch4);
 
+      UpdatePixelShaderModelSpec(programInfo.majorVersion() >= 3);
+
       UpdatePixelBoolSpec(
         m_state.psConsts->bConsts[0] &
         m_consts[uint32_t(D3D9ShaderType::PixelShader)].meta.boolConstantMask);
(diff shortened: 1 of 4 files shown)


K0bin:
We're currently in the process of completely rewriting the shader compiler anyway, so we're not gonna merge this. Especially as it's AI slop.

> This improves compatibility for games that rely on these specific default values when using SM3.

I'm not aware of a single game that actually relies on this.

PR #5641, opened 2026-05-20
DXBC-Spirv based shader compiler for D3D9 (and 8) frontend

K0bin:


src/dxvk/dxvk_shader_ir.cpp
```diff
@@ -1559,6 +1559,8 @@ namespace dxvk {
           return spv::BuiltInPointSize;
         case dxbc_spv::ir::BuiltIn::eTessFactorLimit:
           return std::nullopt;
+        case dxbc_spv::ir::BuiltIn::ePointCoord:
+          return spv::BuiltInPointCoord;
```
doitsujin, line 1563: also needs to be handled in `DxvkIrShader::convertBuiltIn` (i.e. the function that does the reverse lookup), and apparently we're missing PointSize there too so might as well fix that while we're at it (although that's less critical).

Unfortunately this mess needs to stay because we'll still have spirv shaders.

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -0,0 +1,297 @@
...
+#include <utility>
+
+#include "../util/log/log.h"
+
+namespace dxvk {
+
+  D3D9ShaderAnalysis::D3D9ShaderAnalysis(dxbc_spv::util::ByteReader code, bool isSWVP)
+    : m_isSWVP(isSWVP) {
+    if (isSWVP)
+      m_constants.boolMask = ~0u;
```
doitsujin: what exactly does swvp have to do with this?

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+    if (!code) {
+      Logger::err("No code chunk found in shader.");
```
doitsujin, line 35: return? no point in trying to parse the damn thing if there is none

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -0,0 +1,297 @@
...
+    for (uint32_t i = 0u; i < op.getSrcCount(); i++) {
+      const auto& src = op.getSrc(i);
+      auto registerType = src.getRegisterType();
+      uint32_t index = src.getIndex();
+
+      /* The indices start counting at 1 to differentiate between a shader
+       * that doesn't access any constants at all and one that accesses the
+       * first one. */
+
+      if (registerType == RegisterType::eConstInt) {
```
doitsujin: this whole thing would be 10x more readable if it was just a switch-case block, i was scratching my head over this more than i probably should have.

also weird comment above, we store counts rather than indices so naturally count == 0 means nothing is used.

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -0,0 +1,297 @@
...
+      m_constants.floatCount = std::max(
+        m_constants.floatCount,
+        m_isSWVP ? MaxFloatConstantsSoftware : hwvpFloatConstantsCount
+      );
+      m_constants.floatsAccessedDynamically = true;
+    }
+
+    if (GetShaderInfo().getType() == ShaderType::ePixel
+      && op.hasDst()
+      && op.getDst().getRegisterType() == RegisterType::eColorOut) {
```
doitsujin, line 134: does this work for sm1.1 where stuff is just written to r0? although i'd guess we'll just want to set usedRTs to 0x1 for that anyway.

src/d3d9/d3d9_shader_analysis.h
```diff
@@ -0,0 +1,143 @@
...
+
+#include <sm3/sm3_types.h>
+#include <sm3/sm3_parser.h>
+
+#include <vulkan/vulkan.h>
+
+#include <vector>
+
+using namespace dxbc_spv::sm3;
+using dxbc_spv::util::small_vector;
```
doitsujin: Why? dxvk has its own small_vector (more or less the same implementation too, but *not quite* identical).

src/d3d9/d3d9_shader_analysis.h
```diff
@@ -0,0 +1,143 @@
...
+#include <util/util_small_vector.h>
+
+#include <sm3/sm3_types.h>
+#include <sm3/sm3_parser.h>
+
+#include <vulkan/vulkan.h>
+
+#include <vector>
+
+using namespace dxbc_spv::sm3;
```
doitsujin: not a fan, it's okay in the .cpp since it'll be limited to one file, but importing dxbc-spv namespaces into a bunch of unrelated d3d9 code can get very annoying.

src/d3d9/d3d9_shader_analysis.h
```diff
@@ -0,0 +1,143 @@
...
+
+  uint32_t GetInputSignatureSize() const {
+    return m_inputSignature.size();
+  }
+
+  Semantic GetInputSignatureSemantic(uint32_t index) const {
+    return m_inputSignature[index];
+  }
+
+  const D3D9InputSignature& GetInputSignature() const { return m_inputSignature; }
```
doitsujin, line 94: since we kinda need this, why do we have size/get methods in here as well?

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -0,0 +1,297 @@
...
+    if (registerType == RegisterType::eSampler) {
+      switch (dcl.getTextureType()) {
+        case TextureType::eTexture3D:
+          m_imageViewTypes[index] = VK_IMAGE_VIEW_TYPE_3D;
+          break;
+        case TextureType::eTextureCube:
+          m_imageViewTypes[index] = VK_IMAGE_VIEW_TYPE_CUBE;
+          break;
+        case TextureType::eTexture2D:
+        default:
```
doitsujin, line 267: is this `default` case something we ever expect to hit?

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -0,0 +1,297 @@
...
+    if (registerType == RegisterType::eInput && GetShaderInfo().getType() == ShaderType::eVertex) {
+      m_inputSignature.push_back(Semantic { dcl.getSemanticUsage(), dcl.getSemanticIndex() });
+      return true;
+    }
+
+    return true;
+  }
+
+
+  D3D9ShaderAnalysis& D3D9ShaderAnalysis::operator=(const D3D9ShaderAnalysis& other) {
```
doitsujin, line 315: do we need to be able to copy this at all? smells like something where move should suffice

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -255,6 +256,28 @@ namespace dxvk {
...
+      auto ffInputs = GetFixedFunctionIsgn();
+      for (int32_t i = 0; i < int32_t(ffInputs.size()); i++) {
+        const auto& ffInputSemantic = ffInputs[i];
+        if (ffInputSemantic == semantic) {
+          location = i;
+          break;
+        }
+      }
+
+      if (location == -1) {
```
doitsujin: this should be an assert

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -255,6 +256,28 @@ namespace dxvk {
...
     }
 
+    if (GetShaderInfo().getType() == ShaderType::ePixel
+      && dcl.getSemanticUsage() == SemanticUsage::eColor
+      && dcl.getSemanticIndex() < 2u) {
+      Semantic semantic = { dcl.getSemanticUsage(), dcl.getSemanticIndex() };
+
+      int32_t location = -1;
+      auto ffInputs = GetFixedFunctionIsgn();
+      for (int32_t i = 0; i < int32_t(ffInputs.size()); i++) {
```
doitsujin: this could do with a comment or a helper function with a descriptive name, I understand what it does and why it needs to be here but it's not the most intuitive thing ever

src/d3d9/d3d9_shader.cpp
```diff
@@ -4,8 +4,34 @@
...
+#include <sm3/sm3_types.h>
+
 namespace dxvk {
 
+  class D3D9SpecializationConstantLayout : public dxbc_spv::sm3::SpecializationConstantLayout {
+
+  public:
+
+    dxbc_spv::sm3::SpecializationConstantBits getSpecConstantLayout(dxbc_spv::sm3::SpecConstantId id) const {
+      D3D9SpecConstantId specConstId = static_cast<D3D9SpecConstantId>(static_cast<uint32_t>(id));
```
doitsujin: the fuck is this static_cast soup

src/d3d9/d3d9_shader.cpp
```diff
@@ -66,6 +179,25 @@ namespace dxvk {
   }
 
 
+  void D3D9CommonShader::CreateIrShader(
+          D3D9DeviceEx*           pDevice,
+    const DxvkShaderHash&         ShaderKey,
+      const D3D9ShaderCreateInfo& ModuleInfo,
```
doitsujin: weird indent

src/d3d9/d3d9_shader.cpp
```diff
@@ -32,6 +32,116 @@ namespace dxvk {
...
+
+    void convertShader(
+            dxbc_spv::ir::Builder&    builder) {
+      auto debugName = m_key.toString();
+
+      dxbc_spv::sm3::Converter::Options options = { };
+      options.name = debugName.c_str();
+      options.includeDebugNames = true;
+      options.fastFloatEmulation = m_options.d3d9FloatEmulation == D3D9FloatEmulation::Enabled;
+      options.isSWVP = m_options.isSWVP;
```
doitsujin: hnnnngh, this is problematic because the backend is not aware of these options (float stuff and swvp), so this will fall apart with the shader cache.

I think what needs to happen here is that
a) `D3D9ShaderOptions` needs to have a well-defined memory layout (i.e. change the way `D3D9FloatEmulation` is defined to be explicitly uint8 since right now it doesn't have an explicit type for some reason, unlike any other enum) and throw a static_assert at it to make sure it has no padding,
b) change the hasher in CreateShaderModule to dxbc-spirv's MD5 impl since that can actually hash more than one blob
c) feed **both** the code and the `D3D9ShaderOptions` struct into the hash so that the shader key changes together with the options. Not super clean given that one of these things is a game / driver-specific config option, but I don't have any great ideas otherwise and I don't want this mess in the backend as-is (we can talk about moving the float emu stuff to a separate pass, but that's a nontrivial amount of work). For SWVP we actually want different hashes anyway.

Think in the long term we should aim to get rid of the sha1 code altogether.

doitsujin (commented):
:frog:

src/d3d9/d3d9_shader_analysis.cpp
K0bin: I don't really remember, why I did that.
The bool mask is not used when SWVP is enabled because SWVP raises the constant limits far above 32.

src/d3d9/d3d9_shader_analysis.cpp
K0bin: The comment is a leftover from when those variables were still called `index`.

src/d3d9/d3d9_shader_analysis.cpp
K0bin: I don't think so but the old compiler has had this exact default case for years so I'm not 100% sure.

src/d3d9/d3d9_shader_analysis.cpp
K0bin: `D3D9CommonShader` holds an instance of `D3D9ShaderAnalysis` and `D3D9CommonShader` needs to get copied because `D3D9ShaderModuleSet` holds a copy of each shader.

src/d3d9/d3d9_shader.cpp
doitsujin: SWVP also needs to check that this is actually a vertex shader. This breaks Anno 1701 currently.

src/d3d9/d3d9_device.cpp
```diff
@@ -8103,11 +8103,20 @@ namespace dxvk {
...
+
     m_dxvkShaderOptions = m_dxvkDevice->getShaderCompileOptions();
+    m_dxvkShaderOptions.flags.set(DxvkShaderCompileFlag::SemanticIo);
 
     if (m_d3d9Options.forceSampleRateShading)
       m_dxvkShaderOptions.flags.set(DxvkShaderCompileFlag::EnableSampleRateShading);
 
+    if (m_d3d9Options.useFP16) {
+      Logger::info("Using FP16.");
+      m_dxvkShaderOptions.flags.set(DxvkShaderCompileFlag::Supports16BitArithmetic);
```
doitsujin: this check needs to be the inverted, if FP16 is disabled we need to opt out of FP16.   Otherwise, the only thing this does is enable FP16 on devices that don't support that at all.

src/d3d9/d3d9_shader_analysis.cpp
K0bin: DXVK doesn't really use asserts, does it? (Aside from `static_assert`)
So just a regular `assert()`?

src/d3d9/d3d9_shader_analysis.cpp
K0bin: Now that I'm looking at it again, I'm realizing that what I'm doing there makes no sense anyway. `GetFixedFunctionIsgn` is the fixed function vertex **input**.

src/d3d9/d3d9_shader_analysis.cpp
doitsujin: well the code that interacts with dxbc_spirv is already full of dxbc_spirv_assert()s anyway, might as well do that here.

src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -0,0 +1,297 @@
...
+      auto registerType = src.getRegisterType();
+      uint32_t index = src.getIndex();
+
+      switch (registerType) {
+        case RegisterType::eConstInt:
+          m_constants.intCount = std::max(m_constants.intCount, index + 1u);
+          break;
+
+        case RegisterType::eConstBool: {
+          m_constants.intCount = std::max(m_constants.intCount, index + 1u);
```
doitsujin: should this be `boolCount`?

Issue #5133, opened 2025-08-07
Bring back d3d9.forceSwapchainMSAA !!! Removal was huge mistake!


v00d00m4n:
Seriosly, whats up with removal of features nobody asked to removed? Why do you decided for everyone whats useful, whats not?

Big amount of old D3D9 and wrapperd D3D8toD3D9 games NEVER HAD ANY AA OPTIONS BUILT IN GAME, some that has such options had them often limited to fixed amount or technique of AA and in most of cases it was better idea to force MSAA or SSAA via Nvdia or AMD controls panel to most of games.

This setting was effectively imitation of driver forced MSAA

Check your "facts" about MSAA popularity and availability in old games!

MSAA started to become popular in games somewhere around 2002-2003, most of games prior this era did not have any form of AA, and even in that era many games still did not have! It become popular after Far Cry, DOOM 3 and Half Life 2 setting it as new standart, but only for about half of popular releases, other half still missed it.

Also somewhere around 2004-2005 year a lot of games still using D3D9 stoppped to support MSAA due of HDR and some incompatibility between two.

Soon after xbox360 with next gen graphics arrived and we entered quite a dark and jaggy era for almost half of decade, where many games ported from Xbox 360 or PS3 had no AA at all due of defered shading, or had at best some custom edge blur filter, or even if MSAA was present it was too slow and barely used, until in around 2011 FXAA and SMAA arrived that for most of D3D9 games replaced any other AA for a while.

However around same time Nvidia somehow managed to hack in MSAA or some other forms of AA like SGSSAA into games by forcing it via drivers for games that did not have AA support 

Until the point when some short renessaince of MSAA started to happen around same time some games started to provide TAA or nvidias TXAA somewhere around 2010-2013, but then again at least half of games did not have any MSAA or any real AA besides FXAA at all while still was using D3D9!

Basically what you just did by removing legacy option (jeez i hate this trend - if something aint broken dont fixed, if it works just leave it alone for backwards compatibility!) is stripped of any form of good AA hundreds if not thousands of games!

Just to name a few games that now cant be played with MSAA because they never had such option:

[...]

PR #5646, opened 2026-05-25
[dxvk] [ai] [poc] Selectively apply VRS coarse rate shading and remove sample rate shading for additive blending

TRPB:
This is a proof of concept by calude opus 4.7. It was built to fix a specific performance issue I have and I'm not even sure this belongs in DXVK upstream. Sharing here as it might be useful to someone for reference, I am not intending this PR to be merged as-is and only wanted to fix an annoying performance drop I had in some games. The fix works well so may be useful for others.

### Initial problem

In D3D9 titles when `forceSampleRateShading` is enabled, glow/spell/particle effects often introduce a noticable FPS drop, especially when they take up a large portion of the screen.

`forceSampleRateShading` looks fantastic and modern GPUs can handle it easily on old D3D9 games... except in this scenario with lots of transparency and blending on the screen. Which don't benefit from SRS anyway.

It does the following:

1. Disables per-sample shading for blend enabled pipelines
2. Enables 2x2 VRS coarse rate shading for additive/multiplicative blending 

### Result

DXVK 2.7.1 - 75fps
With this patch (against git master) - 196fps (hitting in-game 200fps cap)

My questions are:

1. If this was tidied up would it even be welcome in DXVK? I know you [removed the option to force MSAA](https://github.com/doitsujin/dxvk/issues/5133) citing that DXVK is not focussed on adding new features and this feels similar
2. If it was to be accepted, should it be behind a config option? I assume so, but again, all the arguments against the `forceSwapChainMsaa` options apply here

### DIFF
--- a/src/dxvk/dxvk_device_info.cpp
+++ b/src/dxvk/dxvk_device_info.cpp
@@ -27,6 +27,7 @@ namespace dxvk {
     HANDLE_EXT(extExtendedDynamicState3);          \
     HANDLE_EXT(extFragmentShaderInterlock);        \
     HANDLE_EXT(extFullScreenExclusive);            \
+    HANDLE_EXT(khrFragmentShadingRate);            \
     HANDLE_EXT(extGraphicsPipelineLibrary);        \
     HANDLE_EXT(extHdrMetadata);                    \
     HANDLE_EXT(extLineRasterization);              \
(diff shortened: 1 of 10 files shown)


doitsujin:
I'm categorically uninterested in any sort of AI drivel, but this also just sounds precisely like the kind of feature creep that we need to avoid. Having blending enabled all the time isn't exactly uncommon for older games, it's just going to fall apart there and cause issues, which means people will want more heuristics etc etc etc. No.

Also no good reason to force VRS really.

PR #5652, opened 2026-05-27
Enable dxbc-spirv for D3D9 by default

doitsujin:
Given that everything we know of compiles cleanly and produces valid spirv, this is kinda tempting now.


K0bin (approved):
Fine by me :)

PR #5653, opened 2026-05-28
Nuke legacy fixed-function shader generator

doitsujin:
Think we're confident enough in the übershader these days.


K0bin (approved):
The best kind of PRs :)

PR #5655, opened 2026-05-28
[d3d9] Do not enable FP16 by default

doitsujin:
Games are dumb and require partial precision to actually be FP32 in too many places. We can opt in via app profiles for known-good games.


K0bin (approved):
Unfortunate but not much else we can do.

PR #5657, opened 2026-05-29
Simplify shader creation

doitsujin:
The only reason the double-lock was ever a good idea is because we did the SPIR-V translation on the app thread with the old compilers, but with dxbc-spirv that is no longer the case. Accessing the shader cache serializes anyway and there's nothing else happening here most of the time.


K0bin (approved):
Nice cleanup.

PR #5659, opened 2026-05-29
Nuke dxso

doitsujin:
Draft because this depends on the SWVP rewrite, and we do want to wait for some feedback to arrive in general.

I don't know why MSVC is suddenly throwing linker errors on unrelated code that is **already there**. Would be nice to figure that out, but if we can't then I'd rather just drop MSVC from CI until somebody does.


misyltoad:
rip

pchome:
> I don't know why MSVC is suddenly throwing linker errors on unrelated code that is already there.

Lazy guess: link order?
I remember we had linker errors a shitton years ago, so dependencies was tailored.
Maybe should be `dependencies: [ dxvk_dep, dxbc_spirv_dep ],`

doitsujin:
I already tried changing the order around, hasn't changed anything.

The part that breaks is calling from D3D9 into the DXVK backend, i.e. *the one thing* that this PR doesn't even change, and it complains about code that is *already there* and hasn't really changed besides removing an `if`. It's weird.

Nukem9:
https://github.com/doitsujin/dxvk/blob/be4ac08faa1ba7546193995ac6e45ca8c07763d4/src/dxvk/dxvk_shader_ir.h#L25-L26
https://github.com/doitsujin/dxvk/blob/be4ac08faa1ba7546193995ac6e45ca8c07763d4/src/dxvk/dxvk_device.h#L35-L36

There's a mismatch with `struct` and `class`. Not sure why it showed up now. *Probably* weird header inclusion order?

doitsujin:
Oh, huh, weird that GCC doesn't warn about that.

I'll see if this fixes MSVC later, kinda busy with something else right now.

doitsujin:
That was in fact it, thanks.

doitsujin:
Tempted to do this sooner rather than later actually because dear god is the old code blocking any kind of rework.

PR #5678, opened 2026-06-03
[d3d11] Fail device creation for some cases unsupported on Windows

gofman:


gofman:
That helps Holo Hockey which otherwise crashes on start. Unity engine tries to create d3d11 device for video processing, previously failing to find dxgi adapter by luid for some reason and using NULL one  (that also happens on Windows up to here). It will use D3D_DRIVER_TYPE_WARP and with D3D11_CREATE_DEVICE_VIDEO_SUPPORT that fails on Windows letting it to fallback to software video processing. With Proton it will succeed device creation but only after that Unity will check for no dxgi adapter previously found and take a broken path dereferencing NULL pointer.

I am attaching an ad-hoc addition to Wine test which my changes are based on.

[test.patch](https://github.com/user-attachments/files/28569249/test.patch)

gofman:
Looks like some cases with D3D_DRIVER_TYPE_UNKNOWN here depend on dxgi adapter specified, I will update the patch.

gofman:
Nvm, I added more tests (attached) and the patch still passes those.
[test.patch](https://github.com/user-attachments/files/28570953/test.patch)

PR #5686, opened 2026-06-05
Rework D3D9 shader constant management

doitsujin:
Besides getting rid of some legacy cruft, this compacts statically indexed shader constants in such a way that shaders only accessing `c180` through `c183` will now only allocate and copy 64 bytes of buffer memory instead of 2944 bytes. Hello Witcher 2.

Additionally, this leverages the new `PUSH_ADDRESS` support in the backend for statically indexed constant buffers, which means no more uniform buffer descriptors in most cases.

We also get proper debug names in shaders now if the shader defines `ctab`:
```glsl
layout(set = 1, binding = 6, std140) uniform c_buf {
    vec4 scale_matrix0;
    vec4 scale_matrix1;
    vec4 scene_fog_start;
    vec4 scene_fog_enabled;
    vec4 matrix_worldviewproj0;
    vec4 matrix_worldviewproj1;
    vec4 matrix_worldviewproj2;
    vec4 matrix_worldviewproj3;
    vec4 matrix_projinv2;
    vec4 c20;
} c;
```

Dynamically indexed VS floats get their own dedicated buffer now with more or less the same strategy as before, we only actually bind and upload constants that have a non-zero value. Looks like this in shader:

```glsl
layout(set = 1, binding = 6, std140) uniform light_count_buf {
    ivec4 m;
} light_count;

layout(set = 1, binding = 7, std140) uniform cF_buf {
    vec4 m[256];
} cF;
```

Dirty-tracking is *also* changed in such a way that `SetShaderConstant*` will actually check whether any of the constants have actually changed, this does save a fair bit of work in some cases especially when it affects shaders with dynamic floats. If anything, this could probably be made more granular to actually track dynamic and static buffers separately.

As for performance, somewhat disappointingly I'm not seeing much of an improvement in the handful of games that I tested, even though micro-benchmarking the time spent in `SetShaderConstants` and `UpdateShaderConstants` shows clear improvements.

That said, one *real* benefit (besides wasting slightly less memory) is that the descriptor copy worker does next to nothing now in many 32-bit D3D9 games because we no longer use actual descriptors for UBOs, at least on the descriptor heap path and on AMD. Another one is that RE6 no longer runs out of descriptor heap space.

Needs a lot of testing because this do be a lil scawy.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -629,14 +629,15 @@
 
 # Device Local Constant Buffers
 #
-# Enables using device local, host accessible memory for constant buffers in D3D9.
-# This tends to actually be slower for some reason on AMD,
-# and the exact same performance on NVIDIA.
+# Allows streaming shader constants and render state buffers into
+# VRAM, which may improve performance in GPU-bound scenarios.
 #
 # Supported values:
-# - True/False
+# - True: Always enable constant buffer streaming
+# - Auto: Only enable streaming on discrete GPUs
+# - False: Always disable streaming
 
-# d3d9.deviceLocalConstantBuffers = False
+# d3d9.deviceLocalConstantBuffers = Auto
 
 
 # Support cube texture depth formats
--- a/src/d3d11/d3d11_initializer.cpp
+++ b/src/d3d11/d3d11_initializer.cpp
@@ -121,9 +121,10 @@ namespace dxvk {
         cBuffer       = buffer,
         cStagingSlice = std::move(stagingSlice)
       ] (DxvkContext* ctx) {
-        ctx->uploadBuffer(cBuffer,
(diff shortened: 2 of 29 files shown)


Digger1955:
@doitsujin 

Does [commit 2d1202f - [d3d9] Remove non-GPL fallback code](https://github.com/doitsujin/dxvk/pull/5686/commits/2d1202f481f57c3996a20e820e087fd31ccf9ab7) mean that GPL will become mandatory for DXVK 3.x to work?

doitsujin:
No, it just means that we just unconditionally copy some data around now instead of gating it behind a GPL check. Not really harmful.

misyltoad:
May be worth checking perf before/after in TF2 or HL2

PR #5691, opened 2026-06-08
[d3d9] Consolidate SWVP HUD

doitsujin:
Don't really need two classes to display related stuff under the same `DXVK_HUD` item.

### DIFF
--- a/src/d3d9/d3d9_hud.cpp
+++ b/src/d3d9/d3d9_hud.cpp
@@ -47,38 +47,16 @@ namespace dxvk::hud {
     return position;
   }
 
-  HudSWVPShaders::HudSWVPShaders(D3D9DeviceEx* device)
-  : m_device        (device)
-  , m_swvpShaderCount ("") {}
-
-
-  void HudSWVPShaders::update(dxvk::high_resolution_clock::time_point time) {
-    m_swvpShaderCount = str::format(m_device->GetSWVPShaderCount());
-  }
-
-
-  HudPos HudSWVPShaders::render(
-    const Rc<DxvkCommandList>&ctx,
-    const HudPipelineKey&     key,
-    const HudOptions&         options,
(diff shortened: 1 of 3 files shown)


src/d3d9/d3d9_hud.cpp
```diff
@@ -47,38 +47,43 @@ namespace dxvk::hud {
...
     const HudOptions&         options,
           HudRenderer&        renderer,
           HudPos              position) {
     position.y += 16;
-    renderer.drawText(16, position, 0xffc0ff00u, "SWVP Shaders:");
-    renderer.drawText(16, { position.x + 180, position.y }, 0xffffffffu, m_swvpShaderCount);
+    renderer.drawText(16, position, 0xffc0ff00u, "Vertex Processing:");
+    renderer.drawText(16, { position.x + 240, position.y }, 0xffffffffu, m_isSWVPText);
+
+    if (m_device->CanSWVP()) {
```
WinterSnowfall: Reminder ProcessVertices can (and has been known to be) called from a HWVP device, so this bit should always be shown really.

src/d3d9/d3d9_hud.cpp
```diff
@@ -47,38 +47,43 @@ namespace dxvk::hud {
...
+  HudSWVPState::HudSWVPState(D3D9DeviceEx* device)
+  : m_device(device)  {
+    updateSwvpState();
+  }
 
-  void HudSWVPShaders::update(dxvk::high_resolution_clock::time_point time) {
-    m_swvpShaderCount = str::format(m_device->GetSWVPShaderCount());
+
+  void HudSWVPState::update(dxvk::high_resolution_clock::time_point time) {
+    if (m_device->CanSWVP())
```
WinterSnowfall: Ditto.

PR #5694, opened 2026-06-09
Misc stuff

doitsujin:
Adds a `dxvk.maxFrameRate` option that takes precedence over the two API-specific variants (and thus, app profiles). Supersedes #5611.

Also shows the current size of the shader cache on startup.

### DIFF
--- a/dxvk.conf
+++ b/dxvk.conf
@@ -70,6 +70,7 @@
 
 # dxgi.maxFrameLatency = 0
 # d3d9.maxFrameLatency = 0
+# dxvk.maxFrameLatency = 0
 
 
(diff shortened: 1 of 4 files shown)


dxvk.conf
```diff
@@ -70,6 +70,7 @@
 
 # dxgi.maxFrameLatency = 0
 # d3d9.maxFrameLatency = 0
+# dxvk.maxFrameLatency = 0
```
pchome, line 73: Wrong option? `dxvk.maxFrameLatency` !

doitsujin: yeah whatever.

PR #5695, opened 2026-06-09
[d3d9] Ignore fixed function lights with invalid types

WinterSnowfall:
Apparently these exist in the wild (NFS Most Wanted, allegedly), and current fixed function code more or less treated them as point lights.

WineD3D outright rejects invalid/unknown types for all API levels, however I checked on Windows XP and D3D8/9 let set/get/enable work just fine with *any* type (unlike D3D7 and earlier), but most likely they don't affect lighting in any meaningful way. Also saved some CPU cycles to make up for adding some :frog:.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -8123,7 +8123,7 @@ namespace dxvk {
   }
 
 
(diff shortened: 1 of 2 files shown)


doitsujin:
Gonna hold off on this for now because @K0bin is currently working on stuff in this area, don't want to cause pain there.

Edit: nvm.

PR #5697, opened 2026-06-10
Clean up FF PS

K0bin:
- Remove the FF PS Data CBV and move the `textureFactor` into push constants
- `textureFactor` is the only remaining thing we need from that CBV
- Remove the `FFPixelData` dirty flag, as we no longer need it
- Remove the special fallback and just rely on the spec const fallback
- Remove the 'key' and set the spec constants directly

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -148,7 +148,6 @@ namespace dxvk {
     m_dirty.set(D3D9DeviceDirtyFlag::FFVertexShader);
     m_dirty.set(D3D9DeviceDirtyFlag::FFPixelShader);
     m_dirty.set(D3D9DeviceDirtyFlag::FFViewport);
-    m_dirty.set(D3D9DeviceDirtyFlag::FFPixelData);
     m_dirty.set(D3D9DeviceDirtyFlag::FFGlobalSpecular);
     m_dirty.set(D3D9DeviceDirtyFlag::SharedPixelShaderData);
     m_dirty.set(D3D9DeviceDirtyFlag::DepthBounds);
@@ -2452,7 +2451,7 @@ namespace dxvk {
           break;
 
         case D3DRS_TEXTUREFACTOR:
-          m_dirty.set(D3D9DeviceDirtyFlag::FFPixelData);
+          UpdatePushConstant<D3D9RenderStateItem::TextureFactor>();
           break;
 
         case D3DRS_DIFFUSEMATERIALSOURCE:
@@ -6183,6 +6182,10 @@ namespace dxvk {
 
       UpdatePushConstant<offsetof(D3D9RenderStateInfo, pointScaleC), sizeof(float)>(&scale);
     }
+    else if constexpr (Item == D3D9RenderStateItem::TextureFactor) {
+      uint32_t textureFactor = bit::cast<uint32_t>(rs[D3DRS_TEXTUREFACTOR]);
+      UpdatePushConstant<offsetof(D3D9RenderStateInfo, textureFactor), sizeof(uint32_t)>(&textureFactor);
+    }
     else
       Logger::warn("D3D9: Invalid push constant set to update.");
   }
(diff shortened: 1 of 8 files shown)


src/d3d9/shaders/d3d9_fixed_function_common.glsl
```diff
@@ -364,3 +365,12 @@ bool specBool(uint specConstIdx, uint bitOffset) {
...
     return specUint(specConstIdx) != 0u;
 }
+
+vec4 decodeD3DColor(uint color) {
+    vec4 decoded;
+    decoded.a = float((color & 0xff000000) >> 24) / 255.0f;
+    decoded.r = float((color & 0x00ff0000) >> 16) / 255.0f;
+    decoded.g = float((color & 0x0000ff00) >> 8)  / 255.0f;
+    decoded.b = float((color & 0x000000ff))       / 255.0f;
+    return decoded;
```
doitsujin: this entire function could just be `unpackUnorm4x8(color).bgra`

src/d3d9/d3d9_device.cpp
```diff
@@ -8311,129 +8311,116 @@ namespace dxvk {
...
+          uint32_t alphaArg1 = data[DXVK_TSS_ALPHAARG1];
+          uint32_t alphaArg2 = data[DXVK_TSS_ALPHAARG2];
+          uint32_t alphaArg0 = data[DXVK_TSS_ALPHAARG0];
+
+          if (i == 0
+            && resultIsTemp
+            && colorOp != D3DTOP_DISABLE
+            && alphaOp == D3DTOP_DISABLE) {
+            alphaOp   = D3DTOP_SELECTARG1;
+            alphaArg1 = D3DTA_DIFFUSE;
```
doitsujin, line 8390: what does this do exactly, why, and couldn't we just handle this in shader?

src/d3d9/d3d9_device.cpp
```diff
@@ -8311,129 +8311,116 @@ namespace dxvk {
...
+        if (m_state.textures[i] == nullptr) {
+          // Strip modifiers from arguments
+          uint32_t pureColorArg1 = data[DXVK_TSS_COLORARG1] & D3DTA_SELECTMASK;
+          uint32_t pureColorArg2 = data[DXVK_TSS_COLORARG2] & D3DTA_SELECTMASK;
+          uint32_t pureColorArg0 = data[DXVK_TSS_COLORARG0] & D3DTA_SELECTMASK;
+
+          if ((pureColorArg0 == D3DTA_TEXTURE && usesArg(data[DXVK_TSS_COLOROP], 0u))
+           || (pureColorArg1 == D3DTA_TEXTURE && usesArg(data[DXVK_TSS_COLOROP], 1u))
+           || (pureColorArg2 == D3DTA_TEXTURE && usesArg(data[DXVK_TSS_COLOROP], 2u)))
+            stageActive = false;
```
doitsujin, line 8370: I know this is old code, but this whole check doesn't really seem complete (see the `ugh` commit) since some ops implicitly sample texture and alpha op is also *completely* ignored here.

Not important for this PR in particular, but we should probably verify native behaviour at some point, check if there is a wine test, and if anything behaves unexpectedly, actually write a comment documenting it.

src/d3d9/d3d9_device.cpp
K0bin: I thought about adding a comment but I have honestly no idea what it does. It's a straightforward port of the old code. :/

src/d3d9/d3d9_device.cpp
K0bin: Looks like you're right but can we handle this later, outside of a cleanup PR?

src/d3d9/shaders/d3d9_fixed_function_common.glsl
K0bin: Done

src/d3d9/d3d9_device.cpp
doitsujin: yes

PR #5700, opened 2026-06-11
[d3d9] Slightly clean up FF VS data

K0bin:
~~Much less significant than the PS one.~~ That didn't really last...

Rename it and remove the extra struct as it's no longer used as a lookup key.

Besides that, it removes the super tight packing for what used to be the key of FF vertex shaders as apparently that was UB.
Unfortunately we can't get rid of it on the shader side because single byte types arent supported and replacing all of it with 4 byte unsigned ints would be a bit excessive size-wise.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -8140,63 +8140,6 @@ namespace dxvk {
   }
 
 
-  D3D9PackedFFVSData D3D9DeviceEx::BuildPackedFFVSData(D3D9FF_VertexBlendMode vertexBlendMode, bool indexedVertexBlend, uint32_t lightCount) const {
-    D3D9PackedFFVSData data = {};
-
-    data.VertexHasPositionT = m_state.vertexDecl != nullptr && m_state.vertexDecl->TestFlag(D3D9VertexDeclFlag::HasPositionT);
-    data.VertexHasColor0    = m_state.vertexDecl != nullptr && m_state.vertexDecl->TestFlag(D3D9VertexDeclFlag::HasColor0);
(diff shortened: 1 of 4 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -8291,7 +8235,58 @@ namespace dxvk {
...
+      data->LocalViewer      = m_state.renderStates[D3DRS_LOCALVIEWER] && lighting;
+
+      data->RangeFog         = m_state.renderStates[D3DRS_RANGEFOGENABLE];
+
+      data->DiffuseSource    = m_state.renderStates[D3DRS_DIFFUSEMATERIALSOURCE]  & mask;
+      data->AmbientSource    = m_state.renderStates[D3DRS_AMBIENTMATERIALSOURCE]  & mask;
+      data->SpecularSource   = m_state.renderStates[D3DRS_SPECULARMATERIALSOURCE] & mask;
+      data->EmissiveSource   = m_state.renderStates[D3DRS_EMISSIVEMATERIALSOURCE] & mask;
+
+      uint32_t lightCount = lightIdx;
```
WinterSnowfall, line 8264: Might as well just derp lightIdx directly in data->LightCount when needed, if you've moved the former key stuff in here.

K0bin: No, we need to read lightCount somewhere and that will go terribly wrong *if* the buffer gets allocated in uncached memory in the future. Besides that, I think the extra variable called `lightCount` makes this a lot more understandable than `lightIdx`.

WinterSnowfall: I was only referring to the extra variable (and the row below), but if you want to keep it for extra clarity, then sure.

PR #5704, opened 2026-06-14
Optimize D3D9 vertex buffer binding

doitsujin:
We've had a fast path for re-binding the same vertex buffer at a different offset for a while, but never really updated D3D9 to actually use it.

This actually seems to improve lows in the RE6 benchmark a fair bit, which makes sense given that `bindVertexBuffers` is *the* #1 function in a CPU profile that we actually have control over.

@K0bin let me know if I missed anything wrt state tracking, this *should* be functionally identical to the previous code wrt unbinding etc.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -474,11 +474,11 @@ namespace dxvk {
       m_autoDepthStencil = nullptr;
 
       // Unbind all buffers that were still bound to the backend to avoid leaks.
-      EmitCs([](DxvkContext* ctx) {
+      EmitCs([] (DxvkContext* ctx) {
         ctx->bindIndexBuffer(DxvkBufferSlice(), VK_INDEX_TYPE_UINT32);
-        for (uint32_t i = 0; i < caps::MaxStreams; i++) {
+
+        for (uint32_t i = 0; i < caps::MaxStreams; i++)
           ctx->bindVertexBuffer(i, DxvkBufferSlice(), 0);
-        }
       });
 
       // Tests show that regular D3D9 ends the scene in Reset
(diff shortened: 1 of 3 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -3685,47 +3685,53 @@ namespace dxvk {
...
+          vbo.offset, vbo.length, vbo.stride);
+      } else {
+        // D3D9 doesn't actually unbind any vertex buffer when passing null.
+        // Operation Flashpoint: Red River relies on this behavior.
+        m_vbSlotTracking.bound &= ~bit;
+
+        vbo.vertexBuffer = nullptr;
+        vbo.offset = 0u;
+        vbo.length = 0u;
+        vbo.stride = 0u;
```
K0bin, line 3725: This makes me nervous, especially since I can't get Operation Flashpoint Red River to work to test this.

src/d3d9/d3d9_device.cpp
```diff
@@ -7752,34 +7759,63 @@ namespace dxvk {
...
+      });
+    } else {
+      EmitCs([cSlotId = Slot] (DxvkContext* ctx) mutable {
+        ctx->bindVertexBuffer(cSlotId, DxvkBufferSlice(), 0);
+      });
+    }
+  }
+
+
+  void D3D9DeviceEx::BindVertexBufferRange(
```
K0bin, line 7785: This seems a little redundant when we already have `D3D9DeviceEx::BindVertexBuffer()`.

If there's a good reason to have both, maybe throw in a short comment.

src/d3d9/d3d9_device.cpp
```diff
@@ -7752,34 +7759,63 @@ namespace dxvk {
...
 
   void D3D9DeviceEx::BindIndices() {
-    D3D9CommonBuffer* buffer = GetCommonBuffer(m_state.indices);
-
-    D3D9Format format = buffer != nullptr
-                      ? buffer->Desc()->Format
-                      : D3D9Format::INDEX32;
-
-    const VkIndexType indexType = DecodeIndexType(format);
+    if (likely(m_state.indices)) {
```
K0bin, line 7806: I guess there's no point in the backend to ever explicitly unbind the index buffer then.

Mhm, best thing I can come up with now, is reference counting but that's more of a theoretical use case.

K0bin (commented):
Looks fine, just some minor thoughts.

src/d3d9/d3d9_device.cpp
doitsujin: The old code already set `vertexBuffer` to null even when binding null, and from what I understand, the `EndScene` clean-up code requires that to even do anything.

Only potential problem I see here is when we override vertex buffer binding on the dynamic upload path, but once again, that shouldn't regress anything since the old code would also bind null.

src/d3d9/d3d9_device.cpp
doitsujin: The point is to avoid redundant ref-counting, we've done the exact same thing in D3D11 for years.

Can add a comment I guess.

src/d3d9/d3d9_device.cpp
doitsujin: Well we still need to match native behaviour since having a null index buffer bound will obviously affect the way indexed draws work.

K0bin (approved):

PR #5711, opened 2026-06-17
Only hide Intel vendor ID for XeSS 1.x

doitsujin:
What a mess.

Not sure if I'm looking at the *correct* version info here, but Trails beyond the Horizion is detected as XeSS 2.0 whereas Shadow of the Tomb Raider reports 1.1.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -107,6 +107,7 @@ if platform == 'windows'
   lib_d3d11   = cpp.find_library('d3d11')
   lib_dxgi    = cpp.find_library('dxgi')
   lib_gdi32   = cpp.find_library('gdi32')
+  lib_version = cpp.find_library('version')
 
   if dxvk_is_msvc
     res_ext = '.res'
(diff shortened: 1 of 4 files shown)


pchome:
Maybe makes sense to move the version info code to **utils**.

Also, if not too expensive, you could read and store version info from the **exe** file. This could be useful (in theory) for additional hints in profile config when the game currently detected by it's path and name (e.g. `GameName - Extended Edition\Game.exe`). It's possible there is the "OriginalFilename" field in version info which is "GameName.exe". Could be cross-gamestore where directory and file name can differ for the same game.
Just example, maybe there are other useful cases, like Electron based launchers, etc.

doitsujin:
This is only a hack to make XeSS 1.x not crash on Intel while correctly reporting the GPU vendor for other games (which is relevant in D3D12 land), not something I intend to turn into overcomplicated rocket science for literally no reason.

ccallawa-intel:
@doitsujin thanks for this--I was working on a PR but you beat me to it.

Version 2+ is probably fine, but we did a bit of digging in internal repos and found that XeSS 1.3 and newer should work with Intel vendor IDs.

Issue #5165, opened 2025-08-22
[d3d8] Sea Dogs/Age of Pirates 2: City of Abandoned Ships clip plane issues


WinterSnowfall:
There is something wrong with how we apply clip planes, or how they interact with fog in this game, because distant geometry appears unclipped/unfogged when it should be both.

The issue can be reproduced with the attached trace (see the topside contour of the mountain on the distant island). It can also be observed with d3d8to9 + dxvk, so it's not something specific to d3d8.

P.S.: How do I know this is about clip planes? The game has a config option setting for the far clip plane depth. If I increase that by a ridiculous amount, things start to look correctly.

### Software information
Sea Dogs (aka Age of Pirates 2): City of Abandoned Ships

### System information
- GPU: Nvidia 4070
- Driver: 570.181
- Wine version: 10.13
- DXVK version: 2.7+

### Apitrace file(s)
[Sea_Dogs.trace.zstd](https://drive.proton.me/urls/26ZA2AJRT0#p8Is2AnTK9gq)

### Log files
N/A

WinterSnowfall:
P.S.: The first game in the series, Sea Dogs - Caribbean Tales, which uses the same engine, has the same problem.

PR #5714, opened 2026-06-17
[d3d8] Handle odd cases of texture UAF

WinterSnowfall:
The first two games in the Age of Pirates series will crash on startup in most cases because they call SetTexture() with an already released object, and our code blows up when we try to add a private ref to it.

In a moment of absolute irony, Pirates of the Caribbean, a game made by the same studio and using the same bloody engine, but under the obvious license, *doesn't* exhibit such problems, nor does it have any fogging issues for that matter...

In any case, keeping track of valid textures and checking during SetTexture() calls seems to do the trick with no ill effect in the two games. The logic is gated behind a config option due to the obvious overhead, though it's nothing tragic by any means. If anyone has any better ideas, which don't cause a lot of pain and suffering for well behaving games, I'm all ears.

I've taken the liberty to do some code style cleanup while I was at it, split in a preceding commit (with no functional impact).

Related to #5165, but has no impact on it (other than making the affected games playable).

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -311,7 +311,14 @@ namespace dxvk {
     if (unlikely(FAILED(res)))
       return res;
 
-    *ppTexture = ref(new D3D8Texture2D(this, Pool, std::move(pTex9)));
+    IDirect3DTexture8* tex = new D3D8Texture2D(this, Pool, std::move(pTex9));
+
+    if (unlikely(m_d3d8Options.textureUAFGuard)) {
+      D3D8DeviceLock lock = LockDevice();
+      m_validTextures.insert(tex);
+    }
+
+    *ppTexture = ref(tex);
 
     return D3D_OK;
   }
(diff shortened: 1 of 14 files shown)


src/d3d8/d3d8_options.cpp
```diff
@@ -14,7 +14,7 @@ namespace dxvk {
     return value;
   }
 
-  void D3D8Options::parseVsDecl(const std::string& decl) {
+  inline void D3D8Options::parseVsDecl(const std::string& decl) {
```
doitsujin: why `ìnline`? this won't be doing anything useful as long as this is in its own translation unit, and will at worst screw us over with random linker errors at some point in the future.

WinterSnowfall: Fair point. Not sure where I was going with that, probably a leftover from a move to util, which I later decided against.

src/d3d8/d3d8_device.cpp
```diff
@@ -1395,7 +1416,12 @@ namespace dxvk {
     if (unlikely(FAILED(res)))
       return res;
 
-    m_textures[Stage] = tex;
+    if (unlikely(m_d3d8Options.textureUAFGuard
```
doitsujin: shouldn't this check be immediately *before* the `ShouldRecord` above? i.e. something along the lines of `if (a.find() == a.end()) tex = nullptr;`

Otherwise we pass this through to D3D9 and stuff can explode there.

WinterSnowfall: Believe it or not, if I derp out before the call to d3d9 water becomes borked in the game (probably some other released object remains in place), so I assume d3d9 is handling it properly.

PR #5725, opened 2026-06-21
FF texture tracking

### LINKED ISSUE #5285: [d3d9] Performance regression in Dawn of War DE with Ubershader
Dawn of War DE have quite a huge performance regression since the enablement of the Ubershader.
See screenshots below for a comparison without and with the feature disabled.
<details>
  
No config
Image

With `DXVK_CONFIG="d3d9.ffUbershaderVS = False; d3d9.ffUbershaderFS = False"` 
Image
</details>

Disabling just `VS` on its own improves it from 79fps ish -> 145fps ish in that scene. While `FS` on its own is 79fps ish -> 108fps ish.

### Software information
Warhammer 40,000: Dawn of War - Definitive Edition
Game version 2.4.0 and highest settings at 1080p

### System information
- GPU: RX 7900 XTX
- Driver: Day old mesa-git (9ebda88e) 
- Wine version: Proton Bleeding Edge
- DXVK version: Master

### Apitrace file(s)
https://drive.proton.me/urls/VE0XFXQSG8#zvxxrP8LUPFJ

doitsujin:
Fixes #5285.

We could *theoretically* be even more aggressive here and check alpha test, color write masks etc, but that will lead to a) a lot more state dirtying and b) more complications because the order in which we update state actually matters and we have *so much random state* to look at.

I **hope** I got all the state tracking right, the whole thing where all the bitmasks feed into each other is just pain.

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -8253,44 +8253,43 @@ namespace dxvk {
   }
 
 
+  uint32_t D3D9DeviceEx::GetTextureStageArgMask(
+          D3DTEXTUREOP          Op) {
+    switch (Op) {
+      // No args used
+      case D3DTOP_DISABLE:
+      case D3DTOP_BUMPENVMAP:
+      case D3DTOP_BUMPENVMAPLUMINANCE:
+        return 0b000u;
+
+      // Arg 1 only
+      case D3DTOP_SELECTARG1:
+      case D3DTOP_PREMODULATE:
+        return 0b010u;
+
+      // Arg 2 only
+      case D3DTOP_SELECTARG2:
+        return 0b100u;
+
+      // Arg 0, 1, 2
+      case D3DTOP_MULTIPLYADD:
+      case D3DTOP_LERP:
+        return 0b111u;
+
+      // Arg 1, 2
+      default:
+        return 0b110u;
+    }
+  }
+
+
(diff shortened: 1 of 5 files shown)


src/d3d9/d3d9_device.cpp
```diff
@@ -8253,81 +8256,164 @@ namespace dxvk {
...
+
+    if (Op == D3DTOP_BLENDCURRENTALPHA)
+      result.set(D3D9TextureStageStateFlag::UsesCurrent);
+
+    // Check stage arguments actually used by the op
+    std::array<UINT, 3u> args = { Arg0, Arg1, Arg2 };
+
+    for (uint32_t i = 0u; i < args.size(); i++) {
+      if (usedArgMask & (1u << i)) {
+        auto arg = args[i] & D3DTA_SELECTMASK;
```
K0bin, line 8316: Tiny nit but I liked my `// Strip modifiers from arguments.` comment simply because last time I ended up googling what the point of `D3DTA_SELECTMASK` is.

K0bin (commented):
Just one tiny thing otherwise it looks good as far as I can tell.

src/d3d9/d3d9_device.cpp
doitsujin: fair enough, will re-add.

K0bin (approved):

PR #5746, opened 2026-06-28
[dxvk] Adjust legacy descriptor pool allocation

doitsujin:


understyx:
Unfortunately still happening (atleast in wow)
err:   Exception on CS thread!
err:   Failed to allocate descriptor set: VK_ERROR_OUT_OF_POOL_MEMORY_KHR

Edit: I can try granblue in 3 hours + whatever it'll take me to DL if you want, since wow seems bit dodgy

doitsujin:
Nothing we can do on our end then, short of disabling GPL.

Testing different games won't really be useful here since the driver is choking on stuff that's used everywhere.

PR #5764, opened 2026-07-02
Revert "[d3d9] Ignore buffer lock range for static D3DPOOL_DEFAULT buffers

### LINKED ISSUE #5067: Initial load/calculation time on D3D9 demo much slower than WineD3D

### LINKED ISSUE #5638: Kane & Lynch: Dead Men Lagging (after 2.7.1)
With DXVK, starting at some point after 2.7.1's release, the game will lag to an unplayable level (super stuttery at sub 45 FPS) when loading into a mission. Simply going to the latest release (2.7.1) fixes it, and I can easily hit over 600 FPS with no stuttering.

### Software information
Kane & Lynch: Dead Men on Steam

### System information
- OS: Windows 11 Pro
- GPU: Nvidia RTX 5090 
- Driver: 596.49
- DXVK version: I tried going back as far as I could to pinpoint it, but it still occurred on the oldest available action build I was able to try, which is here: https://github.com/doitsujin/dxvk/actions/runs/24615801533

### Log files

[kaneandlynch_d3d9.log](https://github.com/user-attachments/files/27974625/kaneandlynch_d3d9.log)

(I know it says I have an AMD integrated GPU, tried disabling that in the BIOS & DDUing both AMD & Nvidia drivers, then just installing the Nvidia driver and it made no difference, so it's definitely not using the AMD GPU or something along those lines).

K0bin:
The change to ignore the locking range arguments for `D3DPOOL_DEFAULT` buffers was done to fix #5067 because IIRC that wrote outside of the provided range. It severely impacted performance in Kane and Lynch.

A random broken demo is much less important as an actual game so let's just revert it for now.

Fixes #5638

### DIFF
--- a/src/d3d9/d3d9_device.cpp
+++ b/src/d3d9/d3d9_device.cpp
@@ -5482,8 +5482,7 @@ namespace dxvk {
     // (TODO: Apparently this is meant to happen for DYNAMIC too but I am not sure
     //  how that works given it is meant to be a DIRECT access..?)
     const bool respectUserBounds = !(Flags & D3DLOCK_DISCARD) &&
-                                    SizeToLock != 0 &&
-                                    (desc.Pool == D3DPOOL_MANAGED || (desc.Usage & D3DUSAGE_DYNAMIC));
+                                    SizeToLock != 0;
 
     // If we don't respect the bounds, encompass it all in our tests/checks
     // These values may be out of range and don't get clamped.


WinterSnowfall:
For posterity's sake: it doesn't affect only Kane & Lynch, but also the new King's Bounty games, starting with King's Bounty: The Legend and all later standalone addons. The PR will also fix the performance regression in those games, since @Blisto91 bisected it down to the same commit.

PR #5770, opened 2026-07-03
[util] Update Gothic 3 comment after removal of floatEmulation config

LingMan:
568b8836 dropped the floatEmulation config for Gothic 3 but forgot to also remove the related explanation in the comment above.


Blisto91:
Derp

Issue #5766, opened 2026-07-03
Random frametime spikes / microstutters in GTA IV Complete Edition


hosseinghasemi141-prog:
**Update:**
The Google Drive permissions have been fixed. The existing video link is now public and should be accessible.
Thank you for pointing that out.

### Software information

Grand Theft Auto IV: Complete Edition

### System information

- GPU: NVIDIA GeForce RTX 4060 8GB
- CPU: Intel Core i5-12400F
- RAM: 16 GB
- OS: Windows 11
- DXVK version: 3.0
- FusionFix: 5.0.1

### Bug description

I'm experiencing random frametime spikes / microstutters in Grand Theft Auto IV: Complete Edition.

The game runs at a stable frame rate, but the frametime graph frequently shows short random spikes that cause noticeable stuttering.

The issue occurs:
- while driving,
- while walking,
- while standing completely still without moving the camera,
- and occasionally even in the game menus.

The behaviour is consistent and can be seen in the attached video and frametime graph.

I noticed that the frametime spikes look very similar to those shown in Issue #1419, although I'm not sure if it's related.

### Reproduction

1. Launch GTA IV: Complete Edition.
2. Play normally or simply stand still.
3. Observe the frametime graph.
4. Random frametime spikes appear.

### Additional information

- VSync enabled
- FusionFix 5.0.1
- DXVK 3.0
video:
https://drive.google.com/file/d/1qMaFNQashxs1cktBHZDmfQbLjrvDA8gO/view?usp=drive_link

**Update:**
The Google Drive permissions have been fixed. The existing video link is now public and should be accessible.
Thank you for pointing that out.

video with DXVK hud enabled:
https://drive.google.com/file/d/1YaWtd83DrLLw6JAtMT17T1umdKdLfCJj/view?usp=drive_link

doitsujin:
I  mean, isn't stuttering literally normal for this game? Game logic, resource streaming and things being mostly single-threaded aren't things that DXVK is magically going to improve.

Your google drive is private.

Blisto91:
Would be more interesting with a video where dxvk hud is turned on.
Can be enabled by putting `dxvk.hud = full` in a file called `dxvk.conf` next to the exe

K0bin:
The problem here seems to be shader compilation stutter due to how the game manages its shaders combined with address space limits due to the fact that it's a 32 bit game.

hosseinghasemi141-prog:
[...]

PR #5772, opened 2026-07-03
Use KHR_pipeline_binary instead of relying on driver cache for 32-bit pipeline libraries.

doitsujin:
May or may not improve #5766.

Marking as draft because the unmappnig stuff needs work.


doitsujin:
Doesn't work. Unfortunately this increases virtual memory usage on NV by so much that games become unstable again.

PR #5794, opened 2026-07-14
dxgi: Support modifying swapchain flags in DxgiSwapChain::ResizeBuffers1().

zzhiyi:
Overwatch2 (SteamID: 2357570) doesn't create its swapchain with DXGI_SWAP_CHAIN_FLAG_ALLOW_MODE_SWITCH. It relies on IDXGISwapChain::ResizeBuffers() to add the flag so that it can change the display resolution after the swapchain creation.

The tests and corresponding Wine MR are at https://gitlab.winehq.org/wine/wine/-/merge_requests/11391.

### DIFF
--- a/src/dxgi/dxgi_swapchain.cpp
+++ b/src/dxgi/dxgi_swapchain.cpp
@@ -428,14 +428,28 @@ namespace dxvk {
     if (m_window && !wsi::isWindow(m_window))
       return DXGI_ERROR_INVALID_CALL;
 
-    constexpr UINT PreserveFlags = DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT;
+    /* The following six flags are not allowed in ResizeBuffers() */
+    if (SwapChainFlags & (DXGI_SWAP_CHAIN_FLAG_NONPREROTATED | DXGI_SWAP_CHAIN_FLAG_DISPLAY_ONLY
+      | DXGI_SWAP_CHAIN_FLAG_FULLSCREEN_VIDEO))
+        return DXGI_ERROR_INVALID_CALL;
 
-    if ((m_desc.Flags & PreserveFlags) != (SwapChainFlags & PreserveFlags))
-      return DXGI_ERROR_INVALID_CALL;
-    
+    if (SwapChainFlags & (DXGI_SWAP_CHAIN_FLAG_FOREGROUND_LAYER | DXGI_SWAP_CHAIN_FLAG_YUV_VIDEO
+      | DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING))
+        return E_INVALIDARG;
+
+    /* DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT can't be changed after the swapchain creation */
+    if ((m_desc.Flags ^ SwapChainFlags) & DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT)
+        return E_INVALIDARG;
+
+    /* DXGI_SWAP_CHAIN_FLAG_HW_PROTECTED can't be turned off once once it's set */
+    if (m_desc.Flags & DXGI_SWAP_CHAIN_FLAG_HW_PROTECTED
+      && !(SwapChainFlags & DXGI_SWAP_CHAIN_FLAG_HW_PROTECTED))
+        return E_INVALIDARG;
+
     std::lock_guard<dxvk::mutex> lock(m_lockBuffer);
     m_desc.Width  = Width;
     m_desc.Height = Height;
+    m_desc.Flags = SwapChainFlags;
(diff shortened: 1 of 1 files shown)


src/dxgi/dxgi_swapchain.cpp
```diff
@@ -428,14 +428,25 @@ namespace dxvk {
     if (m_window && !wsi::isWindow(m_window))
       return DXGI_ERROR_INVALID_CALL;
 
-    constexpr UINT PreserveFlags = DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT;
+    if (SwapChainFlags & (DXGI_SWAP_CHAIN_FLAG_NONPREROTATED | DXGI_SWAP_CHAIN_FLAG_DISPLAY_ONLY
+      | DXGI_SWAP_CHAIN_FLAG_FULLSCREEN_VIDEO))
+        return DXGI_ERROR_INVALID_CALL;
+
+    if (SwapChainFlags & (DXGI_SWAP_CHAIN_FLAG_FOREGROUND_LAYER | DXGI_SWAP_CHAIN_FLAG_YUV_VIDEO
```
doitsujin, line 436: can we have some comments here on why this check and the one above are needed?

src/dxgi/dxgi_swapchain.cpp
```diff
@@ -428,14 +428,25 @@ namespace dxvk {
...
+      | DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING))
+        return E_INVALIDARG;
+
+    if ((m_desc.Flags ^ SwapChainFlags) & DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT)
+        return E_INVALIDARG;
+
+    if (m_desc.Flags & DXGI_SWAP_CHAIN_FLAG_HW_PROTECTED
+      && !(SwapChainFlags & DXGI_SWAP_CHAIN_FLAG_HW_PROTECTED))
```
doitsujin, line 446: so adding HW_PROTECTED after the fact is allowed, but removing it isn't? Not that we implement this anyway.

doitsujin (commented):
I guess another question is, is there any difference between D3D11 and D3D12 here?

zzhiyi:
> I guess another question is, is there any difference between D3D11 and D3D12 here?

For ResizeBuffers(), there is no difference between D3D11 and D3D12. Please see [ResizeBuffers() tests](https://gitlab.winehq.org/wine/wine/-/merge_requests/11391/diffs?commit_id=05b06a416a6ed98e104ec03e384590fa3a6304bc#06f6a7decc34b79b8d444693d16a3ee0c5a5f96a_4717_4720). For ResizeBuffers1() on D3D12, it works just like ResizeBuffers(). For ResizeBuffers1() on D3D11, it seems to always return DXGI_ERROR_INVALID_CALL or E_INVALIDARG, even when the flag is 0. Please see [ResizeBuffers1() tests](https://gitlab.winehq.org/wine/wine/-/merge_requests/11391/diffs?commit_id=05b06a416a6ed98e104ec03e384590fa3a6304bc#06f6a7decc34b79b8d444693d16a3ee0c5a5f96a_4717_4766). DXVK already allow ResizeBuffers1() to succeed on D3D11 before this MR so if we want to do it exactly the same as native, we could do it in another patch. It's probably ok if it succeeds.

src/dxgi/dxgi_swapchain.cpp
zzhiyi: Yes. Please see [DXGI_SWAP_CHAIN_FLAG_HW_PROTECTED can't be turned off](https://gitlab.winehq.org/wine/wine/-/merge_requests/11391/diffs?commit_id=05b06a416a6ed98e104ec03e384590fa3a6304bc#06f6a7decc34b79b8d444693d16a3ee0c5a5f96a_4721_4900). I will add a comment.

src/dxgi/dxgi_swapchain.cpp
zzhiyi: Sure.

zzhiyi:
> Causes a lot of games to error out and crash.

Sorry, could you share a list of games that crash with this PR?

zzhiyi:
> > Causes a lot of games to error out and crash.
> 
> Sorry, could you share a list of games that crash with this PR?

Oh, there is https://github.com/doitsujin/dxvk/issues/5796. I will take a look. Sorry for the trouble.

zzhiyi:
The reason this PR breaks other games is that ResizeBuffers() should accept flags added during swapchain creation. For example, if DXGI_SWAP_CHAIN_FLAG_ALLOW_TEARING is used when creating a swapchain, then the flag is allowed in ResizeBuffers(). We should only refuse the flag when it is not used at swapchain creation. Anyway, https://github.com/doitsujin/dxvk/pull/5797 looks good to me. Sorry for breaking everything.

PR #5803, opened 2026-07-20
Remove DXVK_HUD env var and frame rate limiter options

doitsujin:
Apparently some software thinks it's a great idea to enable the HUD so let's just remove the env var. The dxvk.conf way will keep working.

Also makes the frame rate limiter options unusable outside of app profiles. I'm tired.


K0bin:
Not a fan but I saw the issue, so go ahead I guess.

Just a matter of time before that software or confused users create a dxvk.conf in the user directory though.

WinterSnowfall:
I'd just leave the _maxFrameRate_ options as they are, but in case you want to keep them as per PR maybe nuke the dxvk.conf section about that part too.

doitsujin:
I just want people to stop fucking moaning about the limiter when we *already say* not to use it.

PR #5814, opened 2026-07-28
[dxvk] Load Intel vendor hack DLL

damian-tomczak-intel:
This vendor hack allows the Intel Extensions Loader to locate and load the igdext.dll library. This is because the loader searches for the library based partly on the location of the loaded fake DirectX 11 Intel UMD.

### DIFF
--- a/src/d3d11/d3d11_device.cpp
+++ b/src/d3d11/d3d11_device.cpp
@@ -3764,6 +3764,27 @@ namespace dxvk {
   
 
 
+  static HMODULE d3d11_device_init_vendor_hacks(
+      const Rc<DxvkAdapter>& Adapter) {
+    if (Adapter->deviceProperties().core.properties.vendorID == uint32_t(DxvkGpuVendor::Intel)) {
+      char sysdir[MAX_PATH], path[MAX_PATH];
+      GetSystemDirectoryA(sysdir, sizeof(sysdir));
+      snprintf(path, sizeof(path),
+               "%s\\DriverStore\\FileRepository\\igd_faux.inf_1\\igd10iumd64.dll", sysdir);
+      HMODULE igd10iumd = ::LoadLibraryA(path);
+      if (igd10iumd)
+        Logger::info("Loaded igd10iumd64.dll for Intel driver workarounds");
+      return igd10iumd;
+    }
+    return nullptr;
+  }
+
+  static void d3d11_device_cleanup_vendor_hacks(HMODULE igd10iumd64) {
+    if (igd10iumd64)
+      ::FreeLibrary(igd10iumd64);
+  }
+
+
   D3D11DXGIDevice::D3D11DXGIDevice(
           IDXGIAdapter*       pAdapter,
           ID3D12Device*       pD3D12Device,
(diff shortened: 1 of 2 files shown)


src/d3d11/d3d11_device.cpp
```diff
@@ -3764,6 +3764,27 @@ namespace dxvk {
   
 
 
+  static HMODULE d3d11_device_init_vendor_hacks(
+      const Rc<DxvkAdapter>& Adapter) {
+    if (Adapter->deviceProperties().core.properties.vendorID == uint32_t(DxvkGpuVendor::Intel)) {
+      char sysdir[MAX_PATH], path[MAX_PATH];
+      GetSystemDirectoryA(sysdir, sizeof(sysdir));
+      snprintf(path, sizeof(path),
+               "%s\\DriverStore\\FileRepository\\igd_faux.inf_1\\igd10iumd64.dll", sysdir);
```
HansKristian-Work, line 3773: That's not the path in the vkd3d-proton patch. Did I apply the wrong patch?

src/d3d11/d3d11_device.cpp
```diff
@@ -3764,6 +3764,27 @@ namespace dxvk {
   
 
 
+  static HMODULE d3d11_device_init_vendor_hacks(
```
doitsujin: Style nit: Ideally, these new function would just be private methods of `D3D11DXGIDevice` (and generally use camel case like everything else in the code).

src/d3d11/d3d11_device.cpp
```diff
@@ -3764,6 +3764,27 @@ namespace dxvk {
...
+      const Rc<DxvkAdapter>& Adapter) {
+    if (Adapter->deviceProperties().core.properties.vendorID == uint32_t(DxvkGpuVendor::Intel)) {
```
doitsujin: This entire `if` block needs to be inside an `#ifdef _WIN32` to fix native.

src/d3d11/d3d11_device.cpp
doitsujin, line 3773: Nit, we have `str::format(sysdir, "\\DriverStore...")` for this which returns an std::string.

Also, see Hans-Kristian's comment above.

src/d3d11/d3d11_device.cpp
doitsujin: This was changed and doesn't compile anymore, this should do:
```suggestion
    if (Adapter->info().vendorId == uint32_t(DxvkGpuVendor::Intel)) {
```

doitsujin (changes requested):

src/d3d11/d3d11_device.cpp
ivyl: I think the draft version that was shared around was using `intc_wine`. I went with `igd_faux.inf_1` in Proton but I can adjust that if needed.

https://github.com/ValveSoftware/wine/commit/eda69af06f8cbbd6f9a5bc3a3b4e34095f957ace

HansKristian-Work: So the vkd3d-proton thing I merged is broken?

ivyl: For context: `igd_faux.inf_1` matches drivers backed by actual inf files that we currently have in the tree for real drivers a bit better, e.g. `system32/DriverStore/FileRepository/winebus.inf_1`. I went with this convention as it's not unlikely that we'll migrate it to a true, installable `.inf` down the line.

HansKristian-Work: Could you make a vkd3d-proton PR that uses the proper path that Wine settled on?

ivyl: Sure. I'll send a patch once everything is settled on and we know more about the actual DLLs. The Wine patches are hacks that are in bleeding-edge only for now.

PR #5815, opened 2026-07-28
[d3d8] Rework legacy D3D compatibility mode handling

WinterSnowfall:
Refactors the D3D9 bridge with regards to naming and generalizes both the interfaces as well as the setting/storing/checking of D3D compatibility modes, using a `util_flags` bitmap. Very meh when considering D3D8 by itself, however.... you can imagine how "nice" it is to have 5 bools and 5 separate compatibility calls in D7VK. This PR provides a framework on which that entire mess can be simplified.


K0bin (approved):
LGTM

More of a renaming than a big rework.

WinterSnowfall:
> More of a renaming than a big rework.

Well yes, the upstreamable part is functionally identical.

PR #5817, opened 2026-07-29
[d3d8] Assorted debug loggers cleanup

WinterSnowfall:
I've seen these can get spammed quite a bit even during rendering in certain cases, and they're not exactly free real estate even when they don't get printed, as I found out the hard way in D7VK. If they're not generally useful for some odd reason I think it's best to remove them altogether, since they can always be brought back temporarily if any testing/debugging is needed.

Edit: Added 2 more things:
- A cleanup of CopyRects logging, since it is now fully implemented and doesn't have any known edge cases. Moreover, some games expect it to fail at times, and it's a fairly hot path.
- Explicitly log unhandled token types in TranslateVertexShader8 (and comment out the verbose debug loggers by default)

### DIFF
--- a/src/d3d8/d3d8_device.cpp
+++ b/src/d3d8/d3d8_device.cpp
@@ -41,8 +41,6 @@ namespace dxvk {
   }
 
   HRESULT STDMETHODCALLTYPE D3D8Device::GetInfo(DWORD DevInfoID, void* pDevInfoStruct, DWORD DevInfoStructSize) {
-    Logger::debug(str::format("D3D8Device::GetInfo: ", DevInfoID));
-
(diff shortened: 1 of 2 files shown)


Digger1955:
@WinterSnowfall 

I am sorry to intervene, but isn't it better to comment them out? If they are needed - uncomment them.

WinterSnowfall:
That's also an option, can do either. I'll leave it up to K0bin/doitsujin.

WinterSnowfall:
Found a couple of more things I'd like to clean up a bit in d3d8 code, so drafting for now.

doitsujin:
is there any particular reason to nuke the map/unmap log? That really be spammed a lot in practice.

PR #5824, opened 2026-08-02
[build] fix native build on macos

theoparis:


doitsujin:
main question / concerns here are:
- why is WSI hard-coded to only support SDL2?
- why are WSI and Vulkan statically linked when we load everything dynamically everywhere else?

src/dxvk/dxvk_device_info.cpp
```diff
@@ -827,10 +827,10 @@ namespace dxvk {
       ENABLE_FEATURE(core.features, depthClamp, true),
       ENABLE_FEATURE(core.features, drawIndirectFirstInstance, false),
       ENABLE_FEATURE(core.features, dualSrcBlend, true),
-      ENABLE_FEATURE(core.features, fillModeNonSolid, true),
+      ENABLE_FEATURE(core.features, fillModeNonSolid, false),
```
theoparis: Before this change my demo program using d3d9 from dxvk native would crash with both MoltenVK and KosmicKrisp due to the unsupported features. ~~I'm not sure that this is the best approach though as it might break other projects that need the feature...  The same goes for the changes in dxvk_graphics.cpp.~~ A cleaner approach might be overriding it in checkDeviceCompatibility actually

PR #5825, opened 2026-08-02
[d3d9] Handle D3D9Ex checks with the D3D compatibility flags

WinterSnowfall:
@CkNoSFeRaTU gave me the idea to also merge the D3D9Ex/IsExtended() checks into our recently added D3D compatibility flags logic. I think it cleans up quite a bit of extra handling, and makes more sense this way - I know it's not a "compatibility mode" per se, but it very much behaves like one anyway for the most part.


doitsujin:
Removing an indirection certainly won't hurt.

For future reference though, I'd prefer if commits for such reworks were split into something like
- Add new stuff (i.e. add and set 9Ex flag in this case)
- Refactor code to use new stuff instead of old stuff (potentially multiple if it actually functionally changes anything)
- Nuke old stuff

Kinda painful to review otherwise when the code that sets the flag is drowned in trivial changes.

WinterSnowfall:
> For future reference though, I'd prefer if commits for such reworks were split into something like

Noted, but I think that's about it anyway (for the foreseeable future).

PR #5828, opened 2026-08-05
[d3d11] Implement NVAPI multiview rendering

DanFraserUK:
These changes receive information from dxvk-nvapi to generate and position geometry correctly and quickly for Simultaneous Multi-Projection implementation.

iRacing is the main program using this SMP/MVP technology, and as far as I can tell the only game to actually do so.  These changes only affect users with an NVIDIA card of the Turing generation or newer and nothing changes for any users not using this feature either with an incompatible graphics card or SMP simply being disabled in game.

Multiple test runs and the DXVK_HUD and Mangohud data collected from them are in this graphic:

smp-triple-screen

The paired dxvk-nvapi PR that this PR needs to work correctly is
https://github.com/jp7677/dxvk-nvapi/pull/388 and both need to be
merged for either to do anything.

I'm happy to do further testing that may be needed.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -94,7 +94,8 @@ namespace dxvk {
     }
 
     if (riid == __uuidof(ID3D11VkExtContext)
-     || riid == __uuidof(ID3D11VkExtContext1)) {
+     || riid == __uuidof(ID3D11VkExtContext1)
+     || riid == __uuidof(ID3D11VkExtContext2)) {
       *ppvObject = ref(&m_contextExt);
       return S_OK;
     }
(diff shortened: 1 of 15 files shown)


doitsujin:
As with any massive feature dump, it would have been nice to discuss the need for this and technical direction etc beforehand and then upstream this in small chunks? I can't meaningfully review a single +1k LoC patch.

Honestly I don't know how interested I am in a feature that works in exactly *one* game on exactly *one* class of GPUs *and* requires a multi-monitor setup, because all that's really going to do is create a *massive* maintenance burden for us or just randomly break when we decide to change anything ever. How exactly are we supposed to test this going forward?

Also, why exactly does this require Nvidia-only Vulkan extensions anyway? Multiview is *not* limtied to Nvidia, and at least AMD have their own D3D11 vendor extension for this.

If this was a cross-vendor implementation and if there were more (VR?) apps known to actually use it then sure, could be worth discussing, but the reason why we haven't bothered with this is precisely *because* the ratio between implementation/maintenance effort and the number of apps known to benefit from it is really not great.

DanFraserUK:
Fair points, and sorry for dumping it so hard on you like this.  I didn't realise I should have been in contact to discuss directions first.  I agree that a single 1k patch isn't easily reviewable, and you're not the only person pointing it out.

I'm new to submitting work to others.  One Wine MR is the sum of my experience with pull requests and I'm learning the process as I go.

On the Nvidia-only point I got that wrong.  I had it checking for VK_NV_viewport_array2 when nothing in the path actually uses it.  The shader it builds runs one invocation per view and writes ViewportIndex, so core geometryShader and multiViewport is all it needs, and that's what it checks for now.  I'd also left two NV extensions declared in dxvk core from earlier work that nothing ever read, and those have gone with it.  So it should run anywhere those core features are available now, not just Nvidia.

The maintenance and testing side I'd like to work through with you rather than guess at.  If there's a shape that would make this less of a burden to carry, I'd rather build it that way now than argue for what I've already written.

I've been reading up on how things like this have been handled here before, and there's clearly a better way to structure this than what I did.  Happy to do the legwork on the testing side too, I just don't know yet what would actually be useful to you.

If you've got the time to point me in the right direction on what a sensible split into chunks looks like from your side, and what you'd count as a decent testing story for something you may not be able to easily run yourself, that'd help a lot.

doitsujin:
I don't really have a lot of time to really deal with this right now and this is going to be low priority for now, but I think the first step here would be to outline how exactly the AGS and NVAPI extensions for multiview actually *work* so I can even have an idea what the implementation needs to do and could look like.

FWIW, I don't think injecting a GS is going to work when the pipeline itself already has a GS; if anything I think we're going to have to patch the last active geometry stage to emit the proper viewport/layer based on the current view ID and figure out how to actually do the actual multiview part (probably has to involve draw looping + passing view id via push constant rather than Vulkan multiview though because Vulkan multiview is rather... restrictive)

DanFraserUK:
No problem.  I'll go away and put together a proper write up of how the NVAPI and AGS multiview extensions work.

I've had a quick look and the two vendors seem to go about it differently, so it'll take me a bit to get that down properly.  I'll come back when I've got something worth reading rather than maybe muddling things bit by bit.

Issue #1241, opened 2019-11-08
Steam Overlay won't hook with DXVK in Windows


DaRkL3AD3R:
When using DXVK (and D9VK), Steam Overlay will not hook to any game using the wrapper.

### System information
- GPU: 1080 Ti
- Driver: 441.12
- DXVK version: 1.4.4
- OS: Windows 10 1909 x64

doitsujin:
Sounds more like a Steam Overlay issue than a DXVK issue to be honest. Even if it doesn't hook D3D11, it should at least enable its Vulkan layer.

In any case, there's nothing I can do here.

PR #5832, opened 2026-08-08
[dxgi] Reserve a hot-patchable entry on the DXGI entry points overlays hook

anishgoyal1108:
### Problem

Inline-hook engines (the Steam overlay, RTSS, Discord, ReShade's loader) install hooks by copying a function's first bytes into a trampoline and writing a 5-byte `jmp rel32` over them. To do that they must length-decode the prologue, and in practice their decoders only cover the instruction forms MSVC emits — on Windows that has always been sufficient.

Built with frame pointers, clang gives `DxgiFactory::CreateSwapChainForHwnd`:

```
55              push rbp
48 89 e5        mov  rbp, rsp
48 8b 45 40     mov  rax, [rbp+0x40]     <- ModRM mod=01, rm=101
```

`mov r64,[rbp+disp8]` is effectively absent from MSVC x86-64 output, which omits frame pointers and reaches stack arguments through `rsp`+SIB. Steam's decoder stops there holding 4 decoded bytes, one short of the 5 it needs, and skips the hook:

```
Unknown opcodes for AMD64 at 4 bytes: 55 48 89 E5 48 8B 45 40 ...,
    module=dxgi.dll,IDXGIFactory2_CreateSwapChainForHWND
```

`CreateSwapChainForHwnd` is the entry point every modern D3D11/D3D12 title uses, so nothing downstream is wrapped: no `Present` hook, no renderer, and the overlay silently never draws. D3D9 titles are unaffected because they never touch DXGI. This is the same user-visible symptom as #1241.

It is exactly **one function out of 7542** in `dxgi.dll` — and it happens to be the one that matters.

### Why not just change build flags

Frame pointers are what trigger it here, and several downstream distributions build DXVK that way. But turning the flag off is a fix that depends on the compiler continuing to make a favourable choice for this particular function; at `-O2` without frame pointers clang emits `48 8b 44 24 38` and it happens to work. Nothing guarantees that stays true across compilers, versions or optimisation levels — and DXVK currently offers no guarantee that its COM entry points are hot-patchable at all, while inline-hook engines have been hooking DXGI on Windows for a decade.

### Fix

Do what MSVC's `/hotpatch` does — guarantee decodable, relocatable bytes at the function entry — using the GCC/clang `patchable_function_entry` attribute, applied only to the virtual entry points overlays actually hook. This is codegen-independent: it holds under any flags, compiler or optimisation level. The cost is a few NOP bytes per function, on swapchain-creation and present paths that already do far more work.

### Verification

Emitted padding, checked with clang 22 and gcc 16 targeting `x86_64-windows`:

| compiler | bytes |
|---|---|
| clang | `0f 1f 44 00 08` (one 5-byte NOP) |
| gcc | `90 90 90 90 90` (five 1-byte NOPs) |

Both were confirmed to be **accepted by Steam's decoder**, not assumed:

* gcc's form — the overlay's hook installer contains an explicit `cmp al, 0x90`.
* clang's form — tested end to end by placing `0f 1f 44 00 08` at the real `CreateSwapChainForHwnd` entry in a shipped `dxgi.dll` and running a D3D12 title. The overlay then logs `IWrapDXGIFactory2::IDXGIFactory2_CreateSwapChainForHWND called` and goes on to hook the swapchain vtable and create its renderer, where the unmodified build stops at the `Unknown opcodes` line.

Built with `meson --cross-file build-win64.txt -Dcpp_args=-fno-omit-frame-pointer` (i.e. the configuration that reproduces the bug) and confirmed in the resulting binary — `CreateSwapChainForHwnd` now begins with the padding, and the previously-undecodable instruction sits harmlessly at offset 5:

```
0x7e90  90 90 90 90 90
0x7e95  55              push rbp
0x7e96  48 89 e5        mov  rbp, rsp
0x7e99  48 8b 45 40     mov  rax, [rbp+0x40]
```

I'm happy to trim this to just the four factory methods if the swapchain ones (`Present`, `Present1`, `ResizeBuffers`, `SetFullscreenState`) are more surface than you want — the factory entry points alone fix the reported failure.

### Benefit
This one fix allows Steam Overlay to work within Wayland sessions in DXVK/VKD3D games (with the caveat that the steam.exe stub from within Proton must be an *actual* Wine-ran Steam.exe; this is possible in [modern Proton builds](https://github.com/nanomatters/proton-cachyos)).

### Implementation
Via [dxvk-hotpatchable-entry.patch](https://github.com/user-attachments/files/30862954/dxvk-hotpatchable-entry.patch)

### DIFF
--- a/src/dxgi/dxgi_factory.h
+++ b/src/dxgi/dxgi_factory.h
@@ -8,6 +8,7 @@
 #include "dxgi_options.h"
 
 #include "../dxvk/dxvk_instance.h"
+#include "../util/util_hotpatch.h"
 
 namespace dxvk {
 
@@ -73,11 +74,13 @@ namespace dxvk {
             HMODULE               Module,
             IDXGIAdapter**        ppAdapter) final;
     
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE CreateSwapChain(
             IUnknown*             pDevice,
             DXGI_SWAP_CHAIN_DESC* pDesc,
             IDXGISwapChain**      ppSwapChain) final;
     
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE CreateSwapChainForHwnd(
             IUnknown*             pDevice,
             HWND                  hWnd,
@@ -86,6 +89,7 @@ namespace dxvk {
             IDXGIOutput*          pRestrictToOutput,
             IDXGISwapChain1**     ppSwapChain) final;
 
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE CreateSwapChainForCoreWindow(
             IUnknown*             pDevice,
             IUnknown*             pWindow,
@@ -93,6 +97,7 @@ namespace dxvk {
             IDXGIOutput*          pRestrictToOutput,
             IDXGISwapChain1**     ppSwapChain) final;
     
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE CreateSwapChainForComposition(
             IUnknown*             pDevice,
       const DXGI_SWAP_CHAIN_DESC1* pDesc,
--- a/src/dxgi/dxgi_swapchain.h
+++ b/src/dxgi/dxgi_swapchain.h
@@ -13,6 +13,7 @@
 
 #include "../wsi/wsi_window.h"
 #include "../wsi/wsi_monitor.h"
+#include "../util/util_hotpatch.h"
 
 namespace dxvk {
   
@@ -93,15 +94,18 @@ namespace dxvk {
     
     BOOL STDMETHODCALLTYPE IsTemporaryMonoSupported() final;
 
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE Present(
             UINT                      SyncInterval,
             UINT                      Flags) final;
     
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE Present1(
             UINT                      SyncInterval,
             UINT                      PresentFlags,
       const DXGI_PRESENT_PARAMETERS*  pPresentParameters) final;
 
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE ResizeBuffers(
             UINT                      BufferCount,
             UINT                      Width,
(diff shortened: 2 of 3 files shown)


doitsujin:
Sigh, Windows ecosystems are such a pain in the arse.

Didn't read *all* of the LLM wall of text, I guess this *should* be somewhat safe to do but we really need to make sure that this doesn't break any of this stuff on Proton (mostly thinking of Ubisoft overlays etc here).

anishgoyal1108:
The good news is that we're already testing this change on the newer proton-cachyos fork, which is meant to get the Steam overlay working on Wayland (and other game overlays). Technically, when implementing this change, I only looked into the Steam overlay. But we can update you in a couple of weeks to be absolutely certain about any downstream effects

doitsujin:
Well the main thing I'm interested in is whether the Ubisoft overlay works in e.g. AC:Shadows / Black Flag Remastered or Anno 1800. (Note that this *does* affect D3D12 games)

Can probably test that myself tomorrow (ish) though.

ivyl:
This broke ARM64EC build using [LLVM](https://github.com/ValveSoftware/Proton/blob/bleeding-edge/docker/llvm-mingw.Dockerfile.in) from [Proton SDK](https://gitlab.steamos.cloud/proton/steamrt4/sdk):

```
error: Incorrect size for #_ZN4dxvk11DxgiFactory15CreateSwapChainEP8IUnknownP20DXGI_SWAP_CHAIN_DESCPP14IDXGISwapChain prologue: 36 bytes of instructions in range, but .seh directives corresponding to 16 bytes

error: Incorrect size for #_ZN4dxvk11DxgiFactory28CreateSwapChainForCoreWindowEP8IUnknownS2_PK21DXGI_SWAP_CHAIN_DESC1P11IDXGIOutputPP15IDXGISwapChain1 prologue: 36 bytes of instructions in range, but .seh directives corresponding to 16 bytes

error: Incorrect size for #_ZN4dxvk11DxgiFactory29CreateSwapChainForCompositionEP8IUnknownPK21DXGI_SWAP_CHAIN_DESC1P11IDXGIOutputPP15IDXGISwapChain1 prologue: 40 bytes of instructions in range, but .seh directives corresponding to 20 bytes

6 warnings and 3 errors generated.
[5/123] Compiling C++ object src/dxgi/dxgi.dll.p/dxgi_swapchain.cpp.obj
FAILED: [code=1] src/dxgi/dxgi.dll.p/dxgi_swapchain.cpp.obj 
```

@cjacek may know more if this has anything to do with the recent ARM64X work.

gofman:
I wonder does ```__attribute__((__ms_hook_prologue__))``` like we do similar things in Wine fix both the issue in this bug and the breakage on arm64ec compilation?

gofman:
I sent a fix here: https://github.com/doitsujin/dxvk/pull/5846 (which also makes it to emit a proper hotpatch prologue instead of 'nop' when that it applicable).

PR #5844, opened 2026-08-19
[dxvk] Add a hack to support equal-value fence signals

shaunren:
D3D permits signaling a fence to its current value, including for shared fences, while Vulkan timeline semaphore signals must strictly increase. The Skyrim Upscaling Community Shaders mod, with FSR frame generation enabled, signals a shared fence with equal value from both D3D11 and D3D12, resulting in invalid Vulkan usage and device loss, hence causing the game to hang and crash.

This hack skips the redundant Vulkan signal when the fence has already reached the requested value.

### DIFF
--- a/src/dxvk/dxvk_cmdlist.cpp
+++ b/src/dxvk/dxvk_cmdlist.cpp
@@ -343,9 +343,15 @@ namespace dxvk {
 
       if (isLast) {
         // Signal per-command list semaphores on the final submission
-        for (size_t i = 0; i < m_signalSemaphores.size(); i++) {
-          m_commandSubmission.signalSemaphore(m_signalSemaphores[i].fence->handle(),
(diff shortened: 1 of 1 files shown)


doitsujin:
If it's a community mod then why not fix this properly community mod? This smells like a race condition, not sure what they are trying to synchronize there but it's just blatantly broken and almost certainly won't do what they intend to do.

The suggested hack basically won't work in GPU-bound scenarios either because there is absolutely nothing at all that guarantees that we observe the fence value at the time we record the submission, and I'd rather not fuck around with this on the DXVK side because DXVK knows nothing about the D3D12 side, and all the underlying shared fence code lives inside Proton.

shaunren:
Fix for skyrim-community-shaders: https://github.com/community-shaders/skyrim-community-shaders/pull/2673

PR #5851, opened 2026-08-20
Annotate ResizeTarget() as  __declspec(hybrid_patchable)

javelina-pkwy:
Related to FEX issue here: https://github.com/FEX-Emu/FEX/issues/5833

The modding tool REFramework detours this function at runtime to hook itself into the target game. This fails under emulation unless this method is annotated with ```__declspec(hybrid_patchable)```.

Relevant REFramework code can be found here, https://github.com/praydog/REFramework/blob/master/src/D3D12Hook.cpp line number 606.

### DIFF
--- a/src/dxgi/dxgi_swapchain.h
+++ b/src/dxgi/dxgi_swapchain.h
@@ -112,7 +112,8 @@ namespace dxvk {
             UINT                      Height,
             DXGI_FORMAT               NewFormat,
             UINT                      SwapChainFlags) final;
-    
+
+    DXVK_HOTPATCHABLE
     HRESULT STDMETHODCALLTYPE ResizeBuffers1(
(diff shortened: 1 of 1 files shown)


doitsujin:
This is ResizeTarget, which I assume is intentional given that ResizeBuffers is already annotated?

Also, can we just do `ResizeBuffers1` as well while we're at it? I kind of want to avoid situations where we spend the rest of the year annotating all remaining entry points one PR at a time.

javelina-pkwy:
Yes, my mistake. Sorry about that. I was typing too fast and did not notice.

I am hesitant to annotate functions that we don't know are needed, since there is a small performance cost associated with hybrid_patchable. With that said, I don't see a substantial risk in going ahead and doing that one as well while we are thinking about it.

PR #5852, opened 2026-08-21
[util] Only spoof vendor id in New Vegas

Blisto91:
Prevents the confusion from listing every GPU as a 6700 XT in the launchers Graphics Adapter list. As seen multiple times by now.

As far as i can recall neither New Vegas itself or the Reloaded mod bases any ingame decisions on the GPU description, but rather vendor id or detected feature support. I've drafted this to begin with anyway until that has been confirmed in practice.


WinterSnowfall:
But people running the game on Intel iGPUs won't be able to show their friends that they have a 6700 XT now :frog:...

Issue #5856, opened 2026-08-23
DCS Crashing with explosions


ggabriel537:
The game crashes reliably with `C0000005 ACCESS_VIOLATION` inside `d3d11.dll` a few minutes into any mission where ordnance detonates, sometimes dropping a single bomb is enough to trigger it. Reproduces in both singleplayer and multiplayer.

Changing Proton version does not seems to work, already tried GE-Proton 11-5, Experimental and 11.0 CachyOS.

Another user has reported the same issue in the game forum (https://forum.dcs.world/topic/391310-dcs-2928-crashes-with-c0000005-in-d3d11dll-on-linuxwine), the fix provided in the forum does not work either.

### Software information
DCS World 2.9.28.26385, x86_64, multithreaded build (`bin-mt/DCS.exe`), launched with `--no-launcher` as a non-Steam shortcut added to Steam. (https://www.digitalcombatsimulator.com/en/downloads/world/)

WINE_SIMULATE_WRITECOPY=1 PROTON_LOG=1 NVPRESENT_ENABLE_SMOOTH_MOTION=1 MANGOHUD=1 STEAM_COMPAT_MOUNTS=/mnt/nvme1tb,/opt/opentrack taskset -c 0-11 %command% --no-launcher

### System information
- GPU: RTX 5060Ti 16GB
- Driver: 610.57.04
- DXVK version: v3.0.2-21-g3a4c6fa3cb1548d0
- Wine version: 11.0

### Log files

[proton-dcs.zip](https://github.com/user-attachments/files/31352639/proton-dcs.zip)

[dcs.log-20260823-180441.zip](https://github.com/user-attachments/files/31352513/dcs.log-20260823-180441.zip)

doitsujin:
Can you please try this with just plain Proton Experimental or something? It's rather dauting when everyone uses different unofficial Proton builds that have god knows what patch sets applied and use different dxvk builds that I have to fish out by hand (and then can't do much with because there's no debug symbols).

I'm also not sure how to really test this game because 237 GB on Steam is just complete insanity, would take literal days to download and I kind of need the disk space for other things too.

That said, there's a) roughly a million d3dcompiler errors in the log and b) a known issue with the game crashing because it simply requires extreme amounts of memory (see #5799).

ggabriel537:
> Can you please try this with just plain Proton Experimental or something? It's rather dauting when everyone uses different unofficial Proton builds that I then have to manually set up and fish the d3d11.dll out of.
[...]

PR #5862, opened 2026-08-25
[d3d11] Guard against binding an already-destroyed resource view

ggabriel537:
DCS World sometimes binds a shader resource, render target, unordered access or depth-stencil view after it has already released its last reference to it, which is a use-after-free bug in the game itself. Since these view objects are deleted as soon as their reference count hits zero, binding the stale pointer makes DXVK dereference freed memory and crash (see #5856).

Add an opt-in liveness registry for these view objects and check it at every bind/clear/discard entry point that takes a raw view pointer from the application, treating a pointer that no longer refers to a live object like a null bind instead of dereferencing it. This is gated behind a new d3d11.viewUAFGuard option that defaults to off, since both the registration on every view's construction/destruction and the lookup on every bind call add measurable overhead, and is only enabled by default for DCS.exe via the app profile table.

### DIFF
--- a/src/d3d11/d3d11_context.cpp
+++ b/src/d3d11/d3d11_context.cpp
@@ -191,6 +191,9 @@ namespace dxvk {
     if (!pResourceView || (NumRects && pRects))
       return;
 
+    if (unlikely(m_parent->GetOptions()->viewUAFGuard) && !D3D11ViewUAFGuard::isViewLive(pResourceView))
+      return;
+
     // ID3D11View has no methods to query the exact type of
     // the view, so we'll have to check each possible class
     auto dsv = dynamic_cast<D3D11DepthStencilView*>(pResourceView);
@@ -414,6 +417,9 @@ namespace dxvk {
     if (!buf || !uav)
       return;
 
+    if (unlikely(m_parent->GetOptions()->viewUAFGuard) && !D3D11ViewUAFGuard::isViewLive(uav))
+      return;
+
     auto counterView = uav->GetCounterView();
 
     if (counterView == nullptr)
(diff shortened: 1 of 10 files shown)


doitsujin:
Littering highly perf-critical functions with all these checks isn't really an option for something that is a game bug that can easily cause crashes even on Windows. Apps can easily call these several hundred thousand times **per frame**.

Probably more useful to see if `PROTON_HEAP_DELAY_FREE` works, that is literally designed to make things more robust against use-after-free bugs.

ggabriel537:
> Littering highly perf-critical functions with all these checks isn't really an option for something that is a game bug that can easily cause crashes even on Windows. Apps can easily call these several hundred thousand times **per frame**.

Fair point. As I mentioned in the issue, I'm not a C++ or game developer, so feel free to change the code however you see fit. In its current state it at least works, and I haven't observed any noticeable performance impact in my game. I can't say whether other titles would be affected, but in DCS the performance is unchanged.

> Probably more useful to see if `PROTON_HEAP_DELAY_FREE` works, that is literally designed to make things more robust against use-after-free bugs.

I tested it via launch options and the issue still occurs: (Proton Experimental)

WINE_SIMULATE_WRITECOPY=1 PROTON_LOG=1 PROTON_HEAP_DELAY_FREE=1 DXVK_HUD=version,fps NVPRESENT_ENABLE_SMOOTH_MOTION=1 MANGOHUD=1 DXVK_CONFIG="dxgi.maxDeviceMemory=12288" STEAM_COMPAT_MOUNTS=/mnt/nvme1tb,/opt/opentrack taskset -c 0-11 %command% --no-launcher

Logs are below

[dcs-proton-log.zip](https://github.com/user-attachments/files/31442367/dcs-proton-log.zip)
[dcs.log-20260825-235457.zip](https://github.com/user-attachments/files/31442374/dcs.log-20260825-235457.zip)

PR #5867, opened 2026-08-30
[meta] Use Arch container directly in CI 

Blisto91:
Use the Arch base-devel container directly and just install the few missing dependencies needed to build.
This saves a bit of CI time and makes us not dependent on a outside action having to update when some package needs to be added or removed.
This is using the daily updated container https://gitlab.archlinux.org/archlinux/archlinux-docker

### DIFF
--- a/.github/workflows/artifacts.yml
+++ b/.github/workflows/artifacts.yml
@@ -5,8 +5,12 @@ on: [push, pull_request, workflow_dispatch]
 jobs:
   artifacts-mingw-w64:
     runs-on: ubuntu-latest
+    container: archlinux/archlinux:base-devel
 
     steps:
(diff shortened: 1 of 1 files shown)


WinterSnowfall:
> This saves a bit of CI time and makes us not dependent on a outside action having to update when some package needs to be added or removed.

...or when GitHub decides to deprecate whatever Node.js version for actions.

P.S.: Don't forget about vkd3d-proton.

Blisto91:
> P.S.: Don't forget about vkd3d-proton.

There was a ominous background voice (wife) that kept telling me to get to bed so I wouldn't be tired for work. So had to postpone that part till today 😛

PR #5868, opened 2026-08-31
[FFXI] d3d8: state blocks use-after-free on Apply

Xaenalt:
I'd been hitting a pretty obscure use-after-free in FFXI. I had Claude take a stab at fixing it, and it figured out that it seems to have been a reuse in ApplyStateBlock. I added the most relevant summary lines below. It's entirely possible this behavior is specific to FFXI though.

`Capture()` stores `m_device->m_indices.ptr()` (and the analogous stream/texture pointers), and `Apply()` re-binds them via `SetIndices`/`SetStreamSource`/ `SetTexture`. If the application releases a captured resource to refcount zero after the capture, the D3D8 wrapper is destroyed while the state block still holds its pointer; the next `ApplyStateBlock` re-binds freed memory. Native D3D8 state blocks hold references on recorded state objects, so applications are entitled to this pattern.

Fix:

Hold private references in the captured state, exactly like the device's own `m_indices`/`m_streams`/`m_textures` bindings (`Com<T, false>`).

### DIFF
--- a/src/d3d8/d3d8_state_block.cpp
+++ b/src/d3d8/d3d8_state_block.cpp
@@ -100,7 +100,7 @@ namespace dxvk {
     HRESULT res = m_stateBlock->Apply();
 
     if (m_captures.flags.test(D3D8CapturedStateFlag::Indices))
-      m_device->SetIndices(m_state.indices, m_state.baseVertexIndex);
+      m_device->SetIndices(m_state.indices.ptr(), m_state.baseVertexIndex);
 
     // This was a very easy footgun for D3D8 applications.
     if (m_captures.flags.test(D3D8CapturedStateFlag::SWVP))
@@ -109,14 +109,14 @@ namespace dxvk {
     if (m_captures.flags.test(D3D8CapturedStateFlag::VertexBuffers)) {
       for (DWORD stream = 0; stream < m_state.streams.size(); stream++) {
         if (m_captures.streams.get(stream))
-          m_device->SetStreamSource(stream, m_state.streams[stream].buffer, m_state.streams[stream].stride);
+          m_device->SetStreamSource(stream, m_state.streams[stream].buffer.ptr(), m_state.streams[stream].stride);
       }
     }
 
     if (m_captures.flags.test(D3D8CapturedStateFlag::Textures)) {
       for (DWORD stage = 0; stage < m_state.textures.size(); stage++) {
         if (m_captures.textures.get(stage))
-          m_device->SetTexture(stage, m_state.textures[stage]);
+          m_device->SetTexture(stage, m_state.textures[stage].ptr());
       }
     }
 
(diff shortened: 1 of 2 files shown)


WinterSnowfall:
Hello there.

> I'd been hitting a pretty obscure use-after-free in FFXI.

Please open an issue and fill in the issue template, rather than trying to have a go at it yourself. Working with these ancient APIs isn't straight forward, not even for the LLMs which will "save our civilization".

> If the application releases a captured resource to refcount zero after the capture, the D3D8 wrapper is destroyed while the state block still holds its pointer; 

This is yet another example of an LLM being technically correct, but it having nothing to do with what's observed in practice.

Would it be better to hold a reference to the objects captured in a state block? Sure. Does native D3D do it? Nope.

So the implementation is as expected, and, ironically, things are expected to blow up here if an application tries to apply a state block which contains already released objects. This is most likely not observable on Windows in most cases however, because it's known to be far more tolerant of UAF in general.

Unless there's a clear indication this particular quirk is responsible for what you are seeing in the game, I don't think this is something worth considering. And given that state block use in D3D8 is rather common and we haven't seen this explode anywhere else, although we've had other state block related issues in the past which have been fixed, I doubt this is actually the problem here.

P.S.: Holding private references to objects which are supposed to be released can also mess with device loss / reset handling, so it can potentially explode in more common situations.

Xaenalt:
To be clear, I'm not making any claims about LLMs as being beneficial or not to society. Separately, about the patch: ok, I figured it was probably specific to this use case, and is obscure enough I'll just carry the patch locally

WinterSnowfall:
> To be clear, I'm not making any claims about LLMs as being beneficial or not to society.

No worries, didn't think/imply you were. It was I being a bit extra salty on the topic.

> Separately, about the patch: ok, I figured it was probably specific to this use case, and is obscure enough I'll just carry the patch locally

Does it actually solve the issue you were seeing, in a reliable way? It might be something worth addressing down the line, but not with the global approach this PR takes. UAFs in general are a pain to deal with, because replicating exact native behavior is outside of our control.

Xaenalt:
Ah, yeah I understand, LLMs are certainly an issue in many communities. I don't blame maintainers for not wanting to deal with them.

Yes, it does resolve it, otherwise every zone crashes the game if there's an imgui window open, at least on my system. I suspect this is some extreme edge case where the game's not handling them right, it's a pretty old game. I also only applied this in the specific prefix for the game.

WinterSnowfall:
> Yes, it does resolve it, otherwise every zone crashes the game if there's an imgui window open, at least on my system. I suspect this is some extreme edge case where the game's not handling them right, it's a pretty old game. I also only applied this in the specific prefix for the game.

Would appreciate it if you could open an issue for it in the tracker, and potentially grab an apitrace of the crash. I have a FFXI trace around, as it happens, but it plays back just fine, always has.

Issue #5069, opened 2025-07-04
Preview not showing in Adobe Premiere Pro 2024


herobrinecat:
I'm trying to work on the project with Adobe Premiere 2024, but when I'm trying to seek to see the preview, nothing shows up. I think it's because of an unimplemented function in the dxgi, which is

"err:   DxgiSwapChain::SetBackgroundColor: Not implemented"

That message pops up every time I seek in the video I'm making in the terminal.

### Software information
Adobe Premiere Pro 2024 and CUDA acceleration enabled.

### System information
- GPU: NVIDIA RTX 3060
- Driver:570.133.07
- Wine version: 10.11-staging
- DXVK version: [24fcf38](https://github.com/doitsujin/dxvk/commit/24fcf3817ed5dd815f2e4d28c2f3cdc24238e60a)

### Log files

[issue.log](https://github.com/user-attachments/files/21048752/issue.log)

![Image](https://github.com/user-attachments/assets/007b84c0-9b68-4281-965b-8094f03a000b)

doitsujin:
More likely to be related to the fact that we do not implement video decoding APIs. No immediate plans for that either since it's an immense (several months) of work that games simply don't require at the moment.

herobrinecat:
(UPDATE: I somehow made the preview show by making sure that the application uses the non-default gpu)

eszlari:
> More likely to be related to the fact that we do not implement video decoding APIs.

WineD3D supports Vulkan H.264 decode now:
https://gitlab.winehq.org/wine/wine/-/merge_requests/7480

PR #5871, opened 2026-09-02
[dxgi] Implement IDXGISwapChain1::SetBackgroundColor

buwryme:
Adobe After Effects 2024 calls `SetBackgroundColor` during viewport initialization. When this returns `E_NOTIMPL`, the viewport renders as a blank window.

~~This only surfaces when vkd3d or vkd3d-proton is installed, which effectively locks Linux users to AE 2024 -- which was the last version of After Effects that didn't mandate DirectX 12.~~

This only surfaces when vkd3d or vkd3d-proton is installed, which effectively locks users out of DirectX12 plugins (AE 2025 and higher still don't work properly on Wine.)

Previously reported in #5069 for Premiere Pro and closed under the assumption it was a missing video decode API. The actual root cause is this unimplemented DXGI method (per my A/B testing, swapping between dxvk-only and dxvk + vkd3d prefixes).

Per [MSDN](https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgiswapchain1-setbackgroundcolor), this sets the background color DWM uses to fill areas outside presented content. Since DXVK bypasses DWM and presents directly via Vulkan, we simulate this by clearing the swap chain image with the specified color before blitting the back buffer.

The color is threaded through `IDXGIVkSwapChain` to the D3D11 presenter, where it's applied as a `VkClearColorValue` in `PresentImage` before the blit. `GetBackgroundColor` is also implemented to return the stored value.

Notably, this required **zero changes to vkd3d**; the fix lives entirely in DXVK's DXGI and D3D11 swapchain layers, which AE routes through even when DX12 plugins are active.

Tested with Adobe After Effects 2024 on Wine 11.16 + vkd3d-proton (Fedora). Viewport renders correctly; no regressions in other D3D11 apps.

References:
- https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgiswapchain1-setbackgroundcolor
- https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_2/nf-dxgi1_2-idxgiswapchain1-getbackgroundcolor

### DIFF
--- a/src/d3d11/d3d11_swapchain.cpp
+++ b/src/d3d11/d3d11_swapchain.cpp
@@ -338,6 +338,18 @@ namespace dxvk {
     return S_OK;
   }
 
+  HRESULT STDMETHODCALLTYPE D3D11SwapChain::SetBackgroundColor(
+          const DXGI_RGBA*                pColor) {
+    if (!pColor)
+      return E_INVALIDARG;
+
+    m_backgroundColor.float32[0] = pColor->r;
+    m_backgroundColor.float32[1] = pColor->g;
+    m_backgroundColor.float32[2] = pColor->b;
+    m_backgroundColor.float32[3] = pColor->a;
+    return S_OK;
+  }
+
 
   void STDMETHODCALLTYPE D3D11SwapChain::GetLastPresentCount(
           UINT64*                   pLastPresentCount) {
@@ -466,7 +478,8 @@ namespace dxvk {
       cLatency        = m_latency,
       cColorSpace     = m_colorSpace,
       cFrameId        = m_frameId,
-      cDirtyRects     = std::move(dirtyRects)
+      cDirtyRects     = std::move(dirtyRects),
+      cBgColor        = m_backgroundColor
     ] (DxvkContext* ctx) {
       // Update back buffer color space as necessary
       if (cSwapImage->image()->info().colorSpace != cColorSpace) {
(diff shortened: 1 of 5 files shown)


doitsujin:
Which part of the series actually fixes this?

I don't think the clear actually does anything because we don't implement the scaling modes currently, nothing has used those so far. If it's just relying on the DXGI methods not returning `E_NOTIMPL` then we should probably focus on that for now and I'll look into implementing the actually missing presenter bits later.

buwryme:
> Which part of the series actually fixes this?
> 
> I don't think the clear actually does anything because we don't implement the scaling modes currently, nothing has used those so far. If it's just relying on the DXGI methods not returning `E_NOTIMPL` then we should probably focus on that for now and I'll look into implementing the actually missing presenter bits later.

yup the clear doesn't fix the missing scaling logic itself, this patch is purely a compatibility shim to prevent `e_notimpl` from blocking after effects' (or premiere's) viewport being blank

without these stubs, the app renders the viewport black regardless of the content. this pr (as of it's current state) gets us past the initialization hurdle so the viewport actually presents *something*, even if stretched incorrectly.

i could rework this into pure method stubs returning `s_ok` + warnings if you prefer that approach for the eventual presenter rewrite

doitsujin:
> i could rework this into pure method stubs returning s_ok + warnings if you prefer that approach for the eventual presenter rewrite

Think having the current background color stored in the presenter works for now, but we can't currently do anything with it, so getting rid of the clear would be best.

doitsujin:
On second thought, there's another issue because this patch changes `IDXGIVkSwapChain`, but that is a public interface also used by vkd3d-proton so we can't just add methods to that, needs an extended interface.

I'll just deal with this and work your commit into all of that, at least we now know that this functionality is actually used in the wild.

buwryme:
> On second thought, there's another issue because this patch changes `IDXGIVkSwapChain`, but that is a public interface also used by vkd3d-proton so we can't just add methods to that, needs an extended interface.
> 
> I'll just deal with this and work your commit into all of that, at least we now know that this functionality is actually used in the wild.

alright, sounds good!

doitsujin:
Please let me know if #5873 works with Adobe.

buwryme:
> Please let me know if #5873 works with Adobe.

<div align="center"">
  image
</div>

Runs wonderfully! 

The patch seems to fix CEP extension panels too, as of my observing (uses Chromium Embedded Framework)

Adobe After Effects 2024, wine 11.16 staging tkg
Results are identical on After Effects version 2020.

If it ruins DXVK Native, you might want to make this configurable, and off by default (?)

PR #5896, opened 2026-09-11
[d3d9] Update window handle during swapchain reset

### LINKED ISSUE #3762: Painkiller Black Edition (the original, not the HD remaster) doesn't size it's game window properly

WinterSnowfall:
The minimally invasive change from a patch set suggested by @GloriousEggroll . It appears to fix the Painkiller WSI jank, by making sure the *actual* window is being used to enter fullscreen mode during a swapchain reset.

It makes sense to me, since we have the same logic applied on Present, however with the current code we end up with a stale window used during Reset, so fullscreen mode isn't signaled properly. As with any WSI change, this has the potential to explode in other places, so I'll leave it as draft until it gets tested thoroughly.

Fixes #3762 at least, could potentially help in other cases.

### DIFF
--- a/src/d3d9/d3d9_swapchain.cpp
+++ b/src/d3d9/d3d9_swapchain.cpp
@@ -628,6 +628,11 @@ namespace dxvk {
 
     bool changeFullscreen = m_presentParams.Windowed != pPresentParams->Windowed;
 
+    if (m_window != pPresentParams->hDeviceWindow) {
+      m_window = pPresentParams->hDeviceWindow;
+      m_displayRefreshRateDirty = true;
+    }
+
     if (pPresentParams->Windowed) {
       if (changeFullscreen)
         this->LeaveFullscreenMode();


DarkSwan86:
@WinterSnowfall thanks a lot for the patch (also thanks @GloriousEggroll !). If you need more games to test here some from a [ticket on Wine bugtracker](https://bugs.winehq.org/show_bug.cgi?id=52959) that has the same behaviour like Painkiller:

-Empire Earth: Gold
-Commandos: Behind Enemy Lines
-Plants vs Zombies 1 
-Cave Story

WinterSnowfall:
> -Empire Earth: Gold -Commandos: Behind Enemy Lines -Plants vs Zombies 1 -Cave Story

All the games you've mentioned are either D3D7 or DDraw, so not relevant for DXVK. That being said Empire Earth and PvZ work just fine with D7VK, and the problem doesn't even apply because D3D7 and earlier doesn't even have the concept of device/swapchain resets.

WSI issues are sadly not uncommon, but that doesn't mean they're all the same either.

WinterSnowfall:
I've done some random testing of games which are known to reset the device "a lot". And by a lot I mean when going between the main menu and the actual game or whatnot. It seems fine.

Device resets themselves aren't that common in D3D9, only 173 games out of 511 that I have traces of do at least one Reset, and I'm fairly sure almost none of them mess with the HWND during the calls. It might just be Painkiller is fairly singular in that regard.

Guess only time will tell.

PR #5902, opened 2026-09-14
[d3d8] Synchronize the device lock helpers

WinterSnowfall:
The d3d9-side rework completely flew under my radar. We basically have a "shadow copy" of it (for convenience) in d3d8, and there's no point for the implementation to be any different to d3d9's really.


doitsujin:
any particular reason why we can't just #include the d3d9 code and `using D3D8Multithread = D3D9Multithread;`? Don't think we need to duplicate this code.

WinterSnowfall:
> any particular reason why we can't just #include the d3d9 code and `using D3D8Multithread = D3D9Multithread;`? Don't think we need to duplicate this code.

The D3D9 code imports "d3d9_device.h" and we do a D3D9 namespace separation in D3D8 which keeps things sane. I've tried that in the past to be honest, but it's just more trouble than it's worth as it spawns a lot of definition conflicts.

P.S.: Let me have another go at it, I guess.

WinterSnowfall:
Yeah, no, it's an absolute mess... and not because of "d3d9_device.h", because it pulls "d3d9_include.h" in the current namespace and that explodes in D3D8.

K0bin:
I don't see a reason why the device include in d3d9_multithread.cpp is necessary at a quick glance. Just remove that (or replace it with more fine grained includes).

WinterSnowfall:
> I don't see a reason why the device include in d3d9_multithread.cpp is necessary at a quick glance. Just remove that (or replace it with more fine grained includes).

I was wrong about that, the problem is pulling in "d3d9_include.h" into the existing namespace (via "d3d9_multithread.h") that we derp all D3D9 interfaces out of in D3D8. I don't see an easy way out, which is why I went with duplicating the class in the first place.

WinterSnowfall:
As discussed on Discord, ideally this will be moved out of d3d9 code and into util, to be shared across "early D3D".

PR #5904, opened 2026-09-14
Flat shading fixes

doitsujin:
Turns out this didn't work *at all* outside of fixed-function.


src/d3d9/d3d9_shader_analysis.cpp
```diff
@@ -147,6 +147,15 @@ namespace dxvk {
           }
         } break;
 
+        case RegisterType::eInput: {
```
K0bin, line 150: I'd prefer if we just reused the code below which does the same thing. SM2 does have DCL instructions for PS inputs. 

```cpp

    Semantic semantic;
    if (registerType == RegisterType::eInput
        && GetShaderInfo().getType() == ShaderType::ePixel
        && GetShaderInfo().getVersion().first < 3)
        semantic = { dxbc_spv::sm3::SemanticUsage::eColor, index };
    else
        semantic = { dcl.getSemanticUsage(), dcl.getSemanticIndex() };

  if (GetShaderInfo().getType() == ShaderType::ePixel
         && semantic.usage == SemanticUsage::eColor
         && semantic.index < 2u) {
```

K0bin: SM1 doesn't have DCLs, so nvm.

K0bin (approved):

PR #5909, opened 2026-09-17
ARM64X cross-file and CI

ivyl:
Sadly llvm-mingw is not packaged in Arch so I had to use their GitHub release. I've tried to make updates as simple as possible and included the version string only once.

I've modified `package-release.sh` to also build the arm64x version of the artifacts but I'm not sure if this is not going to be too annoying to regular users. You'll have to have the toolchain available.

```
.
├── arm64x
│   ├── d3d10core.dll
│   ├── d3d11.dll
│   ├── d3d8.dll
│   ├── d3d9.dll
│   └── dxgi.dll
├── x32
│   ├── d3d10core.dll
│   ├── d3d11.dll
│   ├── d3d8.dll
│   ├── d3d9.dll
│   └── dxgi.dll
└── x86
    ├── d3d10core.dll
    ├── d3d11.dll
    ├── d3d8.dll
    ├── d3d9.dll
    └── dxgi.dll
```

If it is I can hide bit behind `--build-arm64x` instead. I can also make it a separate artifact / tarball if you prefer that.


Blisto91:
> I've modified package-release.sh to also build the arm64x version of the artifacts but I'm not sure if this is not going to be too annoying to regular users. You'll have to have the toolchain available.
> 
> If it is I can hide bit behind --build-arm64x instead.

Personally i think that would be better as to not have it be a breaking change for anyone that uses the script currently.
But it is more up to @doitsujin ofc.

PR #5912, opened 2026-09-18
[Present timing] added fifo latest ready as fifo mode

BalisticNick:
Add fifo_latest_ready as a fifo mode for present timing for people using nvidia gpu's on x11.


doitsujin:
This does absolutely nothing useful on its own, we never pick that present mode.

Not sure what this would do for us anyway, we either want FIFO (vsync on) or IMMEDIATE (vsync off, no timing).

BalisticNick:
True there isn't anything directly in dxvk to change the present mode but still you can change the present mode to FIFO_LATEST_READY using mangohud. 

This would more be for those who want a low-latency alternative to fifo, basically this is just mailbox with timing.

doitsujin:
The override happens at a Mangohud level though, DXVK isn't aware of it in the slightest. Again, this patch does literally nothing.

In general I'm extremely tired of this low latency snake oil everyone is throwing around as of late. If you want low latency, you **do not want timing**. If you want timing for smooth presentation, you **do not want mailbox** because it literally defeats the purpose. 

`FIFO_LATEST_READY` is in theory usable when running above maximum refresh, but requires more backbuffers and you get the latency **increase** from timing in combination with uneven pacing because you're running above maximum refresh rate. I don't see a use case for gaming here.

BalisticNick:
okay I read the code for a span of more than 5 seconds, I see, sorry for wasting your time. 
I'm still going to work on this and actually make sure it works before making another pull request, but from the sounds of it, I think it would be best I just make this into another fork.

PR #5913, opened 2026-09-18
[util] Use one-arg pthread_setname_np on macOS

OmniTroid:
macOS's pthread_setname_np takes only the name (it always targets the calling thread), unlike the Linux/BSD two-arg form.


doitsujin (commented):
of coooooooooooourse this fucking platform keeps being annoying.

Only reason I'll take this is because it's a literal one-liner, anything more complex than that isn't maintainable given that nobody here runs Apple software.

Issue #5919, opened 2026-09-22
Present1 with dirty rects: the first incremental present after a full one composes over black


giang17:
The incremental present path (#5854) initialises the composition buffer from
`m_backBuffers.back()` on the assumption that it holds the frame last presented
(`src/d3d11/d3d11_swapchain.cpp`, `CompositeIncrementalPresent`, "we know for sure that
the last back buffer contains the full contents last presented"). After a single full
`Present` that buffer holds no frame (it reads `000000`). The first `Present1` with a dirty rect then composes
the rect over black, and nothing outside later dirty rects is ever drawn again.

### Reproducer

`repro-dxvk-dirty-rect-black.c` (https://gist.github.com/giang17/eb7cac79fe2470c5741684f282a4a07d): an HWND swapchain, `FLIP_SEQUENTIAL`, two
buffers. Frame 1 clears the whole buffer to orange and calls `Present`. From frame 2 on,
once a second, the top-left quarter is cleared (green / orange alternating) and presented
through `Present1` with that quarter as the only dirty rect. Each frame prints what
buffer 0 holds inside and outside the quarter.

```
x86_64-w64-mingw32-gcc -O2 -Wall -o repro-dxvk-dirty-rect-black.exe repro-dxvk-dirty-rect-black.c -ld3d11 -ldxgi -ldxguid -luser32
```

| | window from frame 2 on | buffer 0 outside the rect, frames 2 / 3 |
|---|---|---|
| DXVK 3.1.1, llvmpipe and NVIDIA RTX 2000 Ada | **black** with the quarter | `000000` / `ff8000` (`ff7f00` on NVIDIA) |
| Windows 10 19044, Basic Render Driver | — (buffer numbers only) | `ff8000` / `ff8000` |

The frame-3 reading on DXVK shows the buffer rotation: buffer 0 is the buffer of frame 1
again, orange outside the rect, with the green quarter drawn into it at frame 2. On Windows buffer 0 reads the previous frame after every present, orange
outside the rect throughout — the buffers stay complete there, so anything that reads a
back buffer (`IDXGISurface1::GetDC`, a copy for composition) sees whole frames, while on
DXVK it sees alternating partial ones — `D3D11GDISurface::Acquire` copies the raw buffer
(`src/d3d11/d3d11_gdi.cpp`). That second point is the one #5853 described from
the application side.

Same result with Wine 11.0 stable and with a Wine fork carrying unrelated graphics
[...]

PR #5920, opened 2026-09-22
[d3d11] Keep the back buffer complete across incremental presents

giang17:
Fixes the black frame around the first dirty rect after a full present (#5919), and keeps the raw back buffers complete across incremental presents.

Measured with the reproducer from that issue (HWND swapchain, `FLIP_SEQUENTIAL`, full
present, then `Present1` with one quarter as the dirty rect):

| | window from frame 2 on | buffer 0 outside the rect after each present |
|---|---|---|
| 3.1.1 (this code unchanged on master) | black with the quarter | `000000`, `ff8000`, alternating |
| this PR | orange with the quarter | `ff8000` |
| Windows 10 19044 | — | `ff8000` |

The second change costs one image copy per incremental present, on the path that already
composites the dirty rects; it is what makes `GetDC` on the back buffer, or any copy taken
from it, return whole frames. Without it, a consumer reading the back buffer alternates
between two partial frames — seen at 60 Hz with KORG Trinity, a JUCE 8 GUI (full first
frame, dirty rects afterwards), once its composition swapchain was read back for
composition; with this PR the window is stable.


doitsujin:
We *already do* a copy from the last presented back buffer.

doitsujin:
yeah idk what this is trying to do and why, please explain *in the bug report* what the problem is in something that actually resembles somewhat coherent English.

